PR #45, opened 2021-01-29
Add more block store tests

jespino:


server/services/store/storetests/blocks.go
```diff
@@ -310,3 +317,128 @@ func testGetRootID(t *testing.T, store store.Store, tearDown func()) {
...
+	InsertBlocks(t, store, blocksToInsert)
+
+	blocks, err = store.GetAllBlocks()
+	require.NoError(t, err)
+	require.Len(t, blocks, initialCount+6)
+
+	t.Run("from root id", func(t *testing.T) {
+		parentID, err := store.GetParentID("parent")
+		require.NoError(t, err)
+		require.Equal(t, "", parentID)
```
jespino, line 375: @chenilim is this the expected behavior?

server/services/store/storetests/blocks.go
```diff
@@ -310,3 +317,128 @@ func testGetRootID(t *testing.T, store store.Store, tearDown func()) {
...
+
+	t.Run("exiting id multiple times", func(t *testing.T) {
+		// Wait for not colliding the ID+insert_at key
+		time.Sleep(1 * time.Millisecond)
+		err := store.DeleteBlock("block1", userID)
+		require.NoError(t, err)
+		// Wait for not colliding the ID+insert_at key
+		time.Sleep(1 * time.Millisecond)
+		err = store.DeleteBlock("block1", userID)
+		require.NoError(t, err)
```
jespino, line 435: @chenilim Is this the expected behavior?

server/services/store/storetests/blocks.go
```diff
@@ -310,3 +317,128 @@ func testGetRootID(t *testing.T, store store.Store, tearDown func()) {
...
+	})
+
+	t.Run("from not existing id", func(t *testing.T) {
+		// Wait for not colliding the ID+insert_at key
+		time.Sleep(1 * time.Millisecond)
+		err := store.DeleteBlock("not-exists", userID)
+		require.NoError(t, err)
```
jespino, line 442: @chenilim Is this the expected behavior?

server/services/store/storetests/blocks.go
chenilim: Yes, the parentID is null (returned as "") for root blocks.

server/services/store/storetests/blocks.go
chenilim: That's the current expected behavior - you can delete a block multiple times, since delete is an insert.

Off the top of my head, I think the current behavior is ok, optimizing for fast updates and deletes. Have to think of the downstream implications more.

server/services/store/storetests/blocks.go
chenilim: Yes, same as above. Deleting something that doesn't exist won't impact the model, except to bloat the database. It's a "silent failure" on the client. Have to think more about unintended effects, but I think this is the right behavior for now.

chenilim (approved):

PR #46, opened 2021-02-02
Websocket auth

chenilim:
Added an auth op to websocket. Main structural changes:
* Moved GetSession from App into a new Auth class
* Pass Auth class to webSocket and App
* OctoListener passes session token on open

This probably needs refactoring in the future, but I can see more code going into an auth layer above app later on.

### DIFF
--- a/server/app/blocks_test.go
+++ b/server/app/blocks_test.go
@@ -26,7 +26,7 @@ func TestGetParentID(t *testing.T) {
 	auth := auth.New(&cfg, store)
 	wsserver := ws.NewServer(auth, true)
 	webhook := webhook.NewClient(&cfg)
-	app := New(&cfg, store, wsserver, &mocks.FileBackend{}, webhook)
+	app := New(&cfg, store, auth, wsserver, &mocks.FileBackend{}, webhook)
 
 	t.Run("success query", func(t *testing.T) {
 		store.EXPECT().GetParentID(gomock.Eq("test-id")).Return("test-parent-id", nil)
--- a/server/app/blocks_test.go
+++ b/server/app/blocks_test.go
@@ -14,10 +14,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func isValidSessionToken(token string) bool {
-	return true
-}
-
 func TestGetParentID(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
--- a/server/app/blocks_test.go
+++ b/server/app/blocks_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
+	"github.com/mattermost/focalboard/server/auth"
 	"github.com/mattermost/focalboard/server/services/config"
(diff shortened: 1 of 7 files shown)


server/ws/websockets.go
```diff
@@ -154,20 +109,123 @@ func (ws *Server) handleWebSocketOnChange(w http.ResponseWriter, r *http.Request
...
+
+	// Authenticated
+	wsSession.isAuthenticated = true
+	log.Printf("authenticateListener: Authenticated")
+}
+
+// AddListener adds a listener for a block's change.
+func (ws *Server) addListener(wsSession *websocketSession, blockIDs []string) {
+	if !wsSession.isAuthenticated {
+		log.Printf("addListener: NOT AUTHENTICATED")
```
jespino, line 158: Maybe sending a response like `{"error": "NOT AUTHENTICATED"}`?

server/ws/websockets.go
```diff
@@ -154,20 +109,123 @@ func (ws *Server) handleWebSocketOnChange(w http.ResponseWriter, r *http.Request
...
+
+		ws.listeners[key] = listeners
+	}
+	ws.mu.Unlock()
+}
+
+// removeListenerFromBlocks removes a webSocket listener from a set of block.
+func (ws *Server) removeListenerFromBlocks(wsSession *websocketSession, blockIDs []string) {
+	if !wsSession.isAuthenticated {
+		log.Printf("removeListenerFromBlocks: NOT AUTHENTICATED")
```
jespino, line 193: Same here

server/ws/websockets.go
```diff
@@ -154,20 +109,123 @@ func (ws *Server) handleWebSocketOnChange(w http.ResponseWriter, r *http.Request
...
+				break
+			}
+		}
+	}
+
+	ws.mu.Unlock()
+}
+
+// GetListeners returns the listeners to a blockID's changes.
+func (ws *Server) GetListeners(blockID string) []*websocket.Conn {
```
jespino, line 221:
```suggestion
// getListeners returns the listeners to a blockID's changes.
func (ws *Server) getListeners(blockID string) []*websocket.Conn {
```
and of course is needed to change the call to `GetListeners` called later.

server/app/blocks_test.go
```diff
@@ -13,14 +14,19 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func isValidSessionToken(token string) bool {
```
jespino, line 17: Is this use anywhere?

jespino (approved):
a couple of tiny comments, but otherwise looks good to me.

server/app/blocks_test.go
chenilim: Ah, no. Was trying a different approach. Will clean this up.

server/ws/websockets.go
chenilim: Good call. Added an error message to the client.

server/ws/websockets.go
chenilim: Thanks. Cleaned up the function visibility.

PR #48, opened 2021-02-03
Add X-Requested-With header to prevent CSRF

chenilim:
Add X-Requested-With=XMLHttpRequest header to all REST API requests, and check on the server.
* [Reference article](https://markitzeroday.com/x-requested-with/cors/2017/06/29/csrf-mitigation-for-ajax-requests.html)

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -18,6 +18,11 @@ import (
 	"github.com/mattermost/focalboard/server/utils"
 )
 
+const (
+	HEADER_REQUESTED_WITH     = "X-Requested-With"
+	HEADER_REQUESTED_WITH_XML = "XMLHttpRequest"
+)
+
 // ----------------------------------------------------------------------------------------------------
 // REST APIs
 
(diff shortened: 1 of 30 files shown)


server/api/api.go
```diff
@@ -35,35 +40,61 @@ func (a *API) app() *app.App {
...
-	r.HandleFunc("/api/v1/workspace/regenerate_signup_token", a.sessionRequired(a.handlePostWorkspaceRegenerateSignupToken)).Methods("POST")
+	a.addHandler(r, "/api/v1/workspace", "GET", a.sessionRequired(a.handleGetWorkspace))
+	a.addHandler(r, "/api/v1/workspace/regenerate_signup_token", "POST", a.sessionRequired(a.handlePostWorkspaceRegenerateSignupToken))
 }
 
 func (a *API) RegisterAdminRoutes(r *mux.Router) {
 	r.HandleFunc("/api/v1/admin/users/{username}/password", a.adminRequired(a.handleAdminSetPassword)).Methods("POST")
 }
 
+func (a *API) addHandler(r *mux.Router, path string, method string, f func(http.ResponseWriter, *http.Request)) {
```
jespino, line 72: This looks a look like a middleware

server/api/api.go
```diff
@@ -35,35 +40,61 @@ func (a *API) app() *app.App {
...
-	r.HandleFunc("/api/v1/users/me", a.sessionRequired(a.handleGetMe)).Methods("GET")
-	r.HandleFunc("/api/v1/users/{userID}", a.sessionRequired(a.handleGetUser)).Methods("GET")
-	r.HandleFunc("/api/v1/users/{userID}/changepassword", a.sessionRequired(a.handleChangePassword)).Methods("POST")
+	a.addHandler(r, "/api/v1/users/me", "GET", a.sessionRequired(a.handleGetMe))
+	a.addHandler(r, "/api/v1/users/{userID}", "GET", a.sessionRequired(a.handleGetUser))
+	a.addHandler(r, "/api/v1/users/{userID}/changepassword", "POST", a.sessionRequired(a.handleChangePassword))
 
-	r.HandleFunc("/api/v1/login", a.handleLogin).Methods("POST")
-	r.HandleFunc("/api/v1/register", a.handleRegister).Methods("POST")
+	a.addHandler(r, "/api/v1/login", "POST", a.sessionRequired(a.handleLogin))
```
jespino, line 52: You are adding `sessionRequired` to the login and register ;)

server/api/api.go
jespino: look here: https://github.com/gorilla/mux#middleware

jespino (changes requested):
A couple of changes proposed but otherwise looks good to me.

jespino (changes requested):
I think is better to use a middleware in this case. and you moved the login/register api endpoints behind a sessionRequired.

server/api/api.go
```diff
@@ -35,35 +40,65 @@ func (a *API) app() *app.App {
...
-	r.HandleFunc("/api/v1/sharing/{rootID}", a.sessionRequired(a.handleGetSharing)).Methods("GET")
+	// Files API
 
-	r.HandleFunc("/api/v1/workspace", a.sessionRequired(a.handleGetWorkspace)).Methods("GET")
-	r.HandleFunc("/api/v1/workspace/regenerate_signup_token", a.sessionRequired(a.handlePostWorkspaceRegenerateSignupToken)).Methods("POST")
+	files := r.PathPrefix("/files/").Subrouter()
+	files.Use(a.requireCSRFToken)
+
+	files.HandleFunc("/", a.sessionRequired(a.handleUploadFile)).Methods("POST")
+	files.HandleFunc("/{filename}", a.sessionRequired(a.handleServeFile)).Methods("GET")
```
jespino, line 73: I'm not 100% sure about the need of the CSRFToken check for the static files, actually I think it should be only for the API.

jespino: I think we don't have much control over the image loading and files downloading in the browser to add that header.

server/api/api.go
```diff
@@ -35,35 +40,65 @@ func (a *API) app() *app.App {
 }
 
 func (a *API) RegisterRoutes(r *mux.Router) {
-	r.HandleFunc("/api/v1/blocks", a.sessionRequired(a.handleGetBlocks)).Methods("GET")
-	r.HandleFunc("/api/v1/blocks", a.sessionRequired(a.handlePostBlocks)).Methods("POST")
-	r.HandleFunc("/api/v1/blocks/{blockID}", a.sessionRequired(a.handleDeleteBlock)).Methods("DELETE")
-	r.HandleFunc("/api/v1/blocks/{blockID}/subtree", a.attachSession(a.handleGetSubTree, false)).Methods("GET")
+	apiv1 := r.PathPrefix("/api/v1").Subrouter()
```
jespino, line 43: nice touch

server/api/api.go
chenilim: Thanks Jesus! I knew there was a cleaner way. Refactored to use Middleware.

server/api/api.go
chenilim: Ack, fixed that. Much easier to see with the middleware pattern.

server/api/api.go
chenilim: Actually, I had refactored contentBlock to get the images with fetch, passing the session token and CSRF header. However, I removed the CSRF check for GET /files for now, as that (in theory) shouldn't be a CSRF vulnerability as no state is changed. We can think about this one more, and change later if needed.

server/api/api.go
chenilim: Thanks. I smoke tested this and it looks ok, but please LMK if you see anything off.

jespino: Nope, everything is looking great now

jespino (approved):

PR #52, opened 2021-02-10
Single user token

chenilim:
I changed the -single-user flag to accept a string token. If specified, this is the global session token we check against. The desktop apps then inject this as the localStorage sessionId, which is used normally.
* The benefits of this approach is that we use the same auth system for single-user, so virtually no change in the webapp code (I cleaned up octoClient to pull the tokens at request-time).
* The downside is we no longer support a single-user server scenario. I think that is a good idea for now, as that reduces our security attack surface. We can add an open-server scenario later if we need to.


.vscode/launch.json
```diff
@@ -12,6 +12,22 @@
 			"program": "${workspaceFolder}/server/main",
 			"cwd": "${workspaceFolder}"
         },
+        {
```
jespino, line 15: The indentation of this file looks broken.

chenilim: Thanks, pushed a fix, and also cleaned up tab consistency in a few files.

webapp/src/octoClient.ts
```diff
@@ -11,16 +11,17 @@ import {Utils} from './utils'
 //
 class OctoClient {
     readonly serverUrl: string
-    token?: string
-    readonly readToken?: string
+    get token(): string {
+        return localStorage.getItem('sessionId') || ''
```
jespino, line 15: I think the problem of this is, it is going to read to token every single time from the local storage, that is not the fastest thing ever. I would recommend to cache it in a private `_token` variable.

win/main.go
```diff
@@ -49,20 +53,38 @@ func main() {
...
 
-	ui, err := lorca.New("http://localhost:8088", "", 1024, 768)
+	defer func() {
+		log.Println("Cleanup")
+		cancel()
+		if err := cmd.Process.Kill(); err != nil {
+			log.Fatal("failed to kill server process: ", err)
+		}
+
+		if r := recover(); r != nil {
```
jespino, line 63: This new recover here is capturing the panics at lorca.New and ui.Load? for example? makes sense?

win/main.go
```diff
@@ -49,20 +53,38 @@ func main() {
...
+			log.Fatal("ERROR: ", r)
+		}
+	}()
+
+	ui, err := lorca.New("", "", 1024, 768)
 	if err != nil {
+		log.Panic(err)
```
jespino, line 70: Can we "fail more gracefully here?, something like printing a regular error?

webapp/src/octoClient.ts
chenilim: I did this originally because I was worried about timing (when sessionIds change after init), so wanted to push the "source of truth" to the localStorage. There's an issue when sessionIds get stale (e.g. on refresh) in the OctoListener too.

Is localStorage perf a concern? I did a [quick search](https://stackoverflow.com/questions/8074218/speed-cost-of-localstorage) and it should be on the order of microseconds.

Might need to refactor to a central "sessionManager" later.

win/main.go
chenilim: tbh, I'm not sure what the best pattern here is. The original issue was cleanup wasn't being called on crashes, so I deferred it. In the cleanup, the app is exiting anyway, so I added the recover and log to hide the stack trace. What would you recommend instead?

win/main.go
chenilim: The idea here was to use Panic like a throw, so we can catch and run cleanup (Fatal was bypassing cleanup). Cleanup then recovers, then prints the error. Is there a better pattern to use?

jespino:
I have some concerns about security here. I think the command line argument for the token is a problem, because can be readed as part of the process list. I think we should use environment variables for this, that are more secure (not in every operating system, but for sure in most of them). @jupenur do you have any advice around that?

jupenur:
> I think the command line argument for the token is a problem, because can be readed as part of the process list.

I don't *think* there's an issue here, since the command line in the process listing should only be readable to the owner of the process (and root) anyway. The root user having access to the token is not a problem since the token is randomly generated.

I'm not opposed to passing it in an environment variable, though. I think it'd be just a little bit cleaner.

chenilim:
Thanks Jesus and @jupenur. I've moved the token to an environment variable. It's cleaner and also a better practice.

PR #54, opened 2021-03-03
Archive in JSONL format

chenilim:
Changed the .focalboard archive format to JSONL (https://jsonlines.org/)
* The first line is a header, followed by lines for each block, interfaces defined in Archive.ts
* Implemented streaming line-by-line parser for import, but also checked in non-streaming one

### DIFF
--- a/webapp/src/blocks/archive.ts
+++ b/webapp/src/blocks/archive.ts
@@ -39,6 +39,43 @@ class ArchiveUtils {
 
         return content
     }
+
+    static parseBlockArchive(contents: string): IBlock[] {
+        const blocks: IBlock[] = []
+        const allLineStrings = contents.split('\n')
+        if (allLineStrings.length >= 2) {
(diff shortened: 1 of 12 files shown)


webapp/src/blocks/archive.ts
```diff
@@ -2,10 +2,80 @@
...
+        }
+
+        const headerString = JSON.stringify(header)
+        let content = headerString + '\n'
+        for (const block of blocks) {
+            const line: IBlockArchiveLine = {
+                type: 'block',
+                data: block,
+            }
+            const lineString = JSON.stringify(line)
```
chenilim, line 35: JSON.stringify in Chrome outputs a single-line by default, although it's not clear from [the spec](https://tc39.es/ecma262/#sec-json.stringify) if this is true for all JS engines. May need to keep an eye on this.

jespino (approved):
LGTM

PR #58, opened 2021-03-05
Proposing tweak for SEO

it33:
Just a thought on positioning, we can always change it,


chenilim:
Looks good. Thanks @it33!

PR #68, opened 2021-03-09
Refactor content blocks

chenilim:
#### Update 3/10/21
* Reworked the contentRegistry so the contentElements register themselves
* Moved all "contentType"-specific code to the handlers (mainly createBlock for images)
* Kept the ordering logic in the container (contentBlock, cardDetail) - expect this to change later (e.g. grid layout)
* Kept the strong typing in the block (model) layer
  * I can see a way to support fully dynamic contentTypes in the future (e.g. 3rd party plugins), but we should enforce static checking on first-party content

I expect we'll extend and clean up the handler interface much more, but LMK what you think of this structure as a first step.

#### Summary
Cleaned up and consolidated content blocks
* Made type checking dynamic
* Separated out each content element into its own component
* Most logic is in new ContentRegistry

Left hydrateBlocks in OctoUtils - in a future step we can refactor that and related functions into the block (model) layer.


webapp/src/components/contentBlock.tsx
```diff
@@ -168,24 +90,64 @@ class ContentBlock extends React.PureComponent<Props, State> {
...
             </div>
         )
     }
+
+    private addContentMenu(type: BlockTypes): JSX.Element {
+        const {intl, card, contents, block} = this.props
+        const index = contents.indexOf(block)
+
+        switch (type) {
+        case 'image': return (
```
jespino, line 106: Looks like every new content that needs a bit of special treatment end ups needed an entry here. I would expect the `onClick` to be handled somehow as part of the content definition.

jespino: Same for the icon.

jespino: I think you are oversimplifing it saying that a new content type is just a Component. That is only the rendering part, I would love to have the whole thing self-contained. That would help to allow even add plugins completely client side.

chenilim: Agree that's the direction. Roughly encapsulating that in "RegistryEntry" right now. In your mind:
* Would the plugins generate React components for display? What's the alternative?
* Would they handle the onClicks, or provide an interface of actions (create, update, etc.)?

Have to think more about what's needed for fully dynamic extensions via plugins. E.g. namespacing. In theory, a plugin would implement (a future iteration of) RegistryEntry.

webapp/i18n/en.json
```diff
@@ -24,13 +24,8 @@
   "CommentsList.send": "Send",
   "ContentBlock.Delete": "Delete",
   "ContentBlock.DeleteAction": "delete",
-  "ContentBlock.Text": "Text",
-  "ContentBlock.addDivider": "add divider",
+  "ContentBlock.addElement": "add {type}",
```
jespino, line 27: I think we can just leave this as part of the content type definition, clearly i can see here "Add text" or "upload image" or "embed link"...

webapp/src/blocks/block.ts
```diff
@@ -2,7 +2,10 @@
 // See LICENSE.txt for license information.
 import {Utils} from '../utils'
 
-type BlockTypes = 'board' | 'view' | 'card' | 'text' | 'image' | 'divider' | 'comment'
+const contentBlockTypes = ['text', 'image', 'divider'] as const
```
jespino, line 5: I guess this is about type checking, but I think this is another place that you have to edit to make it work. Is there any way to not handle this cases and make it "dynamic".

webapp/src/components/content/contentRegistry.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+class ContentRegistry {
+    private registry: Map<BlockTypes, RegistryEntry> = new Map()
+
+    get contentTypes(): BlockTypes[] {
+        return [...this.registry.keys()]
+    }
+
+    constructor() {
+        this.registerContentType(
```
jespino, line 38: I would expect the register calls being from outside. Something like `ContentRegistry` is a static class or there is a global `ContentRegistry` instance, and the ContentTypes defines themself and register themself at the end of the file.

webapp/src/blocks/block.ts
chenilim: I can see dynamic plugins in the future, but for the short term, we expect extensions to happen in the code, and we should rely on static type checking as much as possible. Have to think about this more - maybe there will be a special "custom" type that has a dynamic "subtype" which is 3rd party extendable.

webapp/i18n/en.json
chenilim: Need to clean up the i18n later. Right now I'm passing in the intl object to the content handler. Not sure if there's a better way (e.g. construct one in the handler). Also looks like `formatjs extract` is not handling the formatMessage calls in the handlers.

webapp/src/components/content/contentElement.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {IContentBlock} from '../../blocks/contentBlock'
+import {Utils} from '../../utils'
+
+import {contentRegistry} from './contentRegistry'
+
+// Need to require here to prevent webpack from tree-shaking these away
+// TODO: Update webpack to avoid this
+require('./textElement')
+require('./imageElement')
+require('./dividerElement')
```
chenilim, line 16: @jespino, do you happen to know how to add a folder to webpack so we don't need to explicitly import them here?  I found a few references to using chunks, but thought you might know better. More generally, we probably should take a pass at optimizing the webpack settings too.

webapp/src/components/content/contentRegistry.tsx
chenilim: Moved registerContentType to the individual elements.

webapp/src/components/contentBlock.tsx
chenilim: Refactored this based on your comments.

webapp/src/blocks/block.ts
jespino: Yes, as far as it gets caught by the type checker whenever y try to register the new component, works for me. That way, at least we have a explicit error that says "change this here".

webapp/src/components/content/contentElement.tsx
jespino: I'll check that, but shouldn't be hard. We can keep it as is now, because we have the TODO there.

jespino: I rather prefer the `import` syntax

webapp/src/components/content/contentRegistry.tsx
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+/* eslint-disable react/require-optimization */
+import {IntlShape} from 'react-intl'
+
+import {BlockTypes} from '../../blocks/block'
+import {IContentBlock, MutableContentBlock} from '../../blocks/contentBlock'
+import {Utils} from '../../utils'
+
+type ContentHandler = {
```
jespino, line 10: nice

webapp/src/components/content/dividerElement.tsx
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import React from 'react'
+
+import {MutableDividerBlock} from '../../blocks/dividerBlock'
```
jespino, line 5: this is still outside this file, I think is ok for now, and maybe forever, but something to think about.

jespino (approved):
LGTM

PR #72, opened 2021-03-13
Fix setup command

kaakaa:
#### Summary

`ln` command in server setup was failed. It should be `ln -s ${source_file} ${target_file}`.

```bash
$ sudo ln -s /etc/nginx/sites-enabled/focalboard /etc/nginx/sites-available/focalboard
ln: failed to create symbolic link '/etc/nginx/sites-available/focalboard': File exists
```

#### Ticket Link
N/A


chenilim:
Thanks for catching and fixing this @kaakaa!

PR #76, opened 2021-03-17
Add Dockerfile to run service in a container

proffalken:
#### Summary

Add a Dockerfile to launch the application within a container.

*NOTE:* At the moment, the focalboard version is hard-coded and there is
a bug that means the download is not actually a .tar.gz file, but a
.tgz.gz file inside a .zip file.

This also assumes the use of SQLite as the database rather than PgSQL.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/75

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu:latest
+
+RUN apt update 
+RUN apt upgrade -y
+RUN apt install -y wget tar gzip unzip file
+RUN wget https://releases.mattermost.com/focalboard/0.5.0/focalboard-server-linux-amd64.tar.gz
+RUn unzip -o focalboard-server-linux-amd64.tar.gz
+RUN tar -xvzf focalboard-server-linux-amd64.tar.gz
+RUN mv focalboard /opt
+
+EXPOSE 8000
+
+WORKDIR /opt/focalboard
+
+CMD /opt/focalboard/bin/focalboard-server
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,15 @@
 FROM ubuntu:latest
 
-RUN apt update 
-RUN apt upgrade -y
-RUN apt install -y wget tar gzip unzip file
-RUN wget https://releases.mattermost.com/focalboard/0.5.0/focalboard-server-linux-amd64.tar.gz
-RUn unzip -o focalboard-server-linux-amd64.tar.gz
-RUN tar -xvzf focalboard-server-linux-amd64.tar.gz
-RUN mv focalboard /opt
+# Make sure that the underlying container is patched to the latest versions
+RUN apt update && \
+    apt upgrade -y && \
+    apt install -y wget tar gzip unzip file
+
+# Now install Focalboard as a seperate layer
+RUN wget https://releases.mattermost.com/focalboard/0.5.0/focalboard-server-linux-amd64.tar.gz && \
+    unzip -o focalboard-server-linux-amd64.tar.gz && \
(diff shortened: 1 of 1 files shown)


laundmo:
Why does this do a full system upgrade in the container? Doesn't that bloat up the container size extremely?

metanerd:
Hi @proffalken ! Thank you for your PR! Sorry about the spam to set up the check-cla command from me.

proffalken:
> Why does this do a full system upgrade in the container? Doesn't that bloat up the container size extremely?

Nope, it updates a single package at the moment and brings it inline with the latest security releases.

If I were installing Ubuntu-Desktop or similar then yes, that would absolutely add bloat, but this purely ensures that we are running the latest versions of any package that is already installed to ensure we're patched and up to date.  At most, I'd expect it to add a couple of Kb to the image.

proffalken:
> Hi @proffalken ! Thank you for your PR! Sorry about the spam to set up the check-cla command from me.

@metanerd No worries, is there a link to the CLA somewhere so I can sign it? I'm more than happy to do so!

metanerd:
Glad to hear that @proffalken ! The CLA form can be found here: https://mattermost.org/mattermost-contributor-agreement/

proffalken:
> Glad to hear that @proffalken ! The CLA form can be found here: https://mattermost.org/mattermost-contributor-agreement/

Great! All signed @metanerd - looking forward to getting involved!

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu:latest
+
+RUN apt update 
+RUN apt upgrade -y
+RUN apt install -y wget tar gzip unzip file
+RUN wget https://releases.mattermost.com/focalboard/0.5.0/focalboard-server-linux-amd64.tar.gz
+RUn unzip -o focalboard-server-linux-amd64.tar.gz
```
jasonraimondi, line 7: Does this `RUn` typo throw a syntax error?

Also, really this should be one run command so the end result isnt a bunch of unnecessary layers

```dockerfile
FROM ubuntu:latest

RUN apt update \
    && apt install -y wget tar gzip unzip file \
    && wget https://releases.mattermost.com/focalboard/0.5.0/focalboard-server-linux-amd64.tar.gz \
    && tar -xvzf focalboard-server-linux-amd64.tar.gz \
    && mv focalboard /opt \
    && rm -f focalboard-server-linux-amd64.tar.gz # cleanup the unnecessary zip to reduce end container size

EXPOSE 8000

WORKDIR /opt/focalboard

CMD /opt/focalboard/bin/focalboard-server
```

proffalken: Good spot on the typo, but no, it doesn't throw an error.

As far as the "single command" is concerned, I'd be happy to condense it down to 2 commands - one for the OS and one for Focalboard?  Seems like that way we only update the OS layer if it changes, regardless of the application version?

proffalken: @jasonraimondi - I've split it out into two RUN statements instead, would appreciate your feedback?

creekorful:
This is really great! Can't wait to try it with Docker 😄

jespino (approved):
Thanks @proffalken! this is awesome. I clearly see adding changes later to support environment variables to configure Postgres instead of SQLite. And clearly we have to fix the "zip/tgz" problem, but we can do it later.

proffalken:
@jespino - thanks for approving.

FWIW, I logged #77 to register the download format, and yes, totally agree that being able to mount the config file in the container on the appropriate path/set versions from ENV vars etc, is needed! :)

jespino:
Thanks! This is a great first iteration, now we have something to improve and make it more configurable.

PR #80, opened 2021-03-18
updated the document to include `dbtype` in dbconfig

mantlecurve:
#### Summary
The documentation doesn't show to change the dbtype to postgres, which when is kept as sqlite will result in failure. This fixes that.


jespino (approved):
Thanks @mantlecurve! :tada:

PR #85, opened 2021-03-18
Register Request email validation

lutfuahmet:
#### Summary

Updated RegisterRequest validation. Added email validation via regex.


jespino (approved):
Thanks @lutfuahmet! We decided to accept emails in the same way we were accepting them in Mattermost. I think we can accept this changes, be restrictive upfront, and be more permissive in the future if is needed.

By the way, good improvements around the validation and the error strings :)

Issue #96, opened 2021-03-18
Add docker-compose definition


creekorful:
- There's now a docker image available for the server: https://github.com/mattermost/focalboard/pull/76
- There's an issue open to add CD to the docker image: https://github.com/mattermost/focalboard/issues/92

once the image is available on docker hub it could be great to have a little `docker-compose.yml` with all dependencies (postresql, nginx/traefik, etc...) bundled, so that end users may get started as simple as running `docker-compose up`.

Making the compose file would be as simple as following this guide: https://www.focalboard.com/download/personal-edition/ubuntu/

creekorful:
Once the images are available on docker hub it would be a pleasure to work on this one :)

jespino:
@creekorful that would be great, let's work in that direction :) I'll let you know when the docker hub image is read. Probably we need to improve the Dockerfile to allow environment variables first.

cbytestech:
#FocalBoard
  focalboard:
    image: mattermost/focalboard
    container_name: focalboard
    ports:
      - 85:8000
    environment:
      - VIRTUAL_HOST=focalboard.local
      - VIRTUAL_PORT=8000

PR #105, opened 2021-03-18
add docker-compose

jbutler992:
This adds docker-compose capabilities

Includes nginx proxy for TLS termination

Includes postgresql per the getting started suggestions


jbutler992:
Addresses #96 . Will leave it up to repository owner to determine if it closes the issue

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    build:
+      context: ./docker
+    container_name: focalboard
+    expose: 
+      - 8000
+    environment:
+      - VIRTUAL_HOST=focalboard.local
+      - VIRTUAL_PORT=8000
+  
+  proxy:
```
obbardc, line 13: the proxy complicates the example tbh, just expose focalboard on some http port :-)

jbutler992: I can do that, would you be open to a ```docker-compose.yml``` and ```docker-compose-nginx.yml``` setup  so that both examples are available?

jbutler992: added additional docker-compose file

jbutler992:
This check in test failure is also occurring on main. Once that's fixed I'll update with main again so the tests re-run

jespino:
@jbutler992 it should be fixed now :)

jbutler992:
@jespino thanks, updated and passing now

docker/config.json
```diff
@@ -0,0 +1,16 @@
+{
+	"serverRoot": "http://localhost:8000",
+	"port": 8000,
+	"dbtype": "postgres",
+	"dbconfig": "postgres://boardsuser:boardsuser-password@focalboard-db/boards?sslmode=disable&connect_timeout=10",
+	"postgres_dbconfig": "dbname=boards sslmode=disable",
+	"useSSL": false,
+	"webpath": "./pack",
+	"filespath": "./files",
+    "telemetry": true,
```
jespino, line 10: nit: a bit of indentation problem here.

jespino (approved):
Looks good to me. Thanks!

PR #106, opened 2021-03-18
Remove unversioned mattermost-server dependency

ebati:
While checking the code i noticed that `github.com/mattermost/mattermost-server/utils` is used instead of `github.com/mattermost/mattermost-server/v5/utils` which leads to depending on two versions of mattermost-server.

I replaced `mattermost-server/utils` with `mattermost-server/v5/utils` and run `go mod tidy` to remove extra dependencies.


jespino:
Thanks @ebati! good caught! :tada:

jespino (approved):

Issue #132, opened 2021-03-19
Add/Improve unit tests for React components.


renjithgr:
**Summary:**
Most of the React components are not covered by any unit tests. This issue is to track adding new unit tests or improving existing ones for those React components.

People can refer to this issue in their PRs for improving the unit tests.

BabbarRaghav:
Hi, is this issue still open and without any assignees?

I would like to solve it

PR #135, opened 2021-03-19
[GH-132] Add unit tests for propertyMenu component.

renjithgr:
#### Summary
Added unit tests for the change name event, change value event and change type event in the propertyMenu component.


jespino (approved):
Looks good to me. Thanks @renjithgr! :sparkles:

PR #145, opened 2021-03-20
Add support for system-theme (dark/default)

jespino:


jespino:
@chenilim Do you mind to test this in windows/mac? I can test it later in linux.

chenilim:
@jespino, I tested this and it works on Windows 10, the latest Edge (89.0.774.57) and Chrome (89). The theme changes when I change the system theme. A few things:
1. In Edge, there is a setting in the browser (... menu) to override the theme. It should be set to "System Default" to follow the OS setting. Mine was on light mode (may be the default), and it took a while to find.
2. I notice that the matchMedia callback is called twice - looks like loadTheme() is called in main.tsx and in sidebar.tsx. Probably should refactor this to only add the listener once.
3. Also, addListener is [marked deprecated](https://stackoverflow.com/questions/56466261/matchmedia-addlistener-marked-as-deprecated-addeventlistener-equivalent), but it appears this is still the most supported way. Might need to change to the new call later.

Other than that, LGTM.

chenilim (changes requested):

jespino:
I guess the point 1. is working fine. if edge have that feature, should behave as the overwritten edge setting says.

The other two things are solved now.

PR #146, opened 2021-03-20
Adding checkbox content type

jespino:


webapp/src/components/content/checkboxElement.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+type State = {
+    active: boolean
+    title: string
+}
+
+class CheckboxElement extends React.PureComponent<Props, State> {
+    constructor(props: Props) {
+        super(props)
+        this.state = {active: Boolean(props.block.fields.value), title: props.block.title}
```
chenilim, line 29: Recommend adding a field accessor to checkboxBlock to get the "value" state as a bool.

chenilim (approved):
Also need to add a line to hydrateBlock in octoUtils.ts for consistency. However, I see a case to refactor that:
* Remove all the child classes of MutableContentBlock, and use MutableContentBlock directly
  * (In that case, continue using the fields directly instead of adding a value accessor)
* Call the createBlock() methods on the contentHandler as a constructor for each content block type

Maybe that can be done as a separate step?

PR #147, opened 2021-03-20
Adding DockerHub, PR#s, release date to changelog

it33:
Also linking to contributor PRs. Also adding release dates, rather than release months. We might release more than once a month given our velocity right now.


jespino (approved):
:tada:

PR #155, opened 2021-03-20
Adding prefix "Bug: " to issue template

it33:
Make it more consistent with feature request template

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,7 +1,7 @@
 ---
 name: Bug report
 about: Create a report to help us improve
-title: ''
+title: 'Bug: '
 labels: bug
 assignees: ''
 


jespino:
I'm not sure about this, looks a bit redundant to me, with the "Bug: " prefix and the label. If for example somebody report a "bug" that it is in reality a enhancement... we would need to change the label plus the title.

it33:
In my mind, the priority is to be welcoming to the community, including people new to open source and new to GitHub, and having a prefix like "Bug: " makes things easier to understand. 

When I look at the issue list there's a lot of different labels and a prefix would make it more clear to new comers that "This is a bug for which we would welcome contributions". 

image

While I agree it's a touch more work if someone misfiles a ticket, I feel it's more important to clear than it is to be efficient in the failure case.

jespino (approved):
Lets try this and see how it goes :)

PR #156, opened 2021-03-21
[GH-132] - Add snapshot test for propertyMenu component

renjithgr:
I didn't cover scenarios related to the HTML structure of the component with my previous PRs . (For example, what options it shows to the users.) 
I think a snapshot test is better suited for that purpose.


jespino (approved):
Looks good to me. Thanks @renjithgr! :fireworks:

PR #165, opened 2021-03-22
Fix typo and use "token" not  translated term

clyang:


jespino:
Hi Clyang, we don't modify directly the translations files, we do all through translate.mattermost.com, because if not it gets complicated.

clyang:
got it. thx.

PR #168, opened 2021-03-22
recreate dockerfile

mrckndt:
This commit recreates the entire dockerfile for several reasons:
* compile focalboard instead of downloading and extracting a tarball
* it uses an alpine image and intermediate containers to obtain a much
  smaller image
* run focalboard as user inside the container instead of root


docker/Dockerfile
```diff
@@ -1,16 +1,38 @@
-FROM ubuntu:20.04
+FROM golang:alpine as gobuild
 
-# Make sure that the underlying container is patched to the latest versions
-RUN apt-get update && \
-    apt-get install -y gzip tar wget
+RUN apk update && \
+    apk add ca-certificates gcc git make musl-dev && \
+    git clone https://github.com/mattermost/focalboard && \
```
jwilander, line 5: Correct me if I'm wrong but won't this always grab the `main` branch when building the docker image? So when we tag and create a docker image it would create the image from master and not the tag. Same issue when building from a branch

I think we need to pass the branch/tag in as an argument and do a `git checkout $TAG` after the clone, or do a COPY instead of a git clone

flexo3001: Yes, that's true. I'll change this; how about using a `ARG` and specify this while building with `--build-arg TAG=v0.6.1`?

jwilander: Per https://docs.docker.com/docker-hub/builds/advanced/ we can use `$SOURCE_BRANCH` for the dockerhub builds

jespino (approved):
This looks good to me. Waiting for the changes about the `$SOURCE_BRANCH`

docker/Dockerfile
laundmo: I think this should not do a git clone, but copy the local files. This would allow for easily modifying something and re-building it, and people will most likely have the repository cloned when building anyway, at which point cloning it again is counterintuitive.

docker/Dockerfile
```diff
@@ -1,16 +1,38 @@
...
-# Now install Focalboard as a seperate layer
-RUN wget https://github.com/mattermost/focalboard/releases/download/v0.6.1/focalboard-server-linux-amd64.tar.gz && \
-    tar -xvzf focalboard-server-linux-amd64.tar.gz && \
-    mv focalboard /opt
+FROM node:alpine as nodebuild
 
-EXPOSE 8000
+RUN apk update && \
+    apk add ca-certificates git && \
+    git clone https://github.com/mattermost/focalboard && \
```
laundmo, line 13: Why have a second clone here? why not just copy the files over from the prior stage?

laundmo (commented):
The second clone leads to additional network overhead and is plain pointless when files can be copied from the prior stage. 
Additionally, i really think (as i've stated in the comments of another review) that cloning in the Dockerfile is unintuitive, and the local version of the repo should be used.

jespino:
I going to merge this and itreate over that with some tickets like the option to use the local directory to build (instead of the clone), and use the SOURCE_BRANCH env variable to set the correct branch.

jespino:
Merging it without the CLA signature because @mrckndt was @flexo3001 before, and his signature is already in the CLA, also, he is mattermost employee, so I guess the CLA is not strictly needed.

PR #175, opened 2021-03-22
Mattermost auth integration

chenilim:
This allows Focalboard to authenticate via Mattermost, and provides a workspace per channel. To set up:
* Set `"authMode": "mattermost"` in config.json
* Set `"mattermostURL"` to your Mattermost test server's root url (e.g. "https://my.mattermost.com")
* Create an OAuth app in Mattermost, and populate `mattermostClientID` and `mattermostClientSecret`
* (If you have an existing install, clear your browser session, and delete the sessions table to avoid infinite loops - may have some to fix later)

To test:
* Get the channelID of a channel on your test server (from view channel info)
* Navigate to `http://localhost:8000/workspace/<channelID>`
* It should go through the OAuth loop and show an empty workspace
* Boards are scoped to each workspace

Breaking changes to the REST API:
* Add a `/workspaces/<workspaceID>/` prefix to the block APIs

We might need a few passes of reviews. To start, please review the overall login flow (in `mmauth.go`).


server/mmauth/mmauth.go
```diff
@@ -0,0 +1,249 @@
...
+			Props:       map[string]interface{}{},
+		})
+		if err != nil {
+			a.handleError("Unable to create the new user", err, w, r)
+			return
+		}
+	}
+
+	// Create session
+	// TODO: Remove any existing sessions with the same token? (Happens on re-login)
```
jwilander, line 197: Token would change on re-login so that shouldn't be necessary

server/mmauth/mmauth.go
```diff
@@ -0,0 +1,249 @@
...
+	channelId := channels[0].Id
+
+	log.Printf("userId: %s, teamId: %s, channelId: %s", user.Id, teamId, channelId)
+
+	// Create native user
+	// TODO: Update user info with any changes?
+	nativeUser, _ := a.store.GetUserById(user.Id)
+	if nativeUser == nil {
+		err = a.store.CreateUser(&model.User{
+			ID:          user.Id,
```
jwilander, line 181: Directly mapping the focalboard user ID to the MM user ID is an interesting method. In the MM server we use the AuthData field for this and the user would get a unique focalboard ID too. This will work but it will tightly couple the user to the oauth service provider (in this case MM). If you ever wanted to change the auth service from MM to Google for example then this would become problematic since the focalboard user ID would need to change. Same problem if you wanted to support multiple login methods for one focalboard user.

server/mmauth/mmauth.go
```diff
@@ -0,0 +1,249 @@
...
+			a.handleError("Unable to create the new user", err, w, r)
+			return
+		}
+	}
+
+	// Create session
+	// TODO: Remove any existing sessions with the same token? (Happens on re-login)
+	session := model.Session{
+		ID:     uuid.New().String(),
+		Token:  token.AccessToken,
```
jwilander, line 200: This is an interesting approach, using the the MM token as the authentication token for focalboard. My initial thought would be to create a focalboard specific session token and store the MM token in the session separately. Like the user ID above this will work but might get a little weird when using non-MM oauth service providers. I think this works more cleanly than the user ID above but you'd need to make sure that any requirements (eg length, characters) you have on the session token are loose enough to fit whatever pattern any oauth service provider might use for a token.

I can't think of any security reasons why this might be bad but @jupenur might have an opinion there

server/mmauth/mmauth.go
```diff
@@ -0,0 +1,249 @@
...
+		UserID: user.Id,
+		Props: map[string]interface{}{
+			"authService":  "mattermost",
+			"expiry":       token.Expiry.Unix(),
+			"refreshToken": token.RefreshToken,
```
jwilander, line 205: Similar comment to an above one, these seem like they might be worth their fields rather than just being in a prop bag

server/mmoauth/mmoauth.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		a.handleError("unable to create session", err, w, r)
+		return
+	}
+
+	// Pass access token to client via cookie
+	// TODO: Review security of this approach
+	var expiration = time.Now().Add(365 * 24 * time.Hour)
+	tokenCookie := http.Cookie{Name: "oauthtoken", Path: "/", Value: token.AccessToken, Expires: expiration, HttpOnly: false}
+	http.SetCookie(w, &tokenCookie)
```
jwilander, line 217: I don't think this part needs to be specific to an oauth token. I would unify this with how non-oauth login works (which looks like via localstorage). Having two different ways of storing the authentication token on the webapp/client increase the attack surface and chance for bad security bugs IMO. And the webapp/client doesn't really need to care about whether it's oauth or not once it has it's session.

webapp/src/app.tsx
```diff
@@ -31,10 +33,27 @@ export default class App extends React.PureComponent<unknown, State> {
             language: getCurrentLanguage(),
             initialLoad: false,
         }
+
+        this.storeTokenFromCookie()
+    }
+
+    private storeTokenFromCookie() {
```
jwilander, line 40: Per above comment I would suggest handling token storage the same way it's done for regular login rather than supporting both cookie and localstorage as options

jwilander (changes requested):
Good stuff so far 👍 Sorry about the barrage of comments 😅

server/api/api.go
```diff
@@ -101,15 +116,49 @@ func (a *API) checkCSRFToken(r *http.Request) bool {
...
+		return container, nil
+	}
+
+	vars := mux.Vars(r)
+	workspaceID := vars["workspaceID"]
+	container := store.Container{
+		WorkspaceID: workspaceID,
+	}
+
+	if workspaceID == "" || workspaceID == "0" {
```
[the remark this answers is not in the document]
chenilim: Yeah, native auth uses workspaceID 0 for the API calls. It's more a developer-case if you switch providers. DoesUserHaveWorkspaceAccess would fail that case anyway, but this avoids a call.

server/api/api.go
```diff
@@ -101,15 +116,49 @@ func (a *API) checkCSRFToken(r *http.Request) bool {
...
+		// When authenticating, a workspace is required
+		return container, errors.New("No workspace specified")
+	}
+
+	ctx := r.Context()
+	session := ctx.Value("session").(*model.Session)
+	if !a.WorkspaceAuthenticator.DoesUserHaveWorkspaceAccess(session, workspaceID) {
+		return container, errors.New("Access denied to workspace")
```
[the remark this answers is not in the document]
chenilim: Updated the function signature to return nil on error. The error was / is logged on the server, and a generic (no workspace) error is returned in any case: if the workspace is invalid or access denied.

server/api/api.go
```diff
@@ -740,10 +878,28 @@ func (a *API) handleGetWorkspace(w http.ResponseWriter, r *http.Request) {
...
+
+		ctx := r.Context()
+		session := ctx.Value("session").(*model.Session)
+		if !a.WorkspaceAuthenticator.DoesUserHaveWorkspaceAccess(session, workspaceID) {
+			errorResponse(w, http.StatusUnauthorized, "", nil)
+			return
+		}
+		workspace = &model.Workspace{
+			ID: workspaceID,
+		}
```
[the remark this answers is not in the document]
chenilim: Currently yes. This reuses the existing workspace model, which was used to return the signup token, not used in mm auth. I envision other properties being added to the workspace in the future.

server/api/auth.go
```diff
@@ -358,9 +358,12 @@ func (a *API) attachSession(handler func(w http.ResponseWriter, r *http.Request)
...
 			session := &model.Session{
-				ID:       "single-user",
-				Token:    token,
-				UserID:   "single-user",
+				ID:     "single-user",
+				Token:  token,
+				UserID: "single-user",
+				Props: map[string]interface{}{
+					"authService": a.authService,
+				},
```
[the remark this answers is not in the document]
chenilim: Expediency / laziness. :) Agree authService should be a field. Not sure about expiry and refreshToken though, as they are service-specific.

server/mmauth/mmauth.go
chenilim: Agree authService should be a field (I'll update the model). @jespino, what do you think about expiry and refreshToken? They are service specific, so maybe should remain as props? Or maybe the prop store was meant for something else?

server/mmoauth/mmoauth.go
chenilim: So, the `oauthtoken` cookie is just used to pass it to the client JS code, which (in app.tsx) copies it into localStorage and deletes it. This is because OAuth does a redirect to the callback, instead of going through a fetch like the login API. Do you, or does anyone else know of a better way to pass the token back to the client code?

server/mmauth/mmauth.go
```diff
@@ -0,0 +1,249 @@
...
+	query := r.URL.Query()
+	redirectUrl := query.Get("r")
+
+	var expiration = time.Now().Add(365 * 24 * time.Hour)
+
+	redirectCookie := http.Cookie{Name: "oauthredirect", Path: "/", Value: redirectUrl, Expires: expiration, HttpOnly: true}
+	http.SetCookie(w, &redirectCookie)
+
+	oauthState := generateSecurityToken()
+	stateCookie := http.Cookie{Name: "oauthstate", Path: "/", Value: oauthState, Expires: expiration, HttpOnly: true}
```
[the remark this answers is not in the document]
chenilim: Good catch. I'll add a todo to add a "useSecureCookie" config setting, and also add a step to the install docs to enable it after setting up TLS.

server/mmoauth/mmoauth.go
jwilander: Got it. Missed that. One way to do it is to have the redirect go to a webapp page that makes the API request to the oauth callback endpoint. That way you could still return a JSON as a response for the webapp to handle the token from like you do for regular login. Other option would be to have regular login use a cookie too. WDYT?

server/api/auth.go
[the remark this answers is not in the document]
chenilim: Updated this, moved AuthService to a field of Session.

server/mmauth/mmauth.go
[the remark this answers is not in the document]
chenilim: Added a SecureCookie config setting.

server/mmoauth/mmoauth.go
chenilim: Thanks for the idea! I realized the callback page can just render JS to set the token and redirect, so reimplemented it that way and removed the unnecessary 'oauthtoken' cookie.

webapp/src/app.tsx
chenilim: With the sessionID set by rendered JS, this code is now gone, and it's much cleaner now.

server/mmauth/mmauth.go
jupenur: This is probably fine for MM auth but for generic OAuth I'd definitely create a new token.

server/mmauth/mmauth.go
```diff
@@ -0,0 +1,249 @@
...
+func (a *MMAuth) handleOAuthLogin(w http.ResponseWriter, r *http.Request) {
+	query := r.URL.Query()
+	redirectUrl := query.Get("r")
+
+	var expiration = time.Now().Add(365 * 24 * time.Hour)
+
+	redirectCookie := http.Cookie{Name: "oauthredirect", Path: "/", Value: redirectUrl, Expires: expiration, HttpOnly: true}
+	http.SetCookie(w, &redirectCookie)
+
+	oauthState := generateSecurityToken()
```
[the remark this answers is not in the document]
jupenur: Not entirely sure what this is referring to. The state param & cookie should prevent CSRF independently of any other session handling, so looks good to me.

server/mmauth/mmauth.go
jwilander: Chatted with Chen in a DM about it. Leaving for now, may revisit in the future.

server-config.json
```diff
@@ -5,6 +5,7 @@
 	"dbconfig": "./focalboard.db",
 	"postgres_dbconfig": "dbname=focalboard sslmode=disable",
 	"useSSL": false,
+    "secureCookie": false,
```
jwilander, line 8: This is probably ok for now but in the future we could automatically set the secure cookie attribute based on HTTP headers like we do in the MM server https://github.com/mattermost/mattermost-server/blob/9cc5089af8b7382c0170417bd2c1808662000573/app/login.go#L243

webapp/src/utils.ts
```diff
@@ -43,6 +43,14 @@ class Utils {
         return String(text).replace(/&amp;/g, '&').replace(/&lt;/g, '<').replace(/&gt;/g, '>').replace(/&quot;/g, '"')
     }
 
+    static getCookieValue(name: string): string {
+        return document.cookie.match('(^|;)\\s*' + name + '\\s*=\\s*([^;]+)')?.pop() || ''
+    }
+
+    static deleteCookie(name: string): void {
+        document.cookie = `${name}=; expires=Thu, 01 Jan 1970 00:00:00 UTC; Max-Age=0; path=/`
+    }
```
jwilander, line 52: These are unused now, right? Can probably delete them if so

jwilander (approved):
Looks good 👍 there's room to improve (as always) but this is a great starting point. Left a couple non-blocking comments

webapp/src/utils.ts
chenilim: Good catch. Deleted.

chenilim:
Thanks for the reviews @jwilander and @jupenur! I've updated the code and refactored this PR into separate checkins. The workspaceID is [checked in here](https://github.com/mattermost/focalboard/commit/725971784b064803850d6e94777cab13788d4b7f), and I'll update the auth separately. Closing this PR.

Issue #174, opened 2021-03-22
Show confirmation that export is completed


chenilim:
**Summary:**
It's not entirely obvious that export completed

**Steps to reproduce the behavior:**
1. Open a board
2. Click on Settings and choose `export archive`
3. File is downloaded, but is not completely obvious (may differ by browser)

**Expected behavior:**
Show a pop-up dialog that export is competed

**Platform:**
 - Chrome 89 on macOS 10.15.7

**Additional context:**
Similar for export CSV

Victor239:
Tested on Kubuntu 20.10 with Linux desktop app, no notification that export was successful.

BharatKalluri:
I can pick this up, looks like this is a small UI change which I should be able to quickly do.

I think there should be "Export complete!" flash message on successfull export or "Export failed!" on a failure.
If this sounds good, I'll make a PR with these changes 😄

chenilim:
Hi @BharatKalluri, yes thanks for offering to contribute a PR! That sounds great. Agree that a temporary message would work best. You can use the [FlashMessages component](https://github.com/mattermost/focalboard/blob/main/webapp/src/components/flashMessages.tsx) to do that. LMK if you run into any issues or have any questions.

jespino:
This is done in the PR #178. Thanks @BharatKalluri!

PR #178, opened 2021-03-24
[GH-174] Added message on export

BharatKalluri:
#### Summary

A small fix to add a flash message for both board export success and failure

#### Ticket Link

https://github.com/mattermost/focalboard/issues/174


webapp/src/components/viewHeader.tsx
```diff
@@ -68,6 +69,23 @@ class ViewHeader extends React.Component<Props, State> {
         }
     }
 
+    onExportCsvTrigger(boardTree: BoardTree, intl: IntlShape) {
```
jespino, line 72: the `intl` object is avaiable in `this.props.intl`, I think is better to use it directly instead of pass this as a parmeter.

jespino (changes requested):
A tiny detail, but other than that looks good to me. Thanks @BharatKalluri! :tada:

webapp/src/components/viewHeader.tsx
BharatKalluri: ah, did not notice. Thanks for the tip!
Made changes https://github.com/mattermost/focalboard/pull/178/commits/a590409ae295cd7f1c0ce309ea37ae1c7cca5176 😄

jespino (approved):
Looks good to me. Thanks @BharatKalluri! :confetti_ball:

PR #213, opened 2021-03-29
Store files per workspace and root block

chenilim:
#### Summary
Changed files API to store and retrieve files per workspace and rootID. This is to allow permissions later. A user has access to a file if and only if one of the following is true:
1. Logged in with a valid session (and has access to the workspace, if workspaces are enabled)
2. Pass a valid read_token in the querystring, that is valid for the specified root block (sharing read-only)

This is a breaking change on the previous file storage location.

### DIFF
--- a/server/app/files.go
+++ b/server/app/files.go
@@ -18,7 +18,7 @@ func (a *App) SaveFile(reader io.Reader, workspaceID, rootID, filename string) (
 	}
 
 	createdFilename := fmt.Sprintf(`%s%s`, utils.CreateGUID(), fileExtension)
-	filePath := fmt.Sprintf(`%s/%s/%s`, workspaceID, rootID, createdFilename)
+	filePath := filepath.Join(workspaceID, rootID, createdFilename)
 
 	_, appErr := a.filesBackend.WriteFile(reader, filePath)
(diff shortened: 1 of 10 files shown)


server/app/files.go
```diff
@@ -10,25 +10,27 @@ import (
...
-func (a *App) SaveFile(reader io.Reader, filename string) (string, error) {
+func (a *App) SaveFile(reader io.Reader, workspaceID, rootID, filename string) (string, error) {
 	// NOTE: File extension includes the dot
 	fileExtension := strings.ToLower(filepath.Ext(filename))
 	if fileExtension == ".jpeg" {
 		fileExtension = ".jpg"
 	}
 
 	createdFilename := fmt.Sprintf(`%s%s`, utils.CreateGUID(), fileExtension)
+	filePath := fmt.Sprintf(`%s/%s/%s`, workspaceID, rootID, createdFilename)
```
jespino, line 21: nit: it should be `filepath.Join`

jespino (approved):
LGTM. Just a tiny detail about path building.

server/app/files.go
chenilim: Thanks! Fixed.

PR #216, opened 2021-03-30
Gallery view

jespino:


webapp/src/components/sidebar/sidebarBoardItem.tsx
```diff
@@ -35,6 +36,7 @@ const SidebarBoardItem = React.memo((props: Props) => {
         switch (viewType) {
         case 'board': return <BoardIcon/>
         case 'table': return <TableIcon/>
+        case 'gallery': return <ImageIcon/>
```
chenilim, line 39: We can fix this later, but let's add a new `GalleryIcon` class, copied from ImageIcon, then assign a new issue to Assad to create a new icon.

webapp/src/components/gallery/galleryCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const {card} = props
+    const [cardTree, setCardTree] = useState<CardTree>()
+    useCardListener(
+        card.id,
+        async (blocks) => {
+            const newCardTree = cardTree ? MutableCardTree.incrementalUpdate(cardTree, blocks) : await MutableCardTree.sync(card.id)
+            setCardTree(newCardTree)
+        },
+        async () => {
+            const newCardTree = await MutableCardTree.sync(card.id)
```
chenilim, line 32: This is fine for now, but some thoughts on how we could reduce the number of API calls in the future:
* Let the Gallery own the state
* Initially we could modify / extend BoardTree to call getSubTree with 3 levels
* I envision adding a `getTree` API that returns all nodes by rootID
* And updating the websocket to notify changes to trees by rootID as well
* Then state can be managed by a `fullBoardTree` view model
* In the future-future, we can look at adding filters to the query (e.g. to limit types of nodes), or build something closer to GraphQL.

Just initial thoughts.

chenilim (approved):
I think folks will be very excited to see this!

webapp/src/components/sidebar/sidebarBoardItem.tsx
jespino: Done

PR #217, opened 2021-03-30
Global template support

chenilim:
#### Summary
* Allows readonly board templates to be used in workspaces
* Board templates are synched from global / root workspace (workspaceID = 0) by sidebar
* I'd like to refactor `duplicateFromRootBoard` and the mutator in general, but want to get this in first


webapp/src/components/sidebar/sidebarAddBoardMenu.tsx
```diff
@@ -115,6 +101,51 @@ class SidebarAddBoardMenu extends React.Component<Props> {
         )
     }
 
+    private renderBoardTemplate(boardTemplate: Board, isGlobal = false): JSX.Element {
```
jespino, line 104: We can do it later, but probably this is a good candidate to create another small component.

jespino (approved):
LGTM

PR #218, opened 2021-03-30
Workspace title

chenilim:
#### Summary
* Show a hint when there is no active view
* Show the workspace name in the sidebar, and in the hint if available

Notes:
* Updated auth API to return workspace info
* Just title for now, but we can extend this to more info later

### DIFF
--- a/webapp/src/components/workspace.tsx
+++ b/webapp/src/components/workspace.tsx
@@ -9,6 +9,7 @@ import {BoardTree} from '../viewModel/boardTree'
 import {WorkspaceTree} from '../viewModel/workspaceTree'
 
 import CenterPanel from './centerPanel'
+import EmptyCenterPanel from './emptyCenterPanel'
 import Sidebar from './sidebar/sidebar'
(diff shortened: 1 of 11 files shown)


webapp/src/components/workspace.tsx
```diff
@@ -21,16 +24,36 @@ type Props = {
     readonly: boolean
 }
 
-const Workspace = React.memo((props: Props) => {
-    const {boardTree, setSearchText, workspaceTree, showBoard, showView, setLanguage} = props
+function centerContent(props: Props) {
```
jespino, line 27: In fact, this is a functional component, we should make it uppercase and use as a component bellow.

jespino (approved):
A tiny detail, but otherwise looks good to me.

jespino:
I guess this needs changes in the enterprise repo too.

chenilim:
Thanks Jesus! The enterprise changes are already in. I'll check this in and clean up the functional component after.

PR #221, opened 2021-03-31
Internationalized hardcoded strings in the frontend codebase

BharatKalluri:
#### Summary

Replaces hardcoded strings with internationlized strings

If anyone happens to find more strings which are hardcoded, just add a comment pointing me to that file and I'll intl them as well 😄 

#### Ticket Link

https://github.com/mattermost/focalboard/issues/33

PS: I think it is a good idea to just add a pre-push hook which runs `npm run i18n-extract`. What do you guys think?


chenilim:
Thanks for the PR @BharatKalluri! Much appreciated!

Thanks also for the suggestion on the pre-push hook. I'll check with the rest of the team.

webapp/src/widgets/propertyMenu.tsx
```diff
@@ -49,6 +49,11 @@ const PropertyMenu = React.memo((props: Props) => {
     const nameTextbox = useRef<HTMLInputElement>(null)
     const [name, setName] = useState(props.propertyName)
 
+    const deleteText = intl.formatMessage({
+        id: 'PropertyType.Delete',
```
chenilim, line 53: This should probably be `PropertyMenu.Delete`. I'll update it.

chenilim (approved):
LGTM

PR #226, opened 2021-04-01
linux: find server executable relative to app

ziprandom:
This PR changes the search path of the server binary in the linux app  to be not `./` (the working directory) but the same path as the apps executable. 

This is part of an effort to better separate program and configuration.

### DIFF
--- a/linux/main.go
+++ b/linux/main.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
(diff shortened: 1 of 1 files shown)


chenilim:
Thanks for the PR @ziprandom! Just a heads up that we have a long weekend coming up, but someone should review this in the next week or so. It LGTM, but I'd like to get some extra eyes on it.

jespino (approved):
LGTM

PR #227, opened 2021-04-01
server: additionally read config settings from env

ziprandom:
The current setup expects `focalboard-app`, `focalboard-server`, `config.json` and the `pack` directory to reside in the same location. 

This PR allows for at least the configuration to be overridden via the execution environment which makes it easier to use the app in a environment where the programs assets and executables reside in a global, not user writeable location as is the case with nix(os) and many other scenarios. 

ideally the app would take a `--config-file` parameter but this is not straight forward with the viper framework and therefore I present this as the smallest and least invasive change to solve the problem.


zacheryph:
👍

chenilim:
Thanks for the PR @ziprandom! Just a heads up that we have a long weekend coming up, but someone should review this in the next week or so. It LGTM, but I'd like to get some extra eyes on it.

jespino (approved):
Great, Thanks @ziprandom! :tada:

PR #84, opened 2021-03-18
[GH-40] Add property type email

renjithgr:
#### Summary

This PR adds the option to choose email as a field type.

image

#### Ticket Link
#40 

Most of the components are not covered by unit tests, including the ones I changed. So I was not sure what strategy the creators have in mind for unit testing components. Please advise. I would be happy to add unit tests.

Thank you!

Issue #37, opened 2021-01-26
Property type: URL


chenilim:
Add a URL property type

BharatKalluri:
Hi,
Made a PR to this one over at https://github.com/mattermost/focalboard/pull/228 since no one is assigned and it looked like a small change 😄

jespino:
Fixed by @BharatKalluri in the PR #228

PR #228, opened 2021-04-01
Added URL as a property on the frontend

BharatKalluri:
Added URL as a proprty option on the frontend

#### Summary

Takes https://github.com/mattermost/focalboard/pull/84 as a reference 😅 and adds a URL prop on the frontend. Seems like its already a part of `propertyValue`

#### Ticket Link

https://github.com/mattermost/focalboard/issues/37

I had one small question, eventually will focalboard have validations on the text in the propertyValue?


chenilim:
Thanks for the PR @BharatKalluri! Let's hold off on this one for now.

Currently the 'text', 'number' and 'email' properties all map to an `Editable` text field. They should instead map to distinct controls with validation, etc. For instance, we should not be able to enter letters into a number field. Emails and URLs should be validated.

jespino:
@BharatKalluri I added a property validator mechanism in master, do you mind to merge master in your branch and add the `url` validator there? With that, I think this PR is ready to go.

BharatKalluri:
Sure, I'll be adding the URL regex from https://www.urlregex.com/

The regex for a URL is given to be 
`/((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/`

jespino (approved):
Now looks good to me. Thanks! :tada:

PR #235, opened 2021-04-06
Clean up temp files in Dockerfile to save image size

PeterDaveHello:
#### Summary

Clean up temporary/cache files in Dockerfile to save image size

Result:

```text
REPOSITORY            TAG             IMAGE ID            CREATED              SIZE
after                 latest          521681716ad5        About a minute ago   107MB
before                latest          c192efaae426        3 minutes ago        147MB
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,11 +2,13 @@ FROM ubuntu:20.04
 
 # Make sure that the underlying container is patched to the latest versions
 RUN apt-get update && \
-    apt-get install -y gzip tar wget
+    apt-get install -y gzip tar wget && \
+    rm -rf /var/lib/apt/lists/*
 
 # Now install Focalboard as a seperate layer
 RUN wget https://github.com/mattermost/focalboard/releases/download/v0.6.1/focalboard-server-linux-amd64.tar.gz && \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -2,11 +2,13 @@ FROM ubuntu:20.04
 
 # Make sure that the underlying container is patched to the latest versions
 RUN apt-get update && \
-    apt-get install -y gzip tar wget
+    apt-get install -y gzip tar wget && \
+    rm -rf /var/lib/apt/lists/*
```
hirusi, line 6: _Might_ be worth adding a note that any `apt-get install`s after this Docker step will require a new `apt-get update` before being able to install a package?

hirusi:
Also worth noting that #168 has been approved already, so we might need to look at clean-ups specifically for Alpine in _that_ Docker image.

docker/Dockerfile
PeterDaveHello: As there is no any other apt-get install yet, I prefer not to have so many comments in Dockerfile, and once another apt-get install was added, contributor will find out, right? :)

jespino:
This is now fixed by the #168 PR

PR #242, opened 2021-04-09
Example moving mutations aside the component

jespino:


chenilim (approved):
While I don't see any issues with this, I'd think of those methods more as "actions" which call "mutations" in the mutator. I'm 0/5, but is it generally a good practice to put action methods in a separate file from the component? So the component only has "render"?

It might help clean up large components, but adds extra complexity for small ones. This particular component seems to be on the mid-to-small side.

Anyway, if you feel this is important, we can make it a style guideline. Suggest maybe calling it `kanban.actions.tsx` to differentiate from the mutator?

I think a bigger issue is splitting the `mutator.ts` up, perhaps by type, e.g. `boardMutator`, `cardMutator`, etc.

jespino:
I think this is going to be discard with the new redux approach.

Issue #41, opened 2021-01-26
Property type: Phone number


chenilim:
Add a phone number property type

PR #286, opened 2021-04-18
Added phone number field

BharatKalluri:
#### Summary

Adds a phone number field and corresponding validation. Uses phone number regex from https://ihateregex.io/expr/phone/.
Tested with Indian and US phone numbers.

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/41

### DIFF
--- a/webapp/src/components/propertyValueElement.tsx
+++ b/webapp/src/components/propertyValueElement.tsx
@@ -30,7 +30,6 @@ const PropertyValueElement = (props:Props): JSX.Element => {
     const finalDisplayValue = displayValue || emptyDisplayValue
 
     const validateProp = (propType: string, val: string): boolean => {
-        const lowerCaseVal = val.toLowerCase()
         if (val === '') {
             return true
         }
@@ -39,15 +38,11 @@ const PropertyValueElement = (props:Props): JSX.Element => {
             return !isNaN(parseInt(val, 10))
         case 'email': {
             const emailRegexp = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
-            return emailRegexp.test(lowerCaseVal)
+            return emailRegexp.test(val)
         }
         case 'url': {
             const urlRegexp = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)/
-            return urlRegexp.test(lowerCaseVal)
-        }
-        case 'phone': {
-            const phoneRegexp = /[+]?[(]?[0-9]{3}[)]?[-\s.]?[0-9]{3}[-\s.]?[0-9]{4,6}/
-            return phoneRegexp.test(lowerCaseVal)
+            return urlRegexp.test(val)
         }
         case 'text':
             return true
--- a/webapp/src/components/propertyValueElement.tsx
+++ b/webapp/src/components/propertyValueElement.tsx
@@ -46,6 +46,8 @@ const PropertyValueElement = (props:Props): JSX.Element => {
         }
         case 'text':
             return true
+        case 'phone':
+            return true
         default:
             return false
         }
--- a/webapp/src/components/propertyValueElement.tsx
+++ b/webapp/src/components/propertyValueElement.tsx
@@ -3,7 +3,7 @@
 
 import React, {useState} from 'react'
 
-import {IPropertyOption, IPropertyTemplate} from '../blocks/board'
+import {IPropertyOption, IPropertyTemplate, PropertyType} from '../blocks/board'
 import {Card} from '../blocks/card'
 import mutator from '../mutator'
 import {OctoUtils} from '../octoUtils'
@@ -30,6 +30,7 @@ const PropertyValueElement = (props:Props): JSX.Element => {
     const finalDisplayValue = displayValue || emptyDisplayValue
 
     const validateProp = (propType: string, val: string): boolean => {
+        const lowerCaseVal = val.toLowerCase()
         if (val === '') {
             return true
         }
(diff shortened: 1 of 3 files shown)


webapp/src/components/propertyValueElement.tsx
```diff
@@ -38,11 +39,15 @@ const PropertyValueElement = (props:Props): JSX.Element => {
...
-            return emailRegexp.test(val.toLowerCase())
+            return emailRegexp.test(lowerCaseVal)
         }
         case 'url': {
             const urlRegexp = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)/
-            return urlRegexp.test(val.toLowerCase())
+            return urlRegexp.test(lowerCaseVal)
+        }
+        case 'phone': {
+            const phoneRegexp = /[+]?[(]?[0-9]{3}[)]?[-\s.]?[0-9]{3}[-\s.]?[0-9]{4,6}/
```
jespino, line 49: @chenilim how do you feel about this? I don't know how international is the way of writing phone numbers, looks more or less accurate, at least for Spain, and probably for US, but I dont't think is going to work always. For example "+34 666666666" doesn't work, and probably is a very common way of writting it (That should be +034 to make it work). Also, in Spain if you write a phone number with dots in it, probably you would use something like "666.666.666" or even "666.66.66.66" but never something like "666.666666". And that is only for Spain, I guess each country have its own slightly different ways of writing it.

jespino: There are other kind of peculiarities to take into consideration, for example "extensions".

webapp/src/components/propertyValueElement.tsx
```diff
@@ -38,11 +39,15 @@ const PropertyValueElement = (props:Props): JSX.Element => {
             return !isNaN(parseInt(val, 10))
         case 'email': {
             const emailRegexp = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
-            return emailRegexp.test(val.toLowerCase())
+            return emailRegexp.test(lowerCaseVal)
```
jespino, line 42: there is no need of making to lowercase for email or url, the regular expresions are allowing uppercase/lowercase characters anyway.

webapp/src/components/propertyValueElement.tsx
Johennes: Sorry for shooting in from the side here but I, too, made the mistake of trying to validate phone numbers with a regex once. Unfortunately, phone numbers are poorly standardized, especially when you factor in different formatting options for the same number. If you target a global audience and if the input has to be validated, I can recommend google/libphonenumber (or one of the forks that exist for various languages).

jespino: Yes, I totally agree with @Johennes here, this is a hard problem and we should use something like the libphonenumber or just not validate it at all. The problem with libphonenumber is really heavyweight, around 420K because requires google clousere compiler, and there is tinier fork but still 110K. I'm not 100% sure what is the best option.

jespino: Yes, probably this is the most sensible option if we want to have phone number validations: https://github.com/catamphetamine/libphonenumber-js

BharatKalluri: Yeah, agreed @jespino , Thanks for the links!
We can either use libphonenumber-js or skip validation for phone number as you were suggesting. Let me know what you think. And I'll make the corresponding changes.

webapp/src/components/propertyValueElement.tsx
BharatKalluri: Makes sense. I'll make changes. Thanks for the review!

webapp/src/components/propertyValueElement.tsx
jespino: @chenilim any strong opinion here? I'm more inclined to not-validate it for now, but if we want validation, looks like the more sane approach is the libphonenumber-js option.

chenilim: Sorry @jespino, I missed your earlier comment. Looked more into this. We should **remove validation** from the phone number. The main value of a phone number field is to (later) add a "call this number" button (makes most sense from mobile). However, the "number" may not actually be a number. E.g. it could be a Facetime contact or something similar.

So, no need to pull in another library for now. P4 (maybe in the far future), we could allow users to pick from a set of regex templates, and maybe show a warning rather than hard validation.

BharatKalluri: Removed phone number validation @ https://github.com/mattermost/focalboard/pull/286/commits/d87b90e6b932e2cf7693871d26e9bad6a3efda42

webapp/src/components/propertyValueElement.tsx
BharatKalluri: also removed `lowercase` for email and url @ https://github.com/mattermost/focalboard/pull/286/commits/d87b90e6b932e2cf7693871d26e9bad6a3efda42

BharatKalluri: also removed `lowercase` for email and url @ https://github.com/mattermost/focalboard/pull/286/commits/d87b90e6b932e2cf7693871d26e9bad6a3efda42

webapp/src/components/propertyValueElement.tsx
```diff
@@ -38,11 +38,11 @@ const PropertyValueElement = (props:Props): JSX.Element => {
...
             const emailRegexp = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
-            return emailRegexp.test(val.toLowerCase())
+            return emailRegexp.test(val)
         }
         case 'url': {
             const urlRegexp = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)/
-            return urlRegexp.test(val.toLowerCase())
+            return urlRegexp.test(val)
         }
         case 'text':
```
jespino, line 47: Because there is no `phone` check at all, it is always returning false in the validation. You have to add the `case 'phone':` here returning true, like the `case 'text':`.

jespino (changes requested):
An small change need, otherwise looks good to me. Thanks @BharatKalluri! :tada:

jespino (approved):
Now looks good to me. Thanks @BharatKalluri!

webapp/src/components/propertyValueElement.tsx
BharatKalluri: Sorry, missed it 😓 
Thanks for the review!

PR #293, opened 2021-04-18
[GH-278] Prevent dragging of gallery cards in read-only mode

### LINKED ISSUE #278: Bug: Should not be able to drag items in shared read-only gallery view
Importance: P3

1. Open or create a gallery view with multiple cards
2. Share the board
3. Open an incognito window and open the readonly shared board link
4. Try to drag a card in the gallery view
5. Able to drag, but dropping has no effect

**Expected behavior:**
Should not be able to start the drag on a readonly view

**Platform:**
 - Chrome 89 on macOS 10.15.7

**Additional context:**
There are likely other related UI issues on readonly shared boards.

Johennes:
#### Summary

This prevents cards on gallery boards from being draggable when the board is read-only.

Comparing with [KanbanCard](https://github.com/mattermost/focalboard/blob/main/webapp/src/components/kanban/kanbanCard.tsx#L44), I noticed that the `draggable` property is also set explicitly there. In my testing for gallery cards this did not seem necessary though.

#### Ticket Link

Fixes #278

PS: When signing the CLA, the form said you'll send me something as a thank you if my contribution is accepted. Please don't mail me anything to Germany for a one-line fix. I would feel embarrassed. 😅

### DIFF
--- a/webapp/src/components/gallery/galleryCard.tsx
+++ b/webapp/src/components/gallery/galleryCard.tsx
@@ -37,7 +37,7 @@ type Props = {
 
 const GalleryCard = React.memo((props: Props) => {
     const {cardTree} = props
-    const [isDragging, isOver, cardRef] = useSortable('card', cardTree.card, props.isManualSort, props.onDrop)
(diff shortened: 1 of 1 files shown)


jespino (approved):
Looks good to me. Thanks @Johennes! :tada:

PR #294, opened 2021-04-18
[GH-277] Hide drag grip on content blocks in read-only mode

### LINKED ISSUE #277: Bug: Content drag grip should not appear on shared read-only boards
Importance: P3

1. Open or create a card
2. Add 2 or more content blocks
2. Share the board
3. Open an incognito window and open the readonly shared board link
4. Open the card
5. Hovering over the content shows the rearrange grip

**Expected behavior:**
Users should not see the drag grip if a board is read-only.

**Platform:**
 - Chrome 89 on macOS 10.15.7

**Additional context:**
Dragging and dropping the content doesn't actually rearrange it if readonly, but the user should not see the grips.

Johennes:
#### Summary

This wraps the `.dnd-handle` drag grip with `!props.readonly &&` so that it doesn't show when viewing card details in read-only mode.

Since all children of the parent `.octo-block-margin` div are excluded in read-only mode now, we could also move the read-only check here. Other than preventing the empty div, that didn't seem to provide any significant advantage though so I left it like this. Happy to change this if you disagree though. 🙂

#### Ticket Link

Fixes #277

### DIFF
--- a/webapp/src/components/contentBlock.tsx
+++ b/webapp/src/components/contentBlock.tsx
@@ -106,12 +106,14 @@ const ContentBlock = React.memo((props: Props): JSX.Element => {
                         </Menu>
                     </MenuWrapper>
                 }
-                <div
-                    ref={gripRef}
-                    className='dnd-handle'
-                >
-                    <GripIcon/>
-                </div>
+                {!props.readonly &&
+                    <div
+                        ref={gripRef}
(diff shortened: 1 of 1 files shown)


jespino (approved):
Looks good to me. Thanks @Johennes! :confetti_ball:

PR #301, opened 2021-04-20
Adding mysql support to the database

jespino:


server/services/store/sqlstore/blocks.go
```diff
@@ -14,32 +15,23 @@ import (
...
 			"id",
 			"parent_id",
 			"root_id",
 			"modified_by",
-			"schema",
+			"`schema`",
 			"type",
 			"title",
-			"COALESCE(\"fields\", '{}')",
+			"COALESCE(`fields`, '{}')",
```
chenilim, line 28: Do you know for sure that backticks also work on SQLite and Postgres, or did you test on those? I just don't know if they do or not, and got burnt by SQL quoting issues on previous commits. If you know this works, that's good with me.

chenilim: Hmmm... a quick search seems to indicate [backticks are non-standard](https://wiki.postgresql.org/wiki/Things_to_find_out_about_when_moving_from_MySQL_to_PostgreSQL)? We should test on all 3 DBs to be sure.

server/services/store/sqlstore/blocks.go
```diff
@@ -337,7 +326,27 @@ func (s *SQLStore) InsertBlock(c store.Container, block model.Block) error {
...
-	_, err = query.Exec()
+	// TODO: migrate this delete/insert to an upsert
+	deleteQuery := s.getQueryBuilder().Delete(s.tablePrefix + "blocks").Where(sq.Eq{"id": block.ID})
+	_, err = sq.ExecContextWith(ctx, tx, deleteQuery)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	_, err = sq.ExecContextWith(ctx, tx, query.Into(s.tablePrefix+"blocks"))
```
chenilim, line 337: Minor: spacing around + operator. Not sure if Go lint would catch this anyway.

server/services/store/sqlstore/migrations/migrations_files/000009_blocks_history.up.sql
```diff
@@ -0,0 +1,30 @@
...
+	title TEXT,
+	fields {{if .postgres}}JSON{{else}}TEXT{{end}},
+	create_at BIGINT,
+	update_at BIGINT,
+	delete_at BIGINT,
+	PRIMARY KEY (id)
+){{if .mysql}}CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci{{end}};
+
+{{if .mysql}}
+INSERT IGNORE INTO {{.prefix}}blocks (SELECT * FROM {{.prefix}}blocks_history ORDER BY insert_at DESC);
```
chenilim, line 22: Not sure if this is actionable here, but it looks like the .prefix setting should be constant for a given instance, otherwise migrations would break in unexpected ways. Maybe more a note for the future that .prefix should ideally be a hardcoded build setting, or something similar.

chenilim (commented):
Looks good, except for my concern about the backtick cross-db support. I'll defer to you if you know this works, but would feel better if we could smoke test this on all 3 DBs. :)

server/services/store/sqlstore/blocks.go
jespino: Yes, clearly the backtick is a problem specially for postgres. I'll fix that, anyway yes, we need to test all the databases before it.

server/services/store/sqlstore/blocks.go
jespino: Yes, weird, I would expect that to be autoformatted.

server/services/store/sqlstore/migrations/migrations_files/000009_blocks_history.up.sql
jespino: maybe yes, maybe not, the behavior if you change the prefix, because the migrations table is prefixed to, is treat that situation as a brand new database, the prefix doesn't change on runtime, only changes on restart, so... shouldn't be a big deal.

server/services/store/sqlstore/blocks.go
jespino: Now is fixed :)

server/services/store/sqlstore/blocks.go
jespino: Looks like the auto formatting of go is removing the spaces, so I think we should ignore this case and keep it as is now.

chenilim (approved):
Changes look good!

server/services/store/sqlstore/blocks.go
```diff
@@ -60,16 +52,17 @@ func (s *SQLStore) GetBlocksWithParent(c store.Container, parentID string) ([]mo
 			"parent_id",
 			"root_id",
 			"modified_by",
-			"schema",
+			s.escapeField("schema"),
```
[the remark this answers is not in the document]
jespino: Yes, exactly, is only needed to escape keywords and things like that.

PR #310, opened 2021-04-21
[GH-190] Add global option to disable randomized card icons

### LINKED ISSUE #190: Feature Idea: Add option to turn off auto-random icon
**Summary:**
Creating a new card automatically sets a random icon. This is by design, as it makes cards immediately distinguishable. However, sometimes we want to turn that off

Importance: Medium

P1: Add an option to the settings menu to turn this off. Store in localStorage per-client for now.
Bonus points: Create a `userSettings.ts` class to manage other settings like this
P3: (Future) consider how to store this and similar settings per server / instance / board

Johennes:
#### Summary

This does the minimum work for #190 by adding a global settings option to disable random icons on newly created cards.

Screenshot 2021-04-21 at 16 41 00

#### Ticket Link

Partially fixes #190

### DIFF
--- a/webapp/src/components/sidebar/sidebarSettingsMenu.tsx
+++ b/webapp/src/components/sidebar/sidebarSettingsMenu.tsx
@@ -1,6 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-import React, {useContext} from 'react'
+import React, {useContext, useState} from 'react'
 import {FormattedMessage, injectIntl, IntlShape} from 'react-intl'
 
 import {Archiver} from '../../archiver'
@@ -8,6 +8,7 @@ import {darkTheme, defaultTheme, lightTheme, setTheme, Theme} from '../../theme'
 import Menu from '../../widgets/menu'
 import MenuWrapper from '../../widgets/menuWrapper'
 import {SetLanguageContext} from '../../setLanguageContext'
+import {UserSettings} from '../../userSettings'
 
 import './sidebarSettingsMenu.scss'
 
@@ -24,6 +25,12 @@ const SidebarSettingsMenu = React.memo((props: Props) => {
         const consolidatedTheme = setTheme(theme)
         const whiteLogo = (consolidatedTheme.sidebarWhiteLogo === 'true')
         props.setWhiteLogo(whiteLogo)
+    }
+
+    const [randomIcons, setRandomIcons] = useState(UserSettings.prefillRandomIcons)
+    const toggleRandomIcons = () => {
+        UserSettings.prefillRandomIcons = !UserSettings.prefillRandomIcons
+        setRandomIcons(!randomIcons)
     }
 
     return (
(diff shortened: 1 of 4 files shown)


webapp/src/components/sidebar/sidebarSettingsMenu.tsx
```diff
@@ -26,6 +27,12 @@ const SidebarSettingsMenu = React.memo((props: Props) => {
         props.setWhiteLogo(whiteLogo)
     }
 
+    const [randomIcons, setRandomIcons] = useState(UserSettings.prefillRandomIcons)
+    const toggleRandomIcons = () => {
+        UserSettings.prefillRandomIcons = !UserSettings.prefillRandomIcons
+        setRandomIcons(!randomIcons)
+    }
```
Johennes, line 34: This is a little odd because the setting state is maintained in two separate spots (`UserSettings` and `randomIcons`) but I couldn't find a better way to cause the component to re-render when changing the setting. Would be happy to hear recommendations on this.

Johennes:
Eliminated the enum in favor of string literals. I think the linting error was a false positive but I couldn't find out how to silence it _and_ keep the enum.

jespino:
I like this idea, but I don't know if this should be per user, or per board, and probably we should store that information on the server. What do you think @chenilim?

chenilim:
I think this is fine as a per-user setting for now. For Personal Desktop it makes no difference, and we haven't implemented any other configurable per-server (actually per-workspace) settings yet.

In the long run, I can imagine multiple layers, i.e.:
* Server admins set the global defaults
* Workspace admins set workspace defaults (maybe for a subset)
* Users can override a subset for themselves

signalwerk:
@Johennes Thank you for your PR this is wonderful! I completely understand your wish and I have the same. But I think all our wishes are different. I'm not certain if this is the right way to cover all the wishes. 

Wouldn't it be more helpful to redefine/edit the default-template for a new card per board? 

## Pro
* Each board can have it's individual default-template
* The default-template can not only control the Icon-Settings

## Con
* No implementation yet (your hard work would be lost)
* If you want to change the behavior for all new boards you have to edit/modify a board-template

Johennes:
@signalwerk being able to change the default card template for a board would be great, agreed. For my personal use case that would be slightly less convenient than a single global "no random icons" setting. But I could live with that since I only have a small number of boards.

@chenilim / @jespino what do you think? Should we rather make the default template per board editable?

For the time being, I converted this PR to a draft because storing the global setting currently doesn't work in the native apps due to #314.

> No implementation yet (your hard work would be lost)

Ah, no worries about that. What I did here so far was pretty basic so I wouldn't have hard feelings if it was discarded for something better. 🙂

chenilim:
Thanks for adding to the discussion. In the long run, I see several buckets of settings / preferences, e.g.:
* Server global
* Board
* User

In general, for non-security related things, the precedence should be bottom up. i.e. a user can override board or server defaults. Auto-emoji is an example of this, because the user can always set a random emoji anyway. The setting just saves them time.

Also, at this point in the project, we should err on the side of speed. It's ok to add desired functionality that can be refined later.

So, I'm happy to merge this change, and keep issue #190 open to add the other levels of settings later.

signalwerk:
@Johennes @chenilim thank you two for your reactions and thoughts. 😌 let's merge it then and think later about next steps.

webapp/src/components/sidebar/sidebarSettingsMenu.tsx
jespino: Yes, probably we can move this into its own hook and make the concept of "userSettings" something that is properly encapsulated. But for now I think is ok as is now.

jespino (approved):
Looks good to me. Thanks @Johennes! :tada:

PR #317, opened 2021-04-22
Add macOS quarantine

Willyfrog:
#### Summary

Let macOS handle quarantine on downloaded files.

I know focalboard still doesn't have files other than images (although non images can be uploaded and they wont display properly). but at some point it will and having the OS handle quarantine files will probably ease security.

#### Ticket Link

No ticket per se, but I recently [did that on the desktop app](https://mattermost.atlassian.net/browse/MM-20677) and thought it would be nice to have it here as well. 

#### Testing

I don't think it can be tested right now, but once you can download a file from Focalboard you can create a sample file out of the terminal app by going to the menu->shell->export settings (export as a `.terminal` file)

### DIFF
--- a/mac/Focalboard/Info.plist
+++ b/mac/Focalboard/Info.plist
@@ -24,18 +24,20 @@
 	<string>public.app-category.productivity</string>
 	<key>LSMinimumSystemVersion</key>
 	<string>$(MACOSX_DEPLOYMENT_TARGET)</string>
-    <key>ITSAppUsesNonExemptEncryption</key>
+	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/>
 	<key>NSAppTransportSecurity</key>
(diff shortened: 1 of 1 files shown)


chenilim (approved):
Thanks for your contribution @Willyfrog! LGTM. Did a quick smoke test, and this didn't appear to affect images, import, or export. xattr shows the com.apple.quarantine flag is applied (and may even have been applied before). In any case, this is a good extra security later.

Willyfrog:
> Thanks for your contribution @Willyfrog! LGTM. Did a quick smoke test, and this didn't appear to affect images, import, or export. xattr shows the com.apple.quarantine flag is applied (and may even have been applied before). In any case, this is a good extra security later.

yes, this mostly affects anything that might carry code (like .terminal files) or applications

PR #323, opened 2021-04-23
Fix readme formatting

signalwerk:
Only some minor formatting issues


chenilim:
Thanks for cleaning this up @signalwerk! Much appreciated!

PR #327, opened 2021-04-24
Add prettier at least for css

signalwerk:
This reverts the deleted `.prettierrc.json` of commit c95ab685e4a77823247886e01fa044223e247a89 and adds code fixes to scss files

I know there are 1000 opinions about code formatting and I saw that prettier was removed at one point (which is somehow standard for many frontend projects).

This commit brings prettier only for .scss files back. Right now there are many formats in the stylesheets (including tab and space indents mixed in the same file) and this should be fixed.


jespino (approved):
I really like this, I going to leave this without merge until we get the @chenilim opinion, because he removed the file in first place. But for me is a :+1: for sure :)

chenilim (approved):
Thanks for the PR @signalwerk! I initially used prettier then removed it when it conflicted with eslint. In this case, it looks like we only use prettier for .scss, which eslint doesn't touch right? If so, that should be fine. It's great to see the .scss files cleaned up!

signalwerk:
@chenilim yes exactly. I tried to keep the changes as sensible as possible with not overruling existing standards. I just applied prettier to SCSS. Furthermore, I also added `npm run fix:scss` for auto-fix in the future. Since you seem to be positive about the change I might also have a look into the eslint ←→ prettier conflict. I would then suggest that in a separate PR.

jespino:
Great, then I going to merge it. Thanks @signalwerk! :confetti_ball:

PR #335, opened 2021-04-26
Changing mattermost-auth method to work based on shared database access

jespino:


config.json
```diff
@@ -11,14 +11,10 @@
...
 	"session_expire_time": 2592000,
 	"session_refresh_time": 18000,
 	"localOnly": false,
 	"enableLocalMode": true,
 	"localModeSocketLocation": "/var/tmp/focalboard_local.socket",
-	"authMode": "native",
-	"mattermostURL": "",
-	"mattermostClientID": "",
-	"mattermostClientSecret": ""
+	"authMode": "mattermost"
```
jespino, line 19: This needs to be "native" by default

mattermost-plugin/README.md
```diff
@@ -0,0 +1,131 @@
+# Plugin Starter Template [![CircleCI branch](https://img.shields.io/circleci/project/github/mattermost/mattermost-plugin-starter-template/master.svg)](https://circleci.com/gh/mattermost/mattermost-plugin-starter-template)
```
jespino, line 1: Probably we should update this README.

mattermost-plugin/plugin.json
```diff
@@ -0,0 +1,26 @@
+{
+    "id": "focalboard",
+    "name": "Focalboard",
+    "description": "This provides focalboard integration with mattermost.",
+    "homepage_url": "https://github.com/mattermost/mattermost-plugin-focalboard",
+    "support_url": "https://github.com/mattermost/mattermost-plugin-focalboard/issues",
+    "release_notes_url": "https://github.com/mattermost/mattermost-plugin-focalboard/releases/tag/v0.1.0",
+    "icon_path": "assets/starter-template-icon.svg",
+    "version": "0.1.0",
```
jespino, line 9: Change the version to 0.2.0

mattermost-plugin/server/imports.go
```diff
@@ -0,0 +1,6 @@
+package main
```
jespino, line 1: Remove this file

server/api/api.go
```diff
@@ -138,14 +133,21 @@ func (a *API) getContainerAllowingReadTokenForBlock(r *http.Request, blockID str
...
+		workspaceID := vars["workspaceID"]
+
 		container := store.Container{
-			WorkspaceID: "0",
+			WorkspaceID: workspaceID,
 		}
 
-		// Has session
-		if session != nil {
+		if workspaceID == "0" {
```
jespino, line 145: verify if this is needed.

server/web/webserver.go
```diff
@@ -94,6 +94,10 @@ func (ws *Server) registerRoutes() {
 // Start runs the web server and start listening for charsetnnections.
 func (ws *Server) Start() {
 	ws.registerRoutes()
+	if ws.port == 0 {
```
jespino, line 97: We can change it to -1

server/ws/websockets.go
```diff
@@ -99,9 +103,11 @@ func (ws *Server) handleWebSocketOnChange(w http.ResponseWriter, r *http.Request
 		client.Close()
 	}()
 
+	userID := r.Header.Get("Mattermost-User-Id")
```
jespino, line 106: Also we need to check if the authentication method is mattermost, if not we can end up with a security hole here.

server/api/api.go
jespino: This is for allowing the users to get the templates from the root workspace.

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -36,13 +36,13 @@ export default class Plugin {
         this.registry = registry;
         this.channelHeaderButtonId = registry.registerChannelHeaderButtonAction(focalboardIcon, () => {
             const currentChannel = store.getState().entities.channels.currentChannelId;
-            window.open(`${window.location.origin}/plugins/focalboard/workspace/${currentChannel}`)
+            window.open(`${window.location.origin}/plugins/focalboard/workspace/${currentChannel}`);
```
jespino, line 39: We dont use `;`

mgdelacroix: I think I added them because of a linter's complain, but will remove them before the merge

Issue #299, opened 2021-04-20
Feature Idea: Manual sorting board views


Johennes:
**Summary:**

Currently the views in a board are sorted alphabetically in the left pane. It would be nice if the views could be sorted manually through drag and drop. Particularly when using emoji as the first character in view names, the alphabetical sorting tends to produce non-intuitive results.

![Screenshot 2021-04-20 at 08 59 12](https://user-images.githubusercontent.com/1137962/115351857-fe72fa80-a1b6-11eb-8d6b-d490481cf2e2.png)

**How important this is to me and why:**

Importance: Low

Low importance because I'm not using terribly many views (yet) so it's a minor inconvenience only.

Use cases:

1. I'm using emojis as the first character in view names which breaks intuitive alphabetical sorting
2. I want certain views to appear at the top / bottom of the list because I use them more / less frequently

chenilim:
Thanks @Johennes! Yes, we do intend to support manual ordering of the sidebar, perhaps with additional organization. If anyone has things they would like to see, or examples of good implementations of this, please add links. Much appreciated!

Johennes:
An easy fix for my particular issue could also be to strip leading emoji from the view names before applying the alphabetical sorting.

chenilim:
That might be a good short-term fix @Johennes. If someone wants to do that first, this ticket is still up for grabs! :)

Johennes:
I am someone so I gave this a try in #341. 😅

chenilim:
@asaadmahmood, assigning this to you to track as part of the sidebar re-design. Thanks.

nqtronix:
I wanted to re-order the entries on the left sidebar, so I tried in order:

1. drag & drop
2. right click
3. 3-dot-menu
4. opening an entry (re-naming an entry requires this)
5. re-naming entries as work around\*
6. reading the help menu (which doesn't mention the order)
7. searching on github issues

IMHO the main problem is that the user (me) may expect that manual sorting *should* work, and may assume he's just too stupid to find it. After all, some common features are not where they usually are (no re-name on right click/ 3-dot-menu).

Any yes, I *did* try rename the entries with a leading number to force sorting, but it didn't work either. After reading this issue I assume not only leading emojis are ignored, but also leading numbers? I've not experience with typescript, so I'd be great if someone can verify and fix this (=re-enable sort by leading numbers).

[...]

PR #341, opened 2021-04-26
[GH-299] Sort views alphabetically after stripping leading emoji

Johennes:
#### Summary

This is a stopgap towards #299 that sorts views in the sidebar alphabetically but only after stripping any leading emoji. This prevents unintuitive results such as putting "📥 Inbox" before "🚀 Active".

*Before*

Screenshot 2021-04-26 at 21 15 36

*After*

Screenshot 2021-04-26 at 21 14 27

Still not sure if maybe this is too specialized a use case to warrant including this logic so happy to hear thoughts. 🙂 

#### Ticket Link

Relates to #299

### DIFF
--- a/webapp/src/components/sidebar/sidebarBoardItem.tsx
+++ b/webapp/src/components/sidebar/sidebarBoardItem.tsx
@@ -79,7 +79,6 @@ const SidebarBoardItem = React.memo((props: Props) => {
 
     const {board, intl, views} = props
     const displayTitle: string = board.title || intl.formatMessage({id: 'Sidebar.untitled-board', defaultMessage: '(Untitled Board)'})
-
     const boardViews = sortBoardViewsAlphabetically(views.filter((view) => view.parentId === board.id))
 
     return (
--- a/webapp/src/components/sidebar/sidebarBoardItem.tsx
+++ b/webapp/src/components/sidebar/sidebarBoardItem.tsx
@@ -79,7 +79,19 @@ const SidebarBoardItem = React.memo((props: Props) => {
 
     const {board, intl, views} = props
     const displayTitle: string = board.title || intl.formatMessage({id: 'Sidebar.untitled-board', defaultMessage: '(Untitled Board)'})
-    const boardViews = views.filter((view) => view.parentId === board.id)
+
+    // Sort views alphabetically after stripping leading emoji
+    const boardViews = views.filter((view) => view.parentId === board.id).map((v) => {
+        return {view: v, title: v.title.replace(/^\p{Emoji}*\s*/u, '')}
+    }).sort((v1, v2) => {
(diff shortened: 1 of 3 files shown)


webapp/src/components/sidebar/sidebarBoardItem.tsx
```diff
@@ -79,7 +79,19 @@ const SidebarBoardItem = React.memo((props: Props) => {
...
+        return {view: v, title: v.title.replace(/^\p{Emoji}*\s*/u, '')}
+    }).sort((v1, v2) => {
+        if (v1.title < v2.title) {
+            return -1
+        }
+        if (v1.title > v2.title) {
+            return 1
+        }
+        return 0
+    }).map((v) => v.view)
```
Johennes, line 94: The linter forced me to spread this out over several lines like this (e.g. no nesting of ternaries allowed).

webapp/src/components/sidebar/sidebarBoardItem.tsx
```diff
@@ -79,7 +79,19 @@ const SidebarBoardItem = React.memo((props: Props) => {
 
     const {board, intl, views} = props
     const displayTitle: string = board.title || intl.formatMessage({id: 'Sidebar.untitled-board', defaultMessage: '(Untitled Board)'})
-    const boardViews = views.filter((view) => view.parentId === board.id)
+
+    // Sort views alphabetically after stripping leading emoji
+    const boardViews = views.filter((view) => view.parentId === board.id).map((v) => {
+        return {view: v, title: v.title.replace(/^\p{Emoji}*\s*/u, '')}
+    }).sort((v1, v2) => {
+        if (v1.title < v2.title) {
```
chenilim, line 87: Thanks for the PR @Johennes! Looks great, but one minor thing here: we probably should use localeCompare here to handle international characters.

chenilim (changes requested):

webapp/src/components/sidebar/sidebarBoardItem.tsx
Johennes: Ah, of course! 🤦 That actually also makes the code more terse, so win-win. 🎉

Johennes:
While changing to `localeCompare` I realized that the sorting logic might benefit from unit tests. I extracted a method for the sorting and added a few basic tests.

chenilim (approved):
Awesome. Thanks for the updates @Johennes! The code looks cleaner now and love the unit tests.

More a note for the future: I can see the need for "sort without leading emoji" as a common operation throughout the app.

Johennes:
> More a note for the future: I can see the need for "sort without leading emoji" as a common operation throughout the app.

Yeah, true. I think refactoring the sorting into a general utility function should be straightforward if we need it elsewhere in the future.

PR #344, opened 2021-04-27
[GH-307] Increase clickable area on sidebar items

### LINKED ISSUE #307: Bug: Clickable area for views in left panel is too small
**Summary:**

Not a functional but rather an accessibility bug: When hovering views in the left panel, they highlight. The area that is clickable and where the cursor turns into a pointer is a lot smaller than the highlighted area though. Occasionally this results in having to double click to activate the view because the first click landed outside the clickable area.

1. Hover over a view name in the left panel
2. Try to click it at the edges
3. Notice how it's only possible to click close to the label

**Expected behavior:**

I think ideally the whole highlighted area should react to clicks.

**Screenshots:**

Screenshot 2021-04-21 at 07 53 25

(Approximate) clickable area marked in red.

**Platform:**

macOS 11.2.1
Focalboard v0.6.0

Johennes:
#### Summary

This moves the click handler from sidebar item titles to the sidebar items themselves. This way, the clickable area matches with the visually highlighted area on hover.

#### Ticket Link

Fixes #307

### DIFF
--- a/webapp/src/components/sidebar/sidebarBoardItem.scss
+++ b/webapp/src/components/sidebar/sidebarBoardItem.scss
@@ -30,6 +30,10 @@
             }
         }
 
+        &:not(.no-views) {
+            cursor: pointer;
+        }
+
         >.Icon {
             stroke: rgba(var(--sidebar-fg), 0.3);
             height: 14px;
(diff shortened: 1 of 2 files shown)


jespino (approved):
Nice one. Thanks @Johennes! :sparkles:

PR #350, opened 2021-04-27
Translate date and datetime with React Translator

leblanc-simon:
#### Summary
Currently dates are displayed in English regardless of the language used by the user.

This PR allows to use the React translation tool already used in focalboard to display the date in the user's language


jespino (approved):
Looks good to me. Thanks @leblanc-simon! :tada:

Issue #353, opened 2021-04-27
TextView - Implement Autosize on columns


sbishel:
Currently there is a handle to grab, drag and resize the column. When that handle is double clicked. Automatically size the column to fit its contents and header.j

PR #355, opened 2021-04-27
[GH-353] initial implmentation of autosize columns

sbishel:
#### Summary
Initial implementation of auto sizing on the columns. 

#### Ticket Link
https://github.com/mattermost/focalboard/issues/353

Apr-27-2021 15-07-57


webapp/src/components/table/table.tsx
```diff
@@ -26,6 +29,39 @@ type Props = {
...
+    fontDescriptor: '14px sans-serif',
+    padding: 30,
+}
+const title = {
+    fontDescriptor: '14px sans-serif',
+    padding: 57,
+}
+
+// re-use canvas object for better performance
+const canvas = document.createElement('canvas') as HTMLCanvasElement
```
chenilim, line 52: Minor: might want to move this to a shared utility class later. I can see this being useful for lots of things.

webapp/src/components/table/table.tsx
```diff
@@ -59,6 +95,59 @@ const Table = (props: Props) => {
...
+                }
+
+                const thisLen = getTextWidth(displayValue, fontDescriptor) + padding
+                if (thisLen > longestSize) {
+                    longestSize = thisLen
+                }
+            })
+        }
+
+        if (longestSize === 0) {
```
chenilim, line 141: I wonder if there should be a min-width for columns? Not sure if there are issues with very narrow columns currently.

chenilim (approved):
Wow, animated GIFs! You've upped our game @sbishel! :)

webapp/src/components/table/table.tsx
sbishel: There is a minimum width. And the existing code already handles that. A completely empty column gets set to the minimum width. (100px)

webapp/src/components/table/table.tsx
sbishel: Yeah, I debated on whether to do it now.

sbishel: Moved it to Utils.

webapp/src/components/table/table.tsx
```diff
@@ -26,6 +29,25 @@ type Props = {
     onCardClicked: (e: React.MouseEvent, card: Card) => void
 }
 
+// Eventually, these will need to come from the theme.
+// Currently, the theme doesn't change any padding or font descriptor
+const header = {
```
jespino, line 34: Do we can extract this information from the `TableHeader` component using the existing content? probably we can calculate it better there, and we can make it more generic here, so... the table Header give you the template id, and the new size, and you don't have to figure out every single use case here.

sbishel: Sorry, I'm not sure I understand which part to move to `TableHeader`. Just the calculation of the header width, right.

jespino: Yes, the thing is, you are taking the value, exposing the font size of the element, and all that, but in the TableHeader you probably already have much of the information inside, and you have all you need to make the change. So instead of passing directly the "onAutoSizeColumn" to the HorizontalGrip, you can have a function in the middle that populates the new width data base on the current title element (you can do data with a ref, and some javascript to get  styles/font-size/font-type) or maybe there is a better way, I don't know, but problably is the place to get that info.

sbishel:
Updated the implementation based on Jesus comments, now pull the font and padding from a ref.

jespino:
Now looks better, lets merge it and we can iterate later :)

Issue #358, opened 2021-04-28
Feature Idea: Allow to pass a docker variable to use a directory to build the image


jespino:
**Summary:**
Currently the docker file is generated from cloning the focalboard repo, having the posibility of use a path in your filesystem to build it would be a great addition. (This idea comes from the #168 PR)

**How important this is to me and why:**

Importance: Medium

Use cases:
1. Some users can want to create the docker file base on its own local source code.

laundmo:
This is most likely not going to be possible or at least very hacky, as docker doesn't allow conditional usage of the `COPY` statement.

I think the best solution is to use multistage builds and to target a specific stage depending on what the source should be.

PR #365, opened 2021-04-30
GH-358 - Make it possible to use sources when building docker image

Szymongib:
#### Summary
This PR makes it possible to build docker image from sources.
To build from sources, run from project root:
```
docker build . -f ./docker/Dockerfile --build-arg USE_LOCAL_SOURCES=true
```

#### Ticket Link
#358


jespino:
@mrckndt what do you think about it?

Issue #332, opened 2021-04-25
Bug: menu -> view -> tabbar should be disabled


Willyfrog:
**Summary:**
clicking on menu -> view -> tabbar shows a single tabbar below the title which looks weird. if tabs are not supported it should be disabled

**Steps to reproduce the behavior:**

1. clicking on menu -> view -> tabbar 

**Expected behavior:**
either the menu is not visible or there should be some sort of help on how to use tabs

**Screenshots:**
<img width="996" alt="Screen Shot 2021-04-26 at 01 16 57" src="https://user-images.githubusercontent.com/1515906/116012802-1a4e2480-a62d-11eb-921d-01d83eb17eb4.png">

**Platform:**
 - OS and version: macOS Big Sur, built from master

chenilim:
In general, the menu items in macOS needs to be cleaned up. Individually, these are fairly easy fixes in the storyboard. Opening it up if someone wants to grab it before I can get to it. :)

chenilim:
@haardikdharma10, to clarify, we should remove that first section of the view menu from the storyboard:
* Show Tab Bar
* Show All Tabs
* The divider

<img width="493" alt="Focalboard mac menu" src="https://user-images.githubusercontent.com/46905241/116725324-a302f280-a996-11eb-838a-1b979432d838.png">

PR #369, opened 2021-05-01
[Fix - #332]: Disable TabBar

haardikdharma10:
#### Summary

Disables Tab Bar in menu -> view -> tab bar.

#### Ticket Link

Fixes: #332

### DIFF
--- a/mac/Focalboard/AutoSaveWindowController.swift
+++ b/mac/Focalboard/AutoSaveWindowController.swift
@@ -6,6 +6,7 @@ import Cocoa
 class AutoSaveWindowController: NSWindowController {
 
 	override func windowDidLoad() {
+        self.window?.tabbingMode = .disallowed
 		super.windowDidLoad()
 
 		// Implement this method to handle any initialization after your window controller's window has been loaded from its nib file.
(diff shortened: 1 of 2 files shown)


mac/Focalboard/AutoSaveWindowController.swift
```diff
@@ -6,6 +6,7 @@ import Cocoa
 class AutoSaveWindowController: NSWindowController {
 
 	override func windowDidLoad() {
+        self.window?.tabbingMode = .disallowed
```
Johennes, line 9: You can actually set this in the storyboard (see the drop-down in the lower right corner of the screenshot below). Setting it in code should have the same effect but since there are already lots of other settings applied in the storyboard, keeping it there could simplify future maintenance.

Screenshot 2021-05-02 at 20 57 38

haardikdharma10: @Johennes, that's convenient, thanks!

chenilim (approved):
Thanks @haardikdharma10! Looks good. Built, tested, and verified that the tab options don't appear in the view menu.

PR #371, opened 2021-05-01
Update de.json

Prophet-Phoenix:
Password is "Passwort" in german and TableRow.open is "Öffnen" as long as it means the action "open that item" and not its status "it is open". Am a native speaker.


Johennes:
@Prophet-Phoenix I just fixed both of these on https://translate.mattermost.com so the translations should show up correctly once they have been imported for a new release.

PR #380, opened 2021-05-03
[GH-314] Export native app user settings on change

Johennes:
#### Summary

This switches from exporting the native app user settings on window close to exporting on settings change. This way the settings export remains independent of native application life-cycle events.

This is a stop-gap towards enabling settings export on the native Linux app. The latter does not have an easy way to catch window close events.

As a positive side-effect this also makes the settings export on the native macOS app more reliable and prevents settings loss when the app crashes or is force-killed. Additionally, the required code on the native app side is reduced.

On the negative side, the settings are still exported as blob which now causes more processing since the export happens more often. I already exposed the changed settings key in the script message so it should be relatively easy to switch to storing the settings individually in a follow-up change.

#### Ticket Link

Relates to #314


mac/Focalboard/ViewController.swift
```diff
@@ -76,34 +71,6 @@ class ViewController:
...
-			}
-			UserDefaults.standard.set(blob, forKey: "localStorage")
-			NSLog("Persisted user settings: \(Data(base64Encoded: blob).flatMap { String(data: $0, encoding: .utf8) } ?? blob)")
-		}
-
-		// During shutdown the system grants us about 5 seconds to clean up and store user data
-		let timeout = DispatchTime.now() + .seconds(3)
-		var result: DispatchTimeoutResult?
-
-		// Busy wait because evaluateJavaScript can only be called from *and* signals on the main thread
```
Johennes, line 96: Busy wait be gone! 🎉

webapp/src/main.tsx
```diff
@@ -2,13 +2,18 @@
...
 import {initThemes} from './theme'
+import {importNativeAppSettings} from './nativeApp'
+import {UserSettings} from './userSettings'
 
 import './styles/variables.scss'
 import './styles/main.scss'
 import './styles/labels.scss'
 
+store.setHandlers({getter: UserSettings.getEmojiMartSetting, setter: UserSettings.setEmojiMartSetting})
+importNativeAppSettings()
```
Johennes, line 17: I moved this here because the `initThemes` call below already reads from localStorage.

webapp/src/userSettings.ts
```diff
@@ -1,23 +1,94 @@
...
     }
 
     static set prefillRandomIcons(newValue: boolean) {
-        localStorage.setItem('randomIcons', JSON.stringify(newValue))
+        UserSettings.set(UserSettingKey.RandomIcons, JSON.stringify(newValue))
     }
-}
 
-const keys = ['language', 'theme', 'lastBoardId', 'lastViewId', 'emoji-mart.last', 'emoji-mart.frequently']
+    static getEmojiMartSetting(key: string): any {
```
Johennes, line 72: This method and the one below are a bit hairy. There's some documentation about the emoji-mart settings keys and values in the [upstream README](https://github.com/missive/emoji-mart) (see the "Storage" section).

Johennes:
For some reason ESLint keeps rejecting the enum as a shadowing declaration.

```
 /home/runner/work/focalboard/focalboard/webapp/src/userSettings.ts
  7:6  error  'UserSettingKey' is already declared in the upper scope on line 7 column 6  no-shadow
```

I suspect this is a false positive as I can't see how this would shadow an existing declaration?

chenilim:
Thanks so much @Johennes! I'm so sorry for the delay. Totally my fault for missing this on my todo list! Reviewing now.

webapp/src/nativeApp.ts
```diff
@@ -19,14 +19,10 @@ export function importNativeAppSettings() {
     NativeApp.settingsBlob = null
 }
 
+export function notifySettingsChanged(key: string) {
+    postWebKitMessage({type: 'didChangeUserSettings', key, settingsBlob: exportUserSettingsBlob()})
+}
+
 function postWebKitMessage(message: any) {
```
chenilim, line 26: Love this construct - for extending webapp-to-native messaging in the future. We'll add a mental todo to also adapt this for Windows / Linux in the future.

webapp/src/userSettings.ts
chenilim: Yeah, thanks for parsing the Emojimart settings! We expect to refactor the emoji picker in the future, so this will likely change (not for a while though).

webapp/src/userSettings.ts
```diff
@@ -1,23 +1,95 @@
...
+    EmojiMartFrequently = 'emoji-mart.frequently',
+    RandomIcons = 'randomIcons'
+}
+
 export class UserSettings {
+    static get(key: UserSettingKey): string | null {
+        return localStorage.getItem(key)
+    }
+
+    static set(key: UserSettingKey, value: string | null) {
```
chenilim, line 24: Love the idea of abstracting UserSettings away from localStorage. One thing to be careful of is localStorage is used for other things too (like the session token). Might be a good idea to check that key is always in the set of allowable keys defined above.

Also, I noticed in `importUserSettings`, that it stamps entries into localStorage. Might be good to check the keys there too. I can't think of any adverse security issues, but this might open holes in the future.

chenilim (changes requested):
Thanks @Johennes, and sorry again for the delay. This looks great, and the only change I think we should make now is to check the key membership for UserSettings.

Adding @sbishel as an FYI. We need to later expand the UserSettings persistence to Windows and Linux, and can continue to refine it.

webapp/src/userSettings.ts
Johennes: Ah, yes that's an excellent point. I added an `if` to check for the allowed keys in both places.

Johennes:
> Thanks so much @Johennes! I'm so sorry for the delay. Totally my fault for missing this on my todo list! Reviewing now.

Ah, no worries at all. I actually forgot that I did this myself until I received the notification about the "stale" label being added last night. 😅 

I made the change to check the local storage keys.

chenilim (approved):
Thanks for making the changes @Johennes! LGTM!

I thought about it a bit more, and it's probably less about being a security issue (if someone can inject JS, there's not much we can do), but this verification would help catch any code or typo errors in the future.

hanzei:
@chenilim Should this PR be reviewed by someone else?

Johennes:
I rebased this on `main` and resolved the conflict but I think it needs another look now. When I start and stop the app multiple times in a row, I sometimes see the previous settings not being retained on relaunch. They're imported at first but then something seems to overwrite them again. I will have to investigate that. 😞

Johennes:
So this was one of these cases where a simple typo makes you go on a two-hour goose chase and lets you question God and the world until you finally realize that you fat-fingeredly used `Object.values(UserSettings)` instead of `Object.values(UserSettingKey)`. 🤦 

This is fixed now and things appear to work correctly again. I left some of the extended logging that I used to analyze the issue around as well. The interaction between the web view and the app is quite tedious to debug so this might come in handy in future. 🙂

jasonblais:
cc @jespino on above

Johennes:
@jespino / @chenilim many apologies for the _huge_ delay. I finally found the time to merge the `main` branch in again. I briefly re-tested the change again and it's now ready for code review. 🙂

jespino:
I'll review it ASAP. Thanks for all the effort and the patience @Johennes

jespino (approved):
LGTM

Issue #31, opened 2021-01-20
Fix server lint issues


chenilim:
Run `make server-lint` to see the list of Go lint issues. We should incrementally fix them.

Snehal1112:
@chenilim I would like to work  on this issue

chenilim:
Thanks @Snehal1112! I recommend tackling this incrementally.  `make server-lint` will show the list of lint errors, but some of them might be false positives. Start with the ones you feel most comfortable with, and we can make multiple PRs. Thank you for contributing to Focalboard!

Snehal1112:
@chenilim @jespino Quick question!! In code, I found there is some code which never used anywhere e.g. https://github.com/mattermost/focalboard/blob/main/server/app/blocks.go#L37 
It intentionally did that or someone forgot to remove it.

chenilim:
Hi @Snehal1112, thanks for catching that! Yup, I forgot to remove them when refactoring. :) Please go ahead and remove `blockIDsToNotify` and `uniqueBlockIDs` from InsertBlocks and DeleteBlock. They aren't used. Thanks!

Snehal1112:
@chenilim thanks will do. I am going to create multiple PR for linter issues so each PR contains a fix for individual modules.
feel free to suggest any other better way.

harshilsharma63:
I'm picking this up due to the inactivity.

PR #381, opened 2021-05-04
#GH-31 Configured Golangci-Lint and fixed errors

harshilsharma63:
#### Summary
1. Moved golangci-lint config from inline command to configuration file.
2. Fixed all reported errors.

#### Ticket Link
Fixed https://github.com/mattermost/focalboard/issues/31


server/server/server.go
```diff
@@ -210,7 +222,9 @@ func (s *Server) Shutdown() error {
 		s.cleanUpSessionsTask.Cancel()
 	}
 
-	s.telemetry.Shutdown()
+	if err := s.telemetry.Shutdown(); err != nil {
```
jespino, line 225: Now if it fails to Shutdown the telemetry is no longer trying to shutdown the store. I think is more correct to log an error there instead of returning the error. Maybe we can use a multi-error here.

jespino (changes requested):
A tiny detail, otherwise looks good to me.

harshilsharma63:
@jespino I've removed presets from `golangci-lint` config as they were forcefully including some not-very-useful linters.

jespino (approved):
LGTM

PR #383, opened 2021-05-05
Update "Open" button size

arjitc:
Updated the "Open" button size to better fill the row

#### Summary
The open button currently does not fill the row completely and makes it a bit odd when trying to click on it

Focalboard_2021-05-05_21-01-47

Which has been updated to,

Focalboard_2021-05-05_21-00-24

#### Ticket Link
 
None I think


jespino:
Thanks @arjitc, nice improvement.

PR #392, opened 2021-05-09
Add missing German translations

Johennes:
#### Summary

This adds missing German translations for all currently existing strings.

#### Ticket Link

None. I hope this is ok to open without a corresponding GitHub issue because it seems like an obvious improvement. I'm a native German speaker.


jespino:
@Johennes thanks, but we add the translations throug our weblate (http://translate.mattermost.com) otherwise the thing get messy pretty quickly.

PR #393, opened 2021-05-10
Clickable Links

harshilsharma63:
#### Summary
Made URL type properties clickable.

#### Ticket Link
N/A

Screenshot -

Screenshot 2021-05-19 at 2 55 43 PM


webapp/package.json
```diff
@@ -22,6 +22,9 @@
 		"cypress:open": "cypress open"
 	},
 	"dependencies": {
+		"@fortawesome/fontawesome-svg-core": "^1.2.35",
```
jespino, line 25: I don't know if we want to add fontawesome as a dependency. Actually, we should be reusing the mattermost icons font.

jespino:
@signalwerk you are right, the only set of icons that we have now are in the `widgets/icons` directory, but in Mattermost we have an icon font and we are moving forward in that direction there, probably we want to use that same icon-font here. My main concern is about using the font-awesome option when we already have one font based option selected for the Mattermost project.

harshilsharma63:
I'll be updating the font library for this PR.

harshilsharma63:
@jespino I've added Compass Icons here and the PR is now ready to go.

harshilsharma63:
@hahmadia added you as a reviewed since Jesus is out this week.

hahmadia (commented):
1. Can we discard / uncommit the package-lock.json file changes? Usually never commit that for mattermost webapp

2. I set a link to `github.com` but when I clicked it took me too `http://localhost:8000/www.github.com` is that expected? Will this only happen locally?

3. Improvement idea: Is it possible to make it so that if I hover over the link, I can click it (it's clickable link without necessarily requiring to click the icon) but if I click on the whitespace beside it, I can edit it and change it? **This doesn't need to be done in this PR, we can note it for future iteration. Just a thought to avoid having to specifically click on the icon itself**

Good job otherwise! 👍

webapp/package.json
harshilsharma63: I've replaced FontAwesome with Compass Icons now.

harshilsharma63:
> 1. Can we discard / uncommit the package-lock.json file changes? Usually never commit that for mattermost webapp

It's changed because I added a new dependency, `image-webpack-loader` for loading SVG font for Compass Icons. I think we need to keep the changes for that.

> 2. I set a link to `github.com` but when I clicked it took me too `http://localhost:8000/www.github.com` is that expected? Will this only happen locally?
> 3. Improvement idea: Is it possible to make it so that if I hover over the link, I can click it (it's clickable link without necessarily requiring to click the icon) but if I click on the whitespace beside it, I can edit it and change it? **This doesn't need to be done in this PR, we can note it for future iteration. Just a thought to avoid having to specifically click on the icon itself**

Awesome catch 👍 This is fixed now. What was happening was that for any link that doesn't start with a protocol or with a `//`, browsers assume it to be a relative URL. I've fixed this by always ensuring a protocol in the underlying anchor tag's `href` value.

> Good job otherwise! 👍

Thanks 🙂

webapp/src/components/properties/link/link.tsx
```diff
@@ -0,0 +1,47 @@
...
+    onCancel: () => void
+    validator: (newValue: string) => boolean
+}
+
+const URLProperty = (props: Props): JSX.Element => {
+    let link: ReactNode = null
+    if (props.value?.trim()) {
+        link = (
+            <a
+                className={'icon-link-variant'}
```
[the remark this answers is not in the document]
harshilsharma63: Done.

webapp/src/components/properties/link/link.tsx
```diff
@@ -0,0 +1,47 @@
...
+                href={props.value.trim()}
+                target={'_blank'}
```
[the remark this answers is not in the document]
harshilsharma63: Done.

webapp/src/components/propertyValueElement.tsx
```diff
@@ -100,6 +103,16 @@ const PropertyValueElement = (props:Props): JSX.Element => {
                 }
             />
         )
+    } else if (propertyTemplate.type === 'url') {
```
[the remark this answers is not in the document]
harshilsharma63: I think logically it should all be `else if`. The one `if` below this is from a community PR and I missed it in the code review. In my other pending PR I've converted the `if` into an `else if`. Either way babe web pack will probably convert it to else if, knowing there is nothing between those condition blocks :P

webapp/src/utils.ts
```diff
@@ -211,6 +211,10 @@ class Utils {
         }
     }
 
+    static assureProtocol(url: string): string {
+        return url.match(/^.{3,9}:\/\//) ? url : `https://${url}`
```
hahmadia, line 215: Nit:

This currently matches antyhing like `afdsffafasd://` as long as it's between 3-9 characters.
A slight improvement would be to do `^(http|https)://` regex

webapp/src/components/propertyValueElement.tsx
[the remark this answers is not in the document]
hahmadia: Yep makes sense. I will convert mine to a else if too

hahmadia (approved):
Looks good. Small nit improvement but otherwise I don't see a problem.

webapp/src/utils.ts
hahmadia: Also I think 
image 
should always be showing up once you enter a valid URL.

It only shows up for me once after I hover over the link area which is confusing.

webapp/src/components/properties/link/link.tsx
```diff
@@ -0,0 +1,48 @@
...
+                className='icon-link-variant'
+                href={Utils.assureProtocol(props.value.trim())}
+                target='_blank'
+                rel='noreferrer'
+            />
+        )
+    }
+
+    return (
+        <div className={'property-link url'}>
```
hahmadia, line 33:
```suggestion
        <div className='property-link url'>
```

webapp/src/components/properties/link/link.scss
```diff
@@ -0,0 +1,26 @@
+.property-link {
+    &.url {
+        padding-right: 2rem;
+    }
+
+    a {
+        padding: 8px; // increases clickable area for better UX
+        display: none;
```
hahmadia, line 8: I think the `display: none` should be removed and just show the icon all the time for link property.

hahmadia: I know this means they can enter a invalid URL and get redirected by clicking the icon but atleast we have a red broder around the link property indicating it's invalid and I feel at that point it's on them.

webapp/src/components/properties/link/link.scss
```diff
@@ -0,0 +1,26 @@
...
+
+    a:hover {
+        background:  unset;
+    }
+
+    &:hover {
+        a {
+            display: inline;
+        }
+    }
```
hahmadia, line 23: If we remove display: none, we can remove this.

hahmadia (commented):
Sorry for double review. Just noticed some small things I want comment on.
Thoughts?

webapp/src/utils.ts
harshilsharma63: > This currently matches antyhing like notright:// as long as it's between 3-9 characters before the ://.
A slight improvement would be to do ^(http|https):// regex

I had the same thought but but then I though why to restrict to those protocols. For example, someone can use FTP, magnet, web socket or anything. Thoughts?

webapp/src/components/properties/link/link.tsx
harshilsharma63: Done :+1:

webapp/src/components/properties/link/link.scss
harshilsharma63: Keeping the icon hidden until hovered on, so need to preserve `display: none`.

webapp/src/components/properties/link/link.scss
harshilsharma63: Keeping the icon hidden until hovered on, so need to preserve this.

webapp/src/utils.ts
hahmadia: I had this thought after but what if my schema is like mattermost:// or something. If it's less than 3 or greater than 9 characters, it will display an error. In my opinion, I would say restrict to http/https protocol _for the time being_ and if in the future for whatever reason we need others, we can remove that limitation. 

This also  comes from testing in notion which doesn't accept protocols other than http/https for it's links.

hahmadia (commented):
Nice changes. One small question about protocol and that's all.

webapp/src/utils.ts
harshilsharma63: Done 👍

webapp/src/components/properties/link/link.tsx
```diff
@@ -0,0 +1,48 @@
...
+                className='icon-link-variant'
+                href={Utils.ensureProtocol(props.value.trim())}
+                target='_blank'
+                rel='noreferrer'
+            />
+        )
+    }
+
+    return (
+        <div className='property-link url'>
```
jespino, line 33: We use by convetion that the className of the root div of a component is the component name, in this case is `URLProperty`, that way uppercased classes are clearly distinguished as component classes. and also is easy to correlate them to the correct react component.

webapp/src/components/propertyValueElement.tsx
```diff
@@ -44,7 +47,7 @@ const PropertyValueElement = (props:Props): JSX.Element => {
             return emailRegexp.test(val)
         }
         case 'url': {
-            const urlRegexp = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)/
+            const urlRegexp = /((((http|https):(?:\/\/)?)(?:[-;:&=+$,\w]+@)?[A-Za-z0-9.-]+|(?:www\.|[-;:&=+$,\w]+@)[A-Za-z0-9.-]+)((?:\/[+~%/.\w\-_]*)?\??(?:[-+=&;%@.\w_]*)#?(?:[.!/\\\w]*))?)/
```
jespino, line 50: I'm ok with this, but I don't know if that is right, a url is a url, and can be anything else, not only http or https. We can decide to not show the link if it is not http or https (something that is also weird to me), but not allow somebody to use a, for example, ftp url here, for me, makes not sense.

webapp/src/components/properties/link/link.tsx
```diff
@@ -0,0 +1,48 @@
...
+    onCancel: () => void
+    validator: (newValue: string) => boolean
+}
+
+const URLProperty = (props: Props): JSX.Element => {
+    let link: ReactNode = null
+    if (props.value?.trim()) {
+        link = (
+            <a
+                className='icon-link-variant'
```
jespino, line 24: This introduces another way to show icons, we were using icon components, and here you are using a class to introduce the icons, in my opinion this is more confusing and hard to track, I prefer to avoid that and create a new icon component that just render a `<span className='icon-link-variant'/>`, that way is easier to update/customize icons in a homogeneous way, and also we have a clear place to find all the supported icons in the application.

webapp/webpack.common.js
```diff
@@ -66,6 +66,22 @@ function makeCommonConfig() {
                     ],
                     exclude: [/node_modules/],
                 },
+                {
+                    test: /\.(png|eot|tiff|svg|woff2|woff|ttf|gif|mp3|jpg)$/,
```
jespino, line 70: Why are we exposing here more than we need? the png, tiff, gif, mp3 and jpg files are not needed yet, why we add them upfront if it is not explicitly needed?

webapp/src/components/propertyValueElement.tsx
harshilsharma63: @jespino Now that you put it that way, I agree with this. IMO we shouldn't even restrict the protocol length between 3 and 9 and allow any arbitrary length. For example, I can create my own desktop app which opens links registered with `foobarbaz://` protocol.

I've created this to weekly agenda, proposing any arbitrary protocol.

hahmadia (approved):
Other than what Jesus pointed out , everything else looks good to me. Thanks

webapp/src/components/properties/link/link.tsx
harshilsharma63: @jespino The existing icon components embedded an SVG, which happened before we added compass icons to Focalboard. With Compass Icons, I think this is the simplest way. If we create a layer of React components over compass icons, we'd have to keep creating new components each time we use. anew icons. Also, with the plan to make compass icons an NPM package and simple add it as a dependency, these components will need maintain wither we update the compass icons.

This is also how Mattermost uses compass icons. React component would make sense if for some reason we wanted to encapsulate the implementation of icons. Otherwise it's unnecessary components IMO.

What's you're rational for using components for individual icons BTW? Maybe I'm missing something?

webapp/src/components/properties/link/link.tsx
harshilsharma63: Got it. Done.

webapp/webpack.common.js
harshilsharma63: Done.

jespino (approved):
Now LGTM.

webapp/src/components/properties/link/link.tsx
harshilsharma63: Added React components for individual icons as discussed.

webapp/src/components/propertyValueElement.tsx
harshilsharma63: This is done now.

Issue #261, opened 2021-04-13
Feature Idea: Export/import board


ZevSwe:
**Summary:**
It would be nice if one could export just one board. Say I designed a killer board for tracking software projects that I would like to share as a template for the community and someone else wanted to import and test that. One would want to do that without effecting any other boards one has.

**How important this is to me and why:**

Importance: Low
This is a nice to have feature. It's not really that important right now. 

Use cases:
1. Share a template with the community
2. Import someone else's template

chenilim:
Just going through issues. Yes, this is a feature we want. What we need is:
* A new API to export all blocks from a rootID (i.e. all the contents of a board)
* UI to export a board into the .focalboard archive format

hahmadia:
Hey guys,

I'll be taking a look at this.

chenilim:
Thanks for picking this up @hahmadia!

PR #403, opened 2021-05-11
Add Export/import board 

hahmadia:
#### Summary

Export/import board 

#### Ticket Link

Fixes GH-261

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -229,7 +229,7 @@ func (a *API) handleGetBlocks(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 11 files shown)


chenilim (approved):
I think @jespino is out, but this LGTM. I'll also loop in @sbishel for a second pair of eyes.

Notes for the future:
* We might want to consider moving the GetBlocksWithRootID function into the handleGetBlocks API route (vs. export).
* A possible perf improvement is for BoardTree to get the entire board tree (vs. just 2 levels) in a single call.

jespino (approved):
LGTM

sbishel (approved):

PR #406, opened 2021-05-12
Person Property Type

harshilsharma63:
#### Summary
Added person property type to fetch dn display list of users from Mattermost


server/services/store/sqlstore/user.go
```diff
@@ -158,3 +158,29 @@ func (s *SQLStore) UpdateUserPasswordByID(userID, password string) error {
 
 	return nil
 }
+
+func (s *SQLStore) GetUsersByWorkspace(workspaceID string) ([]*model.User, error) {
+	// TODO: implement this to fetch users from Mattermost
```
jespino, line 163: Nope, that is wrong, you have to implement this to get the users from Focalboard users table, then I going to put a "layer" on top of that, to get the users from mattermost database whenever this is running as a plugin.

webapp/src/components/properties/user/user.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    return (
+        <Select
+            options={users}
+            isSearchable={true}
+            isClearable={true}
+            backspaceRemovesValue={true}
+            className={'user-picker'}
+
+            styles={{
```
sbishel, line 47: Should these be coming from a constant, rather than setting them here. Could they be reused?

server/swagger/docs/html/index.html
```diff
@@ -2,7 +2,8 @@
 <html>
 <head>
   <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-  <title>Focalboard Server</title>
+  <title>line /Users/harshilsharma/codebase/mattermost/focalboard/server/main/main.go:1:1
+Package classification Focalboard Server</title>
```
sbishel, line 6: I don't know swagger, but these changes look suspicious to me. Although, I'm guessing these were done automatically.

sbishel (approved):

server/swagger/docs/html/index.html
chenilim: Yeah, you need to manually go in and clean up any hardcoded paths when re-generating the Swagger docs.

chenilim: In this case, the title should just be **Focalboard Server**, as before.

server/services/store/sqlstore/user.go
harshilsharma63: Done. I'm now fetching users from Focalboard `users` table.

server/swagger/docs/html/index.html
harshilsharma63: Oh, didn't notice that swagger picked my local paths. I've fixed these now.

webapp/src/components/properties/user/user.tsx
harshilsharma63: Yes, they would serve well as base styles for React Select. I've updated it now.

server/services/store/sqlstore/user.go
```diff
@@ -158,3 +179,44 @@ func (s *SQLStore) UpdateUserPasswordByID(userID, password string) error {
 
 	return nil
 }
+
+func (s *SQLStore) GetUsersByWorkspace(workspaceID string) ([]*model.User, error) {
+	return s.getUsersByCondition(nil)
```
sbishel, line 184: Shouldn't this contain the condition. 
```
sq.Eq{"workspaceID": workspaceID}
```

server/services/store/sqlstore/user.go
```diff
@@ -27,26 +29,45 @@ func (s *SQLStore) GetRegisteredUserCount() (int, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	err = json.Unmarshal(propsBytes, &user.Props)
+	var user *model.User
+
+	if len(users) > 0 {
+		user = users[0]
+	}
```
sbishel, line 41: Previously `.QueryRow` was used which accord to documentation, throws an error if no records are returned. 
https://golang.org/pkg/database/sql/#DB.QueryRow
`.Query` does not throw an error when no rows are returned. This may be find depending on how calling function are working, but thought I'd mention it.

sbishel (changes requested):
Also, if possible, let's add unit tests for user.tsx and propertyValueElement.tsx

server/services/store/sqlstore/user.go
harshilsharma63: This function will only be called for the standalone version of Focalboard, which has just one workspace, qualifying all users as member of that one workspace. Also in Focalboard database, we have no relationship between user and the workspace.

Later on @jespino will be adding a layer of top of this SQL store which would retrieve the Mattermost channel from workspace ID, and then verify user's association with that channel. The `workspaceID` param is only there for the purpose of interface and isn't used in Focalboard's own database.

server/services/store/sqlstore/user.go
harshilsharma63: Great catch! The function's callers are actually assuming a non-nil return value when there is no error. I've fixed this now by manually returning`sql.ErrNoRows` when no data is found.

sbishel (approved):

server/api/api.go
```diff
@@ -1120,6 +1121,52 @@ func (a *API) handleUploadFile(w http.ResponseWriter, r *http.Request) {
...
+	//         "$ref": "#/definitions/User"
+	//   default:
+	//     description: internal error
+	//     schema:
+	//       "$ref": "#/definitions/ErrorResponse"
+
+	vars := mux.Vars(r)
+	workspaceID := vars["workspaceID"]
+
+	users, err := a.app().GetWorkspaceUsers(workspaceID)
```
jespino, line 1155: Probably we should check if you have permissions to see the workspace.

server/services/store/sqlstore/user.go
```diff
@@ -27,26 +29,54 @@ func (s *SQLStore) GetRegisteredUserCount() (int, error) {
...
 
 func (s *SQLStore) getUserByCondition(condition sq.Eq) (*model.User, error) {
+	users, err := s.getUsersByCondition(condition)
+	if err != nil {
+		return nil, err
+	}
+
+	var user *model.User
+
+	if len(users) > 0 {
```
jespino, line 39: I think this can be clearer if you do something like:
```go
if len(users) == 0 {
    return nil, nil
}
return users[0], nil
```

server/services/store/sqlstore/user.go
jespino: I think we should add a comment here to explain that. It is not clear from the context.

webapp/src/components/properties/user/user.tsx
```diff
@@ -0,0 +1,64 @@
...
+type Props = {
+    value: string,
+    readonly: boolean,
+    onChange: (value: string) => void,
+}
+
+const UserProperty = (props: Props): JSX.Element => {
+    const [users, setState] = useState<Array<IUser>>([])
+
+    useEffect(() => {
```
jespino, line 22: This is going to create a getWorkspaceUsers call for each UserProperty instance rendered, I think makes a lot of sense to have the list of users of the workspace in something like context manager. Whenever you switch the workspace, you get the users of the workspace, and wherever you need to show the list of users or get an specific user, you can use it from the context.

webapp/src/widgets/valueSelector.tsx
```diff
@@ -88,50 +89,7 @@ function ValueSelector(props: Props): JSX.Element {
...
-                    paddingTop: 0,
-                    marginBottom: 0,
-                    marginTop: 0,
-                }),
-                menuList: (provided: CSSObject): CSSObject => ({
-                    ...provided,
-                    overflowY: 'unset',
-                }),
-            }}
+            styles={getSelectBaseStyle()}
```
jespino, line 92: nice refactor :)

jespino (changes requested):
Some changes needed :)

server/services/store/sqlstore/user.go
harshilsharma63: Done.

server/api/api.go
harshilsharma63: Done.

webapp/src/widgets/valueSelector.tsx
harshilsharma63: Thanks!

webapp/src/components/properties/user/user.tsx
harshilsharma63: Thanks a lot for that catch. Didn't knew about React Contexts. I've refactored this now.

webapp/src/components/properties/user/user.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const UserProperty = (props: Props): JSX.Element => {
+    let value: IUser | undefined
+
+    if (props.readonly) {
+        return (<div className='UserProperty octo-propertyvalue'>{value ? value.username : props.value}</div>)
+    }
+
+    return (
+        <WorkspaceUsersContext.Consumer>
```
jespino, line 26: Is there any reason to not use https://reactjs.org/docs/hooks-reference.html#usecontext, I think would be more clear to use just:

```typescript
const workspaceUsers = useContext(WorkspaceUsersContenxt);
```

jespino (approved):
LGTM, only a tiny improvement around more "react-hooks-friendly" approach accessing the context data.

webapp/src/user.tsx
```diff
@@ -5,6 +5,8 @@ import React from 'react'
 
 const UserContext = React.createContext(undefined as IUser|undefined)
 
+const WorkspaceUsersContext = React.createContext(undefined as Array<IUser>|undefined)
```
jespino, line 8: Probably is more useful to store in the context a map of `UserID -> User`, that way we can use that map to complete the user information everywhere (avatar, username, etcetera).

webapp/src/components/properties/user/user.tsx
harshilsharma63: Done. Learning a lot about React hooks!

webapp/src/user.tsx
harshilsharma63: Done.

webapp/src/components/properties/user/user.tsx
```diff
@@ -16,24 +16,23 @@ type Props = {
...
+            options={workspaceUsers.users}
             isSearchable={true}
             isClearable={true}
             backspaceRemovesValue={true}
             className={'UserProperty'}
             styles={getSelectBaseStyle()}
             getOptionLabel={(o: IUser) => o.username}
             getOptionValue={(a: IUser) => a.id}
-            value={workspaceUsersById?.get(props.value) || null}
+            value={workspaceUsers.usersById?.get(props.value) || null}
```
jespino, line 35: Because the type syste, I think the `?` is not needed, it should be mandatory. Can be empty, but not `null` or `undefined`.

webapp/src/components/properties/user/user.tsx
```diff
@@ -16,24 +16,23 @@ type Props = {
 }
 
 const UserProperty = (props: Props): JSX.Element => {
-    const workspaceUsers = useContext(WorkspaceUsersContext) as Array<IUser>|undefined
-    const workspaceUsersById = useContext(WorkspaceUsersByIDContext) as Map<string, IUser>|undefined
+    const workspaceUsers = useContext(WorkspaceUsersContext) as WorkspaceUsersContextData
 
     if (props.readonly) {
-        return (<div className='UserProperty octo-propertyvalue'>{workspaceUsersById?.get(props.value)?.username || props.value}</div>)
+        return (<div className='UserProperty octo-propertyvalue'>{workspaceUsers.usersById?.get(props.value)?.username || props.value}</div>)
```
jespino, line 22: same here, the `?` is not needed.

webapp/src/components/properties/user/user.tsx
```diff
@@ -16,24 +16,23 @@ type Props = {
 }
 
 const UserProperty = (props: Props): JSX.Element => {
-    const workspaceUsers = useContext(WorkspaceUsersContext) as Array<IUser>|undefined
-    const workspaceUsersById = useContext(WorkspaceUsersByIDContext) as Map<string, IUser>|undefined
+    const workspaceUsers = useContext(WorkspaceUsersContext) as WorkspaceUsersContextData
```
jespino, line 19: can it be `useContext<WorkspaceUsersContextData>(WorkspaceUsersContext)` instead?

webapp/src/components/properties/user/user.tsx
harshilsharma63: Done.

webapp/src/components/properties/user/user.tsx
harshilsharma63: Done.

webapp/src/components/properties/user/user.tsx
harshilsharma63: You learn something new everyday! Thanks for the tip! Done.

PR #414, opened 2021-05-14
Enable prometheus metrics for Focalboard to enable observability

spirosoik:
#### Summary
As an SRE team we would like to expose standard metrics and grouped by version
of the application. Right now will expose only metrics related to Go but instrumentor
should be used in other parts of the codebase so we can track other metrics, eg.
number of tasks, boards, users etc. Similar as we do in MM.

It will run in port `localhost:9092` and it is a new config `prometheus_address`.

Also in the commit we introduced, `group.Add` which helps us to handle gracefully
errors for goroutines. It's a good practice and there are couple of articles by
Golang for this.

#### Ticket Link
Open source Friday :)

### DIFF
--- a/server/server/server.go
+++ b/server/server/server.go
@@ -21,6 +21,7 @@ import (
 	"github.com/mattermost/focalboard/server/context"
 	appModel "github.com/mattermost/focalboard/server/model"
 	"github.com/mattermost/focalboard/server/services/config"
+	"github.com/mattermost/focalboard/server/services/prometheus"
 	"github.com/mattermost/focalboard/server/services/scheduler"
 	"github.com/mattermost/focalboard/server/services/store"
 	"github.com/mattermost/focalboard/server/services/store/sqlstore"
@@ -30,6 +31,7 @@ import (
 	"github.com/mattermost/focalboard/server/ws"
 	"github.com/mattermost/mattermost-server/v5/services/filesstore"
 	"github.com/mattermost/mattermost-server/v5/utils"
+	"github.com/oklog/run"
 )
 
 const (
@@ -48,6 +50,8 @@ type Server struct {
 	telemetry           *telemetry.Service
 	logger              *zap.Logger
 	cleanUpSessionsTask *scheduler.ScheduledTask
+	promServer          *prometheus.Service
+	promInstrumentor    *prometheus.Instrumentor
 
 	localRouter     *mux.Router
 	localModeServer *http.Server
@@ -165,16 +169,18 @@ func New(cfg *config.Configuration, singleUserToken string) (*Server, error) {
 	})
 
 	server := Server{
-		config:       cfg,
-		wsServer:     wsServer,
-		webServer:    webServer,
-		store:        db,
-		filesBackend: filesBackend,
-		telemetry:    telemetryService,
-		logger:       logger,
-		localRouter:  localRouter,
-		api:          focalboardAPI,
-		appBuilder:   appBuilder,
+		config:           cfg,
+		wsServer:         wsServer,
+		webServer:        webServer,
+		store:            db,
+		filesBackend:     filesBackend,
+		telemetry:        telemetryService,
+		promServer:       prometheus.New(cfg.PrometheusAddress),
+		promInstrumentor: prometheus.NewInstrumentor(appModel.CurrentVersion),
+		logger:           logger,
+		localRouter:      localRouter,
+		api:              focalboardAPI,
+		appBuilder:       appBuilder,
 	}
 
 	server.initHandlers()
@@ -209,6 +215,23 @@ func (s *Server) Start() error {
 		s.telemetry.RunTelemetryJob(firstRun)
 	}
 
+	var group run.Group
+	if s.config.PrometheusAddress != "" {
+		group.Add(func() error {
+			if err := s.promServer.Run(); err != nil {
+				return errors.Wrap(err, "PromServer Run")
+			}
+			return nil
+		}, func(error) {
+			s.promServer.Shutdown()
+		})
+		// expose the build info as metric
+		s.promInstrumentor.ExposeBuildInfo()
+
+		if err := group.Run(); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
--- a/server/services/prometheus/intrumentor.go
+++ b/server/services/prometheus/intrumentor.go
@@ -20,7 +20,7 @@ func NewInstrumentor(version string) *Instrumentor {
 // ExposeBuildInfo exposes a gauge in prometheus for build info
 func (i *Instrumentor) ExposeBuildInfo() {
 	promauto.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "focalboard_build_nfo",
+		Name: "focalboard_build_info",
 		Help: "Build information of Focalboard",
 	}, []string{"Version"},
 	).WithLabelValues(i.Version).Set(1)
--- a/server/services/prometheus/intrumentor.go
+++ b/server/services/prometheus/intrumentor.go
@@ -1,0 +1,27 @@
+package prometheus
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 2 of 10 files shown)


spirosoik:
Few important links to read about :

- `group.Add`: https://blog.gopheracademy.com/advent-2017/run-group/, https://www.youtube.com/watch?v=LHe1Cb_Ud_M&ab_channel=TheGoProgrammingLanguage

- `prometheus build info`:  https://www.robustperception.io/exposing-the-software-version-to-prometheus, https://www.robustperception.io/why-info-style-metrics-have-a-value-of-1

chenilim (approved):
Thank you @spirosoik! This looks great! Minor thing: We should add the `prometheus_address` setting to the [server admin doc page](https://github.com/mattermost/focalboard/blob/main/website/site/content/guide/admin/_index.md), with a note to leave it blank to disable prometheus (@harshilsharma63, we can do this after the merge).

A couple questions:
* Should prometheus be in all editions? (Is it E20-only for Mattermost?) @sbishel, do you know?
* Should it be enabled by default? What do you think, @sbishel?

spirosoik:
@chenilim just added in the admin docs the extra cnfig

server/server/server.go
```diff
@@ -209,6 +215,23 @@ func (s *Server) Start() error {
 		s.telemetry.RunTelemetryJob(firstRun)
 	}
 
+	var group run.Group
+	if s.config.PrometheusAddress != "" {
```
harshilsharma63, line 219: We should check for `s.config.Telemetry` as well since otherwise we'll not be recording metrices.

server/services/prometheus/intrumentor.go
```diff
@@ -0,0 +1,27 @@
...
+func NewInstrumentor(version string) *Instrumentor {
+	return &Instrumentor{
+		Version: version,
+	}
+}
+
+// ExposeBuildInfo exposes a gauge in prometheus for build info
+func (i *Instrumentor) ExposeBuildInfo() {
+	promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "focalboard_build_nfo",
```
harshilsharma63, line 23: Minor typo,

```suggestion
		Name: "focalboard_build_info",
```

server/services/prometheus/intrumentor.go
```diff
@@ -0,0 +1,27 @@
...
+		Help: "Build information of Focalboard",
+	}, []string{"Version"},
+	).WithLabelValues(i.Version).Set(1)
```
harshilsharma63, line 26: I'm not sure if I entirely understand why we're setting the gauge to `1`, but should be use `.Inc()` instead to increment the gauge?

harshilsharma63 (changes requested):

harshilsharma63:
@spirosoik looks awesome! Just some very minor comments.

server/services/prometheus/intrumentor.go
spirosoik: actually we are not increasing right? we just pass the latest version

server/services/prometheus/intrumentor.go
spirosoik: good catch thanks :)

server/server/server.go
spirosoik: could you elaborate? why telemetry is related here?

harshilsharma63:
> Thank you @spirosoik! This looks great! Minor thing: We should add the `prometheus_address` setting to the [server admin doc page](https://github.com/mattermost/focalboard/blob/main/website/site/content/guide/admin/_index.md), with a note to leave it blank to disable prometheus (@harshilsharma63, we can do this after the merge).
> 
> A couple questions:
> 
> * Should prometheus be in all editions? (Is it E20-only for Mattermost?) @sbishel, do you know?
> * Should it be enabled by default? What do you think, @sbishel?

In Mattermost we have metrics enabled by default for both teams and enterprise edition.

server/server/server.go
harshilsharma63: Information sent to Prometheus is about customer's configuration, so I think falls under the tag of tracking metric. We do the same in Mattermost.

server/services/prometheus/intrumentor.go
harshilsharma63: Correct me if I'm wrong here, but you're creating a gauge for how many Focalboard instances are running a specific version, right? I might be entirely wrong here though.

I'm just not sure why the gauge value is being set to 1.

server/server/server.go
spirosoik: good point though prometheus is about operating focalboard right? Telemetry is more analytics oriented right?  @sbishel thoughts?

server/services/prometheus/intrumentor.go
spirosoik: I see your confusion.

> but you're creating a gauge for how many Focalboard instances are running a specific version, right?

No, I am just exposing the version so we can show for example in grafana that these metrics are about X version.

server/server/server.go
spirosoik: up to you guys how we can treat this to avoid confusion for the customer.

server/services/prometheus/intrumentor.go
harshilsharma63: @spirosoik got it.

server/server/server.go
sbishel: I think telemetry and metrics are separate things. Metrics are about performance and telemetry is more about configuration settings. We probably also want the ability to "turn off" metrics. (ie. "enableMetrics") separately from the "PrometheusAddress". Also we should give this a generic name rather than "Prometheus".,
https://docs.mattermost.com/deployment/metrics.html
https://docs.mattermost.com/administration/telemetry.html

sbishel:
> Thank you @spirosoik! This looks great! Minor thing: We should add the `prometheus_address` setting to the [server admin doc page](https://github.com/mattermost/focalboard/blob/main/website/site/content/guide/admin/_index.md), with a note to leave it blank to disable prometheus (@harshilsharma63, we can do this after the merge).
> 
> A couple questions:
> 
> * Should prometheus be in all editions? (Is it E20-only for Mattermost?) @sbishel, do you know?
> * Should it be enabled by default? What do you think, @sbishel?

Metrics are E20 only, we will have to discuss for focalboard. Telemetry is enabled by default.

server/server/server.go
spirosoik: agree. it can be `EnableMetrics`

spirosoik: @sbishel @harshilsharma63 let me know if we will rename it so I can quickly commit the change.

agnivade:
Just wanted to mention that this is okay for now, but it will probably be need to be integrated with the suite once Focalboard starts to behave as a product.

These are some of the common things that need to be shared. It will be difficult if every product starts to open it's own port and scraping endpoint.

spirosoik:
@agnivade agree as long as it's separate process it's ok. Ideally we will expose a `/metrics` per plugin maybe?

agnivade:
To clarify, these are products, not plugins. They just reside in different codebases, but they all give the appearance of being under a single suite. So all products will use the suite API to expose their metrics, and all of it will be just under a single `/metrics` endpoint with different namespaces.

One of our main objectives is to "trojan-horse" these products under the umbrella of Mattermost. If every product starts to have their own metrics endpoint, then the customer needs to change their infrastructure to get the additional metrics, which is not ideal.

So, this PR is good to go for now. But in the near future, this will be consumed by the suite API.

spirosoik:
@agnivade sure I got your point. 

> So, this PR is good to go for now. But in the near future, this will be consumed by the suite API.

yeah exactly defining all the metrics under one endpoint is the way to and we will do the separation part in our side for alerting and grafana dashboards. Thanks for info sharing :)

spirosoik:
@sbishel @chenilim @harshilsharma63 I will merge

harshilsharma63 (approved):

harshilsharma63:
@spirosoik I just approved this. LGTM!

Issue #110, opened 2021-03-18
Feature request: Multiple Select property type


Victor239:
Currently there's only a (Single) Select property type, a Multiple Select property would instead allow you to select multiple of the available options.

chenilim:
Thanks @Victor239! Yup, this is something we expect to support. Have to think about how to best represent this in the object model, i.e. would it be better to represent all select properties as multi-select, but limit some to just one in the UI? This affects the filter and display logic.

Victor239:
Airtable shows them all the options selected in the cell, but it's not wrapped so you might have to adjust the column width manually to show them all.

alerque:
Multi-select appears to be implemented already and this issue status is stale. I believe there is a related issue somewhere about grouping by multi-select fields, but I can't find it. Perhaps that was in one of the project roadmap boards...

PR #415, opened 2021-05-14
Implement Multi Select Values

hahmadia:
#### Summary

Implement multi select values

#### Ticket Link

Fixes GH-110

image

image

image

image

image

### Requirements:
* [x] Able to add a multi-select property type
* [x] Able to set multiple values on a multi-select property type
* [ ] Able to filter by multi-select property types
* [x] Multi-select property types function in all views


chenilim:
Minor thing: add a margin property to the multi-select values in the Card Summary and read-only card detail view. Right now the values appear joined together:
multiselect properties

This can be done in a subsequent commit.

webapp/src/blocks/card.ts
```diff
@@ -7,7 +7,7 @@ import {IBlock, MutableBlock} from './block'
 interface Card extends IBlock {
     readonly icon: string
     readonly isTemplate: boolean
-    readonly properties: Readonly<Record<string, string>>
+    readonly properties: Readonly<Record<string, string | string[]>>
```
chenilim, line 10: One thing to be careful about is supporting changing the type of a property from Select <> Mult-select. Going from single-select to multi-select should appear seamless to the user (though there may be quite involved conversion code).

Going from multi-select to single-select should retain the options, but make the value the first value in the multi-select list.

harshilsharma63: @chenilim on the same topic, I think we should also show a confirmation model when we convert property types between incompatible types. For example, a `person` type to an `updated at` type. This will avoid any surprises and data loss.

harshilsharma63:
It doesn't look very clear to me if an `x` removes the chip on the left to it or the right to it. 

Screenshot 2021-05-18 at 3 42 22 PM

We already have this component in Compass Components actually and we can tweak the CSS to bring that `x` inside the chip - https://zeroheight.com/29be2c109/p/02dffc-chips

This need a design decision though, cc @michaelgamble

webapp/src/blocks/card.ts
harshilsharma63: When switching between `Single Select` and `Multi Select`, should we be retain the options? Right now we're not.

webapp/src/components/propertyValueElement.tsx
```diff
@@ -56,6 +56,58 @@ const PropertyValueElement = (props:Props): JSX.Element => {
         }
     }
 
+    if (propertyTemplate.type === 'multiSelect') {
```
harshilsharma63, line 59: It would be cleaner to have an individual React component for each property type, like I did here - https://github.com/mattermost/focalboard/pull/393

It's a team call though.

chenilim: You guys can make the call, but my advice is to refactor later. i.e. either check in new functionality, or refactor existing functionality. If the change is relatively small or contained, then go ahead and refactor now, otherwise my advice is to do it in a subsequent commit.

webapp/src/blocks/card.ts
chenilim: Yes, to clarify:
* When changing from single-select to multi-select: The options and selected value should remain the same (only change is the user can specify additional values).
* When changing from multi-select to single-select: Retain the options, but make the value the first value in the multi-select list.

I'll leave it to @hahmadia, but both of these can be done in a subsequent checkin if necessary.

hahmadia:
Added margins so it doesn't look stuck
image

sbishel:
A couple things when comparing to Notion. As suggested above. They have the remove "x" within the coloring. 

Also, they display all the options even the ones selected. This allows you to change colors or edit one that is selected. 
Screen Shot 2021-05-21 at 3 48 30 PM

Screen Shot 2021-05-21 at 3 54 09 PM

sbishel (changes requested):
A couple things when comparing to Notion. As suggested above. They have the remove "x" within the coloring. 

Also, they display all the options even the ones selected. This allows you to change colors or edit one that is selected. They also leave the drop down open rather than closing after each selection. 

Screen Shot 2021-05-21 at 3 54 09 PM

Screen Shot 2021-05-21 at 3 48 30 PM

Also the text and button should be center and even with each other.
Screen Shot 2021-05-21 at 3 45 57 PM

webapp/src/blocks/card.ts
hahmadia: Done

webapp/src/components/propertyValueElement.tsx
hahmadia: Done

hahmadia:
To-Do:
- X within label
- Keep menu open after choosing an option
- Text and three dots be aligned in dropdown
- Add test for PropertyValueElement

harshilsharma63:
Some issues I found -

1. The dropdown container doesn't expand all the way to contain the option list if the options are more than a few values. This happens for both single and multi select so can be fixed separate from this ticket.
Screenshot 2021-05-26 at 12 13 26 PM

1. There is no way to delete a really long value. It overflows from the card, making the `x` disappear -
Screenshot 2021-05-26 at 12 15 29 PM

One solution is to truncate the text to a max-length and show the full text as a tooltip on hover-
Screenshot 2021-05-26 at 12 20 34 PM

This is also an existing issue and can be seen in single select, we can fix it separately.

harshilsharma63 (changes requested):

webapp/src/viewModel/boardTree.ts
```diff
@@ -216,7 +216,9 @@ class MutableBoardTree implements BoardTree {
                         if (option?.value.toLowerCase().includes(searchText)) {
                             return true
                         }
-                    } else if (propertyValue.toLowerCase().includes(searchText)) {
+
+                    // TODO: Add search capability for multi-select values BIG BOYY
```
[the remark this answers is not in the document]
hahmadia: I was planning to do it as a separate PR excuse my PR comment lol. I was hoping to get the initial implementation in and worry about the searching separate as a way to not to make PR too big.

webapp/src/components/propertyValueElement.tsx
```diff
@@ -56,6 +59,53 @@ const PropertyValueElement = (props:Props): JSX.Element => {
...
+        const values = Array.isArray(propertyValue) ?
```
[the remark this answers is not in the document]
hahmadia: Good call. I've made more abstractions.

webapp/src/mutator.ts
```diff
@@ -392,10 +396,33 @@ class Mutator {
...
+                    newBlocks.push(newCard)
+                    oldBlocks.push(card)
+                }
+                newTemplate.options = propertyTemplate.options
+            }
+        } else if (propertyTemplate.type === 'select' || propertyTemplate.type === 'multiSelect') { // If the old type was either select or multiselect
             // Map select to their values
             for (const card of boardTree.allCards) {
-                const oldValue = card.properties[propertyTemplate.id]
+                const oldValue = Array.isArray(card.properties[propertyTemplate.id]) ? card.properties[propertyTemplate.id][0] : card.properties[propertyTemplate.id]
```
sbishel, line 425: Do you need to check length here, as done above?

sbishel (changes requested):
I noticed this issue in table view. 

Screen Shot 2021-05-26 at 4 22 41 PM

webapp/src/mutator.ts
hahmadia: Yep! I don't think it hurts to have it. I believe I missed it here. Good catch.

hahmadia: This changed helped me realize and removed the entire if block above!

hahmadia:
@harshilsharma63 

Thanks for the help. I've addressed the long text issue with your help. I also applied our fix to card view because i noticed it overflows in this view as well. Thanks for that.
image


@sbishel 
I've addressed all the concerns except the one where the row is slightly larger due to other ones. It seems to be caused due to padding in between the options and as well as the icon. Do you feel it must be fixed? I personally think it's fine having a slightly bigger row when you have multi select option and don't feel it takes away from anything. 

No longer see the white spacing stuff as shown in your photo above
image


Thanks!

sbishel:
> @sbishel
> I've addressed all the concerns except the one where the row is slightly larger due to other ones. It seems to be caused due to padding in between the options and as well as the icon. Do you feel it must be fixed? I personally think it's fine having a slightly bigger row when you have multi select option and don't feel it takes away from anything.

Yes, I think this should be fixed. I find it distracting and sloppy. My guess is you will need to different "padding" or "margin" settings when displayed in table view to keep it the same size as a single select column.

hahmadia:
> > @sbishel
> > I've addressed all the concerns except the one where the row is slightly larger due to other ones. It seems to be caused due to padding in between the options and as well as the icon. Do you feel it must be fixed? I personally think it's fine having a slightly bigger row when you have multi select option and don't feel it takes away from anything.
> 
> Yes, I think this should be fixed. I find it distracting and sloppy. My guess is you will need to different "padding" or "margin" settings when displayed in table view to keep it the same size as a single select column.

Hi Scott,

Feel free to check now. The table height changing with multiselect should be fixed now. Let me know what you think.

harshilsharma63 (approved):
LGTM 👍

hahmadia:
I've created a card for testing to be done in the future.

https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=baa742b3-f3f1-406e-811a-00f18d723781&c=d43904fb-8d02-4a5b-8892-e9176a1d1193

sbishel (approved):

sbishel:
You may want to update the screenshots, since it has changed.

chenilim (approved):
I did a quick smoke test on a private build, and it looks great! 👍

hahmadia:
@hahmadia 

a note to self:

Update original post to include updated screenshots of work!

PR #422, opened 2021-05-17
add unit tests

sbishel:
#### Summary
Adds Unit tests for ./server/app/auth.go. 
Adds Unit tests for ./server/auth/auth.go. 
These will be used as samples for other server tests.

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-server/issues/421


sbishel:
These will be used as samples for remaining unit test tickets.

harshilsharma63 (approved):

server/app/helper.go
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
jespino, line 1: This helper file is compiling everything inside the app package when is not compiling the tests, so... we are importing here, in the production binary the mocks a some not-needed code.

server/app/auth_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}{
+		{"fail, missing login information", "", "", "", "", true},
+		{"fail, invalid username", "badUsername", "", "", "", true},
+		{"fail, invalid email", "", "badEmail", "", "", true},
+		{"fail, invalid password", "testUsername", "", "badPassword", "", true},
+		{"success, using username", "testUsername", "", "testPassword", "", false},
+		{"success, using email", "", "testEmail", "testPassword", "", false},
+	}
+
+	th.Store.EXPECT().GetUserByUsername(gomock.Eq("badUsername")).Return(nil, errors.New("Bad Username"))
```
jespino, line 41: I'm not sure (I'll check), but I think if you are expecting one specific parameter you can use just `...GetUserByUsername("badUsername")...`

jespino (approved):
Now LGTM.

PR #445, opened 2021-05-19
Add moment as a direct dependency

mgdelacroix:
#### Summary
As part of the `EditableDayPicker` component we're requiring `moment`, which was available as a transitive dependency of `cypress`. This PR makes the dependency explicit and enables building the release pipeline with `npm install --no-optional`, which was failing.


chenilim (approved):
LGTM! Thanks!

harshilsharma63 (approved):
LGTM :+1:

Issue #443, opened 2021-05-18
Bug: App fails to open on dock icon click


jgilliam17:
**Summary:**
When the app is closed using x, it can't be brought back to forefront by clicking on the dock icon

**Steps to reproduce the behavior:**
1. Open the app, dismiss welcome window if present 
2. Close the app by click on the X
3. Click on the dock icon to open again

**Expected behavior:**
App can be brought to forefront by clicking on the dock icon, as the dot indicates that it is still open

**Screenshots:**
<img width="662" alt="Screen Shot 2021-05-18 at 9 11 56 AM" src="https://user-images.githubusercontent.com/52937121/118702156-8feb8180-b7e2-11eb-9440-f550e02d9d09.png">

**Platform:**
 - Focalboard Version 0.6.7 (12)
 - OS and version: macOS 10.15.7 Catalina and macOS 11.3.1 Big Sur

**Additional context:**
App is not installed, opening from unzipped downloaded file. 
Reproduces on App Store installed version 0.6.5
Work-around: File>New opens the app back up, although "new" is misleading as it's opening existing boards

haardikdharma10:
Able to reproduce 👍

chenilim:
Thanks for reporting this @jgilliam17! Setting it as up for grabs. Desired behavior when clicking the dock icon is:
* Bring the first open window to the top, if there is one, or
* Open a new window

PR #451, opened 2021-05-20
[GH-443] Create new window when reopening without existing windows

Johennes:
#### Summary

This change intercepts reopen events and creates a new window in case no windows exist.

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/443

### DIFF
--- a/mac/Focalboard/AppDelegate.swift
+++ b/mac/Focalboard/AppDelegate.swift
@@ -26,6 +26,13 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 		NotificationCenter.default.post(name: AppDelegate.serverStartedNotification, object: nil)
 	}
 
+	func applicationShouldHandleReopen(_ sender: NSApplication, hasVisibleWindows flag: Bool) -> Bool {
+		if !flag {
+			openNewWindow(nil)
+		}
+		return true
+	}
+
 	func applicationWillTerminate(_ aNotification: Notification) {
 		stopServer()
 	}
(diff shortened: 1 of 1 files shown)


mac/Focalboard/AppDelegate.swift
```diff
@@ -26,6 +26,13 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 		NotificationCenter.default.post(name: AppDelegate.serverStartedNotification, object: nil)
 	}
 
+	func applicationShouldHandleReopen(_ sender: NSApplication, hasVisibleWindows flag: Bool) -> Bool {
+		if !flag {
+			openNewWindow(nil)
```
chenilim, line 31: If I understand [the docs](https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428638-applicationshouldhandlereopen?language=objc) correctly, it sounds like we should return false if we handle the open ourselves?

chenilim (commented):
Thanks @Johennes! Only minor thing is per the docs, it appears we should return false if we handle the reopen. It shouldn't make a difference, but would be more future proof?

mac/Focalboard/AppDelegate.swift
Johennes: Hm, I think you're right, yeah. I'm away for the weekend but I'll make the changes next week.

Johennes: Made the change.

chenilim (approved):
Thank you @Johennes! 👍

PR #452, opened 2021-05-21
[GH-410] Remove background from read-only input fields

### LINKED ISSUE #410: Bug: Title cell in read-only table view should not appear to be editable
1. Share a board with a table view
2. Click on the title of a row in the read-only table view
3. It appears to be editable (but is not)

**Expected behavior:**
It should not appear to be editable. Perhaps removing the white border around the input field would be sufficient?

**Screenshots:**
table-view readonly editable

**Platform:**
 - Chrome 89 on macOS 10.15.7

Johennes:
#### Summary

This is an attempt to fix #410 by making the background of text input fields transparent when the field is read-only. Making it transparent will make the parent's background show through which works in conjunction with table cells which change background on hover.

Alternatively, we could also set a static color but I found this solution to work nicely irrespective of the parent's background.

Screenshot 2021-05-21 at 19 41 28   Screenshot 2021-05-21 at 19 41 46

#### Ticket Link

Fixes #410

### DIFF
--- a/webapp/src/components/content/__snapshots__/checkboxElement.test.tsx.snap
+++ b/webapp/src/components/content/__snapshots__/checkboxElement.test.tsx.snap
@@ -56,7 +56,7 @@ exports[`components/content/CheckboxElement should match snapshot on read only 1
       value="off"
     />
     <input
-      class="Editable undefined"
+      class="Editable readonly undefined"
       placeholder="Edit text..."
       readonly=""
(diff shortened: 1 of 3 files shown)


chenilim:
Thanks @Johennes! I tested this change, and it works great! Adding @sbishel to get another set of eyes on it.

sbishel (approved):
Works good, looks good.

Issue #458, opened 2021-05-24
GitHub Actions - Use Node 16


harshilsharma63:
Start using Node v16.1 in GitHub actions, which the same version specified in `.nvmrc`.

harshilsharma63:
This is now done.

PR #459, opened 2021-05-24
Using Node v16.0.1 in CI

harshilsharma63:
#### Summary
Started using Node v16 in GitHub actions, which is the Node version used in web app.

#### Ticket Link
https://github.com/mattermost/focalboard/issues/458

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,11 @@ jobs:
       with:
         go-version: 1.15
 
+    - name: Setup Node
+      uses: actions/setup-node@v2
+      with:
+        node-version: 16.1.0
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -26,6 +26,11 @@ jobs:
       with:
         go-version: 1.15
 
+    - name: Setup Node
```
cpanato, line 29: I think you need to do the same for the `build-*.yml` workflows as well

besides that LGTM

harshilsharma63: Done.

cpanato (approved):
/lgtm

PR #460, opened 2021-05-24
Only injecting the session token when the href match the expected localhost

jespino:


linux/main.go
```diff
@@ -100,10 +100,15 @@ func main() {
 	w.SetTitle("Focalboard")
 	w.SetSize(1024, 768, webview.HintNone)
 
-	script := fmt.Sprintf("localStorage.setItem('focalboardSessionId', '%s');", sessionToken)
+	baseUrl := fmt.Sprintf("http://localhost:%d", port)
+	w.Navigate(baseUrl)
+	script := fmt.Sprintf(`
+		if (window.location.href.startsWith('%s')) {
```
jespino, line 106: @jupenur I think this solves the bug for the linux client, but let me know if you see any problem.

jupenur: @jespino a couple of questions I can't immediately find the answers for:

 1. Is this script guaranteed to execute before any other scripts on a page? The godoc for webview says "before onload", but that's too late if that's the case
 2. Is the script executed in an isolated world or the default script world?
 3. Is the script executed synchronously?

These all impact security in a major way. If the script is not isolated and the answer to either (1) or (3) is "no", then this approach won't cut it because a malicious page could fiddle with prototypes.

The same questions also apply to #465

linux/main.go
```diff
@@ -97,10 +97,15 @@ func main() {
 	w.SetTitle("Focalboard")
 	w.SetSize(1024, 768, webview.HintNone)
 
-	script := fmt.Sprintf("localStorage.setItem('focalboardSessionId', '%s');", sessionToken)
+	baseUrl := fmt.Sprintf("http://localhost:%d", port)
```
jupenur, line 100: `baseUrl` needs a trailing slash to make sure the port number can't have extra digits

Issue #408, opened 2021-05-12
Feature Idea: Grouped Table / List view


chenilim:
**Summary:**
Allow the current Table view to show rows grouped by a property. This will provide similar and improved functionality to the "List" views in reference products.

**How important this is to me and why:**

Importance: Medium

Use cases:
* Allow Table views to group-by a selected property
* Drag and drop rows between groups to change properties
* Drag and drop the order of groups
* Sort by property (columns) within groups

**Additional context / similar features:**
Many other products have a "List" view, but adding a "group-by" capability to the existing Table view might provide a better set of capabilities.

What do you think? Would this meet most of your needs for a "list"? Or does a plain list offer additional functionality to you?

PR #447, opened 2021-05-20
Feature/adjust board component scroll

harshilsharma63:
This PR is a continuation of another PR - https://github.com/mattermost/focalboard/pull/325 . This contains the other PR's changes along with the fixes for cypress tests and a few more enhancements.

I've made two improvements over the work already done by @signalwerk 

The table header in table view and swimlane header in Kannan board also stay in view now when scrolling.

**Kanban View**
kanban

**Table View**
table

cc @signalwerk

PR #463, opened 2021-05-24
GH-408 Implement Table Group

sbishel:
#### Summary
Implements table grouping, being able to group by a field in a table view.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/408

Screen Shot 2021-05-27 at 3 01 07 PM


harshilsharma63 (commented):
@sbishel some observations-

1. We don't have a way to un-group items once they are grouped. I think we should have the option to ungroup as well.
1. After syncing with latest `main`, there is a gap above the second row of each group, and the no group header is not visible. I believe it's from this PR of mine - https://github.com/mattermost/focalboard/pull/447
Screenshot 2021-06-01 at 5 35 11 PM
1. When dragging and dropping items in table view with grouping, it feels a bit wierd. When drag-and-droping from one group to other, I feel like I'm moving the card to a new group but it's actually swapping.

hahmadia (commented):
Some quick observations:

An odd gap betwen first item and second item of categories:
image

Click the status does nothing (I expect them to ungroup).
image

Dragging a category to the top made it go off screen (drag to the same row as the name row) image

sbishel:
> 1. We don't have a way to un-group items once they are grouped. I think we should have the option to ungroup as well.
> 2. After syncing with latest `main`, there is a gap above the second row of each group, and the no group header is not visible. I believe it's from this PR of mine - #447

I have updated. I added "ungroup" and also reworked the table based on the fixed row that got introduced.

sbishel:
> 1. When dragging and dropping items in table view with grouping, it feels a bit wierd. When drag-and-droping from one group to other, I feel like I'm moving the card to a new group but it's actually swapping.

I'm not sure I follow. If dragging a "group" row, you are rearranging the "groups". Essentially groups are always manually sorted. When dragging a card to another group, it does (should) change groups. If you have manual sort it should drop either above or below the card depending on if dragged up or down. If a card is dropped on a group header it should be dropped at the top. It should work the same as without grouping and manual sort. I'm not sure what you mean by it swaps.

sbishel:
@hahmadia I think the issues you found are also handled by the above fixes.

harshilsharma63:
> > 1. When dragging and dropping items in table view with grouping, it feels a bit wierd. When drag-and-droping from one group to other, I feel like I'm moving the card to a new group but it's actually swapping.
> 
> I'm not sure I follow. If dragging a "group" row, you are rearranging the "groups". Essentially groups are always manually sorted. When dragging a card to another group, it does (should) change groups. If you have manual sort it should drop either above or below the card depending on if dragged up or down. If a card is dropped on a group header it should be dropped at the top. It should work the same as without grouping and manual sort. I'm not sure what you mean by it swaps.

Here are the steps -

1. Create a select type property, lets say `Owner`. Assign some values to some cards for this property.
2. Switch to table view and group by `Owner`.
3. Delete the `Owner` property. The grouping will switch to  by `Status`.
4. Assign value to `Status` to some cards. The grouping will look perfectly fine at this point.
5. Now drag-and-drop a card from `No Status` group to some other group. If that other group already had a card in it, the card will not drag into it. If that group was empty, one card will successfully drop into it but the second won't.
6. Also, drawing and dropping from one group to other won't work.

This isn't resolved by refreshing and the only way to get it working is to manually set the Group By to status.

https://user-images.githubusercontent.com/18575143/120614017-697d4580-c474-11eb-8cc8-06437bcb7e5a.mp4

I think the issue is with the auto change of group by not being saved when the group by property is deleted.

hahmadia (commented):
Hi Scott,

Concerns from above are fixed. Some more observations.

Clicking enter when editing last row in category spawns a new row. Clicking outside however does not. I believe this should be a general bug in table view but I thought I point out.
image

2. I was also able to reproduce Harshil's bug. After I delete property, I can drag items into another category if I drag onto the header but not if i drag onto other items. Dragging onto other items causes it to go to no status. As Harshil stated, refreshing did not work and basically no matter what I did I could not revert it to the post property delete behavior unless I started a fresh board,

https://user-images.githubusercontent.com/17804942/120667725-c3bddc80-c45b-11eb-96a5-55f8577e6137.mp4


PS. I just figured we can upload mp4 on github and don't need to convert to GIF!!!

sbishel:
>> Clicking enter when editing last row in category spawns a new row. Clicking outside however does not. I believe this should be a general bug in table view but I thought I point out.

This appears to be a feature...
https://github.com/mattermost/focalboard/blob/019dd3da8f5c374d7d9f2a94e2569c25d64a41d6/webapp/src/components/table/table.tsx#L206

harshilsharma63 (approved):
LGTM!

hahmadia (approved):
Nothing that stood out to me. Nice job! I like this feature a lot 👍

PR #465, opened 2021-05-24
MM-35819 mac win: Only injecting the session token when the href match the expected localhost

chenilim:
Similar to #460 (for Linux), this adds the same check for Mac and Windows.

### DIFF
--- a/mac/Focalboard/ViewController.swift
+++ b/mac/Focalboard/ViewController.swift
@@ -107,8 +107,14 @@ class ViewController:
 
 	private func updateSessionTokenAndUserSettings() {
 		let appDelegate = NSApplication.shared.delegate as! AppDelegate
+		let url = URL(string: "http://localhost:\(appDelegate.serverPort)/")!
+		let sessionScript = """
+		if (window.location.href.toLowerCase().startsWith('\(url)'.toLowerCase())) {
+			localStorage.setItem('focalboardSessionId', '\(appDelegate.sessionToken)');
+		}
+		"""
 		let sessionTokenScript = WKUserScript(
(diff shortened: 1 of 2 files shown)


mac/Focalboard/ViewController.swift
```diff
@@ -107,8 +107,14 @@ class ViewController:
 
 	private func updateSessionTokenAndUserSettings() {
 		let appDelegate = NSApplication.shared.delegate as! AppDelegate
+		let url = URL(string: "http://localhost:\(appDelegate.serverPort)/")!
+		let sessionScript = """
+		if (window.location.href.toLowerCase().startsWith('\(url)'.toLowerCase())) {
```
chenilim, line 112: The .toLowerCase() should not strictly be needed, but is there for future-proofness.

mac/Focalboard/ViewController.swift
```diff
@@ -107,8 +107,14 @@ class ViewController:
 
 	private func updateSessionTokenAndUserSettings() {
 		let appDelegate = NSApplication.shared.delegate as! AppDelegate
+		let url = URL(string: "http://localhost:\(appDelegate.serverPort)/")!
```
Johennes, line 110: This should currently not be a practical problem but I think you don't actually have to create a `URL` object since the only place you're using the result is in the `"...\(url)..."` string interpolation below. I believe what happens in the interpolation is it converts the `URL` object to its `description` property (which happens to be the URL as a `String`). To prevent that implicit conversion, you could just use `let url = "http://localhost:\(appDelegate.serverPort)/"`.

PR #471, opened 2021-05-25
Adds plugin actions to the project's CI

mgdelacroix:
#### Summary

This PR adds a github action manually triggered to compile and upload the artifact for the Focalboard plugin.


.github/workflows/plugin-release.yml
```diff
@@ -0,0 +1,48 @@
+name: Plugin-Release
+
+on: pull_request
```
mgdelacroix, line 3: This will change to `on: workflow_dispatch`, keeping it on `pull_request` temporarily for the CI to trigger the jobs

mattermost-plugin/plugin.json
```diff
@@ -6,8 +6,8 @@
     "support_url": "https://github.com/mattermost/mattermost-plugin-focalboard/issues",
     "release_notes_url": "https://github.com/mattermost/mattermost-plugin-focalboard/releases/tag/v0.1.0",
     "icon_path": "assets/starter-template-icon.svg",
-    "version": "0.2.0",
-    "min_server_version": "5.12.0",
+    "version": "0.6.7",
+    "min_server_version": "5.35.0",
```
jespino, line 10: It actually should be 5.36

mgdelacroix: Right! server `master` has currently `5.35` so I left it here for testing. Will update!

mgdelacroix: Done in e14730d

.github/workflows/plugin-release.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: npm install plugin dependencies
+      run: cd mattermost-plugin/webapp; npm install --no-optional
+
+    - name: Build plugin
+      run: cd mattermost-plugin; make dist
+      env:
+        BUILD_NUMBER: ${{ github.run_id }}
+
+    - name: Rename plugin file
+      run: cd mattermost-plugin/dist; mv focalboard-0.6.7.tar.gz mattermost-plugin-focalboard-0.6.7.tar.gz
```
chenilim, line 45: This probably should not have the version number hardcoded in the file name. Is the version necessary for the plugin infrastructure? If so, that's ok for now. Longer-term, we will need to consolidate our version numbers anyway.

jespino (approved):
LGTM

.github/workflows/plugin-release.yml
mgdelacroix: I was hesitant to add it at first, but all the plugins are released with their version numbers. What we can do is to abstract this job to generate an artifact without the version number, but my understanding is that we would have to rename the artifact before uploading it.

I'll make a quick test on how the no-version-number approach would look like, and we can iterate over it

mgdelacroix: Addressed in c58156c

.github/workflows/plugin-release.yml
```diff
@@ -0,0 +1,51 @@
+name: Plugin-Release
+
+on: pull_request
+
+jobs:
+
+  ubuntu:
```
chenilim, line 7: The job should have a different name, e.g. 'plugin'

mgdelacroix: Addressed in f17112c

chenilim (approved):
LGTM 👍

PR #486, opened 2021-05-27
Fixed Docker build breaking due to webpack-image=loader

harshilsharma63:
#### Summary
Fixed Docker build failure due to some missing optional dependencies.

Certain dependencies of `webpack-image-loader` are marked as optional dependencies, which we don't install during Docker build.

I've specified those as Focalboard's direct dependencies.


cpanato (approved):
thanks for a quick fix!
/lgtm

Issue #418, opened 2021-05-16
Bug: Text search clears when card is updated


chenilim:
**Steps to reproduce the behavior:**
1. Open a table view with multiple cards
2. Enter text in the text search box, such that some cards are filtered out
3. Open a card
4. Update it (e.g. a property or title)
5. The table updates in the background, showing all cards, even though the search text field is still populated

**Expected behavior:**
The table should update, but still maintain the text filter.

**Platform:**
 - Chrome 89 on macOS 10.15.7

PR #508, opened 2021-06-01
GH-418: Fix Resetting Search after adding new Property

hahmadia:
#### Summary

Search restting after adding a new property FIXED!

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/418

### DIFF
--- a/webapp/src/components/viewHeader/viewHeaderSearch.tsx
+++ b/webapp/src/components/viewHeader/viewHeaderSearch.tsx
@@ -14,20 +14,16 @@ type Props = {
 }
 
 const ViewHeaderSearch = (props: Props) => {
-    const {boardTree} = props
+    const {boardTree, setSearchText} = props
     const intl = useIntl()
 
     const searchFieldRef = useRef<{focus(selectAll?: boolean): void}>(null)
-    const [isSearching, setIsSearching] = useState(Boolean(props.boardTree.getSearchText()))
+    const [isSearching, setIsSearching] = useState(Boolean(boardTree.getSearchText()))
     const [searchValue, setSearchValue] = useState(boardTree.getSearchText())
 
     useEffect(() => {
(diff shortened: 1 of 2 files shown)


webapp/src/components/viewHeader/viewHeaderSearch.tsx
```diff
@@ -14,21 +14,17 @@ type Props = {
...
     const [searchValue, setSearchValue] = useState(boardTree.getSearchText())
 
     useEffect(() => {
         searchFieldRef.current?.focus()
     }, [isSearching])
 
-    useEffect(() => {
-        setSearchValue(boardTree.getSearchText())
-    }, [boardTree])
-
```
hahmadia, line 31: Note:

I removed this because this did nothing. Search doesn't get triggered if you don't do `setSearchText` and setting this would just temporarily set state but wouldn't actually do anything. 

Feel free to correct me if I'm wrong but I came to conclusion we don't need this.

harshilsharma63: This seems to be assigned the changed value coming from `Editable` back to `Editable`. But I see we're doing this everywhere we use `Editable`, so there could be an edge case this is covering. @jespino would know best.

I think we should be careful removing this.

hahmadia: Note to self:

Add reason why we keep this part in the code.

jespino: I think it is ok, lets merge it and see where it breaks. clearly makes more sense to remove that block than keeping it. And the new change on the incrementalUpdate, I think is correct from the data perspective, but I would like to investigate it a bit more later.

jespino (approved):
LGTM

Issue #519, opened 2021-06-03
Bug: Windows build failing due to Logr syslog dependency


wiggin77:
**Summary:**
Windows build failing due to Logr syslog dependency since syslog is not supported on Windows.

**Steps to reproduce the behavior:**
Build FocalBoard for Windows.

**Expected behavior:**
Build success

**Platform:**
 - Windows

The fix is to conditionally import syslog within Logr, and update Logr version used in FocalBoard.

wiggin77:
Fixed by https://github.com/mattermost/focalboard/pull/520

PR #520, opened 2021-06-03
Bump Logr version to fix Windows build

wiggin77:
#### Summary
Windows build was failing due to Logr syslog dependency since syslog is not supported on Windows. This PR fixes the issue by bumping the Logr version to the one that includes a fix.

#### Ticket Link
https://github.com/mattermost/focalboard/issues/519


cpanato (approved):
thanks Doug!

PR #536, opened 2021-06-07
Marked selected language and theme

### LINKED ISSUE #385: Feature Idea: Show a visual indicator next to the selected language & theme options
**Summary:**
There's currently no way to tell which theme or language is selected when looking at the list. It would be great to show a visual indicator, like a checkmark, next to the selected option.

Note in the screenshot that there's nothing to indicate which theme is currently selected

focalbard_theme

harshilsharma63:
#### Summary
Marked selected language and theme

#### Ticket Link
https://github.com/mattermost/focalboard/issues/385

Fixed #385 

#### Demo

**Language-**

https://user-images.githubusercontent.com/18575143/120996367-e32e6f80-c7a3-11eb-997e-69b1ab5a037a.mp4

**Theme-**
theme

### DIFF
--- a/webapp/src/components/sidebar/sidebarSettingsMenu.tsx
+++ b/webapp/src/components/sidebar/sidebarSettingsMenu.tsx
@@ -25,6 +25,7 @@ import CheckIcon from '../../widgets/icons/check'
 type Props = {
     setWhiteLogo: (whiteLogo: boolean) => void
     activeTheme: string
+    activeLanguage: string
 }
 
 const SidebarSettingsMenu = React.memo((props: Props) => {
@@ -35,6 +36,7 @@ const SidebarSettingsMenu = React.memo((props: Props) => {
     // on theme change. This can cause props and the actual
     // active theme can go out of sync
     const [themeName, setThemeName] = useState(props.activeTheme)
+    const [lang, setLang] = useState(props.activeLanguage)
 
     const updateTheme = (theme: Theme | null, name: string) => {
         const consolidatedTheme = setTheme(theme)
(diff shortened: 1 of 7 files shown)


webapp/src/theme.ts
```diff
@@ -128,12 +141,26 @@ export function setTheme(theme: Theme | null): Theme {
     return consolidatedTheme
 }
 
+function setActiveThemeName(consolidatedTheme: Theme, theme: Theme | null) {
+    if (theme === null) {
+        activeThemeName = systemThemeName
+    } else if (_.isEqual(consolidatedTheme, darkTheme)) {
```
jespino, line 147: Maybe is a bit overkill to add a whole new library to this tiny functionality.

webapp/src/theme.ts
```diff
@@ -2,6 +2,9 @@
 // See LICENSE.txt for license information.
 
 import {CSSObject} from '@emotion/serialize'
+import * as _ from 'lodash'
```
jespino, line 5: Also, I think this imports everything here, so the bundle size is bigger. Basically we are including all the lodash library to use 1 function. A better approach would be something like:

```js
import isEqual from 'lodash/isEqual'
```

this way is only importing the isEqual code, and the bundle size should be slightly smaller.

harshilsharma63: Done.

webapp/src/components/sidebar/sidebarSettingsMenu.tsx
```diff
@@ -32,6 +51,87 @@ const SidebarSettingsMenu = React.memo((props: Props) => {
         setRandomIcons(!randomIcons)
     }
 
+    const languages = [
```
jespino, line 54: this is nicer :)

harshilsharma63: 👍

webapp/src/components/sidebar/sidebarSettingsMenu.tsx
```diff
@@ -4,26 +4,45 @@ import React, {useContext, useState} from 'react'
...
 const SidebarSettingsMenu = React.memo((props: Props) => {
     const intl = useIntl()
     const setLanguage = useContext(SetLanguageContext)
 
-    const updateTheme = (theme: Theme | null) => {
+    // we need this as the sidebar doesn't always need to re-render
+    // on theme change. This can cause props and the actual
+    // active theme can go out of sync
+    const [themeName, setThemeName] = useState(props.activeTheme)
+    const [lang, setLang] = useState(props.activeLanguage)
```
jespino, line 39: maybe we can replace this with `useIntl` and use the `intl.locale` to identify the current locale. That way we don't need to recieve or track the the language.

harshilsharma63: On it.

harshilsharma63: Done.

webapp/src/theme.ts
harshilsharma63: Continuing using this as its simpler way, but importing only required function instead of the whole package.

jespino (approved):
LGTM

PR #539, opened 2021-06-07
Multi Select Filter + Sort

hahmadia:
#### Summary

Ensures Sorting + Filtering works for MultiSelect

#### Ticket Link

https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=baa742b3-f3f1-406e-811a-00f18d723781&c=d0e40adc-b8d1-41fb-874d-e1c95cc7e21d

image

image

image

image

image


harshilsharma63 (approved):
LGTM

webapp/src/cardFilter.ts
```diff
@@ -50,19 +50,19 @@ class CardFilter {
...
         case 'notIncludes': {
             if (filter.values.length < 1) {
                 break
             }		// No values = ignore clause (always met)
-            return (filter.values.find((cValue) => cValue === value) === undefined)
+            return (filter.values.find((cValue) => (Array.isArray(value) ? value.includes(cValue) : cValue === value)) === undefined)
         }
         case 'isEmpty': {
-            return !value
+            return value.length <= 0
```
jespino, line 62: What would happen with a checkbox, for example?

webapp/src/components/viewHeader/filterValue.tsx
```diff
@@ -31,7 +31,7 @@ const filterValue = (props: Props): JSX.Element|null => {
         displayValue = filter.values.map((id) => {
             const option = template.options.find((o) => o.id === id)
             return option?.value || '(Unknown)'
-        }).join(', ')
+        }).join(' or ')
```
jespino, line 34: This is going to generate problems with the translations (probably the previous one would generate the same kind of problems. We should file a ticket (probably a help wanted one) to see how we tackle this on translations.

webapp/src/viewModel/boardTree.ts
```diff
@@ -381,21 +385,21 @@ class MutableBoardTree implements BoardTree {
...
                         }
-                        if (!aValue && !bValue) {
+                        if (aValue.length <= 0 && bValue.length <= 0) {
                             return this.titleOrCreatedOrder(a, b)
                         }
 
                         // Sort by the option order (not alphabetically by value)
-                        const aOrder = template.options.findIndex((o) => o.id === aValue)
-                        const bOrder = template.options.findIndex((o) => o.id === bValue)
+                        const aOrder = template.options.findIndex((o) => o.id === (Array.isArray(aValue) ? aValue[0] : aValue))
```
jespino, line 401: We decided that we should try to order alphabetically.

webapp/src/cardFilter.ts
hahmadia: value here according to typescript can only be a `string | string[]`.  Checkbox option does not show up. It's impossible to filter by checkbox.

image

You can see prior to my changes, only select options showed up in that dropdown and I've added multiselect.

webapp/src/components/viewHeader/filterValue.tsx
hahmadia: So notion doesn't handle it all (only allowed to select one option at one time) 
image

Jira handles it how we handle it currently
image

Asana doesn't handle it either (similar to notion). It's a open feature request.
https://forum.asana.com/t/sort-and-filter-project-by-multiple-fields-due-date-and-assignee-for-example/15234/63

image


So with all that in mind, I've gone ahead and reverted this change to match Jira's.

webapp/src/viewModel/boardTree.ts
hahmadia: Done!

webapp/src/cardFilter.ts
jespino: :+1:

webapp/src/viewModel/boardTree.ts
hahmadia: I noticed ordering alphabetically makes the if multiselect || select block statement almost idnetical to the else statement so I merged them together to avoid repeatable code

webapp/src/components/viewHeader/filterValue.tsx
jespino: :+1:

jespino (approved):
Awesome, now looks good to me! :tada:

PR #542, opened 2021-06-07
Auditing

wiggin77:
#### Summary
This PR adds auditing to all FocalBoard APIs.  Audit output can be send to file, or compliance tool / log aggregator via TCP.  
Output can also be filtered by authorization APIs, read APIs, and APIs that modify the database.

Auditing leverages the logging engine, using a separate instance and config.

Documentation is needed for the config.

#### Ticket Link
https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=ca1a5441-4c1c-4d0e-856e-88cc744576c8&c=495b8708-4b78-49fe-8715-f6294f9c5aa3


audit.json
```diff
@@ -0,0 +1,37 @@
+{
```
harshilsharma63, line 1: We'll need to include this in Docker build so auditing works in Docker build as well.

cc @cpanato

server/api/api.go
```diff
@@ -797,8 +861,19 @@ func (a *API) handleGetSharing(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	a.logger.Debug("GET sharing", mlog.String("rootID", rootID))
 	jsonBytesResponse(w, http.StatusOK, sharingData)
+
+	if sharing == nil {
+		sharing = &model.Sharing{}
+	}
+	a.logger.Debug("GET sharing",
```
harshilsharma63, line 869: Instead of this, would setting the value as `N/A` be better? It will be clearer in the logs IMO.

server/go.mod
```diff
@@ -27,7 +27,6 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/tidwall/gjson v1.7.3 // indirect
 	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
-	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
```
harshilsharma63, line 30: This is actually used. It's used in enterprise repo and since go mod doesn't follow symlinks, it thinks this is unused and removes it. 

We shouldn't remove this.

harshilsharma63: We can revert `go.mod` and `go.sum files`.

server/server/server.go
```diff
@@ -116,7 +118,13 @@ func New(cfg *config.Configuration, singleUserToken string, logger *mlog.Logger)
 	}
 	metricsService := metrics.NewMetrics(instanceInfo)
 
-	appServices := app.AppServices{
+	// Init audit
```
harshilsharma63, line 121: As of now, auditing is a mandatory thing and the server fails to launch without `audit.json`. Is this intentional?

harshilsharma63 (changes requested):
Looks awesome!

server/go.mod
wiggin77: That means we'll have to be hand editing these files.  I don't recall having to do that for chat server.  I'll revert and raise this next meeting so we can figure out a solution that doesn't require hand editing.

audit.json
wiggin77: Cloud typically configures this via ENV vars instead of file, so I don't think we need to ship the sample config.

server/server/server.go
wiggin77: I don't think so. If the `cfg.AuditFile` and `cfg.AuditJSON` are both empty the configuration still succeeds and the audit calls don't do anything.

wiggin77: I've removed the sample config to make this more clear.

wiggin77: I also removed the sample logger config and added a default one.  Thanks for raising this - the default logger was needed or zero logging would happen by default.

server/api/api.go
wiggin77: I think the issue here is that this API should not allow empty responses.  IMO a 404 should be returned if no sharing record exists to be consistent with other Rest APIs.  I'm not sure if the client can handle that.

server/go.mod
wiggin77: Fixed

audit.json
harshilsharma63: We need to include something so users can run Focalboard from Dockerfile. Without this file, the server won't start.

server/api/api.go
harshilsharma63: Agree on returning 404. I can create a card for it. Can we update the audit code here assuming we're returning a 404? Refactoring to 404 and updating audit aren't blocked on each other IMO, unless they are?

server/go.mod
harshilsharma63: Agree on fixing this t avoid manually editing it.

server/server/server.go
harshilsharma63: Awesome!

audit.json
harshilsharma63: Never mind, I see we updating audit initialization. So we're good here.

harshilsharma63 (approved):

harshilsharma63:
LGTM! We can update that 404 handling if needed but not a blocker.

jespino (approved):
LGTM

Issue #387, opened 2021-05-06
Bug: Comment box doesn't clear after sending


chenilim:
**Steps to reproduce the behavior:**
1. Open a card
2. Enter text in the comment box (e.g. "test")
3. Click send (comment is posted)
4. Click on the comment box again
5. The text ("test") appears in the comment box

**Expected behavior:**
The comment box should be cleared after sending

**Platform:**
 - Chrome 89 on macOS 10.15.7

prakharporwal:
newbie to opensource . Like to pick this issue.

chenilim:
You got it @prakharporwal! Thanks!

prakharporwal:
@chenilim  not able to make a pull request on this repo ?? says access denied!!

prakharporwal:
@chenilim Made a pull request after forking the repo . please review . # 547

jespino:
This was already solved by @prakharporwal.

PR #547, opened 2021-06-08
Solved #387 Bug: Comment box doesn't clear after sending

prakharporwal:
The variable text was not updating on updation of UI and was getting reassigned the old value. I propagated  that by directly using the  text variable from props.


harshilsharma63:
LGTM! Thanks for fixing this @prakharporwal !

Issue #524, opened 2021-06-04
Bug: Card Template > Creating a property & setting type only stores either property name or type


lindy65:
Reported on behalf of contributor `jprusch`

**Summary:**
Card Template > Creating a property & setting type only stores either property name or type

**Steps to reproduce the behavior:**

https://user-images.githubusercontent.com/13020833/120787722-ed483800-c52f-11eb-9def-9ad252cfa9bf.mp4

**Expected behavior:**
Both the property name and type are stored

**Platform:**
 - OS and version: Windows 10  20H2 Build 19042.985

PR #553, opened 2021-06-09
Card Template > Creating a property & setting type only stores either property name or type 

hahmadia:
#### Summary

Set property type and property name at the same time. 
Fixes GH-524

#### Ticket Link

https://github.com/mattermost/focalboard/issues/524

### DIFF
--- a/webapp/src/components/cardDetail/cardDetailProperties.tsx
+++ b/webapp/src/components/cardDetail/cardDetailProperties.tsx
@@ -41,7 +41,7 @@ const CardDetailProperties = React.memo((props: Props) => {
                                     propertyId={propertyTemplate.id}
                                     propertyName={propertyTemplate.name}
                                     propertyType={propertyTemplate.type}
-                                    onTypeAndNameChanged={(newType: PropertyType = propertyTemplate.type, newName: string = propertyTemplate.name) => mutator.changePropertyTypeAndName(boardTree, propertyTemplate, newType, newName)}
+                                    onTypeAndNameChanged={(newType: PropertyType, newName: string) => mutator.changePropertyTypeAndName(boardTree, propertyTemplate, newType, newName)}
                                     onDelete={(id: string) => mutator.deleteProperty(boardTree, id)}
                                 />
                             </MenuWrapper>
--- a/webapp/src/components/cardDetail/cardDetailProperties.tsx
+++ b/webapp/src/components/cardDetail/cardDetailProperties.tsx
@@ -41,8 +41,7 @@ const CardDetailProperties = React.memo((props: Props) => {
                                     propertyId={propertyTemplate.id}
                                     propertyName={propertyTemplate.name}
                                     propertyType={propertyTemplate.type}
-                                    onNameChanged={(newName: string) => mutator.renameProperty(board, propertyTemplate.id, newName)}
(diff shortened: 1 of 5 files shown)


webapp/src/widgets/menu/textOption.tsx
```diff
@@ -23,6 +23,10 @@ function TextOption(props:TextOptionProps): JSX.Element {
                 e.target.dispatchEvent(new Event('menuItemClicked'))
                 props.onClick(props.id)
             }}
+            onMouseDown={(e: React.MouseEvent) => {
+                e.target.dispatchEvent(new Event('menuItemClicked'))
+                props.onClick(props.id)
+            }}
```
hahmadia, line 29: Note: 

I had to add this because the `onBlur` handler on the property name was being called over the onClick when I clicked on a new property Type.

webapp/src/components/cardDetail/cardDetailProperties.tsx
```diff
@@ -41,8 +41,7 @@ const CardDetailProperties = React.memo((props: Props) => {
                                     propertyId={propertyTemplate.id}
                                     propertyName={propertyTemplate.name}
                                     propertyType={propertyTemplate.type}
-                                    onNameChanged={(newName: string) => mutator.renameProperty(board, propertyTemplate.id, newName)}
-                                    onTypeChanged={(newType: PropertyType) => mutator.changePropertyType(boardTree, propertyTemplate, newType)}
+                                    onTypeAndNameChanged={(newType: PropertyType = propertyTemplate.type, newName: string = propertyTemplate.name) => mutator.changePropertyTypeAndName(boardTree, propertyTemplate, newType, newName)}
```
jespino, line 44: I think I prefer to avoid here the default value for the parameters. I don't even know when the default value is set, on call, when the function is define, when the components get render for first time, each render of the component? I think is hard to think about it, and I think I prefer to just make the parameters mandatory.

webapp/src/widgets/propertyMenu.tsx
```diff
@@ -68,10 +67,10 @@ const PropertyMenu = React.memo((props: Props) => {
                 onClick={(e) => e.stopPropagation()}
                 onChange={(e) => setName(e.target.value)}
                 value={name}
-                onBlur={() => props.onNameChanged(name)}
+                onBlur={() => props.onTypeAndNameChanged(undefined, name)}
```
jespino: Again, here, I prefer to not use the default value, looks weird and hard to undrestand and follow, and you already have the propertyType in the properties.

hahmadia: Done! Please refresh

webapp/src/components/cardDetail/cardDetailProperties.tsx
hahmadia: Done, please refresh.

webapp/src/widgets/menu/textOption.tsx
jespino: Maybe other option is to make a debounced version of the `onTypeAndNameChanged` so the last one is the one that is used. That way you don't have to mess with this `onMouseDown`.

hahmadia: Done!

harshilsharma63 (approved):
LGTM

PR #557, opened 2021-06-10
Fix MySQL migration connection when the DSN doesn't contain any query parameters

mgdelacroix:
#### Summary
This PR fixes a server panic that would happen in case the server is trying to connect to a MySQL database with no query parameters, when trying to append the `multiStatements` parameter.

Based on the [server fix applied for this same bug](https://github.com/mattermost/mattermost-server/pull/17723).


chenilim (approved):
I don't have good experience with this code, but LGTM.

jespino (approved):
Great!

mgdelacroix:
Cherry-picked in `v0.7.0` at https://github.com/mattermost/focalboard/commit/2c56d373a54aacc967d0afe9a50eae979f23b96c

PR #564, opened 2021-06-11
Adding the users for workspace in mattermost layer

jespino:


harshilsharma63 (approved):
LGTM

chenilim:
Thanks! I've cherry picked this into the [release-0.7.0 branch](https://github.com/mattermost/focalboard/tree/release-0.7.0).

It looks like this returns a list of users who are members of the associated channel. This is ok as a first step, but we will need some improvements in the future, including:
* Showing a list of all users in the system (including non-channel members)
* Paging through the list (ideally reusing the user picker component, or doing something similar)
* P2: Notify users who have been added / removed from a card

jespino:
@jespino For now the simplest approach is only expose the current channel users, probably to have a better approach we would need a `SearchUsers` endpoint and relay on that for autocompletion of users. So that way I only show you the first 15 results of users that matches with your already introduced text. For now, users in the current workspace is good, specially for things like "assigned to" or "creator". Other thing is we maybe need a special Person field that includes anybody in the instance. For example for something like "Reporter". They are 2 different use cases that I, as a user, expect different behavior.

About paging, I think is better filtering/searching, paging can be ok in certain cases, but for the usecase here paging is almost useless.

About user notifications, we should build a notification framework that can relay on mattermost notifications and maybe in anything else in the future. I think I prefer to abstract the concept of notification, and then create the implementation needed now.

Issue #561, opened 2021-06-11
Bug: Sorting icons are incorrect


harshilsharma63:
**Summary:**
Sorting icons are incorrect on columns in the table view are incorrect. They should the opposite of what they are right now.

**Steps to reproduce the behavior:**
1. Open a table view and sort any column.
2. Notice the icons ti the right of column name.

**Expected behavior:**
Asecnding should have an upward-pointing array and descending should have downward-pointing arrow.

**Screenshots:**
![sorting-icons](https://user-images.githubusercontent.com/18575143/121635941-bf6d7100-caa4-11eb-8cf8-f09511de4f35.gif)

**Platform:**
 - Browser and version: Safar 14.1
 - OS and version: MacOS Big Sur

**Additional context:**
This is platform independent issue.

cc @michaelgamble

harshilsharma63:
@haardikdharma10 assigning this to you.

harshilsharma63:
This is now resolved.

PR #576, opened 2021-06-14
Change sorting icons.

haardikdharma10:
#### Summary
Earlier, the sorting icon for `Sort ascending` was an arrow pointing downwards and opposite for `Sort descending`. Changed it so that the `Sort ascending` has arrow pointing upwards.

#### Ticket Link
Fixes : #561 

#### How has this been tested?

Refer to the screenshots below: (Ascending order shows an upward arrow and reverse for descending)

Screenshot 2021-06-14 at 5 28 43 PM

Screenshot 2021-06-14 at 5 28 54 PM

### DIFF
--- a/webapp/src/components/table/tableHeader.tsx
+++ b/webapp/src/components/table/tableHeader.tsx
@@ -59,8 +59,8 @@ const TableHeader = React.memo((props: Props): JSX.Element => {
             <MenuWrapper disabled={props.readonly}>
                 <Label>
                     {props.name}
-                    {props.sorted === 'up' && <SortUpIcon/>}
-                    {props.sorted === 'down' && <SortDownIcon/>}
+                    {props.sorted === 'up' && <SortDownIcon/>}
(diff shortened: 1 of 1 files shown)


sbishel (changes requested):
Thanks @haardikdharma10 for you submission. Although this change works, it doesn't seem right to me. The logic here seems right. I wouldn't want to display Icons with reversed names, that could cause confusion.

Assuming the icon names are not reversed, I would look for a logic issue when setting `props.sorted`.

haardikdharma10:
> Thanks @haardikdharma10 for you submission. Although this change works, it doesn't seem right to me. The logic here seems right. I wouldn't want to display Icons with reversed names, that could cause confusion.
> 
> Assuming the icon names are not reversed, I would look for a logic issue when setting `props.sorted`.

Thanks @sbishel! I'll close this as @harshilsharma63 already opened #611

PR #583, opened 2021-06-16
New Props: Created By, Created At, Updated By, Updated At

harshilsharma63:
#### Summary
Added following new properties-
1. Created by - username of the user who created the card
2. Updated by - username of user who made the most recent modification to the card
3. Updated at - timestamp of most recent modification to the card
4. Created at - timestamp of when the card was created

#### Ticket Link
https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=392c8889-6c8f-4cb7-8407-f5c57c1acfe8&c=b0111694-b37a-4701-8f0a-ee0a9aa87d17


hahmadia:
Hi Harshil,

I did a functionality test before deciding to do a code review. Please confirm if you experience some of these problems.

1. Created At and Updated By do not show up on KanBan view?
image

2. Created At Overflowing in Table View, Updated By Column shows up with no value displayed.
image

3. Updated At Value in Table View not the same as Card View? Created By did not update even after refresh?
image

image

hahmadia (changes requested):
Some functionality bugs noticed mentioned above

hahmadia:
Just thought I point this out, Notion couples the `Created` and `Updated` together in their dropdown. Should we do that?
Also, they use `Last edited` instead of `update`. I feel their wording is less "developer" and more "general-user-friendly-ish". I think we should change and match their coupling and wording. Your thoughts?

image

hahmadia:
Does it make sense to display the timezone after the Created and Updated times? Do you feel that would be over complicating? I feel it would be explicit and provide more clarity. Notion doesn't do it but this is where I'm on the fence and not 100% sure.

hahmadia:
Deleting a comment did not update the `Update At` field.

harshilsharma63:
I missed handling sort and search for these fields. I'll update the PR once added.

harshilsharma63:
> Some functionality bugs noticed mentioned above

These are now fixed.

harshilsharma63:
> Just thought I point this out, Notion couples the `Created` and `Updated` together in their dropdown. Should we do that?
> Also, they use `Last edited` instead of `update`. I feel their wording is less "developer" and more "general-user-friendly-ish". I'm 3/5 think we should change and match their coupling and wording. Your thoughts?
> 
> image

Makes sense. Update the order now.

harshilsharma63:
> Does it make sense to display the timezone after the Created and Updated times? Do you feel that would be over complicating? I feel it would be explicit and provide more clarity. Notion doesn't do it but this is where I'm on the fence and not 100% sure.

We're storing the time in UTC and displaying it in the user's local time zone, just like we do in Mattermost, for example, for post's creation or update time.

harshilsharma63:
@sbishel @hahmadia I'm done with all the changes. I'll be doing a self-review once tomorrow after which you can review this.

Thanks.

sbishel (changes requested):
When I downloaded to test, it appears that my data did not upgrade correctly. I received the following server errors. 
rror [2021-06-30 11:29:50.985 -06:00] ERROR blocksFromRows                     error="sql: Scan error on column index 3, name "created_by": converting NULL to string is unsupported"
error [2021-06-30 11:29:50.985 -06:00] API ERROR                                code=500 error="sql: Scan error on column index 3, name "created_by": converting NULL to string is unsupported"

I can also duplicate by importing and older export. Let me know if you need help duplicating.

server/services/store/sqlstore/blocks.go
```diff
@@ -341,11 +355,66 @@ func (s *SQLStore) InsertBlock(c store.Container, block model.Block) error {
 			"create_at",
 			"update_at",
 			"delete_at",
-		).Values(
+		)
+
+	if existingBlock != nil {
+		block.ModifiedBy = userID
```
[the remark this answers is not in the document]
harshilsharma63: Done.

server/services/store/sqlstore/blocks.go
```diff
@@ -341,11 +355,66 @@ func (s *SQLStore) InsertBlock(c store.Container, block model.Block) error {
...
+			return err
+		}
+
+		if _, err := tx.Exec(q, args...); err != nil {
+			s.logger.Error(`InsertBlock error occurred while updating existing block`, mlog.String("blockID", block.ID), mlog.Err(err))
+			return err
+		}
+	} else {
+		block.CreatedBy = userID
+		block.ModifiedBy = userID
```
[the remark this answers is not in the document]
harshilsharma63: Done.

server/services/store/storetests/blocks.go
```diff
@@ -722,3 +769,32 @@ func testGetBlocksWithRootID(t *testing.T, store store.Store, container store.Co
...
+			CreatedBy:  "user-id-1",
+		}, fetchedBlock)
+	})
+
+	t.Run("get a non-existing block", func(t *testing.T) {
+		fetchedBlock, err := store.GetBlock(container, "non-existing-id")
+		require.NoError(t, err)
+		require.Nil(t, fetchedBlock)
+	})
+}
```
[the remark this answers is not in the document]
harshilsharma63: Done.

webapp/src/viewModel/boardTree.ts
```diff
@@ -448,18 +473,27 @@ class MutableBoardTree implements BoardTree {
         return cards
     }
 
-    private mutableCopy(): MutableBoardTree {
-        return MutableBoardTree.buildTree(this.board.id, this.allBlocks)!
+    foo(): BoardTree {
+        return new MutableBoardTree(new MutableBoard([]))
+    }
```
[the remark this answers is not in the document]
harshilsharma63: No, removed it. This is why I was doing a self-review. I knew I had some funky debug methods left in :P.

webapp/src/viewModel/boardTree.ts
```diff
@@ -448,18 +473,27 @@ class MutableBoardTree implements BoardTree {
...
+        const boardTree = await this.mutableCopy() as MutableBoardTree
         boardTree.setActiveView(viewId)
         return boardTree
     }
 
-    copyWithSearchText(searchText?: string): BoardTree {
-        const boardTree = this.mutableCopy()
+    bar(): string {
+        return 'ASdas'
+    }
```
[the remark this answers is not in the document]
harshilsharma63: No, removed.

webapp/src/widgets/propertyMenu.tsx
```diff
@@ -91,66 +108,16 @@ const PropertyMenu = React.memo((props: Props) => {
...
+                {
+                    propertyTypes.map((propertyType) => (
+                        <Menu.Text
+                            key={propertyType.type}
+                            id={propertyType.type}
+                            name={typeDisplayName(intl, propertyType.type as PropertyType)}
+                            onClick={() => debouncedOnTypeAndNameChanged(propertyType.type as PropertyType)()}
+                        />
+                    ))
+                }
```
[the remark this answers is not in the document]
harshilsharma63: 👍

harshilsharma63:
> When I downloaded to test, it appears that my data did not upgrade correctly. I received the following server errors.
> rror [2021-06-30 11:29:50.985 -06:00] ERROR blocksFromRows error="sql: Scan error on column index 3, name "created_by": converting NULL to string is unsupported"
> error [2021-06-30 11:29:50.985 -06:00] API ERROR code=500 error="sql: Scan error on column index 3, name "created_by": converting NULL to string is unsupported"
> 
> I can also duplicate by importing and older export. Let me know if you need help duplicating.

@sbishel can you share that focalboard export? I can try importing it locally.

sbishel:
> @sbishel can you share that focalboard export? I can try importing it locally.

I had to zip for github.
[archive-2021-6-30.focalboard.zip](https://github.com/mattermost/focalboard/files/6770502/archive-2021-6-30.focalboard.zip)

webapp/src/components/kanban/kanban.tsx
```diff
@@ -28,159 +32,52 @@ type Props = {
...
-    private propertyNameChanged = async (option: IPropertyOption, text: string): Promise<void> => {
-        const {boardTree} = this.props
+    const propertyValues = groupByProperty.options || []
+    Utils.log(`${propertyValues.length} propertyValues`)
+
+    const {board, activeView, visibleGroups, hiddenGroups} = boardTree
+    const visiblePropertyTemplates = board.cardProperties.filter((template) => activeView.visiblePropertyIds.includes(template.id))
+    const isManualSort = activeView.sortOptions.length < 1
+
+    const [cardTrees, setCardTrees] = useState<{[key: string]: CardTree | undefined}>({a: undefined})
```
hahmadia, line 51: since it can be undefined, would it not work to just do `()` rather than do `{a: undefined}` ? 
```suggestion
    const [cardTrees, setCardTrees] = useState<{[key: string]: CardTree | undefined}>()
```
This wouldn't work?

webapp/src/components/kanban/kanban.tsx
```diff
@@ -278,6 +173,115 @@ class Kanban extends React.Component<Props, State> {
...
+
+                {/* Hidden column header */}
+
+                {hiddenGroups.length > 0 &&
+                <div className='octo-board-header-cell narrow'>
+                    <FormattedMessage
+                        id='BoardComponent.hidden-columns'
+                        defaultMessage='Hidden columns'
+                    />
+                </div>
```
hahmadia, line 206: Personal preference but can we indent the blocks of code within a `{}` to make it more clear that these only spawn under the condition. Otherwise I have to look for the opening and closing braces. I just find it will be more readable if we can indent these.

hahmadia: It's also what's shown in the React docs https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator . Notice the indentation for multiline things within the `{}`. 
Thanks

webapp/src/components/kanban/kanbanCard.tsx
```diff
@@ -16,9 +16,11 @@ import {useSortable} from '../../hooks/sortable'
 
 import './kanbanCard.scss'
 import PropertyValueElement from '../propertyValueElement'
+import {CardTree} from '../../viewModel/cardTree'
 
 type Props = {
     card: Card
+    cardTree: CardTree | undefined
```
hahmadia, line 23: I think you can just do ? operator if it can also be undefined. You don't have to be explicit.
```suggestion
    cardTree?: CardTree
```

https://www.typescriptlang.org/docs/handbook/2/objects.html#optional-properties

You can also apply to this  to anywhere else you may have done a `property | undefined` which I may have missed.

webapp/src/components/properties/lastModifiedAt/lastModifiedAt.tsx
```diff
@@ -0,0 +1,25 @@
...
+import React from 'react'
+
+import {Card} from '../../../blocks/card'
+import {CardTree} from '../../../viewModel/cardTree'
+
+const moment = require('moment')
+
+type Props = {
+    card: Card,
+    cardTree: CardTree | undefined
```
hahmadia, line 13: optional property `?`

webapp/src/components/properties/lastModifiedAt/lastModifiedAt.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const LastModifiedAt = (props: Props): JSX.Element => {
+    let latestBlock = props.cardTree ? props.cardTree.allBlocks.filter((block) => block.parentId === props.card.id || block.id === props.card.id).sort((a, b) => b.updateAt - a.updateAt)[0] : props.cardTree
+    latestBlock = latestBlock || props.card
```
hahmadia, line 18: I'm gonna say lets opt to use if statement. Although more lines of code, I feel the readability will increase.
Alternative is
```suggestion
    let latestBlock = props.card
    if (props.cardTree) {
        latestBlock = props.cardTree.allBlocks
        .filter((block) => block.parentId === props.card.id || block.id === props.card.id)
        .sort((a, b) => b.updateAt - a.updateAt)[0]
   }
```

I was trying to move each function call to a new line for better readability like the .filter and .sort in my revision above.

webapp/src/components/properties/lastModifiedAt/lastModifiedAt.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+    return (
+        <div className='LastModifiedAt octo-propertyvalue'>{moment(latestBlock.updateAt).format('llll')}</div>
```
hahmadia, line 21: nit:

can we multiline this for readability?
```suggestion
        <div className='LastModifiedAt octo-propertyvalue'>
              {moment(latestBlock.updateAt).format('llll')}
       </div>
```

webapp/src/components/properties/lastModifiedBy/lastModifiedBy.tsx
```diff
@@ -0,0 +1,28 @@
...
+import {BoardTree} from '../../../viewModel/boardTree'
+
+type Props = {
+    card: Card,
+    boardTree?: BoardTree,
+}
+
+const LastModifiedBy = (props: Props): JSX.Element => {
+    let latestBlock = props.boardTree?.allBlocks.filter((block) => block.parentId === props.card.id || block.id === props.card.id).sort((a, b) => b.updateAt - a.updateAt)[0]
+    latestBlock = latestBlock || props.card
```
hahmadia, line 17: Same as above. Could we refactor into if statement with each function call being on a new line to avoid having one long line. Thanks

webapp/src/components/propertyValueElement.tsx
```diff
@@ -178,6 +182,32 @@ const PropertyValueElement = (props:Props): JSX.Element => {
                 readOnly={readOnly}
             />
         )
+    } else if (propertyTemplate.type === 'createdBy') {
+        return (
+            <UserProperty
+                value={card.createdBy}
+                readonly={true} // created by is an immutable property, so will always be readonly
+                onChange={() => {}} // since created by is immutable, we don't need to handle onChange
+            />
```
hahmadia, line 191: If we are following the pattern of every property has it's own component, would it be too much proxy if createdBy had it's own component too which did just this? I'm not sure but it came to mind with what we had discussed before.

webapp/src/viewModel/cardTree.ts
```diff
@@ -59,12 +64,24 @@ class MutableCardTree implements CardTree {
...
+
         return cardTree
     }
 
-    // private mutableCopy(): MutableCardTree {
-    //     return MutableCardTree.buildTree(this.card.id, this.allBlocks)!
-    // }
+    public static getMostRecentBlock(cardTree: CardTree): IBlock {
+        let latestBlock: IBlock = cardTree.card
+        cardTree?.allBlocks.forEach((block) => {
```
hahmadia, line 74: Since it's required, is the `?` needed? It's minor but made me think it can be undefined.
```suggestion
        cardTree.allBlocks.forEach((block) => {
```

webapp/src/widgets/propertyMenu.tsx
```diff
@@ -91,66 +108,16 @@ const PropertyMenu = React.memo((props: Props) => {
...
-                    name={typeDisplayName(intl, 'createdTime')}
-                    onClick={() => debouncedOnTypeAndNameChanged('createdTime')()}
-                />
-                <Menu.Text
-                    id='updatedTime'
-                    name={typeDisplayName(intl, 'updatedTime')}
-                    onClick={() => debouncedOnTypeAndNameChanged('updatedTime')()}
-                />
+                {
+                    propertyTypes.map((propertyType) => (
```
hahmadia, line 112: Oh smart! I love it ❤️

hahmadia (changes requested):
Great work! I really like it 😄 

I've made comments on things that caught my eye while reviewing the code. 

I also got the same thing happening to me that Scott above mentioned.

harshilsharma63:
> Great work! I really like it 😄
> 
> I've made comments on things that caught my eye while reviewing the code.
> 
> I also got the same thing happening to me that Scott above mentioned.

@hahmadia you must have switched to my branch, ran Focalboard, then switched to some other branch. Doing that will cause the issue as when you switched to my branch, the new columns were created. Then when you worked on the other branch, those columns were there but the code to populate the columns wasn't there. Since the data migration works just once, switching back to my branch won't update those blocks.

I believe the same happened with @sbishel and I too faced this.

server/services/store/sqlstore/blocks.go
```diff
@@ -341,11 +356,68 @@ func (s *SQLStore) InsertBlock(c store.Container, block model.Block) error {
...
+			block.CreatedBy,
+			block.ModifiedBy,
+			block.Schema,
+			block.Type,
+			block.Title,
+			fieldsJSON,
+			block.CreateAt,
+			block.UpdateAt,
+			block.DeleteAt,
+		)
```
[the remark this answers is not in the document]
harshilsharma63: Done.

server/services/store/sqlstore/blocks.go
```diff
@@ -341,11 +356,68 @@ func (s *SQLStore) InsertBlock(c store.Container, block model.Block) error {
...
+		_, err = sq.ExecContextWith(ctx, tx, query.Into(s.tablePrefix+"blocks"))
```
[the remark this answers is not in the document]
harshilsharma63: I'm actually setting insert query + the params into the `query` variable, so this is fine.

server/services/store/sqlstore/blocks.go
```diff
@@ -356,23 +428,6 @@ func (s *SQLStore) InsertBlock(c store.Container, block model.Block) error {
 		block.DeleteAt,
 	)
```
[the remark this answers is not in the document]
harshilsharma63: Avoided duplication using `SetMap` method.

server/services/store/sqlstore/blocks.go
```diff
@@ -467,3 +522,38 @@ func (s *SQLStore) GetBlockCountsByType() (map[string]int64, error) {
 	}
 	return m, nil
 }
+
+func (s *SQLStore) GetBlock(c store.Container, blockID string) (*model.Block, error) {
```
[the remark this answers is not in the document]
harshilsharma63: I think we can skip the bulk fetch method with this PR as this PR has already gotten pretty big. We can add it later reusing this method's code.

server/services/store/sqlstore/initialize.go
```diff
@@ -39,13 +39,13 @@ func (s *SQLStore) importInitialTemplates() error {
...
 		s.logger.Trace("insert block",
-			mlog.String("blockID", block.ID),
-			mlog.String("block_type", block.Type),
-			mlog.String("block_title", block.Title),
+			mlog.String("blockID", archive.Blocks[i].ID),
+			mlog.String("block_type", archive.Blocks[i].Type),
+			mlog.String("block_title", archive.Blocks[i].Title),
 		)
-		err := s.InsertBlock(globalContainer, block)
+		err := s.InsertBlock(globalContainer, &archive.Blocks[i], "system")
```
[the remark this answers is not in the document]
harshilsharma63: I'm modifying the block in `InsertBlock` method and want the modifications to be preserved beyond that method. For that, we need to pass a pointer, hence using the `&archive.Blocks[i]`. PS `&block` would be incorrect here.

webapp/src/components/kanban/kanban.tsx
```diff
@@ -28,159 +32,52 @@ type Props = {
...
-    }
-
-    private propertyNameChanged = async (option: IPropertyOption, text: string): Promise<void> => {
-        const {boardTree} = this.props
+    const propertyValues = groupByProperty.options || []
+    Utils.log(`${propertyValues.length} propertyValues`)
+
+    const {board, activeView, visibleGroups, hiddenGroups} = boardTree
+    const visiblePropertyTemplates = board.cardProperties.filter((template) => activeView.visiblePropertyIds.includes(template.id))
+    const isManualSort = activeView.sortOptions.length < 1
```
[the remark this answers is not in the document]
harshilsharma63: Done.

webapp/src/components/kanban/kanban.tsx
harshilsharma63: Hey, yes This was added for debugging and I missed removing it.

webapp/src/components/kanban/kanban.tsx
harshilsharma63: Done.

webapp/src/components/kanban/kanbanCard.tsx
harshilsharma63: I was actually using the `?:` in some places but not concisely. I'll keep this in mind.

webapp/src/components/properties/lastModifiedAt/lastModifiedAt.tsx
harshilsharma63: Done.

webapp/src/components/properties/lastModifiedAt/lastModifiedAt.tsx
harshilsharma63: I really like how it looks. Done.

webapp/src/components/properties/lastModifiedAt/lastModifiedAt.tsx
harshilsharma63: Done.

webapp/src/components/properties/lastModifiedBy/lastModifiedBy.tsx
harshilsharma63: Done.

webapp/src/components/propertyValueElement.tsx
harshilsharma63: Makes sense. Done.

webapp/src/viewModel/cardTree.ts
harshilsharma63: Done.

webapp/src/widgets/propertyMenu.tsx
harshilsharma63: Thanks!

sbishel (approved):
I tested again. This time, using a fresh db from main. Created some boards. Exported the boards.
Updated to this PR, restarted server, webapp. App continued to work. Imported earlier export, worked. 

Looks good.

harshilsharma63:
> I tested again. This time, using a fresh db from main. Created some boards. Exported the boards.
> Updated to this PR, restarted server, webapp. App continued to work. Imported earlier export, worked.
> 
> Looks good.

@hahmadia can you also confirm the same? If yes then I can merge this.

harshilsharma63:
> Deleting a comment did not update the `Update At` field. In fact, it started to roll back the Updated At field to an older time an older change was made. Easy to be malicious here without getting caught.

@hahmadia yeah, this is somewhat of a design issue which how we calculate the updated at the timestamp in the client. I've created a bug for fixing this later on as it will need more than some minor changes.

hahmadia (approved):
I tested and works fine!
Thanks for quick changes.

server/services/store/sqlstore/migrations/bindata.go
```diff
@@ -1,446 +1,208 @@
-// Code generated by go-bindata. DO NOT EDIT.
```
wiggin77, line 1: This file is no longer generated by `go-bindata`?

server/services/store/storetests/helpers.go
```diff
@@ -8,9 +8,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func InsertBlocks(t *testing.T, s store.Store, container store.Container, blocks []model.Block) {
+func InsertBlocks(t *testing.T, s store.Store, container store.Container, blocks []model.Block, userID string) {
 	for _, block := range blocks {
-		err := s.InsertBlock(container, block)
+		err := s.InsertBlock(container, &block, userID)
```
wiggin77, line 13: Passing a pointer to a loop var can cause unexpected problems.  I see this was done correctly in other places so this was probably just missed.

server/services/store/sqlstore/blocks.go
```diff
@@ -346,38 +361,74 @@ func (s *SQLStore) InsertBlock(c store.Container, block model.Block) error {
...
+
+		insertQueryValues["created_by"] = block.CreatedBy
+		insertQueryValues["create_at"] = block.CreateAt
+		insertQueryValues["update_at"] = block.UpdateAt
+		insertQueryValues["modified_by"] = block.ModifiedBy
+
+		query := insertQuery.SetMap(insertQueryValues)
+		_, err = sq.ExecContextWith(ctx, tx, query.Into(s.tablePrefix+"blocks"))
+		if err != nil {
+			tx.Rollback()
```
wiggin77, line 424: This error is being ignored.  It should be logged.

server/services/store/sqlstore/blocks.go
```diff
@@ -473,3 +524,38 @@ func (s *SQLStore) GetBlockCountsByType() (map[string]int64, error) {
...
+		Where(sq.Eq{"id": blockID}).
+		Where(sq.Eq{"coalesce(workspace_id, '0')": c.WorkspaceID})
+
+	rows, err := query.Query()
+	if err != nil {
+		s.logger.Error(`GetBlock ERROR`, mlog.Err(err))
+		return nil, err
+	}
+
+	blocks, err := s.blocksFromRows(rows)
```
wiggin77, line 554: This error is being ignored.  I'm not sure if empty result set is considered an error, but either way that can be checked and if a true error then err should be returned.

server/services/store/sqlstore/migrations/bindata.go
harshilsharma63: It is. For me, go-bindata itself removed this comment. Not sure why though.

server/services/store/storetests/helpers.go
harshilsharma63: Nice catch! Fixed.

server/services/store/sqlstore/blocks.go
harshilsharma63: Done. We weren't logging it anywhere so I've updated it everywhere.

server/services/store/sqlstore/blocks.go
harshilsharma63: Yes, this was missed.

harshilsharma63:
@wiggin77 I've created a PR with fixes to your comments - https://github.com/mattermost/focalboard/pull/694

Issue #541, opened 2021-06-07
Bug: Column Headers no longer scroll horizonally


sbishel:
**Summary:**
Column headers have been frozen for vertical scroll.  However, they also don't scroll with columns when scrolling horizontally.

**Steps to reproduce the behavior:**
1. Go to Table View with columns wider than screen
2. Scroll horizontally
3. See column headers do not scroll with column data.

**Expected behavior:**
Column headers should scroll horizontally with column data.

**Screenshots:**

https://user-images.githubusercontent.com/12704875/121084436-e0308080-c79d-11eb-83b7-e1bc8d3b42c0.mp4

**Platform:**
 - Browser and version: Chrome 91.0.4472.77
 - OS and version: macOS 11.4

harshilsharma63:
This has been fixed now.

PR #586, opened 2021-06-17
Fixed horizontal scroll issue with table view

harshilsharma63:
#### Summary
Fixed horizontal scroll issue with table view

#### Ticket Link
https://github.com/mattermost/focalboard/issues/541

horizontal-scroll

### DIFF
--- a/webapp/src/components/table/table.scss
+++ b/webapp/src/components/table/table.scss
@@ -1,7 +1,4 @@
 .Table {
-    .table-row-container {
-        margin-top: 48px;
-    }
     .octo-group-header-cell {
         display: flex;
         flex-shrink: 0;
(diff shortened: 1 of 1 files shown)


webapp/src/components/table/table.scss
```diff
@@ -1,7 +1,4 @@
 .Table {
-    .table-row-container {
-        margin-top: 48px;
-    }
```
sbishel, line 4: 👍 Glad you were able to remove this.

sbishel (approved):

webapp/src/components/table/table.scss
harshilsharma63: Yeah, it wasn't needed now.

PR #588, opened 2021-06-17
Added build sttaus badge to readme

harshilsharma63:
#### Summary
Added build staus badge to readme


jespino (approved):
LGTM. I don't know if makes a lot of sense to expose all of them, I don't know how valuable is that amount of information.

harshilsharma63:
@jespino I think it's fine as it allows us easily view the status.

PR #589, opened 2021-06-17
Remove enterprise symlink

wiggin77:
#### Summary
Including `enterprise` repo as a sym-link within Focalboard is not supported by Go tools such as `go mod tidy`.  Since the enterprise repo is no longer needed, this PR removes the sym-link and updates the `make` file accordingly.

#### Ticket Link
https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=ca1a5441-4c1c-4d0e-856e-88cc744576c8&c=e785daa8-b5db-4fd7-8310-16d1f8517abe


.gitignore
```diff
@@ -35,7 +35,6 @@ coverage
 build/Release
 
 # Enterprise files
-server/main/imports.go
 server/enterprise
```
jespino, line 38: I think this line and the comment are no longer needed.

jespino:
I think we can also remove the `server/server/enterprise.go` file, and the `server/einterfaces/` directory.

jespino (changes requested):
A couple of small changes, but otherwise looks good to me.

.gitignore
wiggin77: I was concerned someone might check in their left-over enterprise directory since this PR does not remove an existing sym-links.  I guess we would catch that in a review.

jespino (approved):
LGTM.

harshilsharma63 (approved):

PR #590, opened 2021-06-17
Refactor App lifecycle

wiggin77:
#### Summary
This PR refactors the App lifecycle by not creating a new App instance each request.  One App is created at server creation time, and passed to any services needing it.

To avoid cyclical dependencies, the few server methods needed by App are provided via interface.

#### Ticket Link
https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=ca1a5441-4c1c-4d0e-856e-88cc744576c8&c=8b6fc880-163b-4f70-8cfa-2be198b487be

### DIFF
--- a/server/app/app.go
+++ b/server/app/app.go
@@ -2,15 +2,20 @@ package app
 
 import (
 	"github.com/mattermost/focalboard/server/auth"
+	"github.com/mattermost/focalboard/server/model"
 	"github.com/mattermost/focalboard/server/services/config"
 	"github.com/mattermost/focalboard/server/services/metrics"
 	"github.com/mattermost/focalboard/server/services/mlog"
 	"github.com/mattermost/focalboard/server/services/store"
 	"github.com/mattermost/focalboard/server/services/webhook"
-	"github.com/mattermost/focalboard/server/ws"
 
 	"github.com/mattermost/mattermost-server/v5/shared/filestore"
 )
+
+type WebsocketServer interface {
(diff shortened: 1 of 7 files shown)


server/app/app.go
```diff
@@ -25,14 +30,14 @@ type App struct {
...
-	wsServer     *ws.Server
+	wsServer     WebsocketServer
 	filesBackend filestore.FileBackend
 	webhook      *webhook.Client
 	metrics      *metrics.Metrics
 	logger       *mlog.Logger
 }
 
-func New(config *config.Configuration, wsServer *ws.Server, services Services) *App {
+func New(config *config.Configuration, wsServer WebsocketServer, services Services) *App {
```
jespino, line 40: Is there any reason to not follow the proposed approach of `type App struct {services Services}`, also is there any reason to pass the `wsServer` outside and inside the `services` param? (I know that are not related to your changes).

wiggin77: We end up with calls like `a.services.logger.Debug(...)` which gets tedious. I think we still achieve the abstraction/separation even though the App has each service as a separate field.

The only reason the wsServer` is not passed in the services struct is because `ws.Server` is not in the services package.

jespino: ok, makes sense.

jespino (approved):
LGTM

mgdelacroix (approved):
Awesome!

PR #591, opened 2021-06-17
rename logger & audit config

wiggin77:
#### Summary
Renamed the config key names for logger and audit config.  `config.json` lets you specify a path to logging/audit config or provide config via embedded JSON.

`logging_file` was used as the key name to specify a path to config file for logging.  It was confusing as it could be mistakenly thought to be the name of the file to output logs to.  It has been changed to `logging_cfg_file`.  Same reasoning for changing `audit_file` to `audit_cfg_file`.

#### Ticket Link
NONE


mgdelacroix (approved):
LGTM! :+1:

PR #594, opened 2021-06-18
Fixed issue with dropdown background being height-limited

harshilsharma63:
#### Summary
Fixed issue with the dropdown background being height-limited

Before:
Screenshot 2021-06-18 at 3 26 30 PM

After: 
Screenshot 2021-06-18 at 3 25 50 PM

### DIFF
--- a/webapp/src/widgets/valueSelector.scss
+++ b/webapp/src/widgets/valueSelector.scss
@@ -56,12 +56,6 @@
             }
         }
     }
-
-    div:nth-child(3) {
-        div {
-            max-height: unset;
-        }
-    }
 }
 
 .label-container > .Label {
--- a/webapp/src/widgets/valueSelector.scss
+++ b/webapp/src/widgets/valueSelector.scss
@@ -56,6 +56,12 @@
             }
         }
     }
+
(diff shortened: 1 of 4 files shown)


webapp/src/widgets/valueSelector.scss
```diff
@@ -56,6 +56,12 @@
             }
         }
     }
+
+    div:nth-child(3) {
```
jespino, line 60: This is flaky and is not how react-select expects you to change the styles, in this case the approach is to use the property `maxMenuHeight` (you can find it here: https://react-select.com/props#select-props), but for changing styles inside react select components normally you change the styles function generator that you can find in the `valueSelector.tsx` file.

You can see more information about modifying the styles in https://react-select.com/styles

jespino (changes requested):
Not the right approach for react-select

webapp/src/widgets/valueSelector.scss
harshilsharma63: Done. I've fixed it in react-select's recommended way.

jespino (approved):
LGTM

PR #597, opened 2021-06-18
Adding support for embed the webapp into the plugin

jespino:


harshilsharma63:
@jespino is this still needed?

jespino:
@harshilsharma63 yes it is, we have to decide if we want to embed the whole application in the frontend, but probably the answer is yes.

harshilsharma63:
@jespino oh okay.

hahmadia (approved):
There wasn't anything that caught my eye that needs changing. I'm approving as it LGTM. Thank you

mgdelacroix (approved):
LGTM! :+1:

chenilim:
cc @ogi-m: This is a very significant infrastructure change. In theory, no functionality should change directly, but please run the appropriate level of testing on this. At a minimum I suggest a basic functionality test pass.

One concern in particular is page load times - can we quantify this? And quantify the impact of this change?
* i.e. what is the "fresh load" download size (html, js, etc.)? And avg. load time? How does this compare with core Mattermost?

wiggin77 (approved):
Code LGTM. :+1:

ogi-m:
@jespino I'm having trouble with the plugin. This happened to me on two servers already:
1. Created a fresh server
2. Uploaded the plugin build from the PR
3. Opened focalboard from a channel
4. Added some boards and cards
5. Refresh - blank screen, can no longer access the server

Tried logging in from a different browser but it hangs at login. This doesn't happen with 0.7.3

I also noticed that the path is _.../plug/focalboard/..._ and not _.../plugins/focalboard/...._

jespino:
@ogi-m yes, the path is the correct one, it needs to change, because the new way of doing it, but I don't care about that. I'll check it and see if I can reproduce your problems.

jespino:
After investigating the problems found by ogi, looks like is some kind of problem with the client config request that gets hanging on mattermost side when the plugin is enabled. Looks like more related to database access problem than a plugin problem.

jespino:
@ogi-m finally, after investigating it, the problem happens on the main branch too, but is not so obvious because is only affecting mattermost. My guess is the problem is related to the database access. Anyway, this is not affected by that, so I going to merge it into master.

PR #598, opened 2021-06-18
Updated lanquage selectors and fixed traditional chinese

### LINKED ISSUE #448: Bug: Language Translations should not also translate the language selector options
**Summary:**
The language selector will also translate the language selector options. This makes it difficult to get back to your original language.

1. Click Settings
2. Set Language
3. Pick any language

**Expected behavior:**
The language options will be translated to the language they belong to. Similar to how Mattermost currently operates.

**Screenshots:**
Screen Shot 2021-05-20 at 9 43 02 AM

**Platform:**
 - Browser and version: v0.6.5
 - OS and version:MacOS Big Sur

coltoneshaw:
Fixes #448 

- Modified the language selectors to remain the base language regardless of the translation.
- Traditional Chinese also had the wrong code and the name/code swapped.
- Modified the split of the `locale` to support Chinese language codes.

### DIFF
--- a/webapp/src/combinedProviders.tsx
+++ b/webapp/src/combinedProviders.tsx
@@ -22,7 +22,7 @@ const CombinedProviders = React.memo((props: Props): JSX.Element => {
     const {language, setLanguage, user} = props
     return (
         <IntlProvider
-            locale={language.split(/[-_]/)[0]}
+            locale={language.split(/[_]/)[0]}
             messages={getMessages(language)}
         >
             <SetLanguageContext.Provider value={setLanguage}>
--- a/webapp/src/components/properties/user/user.test.tsx
+++ b/webapp/src/components/properties/user/user.test.tsx
@@ -3,7 +3,7 @@
 
 import React from 'react'
 import {IntlProvider} from 'react-intl'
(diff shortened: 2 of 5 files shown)


coltoneshaw:
@jespino / @harshilsharma63 - I noticed that the language codes for Chinese traditional / simplified don't add check marks when they're selected. It looks like the issue could be [here](https://github.com/mattermost/focalboard/blob/521028612a68db7941fab096ebf11e382f4a2c74/webapp/src/i18n.tsx#L46) since the supported language is split on the `_` but Chinese is `'zh_Hant', 'zh_Hans'`

Doing some testing to see what resolves it but I'm a bit unsure of where specifically the value the code is matched to comes from.

coltoneshaw:
Screen Shot 2021-06-18 at 2 03 20 PM

I believe I found the issue. Splitting on the `-` did not work for the Chinese language codes unless only one Chinese translation was supported. Modified this to **only** split on the `_`. Confirmed each language selector, checkbox, and translation is working.

I could not find any other references to this code where these changes would also break.

jespino (approved):
Nice fix! Thank you!!

PR #603, opened 2021-06-18
First pass linter cleanup

wiggin77:
#### Summary
This PR is a first pass at cleaning up the 100+ linter errors in Focalboard codebase.  After merge, the number is less than 30.

#### Ticket Link
https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=ca1a5441-4c1c-4d0e-856e-88cc744576c8&c=f17a2e0c-fb55-465c-80b5-b484926239f8

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -719,17 +719,6 @@ func filterOrphanBlocks(blocks []model.Block) (ret []model.Block) {
 	return blocks
 }
 
-/*
-func arrayContainsBlockWithID(array []model.Block, blockID string) bool {
-	for _, item := range array {
-		if item.ID == blockID {
-			return true
-		}
-	}
-	return false
-}
-*/
-
 func (a *API) handleImport(w http.ResponseWriter, r *http.Request) {
 	// swagger:operation POST /api/v1/workspaces/{workspaceID}/blocks/import importBlocks
 	//
@@ -1341,11 +1330,3 @@ func jsonBytesResponse(w http.ResponseWriter, code int, json []byte) { //nolint:
 	w.WriteHeader(code)
 	_, _ = w.Write(json)
 }
-
-/*
-func addUserID(rw http.ResponseWriter, req *http.Request, next http.Handler) {
-	ctx := context.WithValue(req.Context(), "userid", req.Header.Get("userid"))
-	req = req.WithContext(ctx)
-	next.ServeHTTP(rw, req)
-}
-*/
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -22,13 +21,14 @@ import (
 )
 
 const (
-	HEADER_REQUESTED_WITH     = "X-Requested-With"
-	HEADER_REQUESTED_WITH_XML = "XMLHttpRequest"
+	HeaderRequestedWith    = "X-Requested-With"
+	HeaderRequestedWithXML = "XMLHttpRequest"
+	SingleUser             = "single-user"
 )
 
 const (
-	ERROR_NO_WORKSPACE_CODE    = 1000
-	ERROR_NO_WORKSPACE_MESSAGE = "No workspace"
+	ErrorNoWorkspaceCode    = 1000
+	ErrorNoWorkspaceMessage = "No workspace"
 )
 
 // ----------------------------------------------------------------------------------------------------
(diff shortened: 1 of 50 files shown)


server/api/api.go
```diff
@@ -720,6 +719,7 @@ func filterOrphanBlocks(blocks []model.Block) (ret []model.Block) {
 	return blocks
 }
 
+/*
```
jespino, line 722: I'm not a big fan of commenting things, if it is no longer needed, I guess we can delete it.

server/api/api.go
```diff
@@ -1322,27 +1323,29 @@ func (a *API) errorResponseWithCode(w http.ResponseWriter, statusCode int, error
...
 
-func jsonBytesResponse(w http.ResponseWriter, code int, json []byte) {
+func jsonBytesResponse(w http.ResponseWriter, code int, json []byte) { //nolint:unparam
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(code)
-	w.Write(json)
+	_, _ = w.Write(json)
 }
 
+/*
```
jespino, line 1345: same here

server/api/api.go
```diff
@@ -1322,27 +1323,29 @@ func (a *API) errorResponseWithCode(w http.ResponseWriter, statusCode int, error
...
 
-func jsonStringResponse(w http.ResponseWriter, code int, message string) {
+func jsonStringResponse(w http.ResponseWriter, code int, message string) { //nolint:unparam
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(code)
 	fmt.Fprint(w, message)
 }
 
-func jsonBytesResponse(w http.ResponseWriter, code int, json []byte) {
+func jsonBytesResponse(w http.ResponseWriter, code int, json []byte) { //nolint:unparam
```
jespino, line 1339: why is needed the `unparam` here? looks like all the 3 parameters are used.

server/api/api.go
```diff
@@ -1322,27 +1323,29 @@ func (a *API) errorResponseWithCode(w http.ResponseWriter, statusCode int, error
...
+	_, _ = w.Write(data)
 }
 
 func (a *API) noContainerErrorResponse(w http.ResponseWriter, sourceError error) {
-	a.errorResponseWithCode(w, http.StatusBadRequest, ERROR_NO_WORKSPACE_CODE, ERROR_NO_WORKSPACE_MESSAGE, sourceError)
+	a.errorResponseWithCode(w, http.StatusBadRequest, ErrorNoWorkspaceCode, ErrorNoWorkspaceMessage, sourceError)
 }
 
-func jsonStringResponse(w http.ResponseWriter, code int, message string) {
+func jsonStringResponse(w http.ResponseWriter, code int, message string) { //nolint:unparam
```
jespino, line 1333: why is needed the `unparam` here? looks like all the 3 parameters are used.

server/app/files.go
```diff
@@ -62,7 +68,9 @@ func (a *App) GetFileReader(workspaceID, rootID, filename string) (filestore.Rea
 	return reader, nil
 }
 
+/*
```
jespino, line 71: Another thing that I would remove instead of comment.

server/web/webserver.go
```diff
@@ -24,12 +25,12 @@ type RoutedService interface {
...
-	rootPath  string
-	port      int
-	ssl       bool
-	localOnly bool
-	logger    *mlog.Logger
+	baseURL  string
+	rootPath string
+	port     int
+	ssl      bool
+	// localOnly bool
```
jespino, line 32: If we are no longer using it, maybe we should remove it.

jespino (changes requested):
A ton of good stuff here, some tiny comments to clean up a bit more, but other than that, this is great!

jespino:
by the way, the tests are failing because the change on the name of an attribute in the files backend configuration, but wasn't catch in your environment because the compilation of the plugin is independent.

server/api/api.go
wiggin77: `unparam` also checks if any params are always called with the same value.  For example, `jsonStringResponse` is called using code=200 in all cases currently, but the method should remain generic IMO.

server/api/api.go
wiggin77: Same as below.

wiggin77:
> by the way, the tests are failing because the change on the name of an attribute in the files backend configuration, but wasn't catch in your environment because the compilation of the plugin is independent.

Thanks.  I reverted that one until it can be fixed in chat service.

server/api/api.go
jespino: That sounds right, other option is to remove the code parameter and add it when is needed, instead of giving the function a flexibility that we are not using now. Anyway... It is ok to have the `//nolint` for now. I guess sooner than later we are going to need the code there.

jespino (approved):

Issue #535, opened 2021-06-07
Bug: Space missing between "Show/hide Description" on a Board


lindy65:
Submitted on behalf of contributor `jgilliam17`

**Summary:**
Space missing between "Show/hide Description" on a Board

**Steps to reproduce the behavior:**
1. Open a board
2. Hover over the area top left above the board name
3. Observed: space missing between icon and wording (show/hide description) - see screenshot

**Expected behavior:**
A space between icon and "show/hide description"

**Screenshots:**
![image](https://user-images.githubusercontent.com/13020833/120990181-91c0b900-c780-11eb-9d79-fe6c3fc0e6a8.png)

**Platform:**
 - Browser and version: Safari 14
 - OS and version: macOS

chenilim:
Thanks @lindy65 and `jgilliam17`! Looks like a regression. Hopefully a really easy CSS fix - seems to be some padding / margin missing. Opening this up to any takers. cc @sbishel.

dineshkumar12004:
I couldn't see that show description in Board view ? How to reproduce this issue ? I'm using ubuntu 20.04

lindy65:
Hi @dineshkumar12004 - If you hover above the board name, you will see a "Show/Hide Description" appear

dineshkumar12004:
Got it now I can see 👍

PR #604, opened 2021-06-19
[GH-535] Add non-breaking space before 'show description button'

lunaticmonk:
#### Summary
Adds a space between the "show description" description and the icon on a board.

image

#### Ticket Link
https://github.com/mattermost/focalboard/issues/535


lunaticmonk:
@chenilim I have signed the CLA but I don't know how to re-trigger the workflows to let the cla/mattermost workflow pass. Can you please help me out?

lunaticmonk:
image

Here is the screenshot of the fix.

lunaticmonk:
@chenilim Can you please review?

image

webapp/src/components/viewTitle.tsx
```diff
@@ -64,6 +64,7 @@ const ViewTitle = React.memo((props: Props) => {
                         }}
                         icon={<ShowIcon/>}
                     >
+                        &nbsp;
```
jespino, line 67: I don't love the use of `&nbsp;` here, it remove certain control of the styling, do you mind to add a `.ViewTitle .add-buttons .Icon` css that add a `margin-right`, that way we can make more fine grain styling of this.

jespino (changes requested):
I prefer to go with css intead of `&nbsp` for this visual bug.

webapp/src/components/viewTitle.tsx
jespino: by the way, the right place would be the file `webapp/src/components/viewTitle.scss`

harshilsharma63:
@lunaticmonk just checking in if you have any update on this?

lunaticmonk:
@harshilsharma63 @jespino Pushed the css changes. Please review once.

webapp/src/components/viewTitle.scss
```diff
@@ -38,4 +38,8 @@
     > .description > * {
         flex-grow: 1;
     }
+
+    .non-breaking-space {
```
jespino, line 42: This is not what I was expecting, I was thinking more in something like changing the already existing `.add-buttons .Icon` class and add there the `margin-right: 0.5em`, that way you don't need the `span` element at all, and you also fix the same problem with the `Hide` and `Emoji` icons.

jespino (changes requested):

sbishel:
@lunaticmonk Just checking in if you are still working on this.

lunaticmonk:
@sbishel Yes, will update this PR soon, was a bit busy from few days at work!

lunaticmonk:
@sbishel I accidently logged out of my account on my local setup and can't remember the password. Is there a way to reset the password via some hack? I need to test these changes once. I also tried signing up again but getting:

image


Can you help me in this?

sbishel:
@lunaticmonk Sorry, I missed this message. No, we don't have a hack for that. You could hack the change password code (ie, don't require old password) and run via ChangePassword API. 

I'm not sure why you can't register a new user are you seeing any other errors?

 If you want to push it up this PR, I'm happy to test.

lunaticmonk:
@sbishel I tried doing few hacks but the page gets redirected to login and I cannot reset password. Looks like I will need your help in testing this and merging it. Meanwhile, I will set up the repo again in my system.

sbishel:
@lunaticmonk Sure, merge your changes to this PR and I'll test. The only hack would be if your running source code to put a break point in where login fails and go around it.

lunaticmonk:
@sbishel Didn't get you? Which PR should I merge my changes to? Is there any other branch you have where I should merge these?

sbishel:
@lunaticmonk This one. I thought you had additional updates to this PR based on Jesus feedback.

lunaticmonk:
@sbishel These are the only changes I believe. You can test this branch.

sbishel:
Fixed by https://github.com/mattermost/focalboard/pull/765

Issue #533, opened 2021-06-07
Bug: Filter menu does not disappear when clicked for a second time (Board > Filter)


lindy65:
Submitted on behalf of `ewwollesen`

**Summary:**
Filter menu does not disappear when clicked for a second time (Board > Filter)

**Steps to reproduce the behavior:**
If you're on a board, Personal Goals for instance, along the top right of the board there are buttons for Filter, Sort, Group By, etc. If you click on Sort it drops down a menu, and then if you click on Sort a second time the menu goes away. Same with Properties and Group By. 

That functionality isn't there for the Filter option. When you click Filter it brings down the menu as expected, but clicking it a second time does not cause the menu to go away.

chenilim:
Thanks @lindy65 and `ewwollesen`! Great catch! Yes, the filter dialog should close if the filter button is clicked while it is visible, to be consistent with the other menu buttons. Should be an easy fix. Opening this up to any takers. cc @sbishel.

AbdelilahRami:
Hello, How can I assign this issue to me ?

haardikdharma10:
Hey @AbdelilahRami, you can start working on it as it is still `Up for Grabs` :) Probably the staff won't be very active until the weekend ends so you can give it a go. You might even finish before it is assigned to you :P

AbdelilahRami:
Okay, I finished it and made a PR

xNaCly:
Is this issue not fixed, otherwise i would like to work on it?

WhiteHsu:
hi @lindy65 @haardikdharma10 @chenilim 
I'm very interested in this bug, may I take it?  :-)

lindy65:
Hi @WhiteHsu - fantastic :)  I've assigned the issue to you!

WhiteHsu:
Thanks @lindy65 

But when I am trying to build the project, there is prebuild issue as here:
https://github.com/mattermost/focalboard/discussions/4024

My you kindly suggest?

WhiteHsu:
hi @lindy65 @haardikdharma10 @chenilim

I've already submitted my PR to fix the issue on the Filter button of the Header: 
https://github.com/mattermost/focalboard/pull/4050

Please kindly help review and feedback.

In addition to my commits, the PR shows that I need to sign the CLA. May you help suggest how I should do for the next to sign it? Thank you a lot.
[...]

PR #605, opened 2021-06-19
[GH-533] fFix bug : Filter menu does not disappear when clicked for a second time (Board > Filter)

AbdelilahRami:


AbdelilahRami:
Could you please explain what I have to do ?

haardikdharma10:
Hi @AbdelilahRami, please fill in your details after reading the Mattermost [contribution license agreement](https://mattermost.org/mattermost-contributor-agreement/). Once filled and submitted, comment /check-cla on this issue. Hope this helps:)

AbdelilahRami:
Can someone please review my PR?

jespino:
The `package-lock.json` file changes are not needed and actually are changing that to a previous version, do you mind to revert that?

webapp/src/components/viewHeader/viewHeader.tsx
```diff
@@ -37,19 +37,23 @@ type Props = {
...
-
     const [viewTitle, setViewTitle] = useState(activeView.title)
 
     useEffect(() => {
         setViewTitle(activeView.title)
     }, [activeView.title])
 
     const hasFilter = activeView.filter && activeView.filter.filters?.length > 0
+    const onCloseFilter = () => {
+        document.body.addEventListener('click', (event: Event) => {
```
jespino, line 51: This is adding an even listerner on every filter close and is not removing it, also the check of the target as an string looks fragile.

I think there must be a better way to do this. What is the thinking behind this block of code?

jespino (changes requested):
Some changes needed and I need some clarification about the intention of something.

webapp/src/components/viewHeader/viewHeader.tsx
AbdelilahRami: the problem here is that when we click Filter Button it triggers 2 events onClick and onClose (because filter button is outside the dropdown menu) and that's why the dropdown remains open.
I agree about the use of string

harshilsharma63:
@AbdelilahRami are there any changes pending on this?

AbdelilahRami:
@harshilsharma63 changes requested by @jespino are already made

webapp/src/components/viewHeader/viewHeader.tsx
```diff
@@ -51,6 +51,15 @@ const ViewHeader = React.memo((props: Props) => {
...
     const hasFilter = activeView.filter && activeView.filter.filters?.length > 0
 
+    const onCloseFilter = () => {
+        const myCall = (event: Event) => {
+            if (showFilter && !buttonRef.current?.contains(event.target as Node)) {
+                setShowFilter(false)
+            }
+        }
+        document.addEventListener('click', myCall)
+        return () => document.removeEventListener('click', myCall)
```
jespino, line 61: I don't get how this un-register the event listener. This is returning the function that unregister the event listener, but this is not an effect, so... who is calling the returned function?

chenilim:
Trying to clean up stale PRs. @jespino, what is the next step here? Are there still changes required? Or can those be fixed later? Thanks.

jespino:
@chenilim I was waiting for @AbdelilahRami explanation about the unregistering of the handler.

PR #611, opened 2021-06-21
Fixed sorting icons

harshilsharma63:
#### Summary
Fixed ascending and descending sort icons.

1. Changed ascending icon to up arrow.
2. Changed descending arrow to down arrow.

Before-
before

After-
after

#### Ticket Link
https://github.com/mattermost/focalboard/issues/561

### DIFF
--- a/webapp/src/components/table/table.tsx
+++ b/webapp/src/components/table/table.tsx
@@ -215,7 +215,7 @@ const Table = (props: Props) => {
     const titleSortOption = activeView.sortOptions.find((o) => o.propertyId === Constants.titleColumnId)
     let titleSorted: 'up' | 'down' | 'none' = 'none'
     if (titleSortOption) {
-        titleSorted = titleSortOption.reversed ? 'up' : 'down'
+        titleSorted = titleSortOption.reversed ? 'down' : 'up'
     }
 
     return (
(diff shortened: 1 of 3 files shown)


harshilsharma63:
cc @haardikdharma10

webapp/src/components/table/table.tsx
```diff
@@ -215,7 +215,7 @@ const Table = (props: Props) => {
     const titleSortOption = activeView.sortOptions.find((o) => o.propertyId === Constants.titleColumnId)
     let titleSorted: 'up' | 'down' | 'none' = 'none'
     if (titleSortOption) {
-        titleSorted = titleSortOption.reversed ? 'up' : 'down'
+        titleSorted = titleSortOption.reversed ? 'down' : 'up'
```
harshilsharma63, line 218: `reversed: false` indicates descending and `reversed: true` indicated ascending across the sorting logic. So I haven't changed as there's no right or wrong answer as to whether reversed means ascending or descending.

haardikdharma10:
> cc @haardikdharma10

Thanks @harshilsharma63, I got the correct logic behind it. I'll close #576

webapp/src/components/table/table.tsx
sbishel: Yes, this is driving me nuts. We have 'reversed', 'up/down', 'asc/dsnd'.

But I agree, I think of ascending as "normal" or "up".

sbishel (approved):

PR #619, opened 2021-06-23
Add RPC API support to plugin

agnivade:
We use the mattermost-plugin-api client to create the sql.DB object
and pass it to the store.

To keep a common point of entry for both the standalone server
and plugin, we refactor the store creation part out of server.New
and pass the DB as a dependency to the server constructor.

This allow us to create different stores in plugin and standalone,
so that the server code remains unaware of any differences.

https://focalboard-community.octo.mattermost.com/workspace/zyoahc9uapdn3xdptac6jb69ic?id=285b80a3-257d-41f6-8cf4-ed80ca9d92e5&v=495cdb4d-c13a-4992-8eb9-80cfee2819a4&c=c7386db7-65fd-469b-8bcf-8dc8f8e61e4f


mattermost-plugin/server/plugin.go
```diff
@@ -103,7 +113,21 @@ func (p *Plugin) OnActivate() error {
...
+	if err != nil {
+		return fmt.Errorf("error initializing the DB: %v", err)
+	}
+	if cfg.AuthMode == server.MattermostAuthMod {
+		layeredStore, err2 := mattermostauthlayer.New(cfg.DBType, cfg.DBConfigString, db)
+		if err2 != nil {
+			return fmt.Errorf("error initializing the DB: %v", err2)
+		}
+		db = layeredStore
+	}
```
agnivade, line 128: This part is getting repeated. I think we can combine the `sqlstore.New` and `mattermostauthlayer.New` inside a single store constructor `store.New`. LMK if you want me to do that in this PR itself.

jespino: Yes that can work, I kinda prefer the manual approach, because is explicit that the store layers doesn't need to have an specific order, and you can enable/disable them bases on certain logic. But I'm not against that approach, maybe something like `NewStore`? at the server package?

agnivade: Yep, did exactly that!

linux/main.go
```diff
@@ -116,3 +129,21 @@ document.addEventListener('click', function (e) {
...
+
+func getStore(config *config.Configuration, logger *mlog.Logger) (store.Store, error) {
+	var db store.Store
+	var err error
+	db, err = sqlstore.New(config.DBType, config.DBConfigString, config.DBTablePrefix, logger, nil)
+	if err != nil {
+		return nil, err
+	}
+	if config.AuthMode == server.MattermostAuthMod {
+		layeredStore, err2 := mattermostauthlayer.New(config.DBType, config.DBConfigString, db)
```
jespino: If you haven't modify the mattermostauthlayer, it is going to create another connection to the database, so probably instead of passing the DBType and the DBConfigString, it should pass the connection already in use. (This is this way because you can use different databases, in theory to have mattermost and focalboard in personal server connected, but that is a corner case).

agnivade: Ah interesting! Wasn't aware of that. Fixing.

mattermost-plugin/server/plugin.go
```diff
@@ -103,7 +113,21 @@ func (p *Plugin) OnActivate() error {
 		EnableLocalMode:         false,
 		LocalModeSocketLocation: "",
 		AuthMode:                "mattermost",
-	}, "", logger)
+	}
+	var db store.Store
+	db, err = sqlstore.New(cfg.DBType, cfg.DBConfigString, cfg.DBTablePrefix, logger, sqlDB)
```
jespino, line 118: Maybe we can pass to the `NewStore` constrator the `sqlDB` and that way this gets simplified too.

agnivade: Hmm .. yeah I thought about that. It would cause "sql.Open + db.Ping" code to be repeated in 4 places. I think a reasonable cleanup would be to always pass `sql.DB` to sqlstore.New and move the db creation into `NewStore`. That way the `if db == nil` in `sqlstore.New` logic gets cleared up.

So the only difference remains in the plugin path and server path. That keeps things separate enough.

agnivade:
Ok, all cleaned up now. PTAL. It would be also good for someone to run a quick sanity check as a standalone + plugin mode. I tested it myself, but was getting the ws path errors. Perhaps there's a test setup I am missing. Would be good to have someone from the team verify.

jespino (approved):
LGTM

spirosoik (approved):
good job 🚀

agnivade:
@jespino - Who would be a good person to QA this?

spirosoik:
> @jespino - Who would be a good person to QA this?

I think @agnivade we should test this too cc @stylianosrigas

mgdelacroix (approved):
LGTM!! :+1:

agnivade:
Hey @mgdelacroix / @jespino - I'd like someone to just give a 2 min quick sanity check in standalone + plugin mode before I merge this. Can you point me to someone who can help me with this?

I tested the plugin path myself, but it wasn't very thorough. Just want a second set of eyes for a more thorough check :)

jespino:
@agnivade I can give it a try during my morning :)

jespino:
It is working fine, I can access the database properly on standalone and plugin. I'm accesing mattermost information and focalboard information (read and write) correctly from the plugin, so, looks OK. Also the same for standalone, I can do things in the main tables :)

jespino:
The Only problem I see is, this is going to need 5.37, right?

agnivade:
Awesome, thanks @jespino !

> The Only problem I see is, this is going to need 5.37, right? I think we should add that right away to the plugin.json

Yep, correct :) I tried to do that in the IC plugin. But unfortunately, 5.37 branch hasn't been cut, so I cannot add that yet. What I did was merge the commit to master, with the promise that once the branch is cut, I will update the plugin.json. It shouldn't be a problem either way because it will only be released along with 5.37. And for master, developers can just use master.

Does that work?

jespino:
Yes, that works!, lets merge it! :tada:

agnivade:
Awesome, really excited to see this go live!

PR #637, opened 2021-06-26
Description Section of Card can now Have Columns

hahmadia:
#### Summary

Make description section be able to have columns similar to notion

#### Ticket Link

https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=baa742b3-f3f1-406e-811a-00f18d723781&c=8737ff17-988f-49fd-87b3-178b38315dd8

# This works the same way as Notion's grid system with the exception of not having columns within a row. Currently only 2 dimensions (Columns and Rows)

Drag to Location

image

And then ... BAM

image


harshilsharma63:
@hahmadia this is still WIP, right?

hahmadia:
@harshilsharma63 Yes

harshilsharma63:
@hahmadia okay. I've marked this PR as draft as it was showing up in pending review forever I hope its fine.

webapp/src/blocks/card.ts
```diff
@@ -8,7 +8,7 @@ interface Card extends IBlock {
     readonly icon: string
     readonly isTemplate: boolean
     readonly properties: Readonly<Record<string, string | string[]>>
-    readonly contentOrder: readonly string[]
+    readonly contentOrder: readonly (string|string[])[]
```
jespino, line 11: I guess this is because backward compatibility, but this type looks weird.

Maybe would be simpler if we write that as `string[]|string[][]`

webapp/src/components/contentBlock.tsx
```diff
@@ -27,29 +28,35 @@ import './contentBlock.scss'
...
+    cords: {x: number, y?: number, z?: number}
 }
 
 const ContentBlock = React.memo((props: Props): JSX.Element => {
-    const {card, contents, block, readonly} = props
+    const {card, block, readonly, cords} = props
     const intl = useIntl()
-    const [isDragging, isOver, gripRef, itemRef] = useSortableWithGrip('content', block, true, props.onDrop)
+    const [, , gripRef, itemRef] = useSortableWithGrip('content', {block, cords}, true, () => {})
+    const [, isOver2,, itemRef2] = useSortableWithGrip('content', {block, cords}, true, (src, dst) => props.onDrop(src, dst, 'right'))
```
jespino, line 41: why are we using the `useSortableWithGrip` instead of `useSortable` if we are not using the Grip?

webapp/src/viewModel/cardTree.ts
```diff
@@ -12,19 +12,19 @@ import {OctoUtils} from '../octoUtils'
 interface CardTree {
     readonly card: Card
     readonly comments: readonly CommentBlock[]
-    readonly contents: readonly IContentBlock[]
+    readonly contents: readonly (IContentBlock|IContentBlock[])[]
```
jespino, line 15: same type problem here, it is hard to understand, maybe we can express this in other way.

jespino (changes requested):
Some questions, more than suggestions.

webapp/src/blocks/card.ts
hahmadia: Done!

webapp/src/viewModel/cardTree.ts
hahmadia: Done

webapp/src/components/contentBlock.tsx
jespino: Talked this on mattermost, this is right.

jespino (approved):
LGTM

harshilsharma63 (approved):

harshilsharma63:
One minor fix.

There is some extra spacing around the "Insert Above" text and icon.
Jul-15-2021 11-06-00

hahmadia:
@harshilsharma63 

Hey Harshil,

Thanks for pointing that out. I looked into it and noticed it was also happening on master. For that reason, I think it's best for us to create a separate ticket and address this separately.

Issue #629, opened 2021-06-24
Feature Idea: Property Popover in Kanban view


hahmadia:
**Summary:**
What the new capability is.
Hovering over a property in Kanban view and having it display what property it is. Currently there is no way of knowing what property is what without explicitly opening the card and inspecting the properties.

**How important this is to me and why:**

Importance: Medium

Use cases:
1. Getting more detailed view while in Kanban view

**Additional context / similar features:**
Any examples of good implementations of this capability.

Focalboard UX member, Michael Gamble, has proposed that the popover should look like this (where it shows priority). 
![image](https://user-images.githubusercontent.com/17804942/123264308-e75acc80-d4c7-11eb-8b98-3323029dc5dd.png)

For example, if I hover on the `TODO` in the image below, a popover should display showing `Status` since that's the property.

![image](https://user-images.githubusercontent.com/17804942/123264971-8ed7ff00-d4c8-11eb-9024-4e29fb255040.png)

chenilim:
Thanks @hahmadia! Opening this up for grabs. The easy fix would be to add a `title` property to the property values / tags in the card preview. Bonus points if the popover can be styled as above.

lordinkavu:
Can I take this up ?

hahmadia:
@lordinkavu 

> Can I take this up ? 

Sure thing! Good luck and have fun!

lordinkavu:
@ditsemto No issues. I've unassigned myself @hahmadia

hahmadia:
Sorry about that @lordinkavu . Thanks for the understanding.

PR #643, opened 2021-06-28
GH-629 - Property Popover in Kanban view

ditsemto:
#### Summary

Adds a tool-tip popover above the elements of a Kanban card.
The property name is shown in a tool-tip above the element.
Only readonly elements show the tooltip.

The style was added to the main project stylesheet.
A data attribute is used as the tooltip text.

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/629


hahmadia:
Hi @ditsemto ,

Could you please fix the broken tests that are failing in the CI pipeline? Thanks!

ditsemto:
Hi @hahmadia, the power has been out in my area for the last 6 hours. I apologize for the inconvenience.

I'll get on at as soon as it is back.

webapp/src/components/properties/link/link.tsx
```diff
@@ -15,10 +15,13 @@ type Props = {
     onSave: () => void
     onCancel: () => void
     validator: (newValue: string) => boolean
+    tooltip?: string
 }
 
 const URLProperty = (props: Props): JSX.Element => {
     let link: ReactNode = null
+    const tooltipClassName = props.tooltip ? 'octo-tooltip tooltip-top' : ''
```
harshilsharma63, line 23: This code would repeat over and over in any new property component that gets added in the future. Also since the logic for adding and generation tooltip is independent of property type, I'd like to remove it from each individual component.

Instead, I think we can add a wrapper `div` with tooltip over here - 

https://github.com/mattermost/focalboard/blob/6cd5003a7aa67ff11fb1bcaca4a46914a8730428/webapp/src/components/kanban/kanbanCard.tsx#L78

This is where the visible properties are added to the kanban card component. That part of the code is only used for readonly view so we can skip the logic of deciding whether to add the tooltip or not. This will have the following advantages -

1. No repeating code in each property component (there are a bunch of new ones being added in the open PRs)
2. Tooltips for new types gets added automatically without any code change.

This approach can also be applied to the gallery view which suffers from the same problem.

webapp/src/styles/main.scss
```diff
@@ -240,3 +240,123 @@ h1 {
         width: 126px;
     }
 }
+
+// Tooltip arrow width
```
harshilsharma63, line 244: Great work here!

Can we move this to a separate file to keeps things better structured and managable?

harshilsharma63 (changes requested):
Awesome work @ditsemto 🎉 I really love how the cards look with the tooltip.

I've added a change request for the approach we've taken here for adding the tooltips. I think with the newly suggested approach the code changes will be less than half what we have right now.

Let me know if you have any questions.

webapp/src/components/properties/link/link.tsx
ditsemto: Thanks @harshilsharma63, this is a better way of handling it, I'll will update it.

webapp/src/styles/main.scss
ditsemto: Will do :+1:

webapp/src/components/properties/link/link.tsx
ditsemto: Hi @harshilsharma63,

With regards to adding the same functionality to the gallery view, should one create a new ticket for it so there is a track record of the change or should I add it while I am busy?

Currently I only updated the current Kanban card.

webapp/src/widgets/tooltip.tsx
```diff
@@ -0,0 +1,28 @@
...
+// be positioned based on the specified placement
+// Default position is 'top'
+function Tooltip(props: Props): JSX.Element {
+    const placement = props.placement ? props.placement : 'top'
+    const className = `octo-tooltip tooltip-${placement}`
+    return (
+        <div
+            className={className}
+            data-tooltip={props.tooltip}
+        >{props.children}
```
hahmadia, line 23: Could we move `{props.children}` to the next line?

webapp/src/widgets/tooltip.tsx
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import React from 'react'
+
+import './tooltip.scss'
+
+type Props = {
+    tooltip: string
```
hahmadia, line 8: Let's rename this to `title`
```suggestion
    title: string
```

webapp/src/components/kanban/kanbanCard.tsx
```diff
@@ -75,13 +76,17 @@ const KanbanCard = React.memo((props: Props) => {
...
             {visiblePropertyTemplates.map((template) => (
-                <PropertyValueElement
+                <Tooltip
                     key={template.id}
-                    readOnly={true}
-                    card={card}
-                    propertyTemplate={template}
-                    emptyDisplayValue=''
-                />
+                    tooltip={template.name}
```
hahmadia, line 81:
```suggestion
                    title={template.name}
```

webapp/src/widgets/tooltip.tsx
```diff
@@ -0,0 +1,28 @@
...
+    children: any
```
hahmadia, line 9: I believe this shouldn't be `any` but instead `React.ReactNode`.

webapp/src/widgets/tooltip.tsx
```diff
@@ -0,0 +1,28 @@
...
+    placement?: 'top'|'left'|'right'|'bottom'
+}
+
+// Adds tooltip div over children elements, the popup will
+// be positioned based on the specified placement
+// Default position is 'top'
+function Tooltip(props: Props): JSX.Element {
+    const placement = props.placement ? props.placement : 'top'
```
hahmadia, line 17: Can be simplified
```suggestion
    const placement =  props.placement || 'top'
```

hahmadia (changes requested):
I tried it and it's great! I love it! Thanks a lot!
I just have some minor nit changes to make but usability is awesome. Thanks!

webapp/src/components/properties/link/link.tsx
hahmadia: If it's not too much extra work, I think getting gallery view done within this same PR would be great unless you are busy and don't have the time to do it which we can create a new PR for it. Thoughts @harshilsharma63? I'll make it your call but I ultimately think it makes sense to do it one PR.

harshilsharma63: @ditsemto sorry I missed your reply. If it's a quick thing then we can add it for gallery view as well. If not, we can handle it separately.

webapp/src/widgets/tooltip.tsx
ditsemto: Will do :+1:

webapp/src/widgets/tooltip.tsx
ditsemto: Sure thing

webapp/src/widgets/tooltip.tsx
ditsemto: Thanks for this

webapp/src/widgets/tooltip.tsx
ditsemto: Definitely cleaner thanks :smile:

webapp/src/components/properties/link/link.tsx
ditsemto: @harshilsharma63 No worries. 

I made the requested changes on one commit and updated the gallery view in the second commit.

harshilsharma63 (approved):
Awesome work!

LGTM 🎉

harshilsharma63:
@hahmadia the changes look good to me. We can merge once you approve.

hahmadia (approved):

hahmadia:
Thanks!!

PR #644, opened 2021-06-28
Adding initial usage of compass-components

jespino:
This is an initual proof of concept using the compass-components.


michelengelen (approved):
LGTM

nevyangelova (approved):
As per discussed earlier we will make the onClick prop optional so we have to keep in mind to change it here as well. Will let you know 👍

harshilsharma63:
@jespino can you fix the error in the Ubuntu CI job?

jespino:
We are no longer moving forward with this for now, maybe we can redo this work in the future.

Issue #616, opened 2021-06-22
Bug: Mattermost SqlSetting.MaxOpenConns is not respected


Vorlif:
**Summary:**
We have a Mattermost instance with few users and the Focalboard plugin running. We use PostgreSQL as database and allow a maximum of 100 connections. Since we still use other applications, the number of connections for Mattermost is limited to 50. Unfortunately, Mattermost does not respect this limit and often uses all connections. This results in other applications not being able to connect to the database. 

If I understand the program code correctly, Focalboard establishes its own connection to the database. I therefore think that this is where the problem arises. 

**Steps to reproduce the behavior:**
1. Install Mattermost with Focalboard Plugin
2. Set the number of database connections to a low value
3. Wait a few hours

**Expected behavior:**
The Mattermost plugin respects the limit set in the Mattermost configuration

**Platform:**
 - Mattermost-Version: 5.36.0
 - Focalboard (focalboard - 0.7.0)
 - PostgreSQL 13
 - Server: Debian GNU/Linux 10 (buster)

**Additional context:**

```sql
select count(*) from pg_stat_activity where usename = 'mattermost';

95

select query, state from pg_catalog.pg_stat_activity where usename = 'mattermost'
```

```
query                                                                               |state|
------------------------------------------------------------------------------------+-----+
SELECT COUNT(*) AS count FROM focalboard_workspaces                                 |idle |
SELECT COUNT(*) AS count FROM focalboard_workspaces                                 |idle |
SELECT COUNT(*) AS count FROM focalboard_workspaces                                 |idle |
SELECT COUNT(*) AS count FROM focalboard_workspaces                                 |idle |
SELECT COUNT(*) AS count FROM focalboard_workspaces                                 |idle |
SELECT COUNT(*) AS count FROM focalboard_workspaces                                 |idle |
SELECT COUNT(*) AS count FROM focalboard_workspaces                                 |idle |
...
```

chenilim:
Thanks @Vorlif! Is this something you can help investigate @jespino? Or redirect as needed? Thanks. cc @sbishel.

chenilim:
Updating with the [current plan of record](https://community.mattermost.com/core/pl/h3c45szb7p8ttm5d1mrkkirwao):

Add the following settings and defaults to `config.json`:
* MaxOpenConns: 300
* MaxIdleConns: 20
* ConnMaxLifetimeMilliseconds: 3600000 (60 minutes)
[...]

PR #645, opened 2021-06-28
Adding max-db-conns and max-idle-conns through environment variables

jespino:
This adds the environment variables to limit the number of connections to the database. This is a temporary fix, and this works just fine in regular on prem service, but in mattermost-plugin, because the database connection is "duplibcated" the value added in the environment variable is going to be duplicated.

This is only for the 0.7.1 version, the 0.8.0 version is going to come with the new shared db connections provided by @agnivade, so this can be perfectly deprecated.

### DIFF
--- a/server/services/store/mattermostauthlayer/mattermostauthlayer.go
+++ b/server/services/store/mattermostauthlayer/mattermostauthlayer.go
@@ -42,14 +42,24 @@ func New(dbType, connectionString string, store store.Store) (*MattermostAuthLay
 	}
 	maxDBIdleConns, err := strconv.Atoi(os.Getenv("FOCALBOARD_DB_MAX_IDLE_CONNS"))
 	if err != nil {
-		maxDBIdleConns = 0
+		maxDBIdleConns = 20
 	}
 	maxDBOpenConns, err := strconv.Atoi(os.Getenv("FOCALBOARD_DB_MAX_OPEN_CONNS"))
 	if err != nil {
-		maxDBIdleConns = 0
+		maxDBOpenConns = 300
+	}
+	maxDBIdleTime, err := strconv.Atoi(os.Getenv("FOCALBOARD_DB_MAX_IDLE_TIME"))
+	if err != nil {
+		maxDBIdleTime = 300
+	}
+	maxDBLifetime, err := strconv.Atoi(os.Getenv("FOCALBOARD_DB_MAX_LIFETIME"))
+	if err != nil {
+		maxDBLifetime = 3600
 	}
 	db.SetMaxIdleConns(maxDBIdleConns)
 	db.SetMaxOpenConns(maxDBOpenConns)
+	db.SetConnMaxIdleTime(time.Duration(maxDBIdleTime) * time.Second)
+	db.SetConnMaxLifetime(time.Duration(maxDBLifetime) * time.Second)
 
 	err = db.Ping()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -38,6 +40,16 @@ func New(dbType, connectionString string, store store.Store) (*MattermostAuthLay
 
 		return nil, err
 	}
+	maxDBIdleConns, err := strconv.Atoi(os.Getenv("FOCALBOARD_DB_MAX_IDLE_CONNS"))
+	if err != nil {
+		maxDBIdleConns = 0
+	}
+	maxDBOpenConns, err := strconv.Atoi(os.Getenv("FOCALBOARD_DB_MAX_OPEN_CONNS"))
+	if err != nil {
+		maxDBIdleConns = 0
```
chenilim, line 49: This should set `maxDBOpenConns`. Can we use a reasonable default instead of 0 (unlimited)? Per #616, suggest the following defaults:
* MaxOpenConns: 300
* MaxIdleConns: 20
* ConnMaxLifetimeMilliseconds: 3600000 (60 minutes)
* ConnMaxIdleTimeMilliseconds: 300000 (5 minutes)

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -38,6 +40,16 @@ func New(dbType, connectionString string, store store.Store) (*MattermostAuthLay
 
 		return nil, err
 	}
+	maxDBIdleConns, err := strconv.Atoi(os.Getenv("FOCALBOARD_DB_MAX_IDLE_CONNS"))
+	if err != nil {
+		maxDBIdleConns = 0
```
chenilim, line 45: Recommend default of 20 here, unless you see an issue?

server/services/store/sqlstore/sqlstore.go
```diff
@@ -33,6 +35,16 @@ func New(dbType, connectionString string, tablePrefix string, logger *mlog.Logge
 
 		return nil, err
 	}
+	maxDBIdleConns, err := strconv.Atoi(os.Getenv("FOCALBOARD_DB_MAX_IDLE_CONNS"))
+	if err != nil {
+		maxDBIdleConns = 0
+	}
+	maxDBOpenConns, err := strconv.Atoi(os.Getenv("FOCALBOARD_DB_MAX_OPEN_CONNS"))
+	if err != nil {
+		maxDBIdleConns = 0
```
chenilim, line 44: Should set `maxDBOpenConns` here, using the same defaults as above.

chenilim (changes requested):
Thanks Jesus! Can we set reasonable defaults unless you see an issue? Do we need to specify the env vars anywhere (e.g. dockerfile) for the Cloud team to set up? cc @gabrieljackson.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
jespino: :+1:

chenilim (approved):
Thanks Jesus. LGTM 👍!

gabrieljackson (approved):
This should hold us over well!

sbishel (approved):

mgdelacroix (approved):

PR #646, opened 2021-06-29
Sample using @reduxjs/toolkit

jespino:
this is a POC and needs some extra work before is merged.


webapp/src/store/currentWorkspace.ts
```diff
@@ -0,0 +1,36 @@
...
+const currentWorkspaceSlice = createSlice({
+    name: 'currentWorkspace',
+    initialState: {value: null} as {value: IWorkspace|null},
+    reducers: {
+        setWorkspace: (state, action: PayloadAction<IWorkspace>) => {
+            state.value = action.payload
+        },
+    },
+    extraReducers: (builder) => {
+        builder.addCase(fetchCurrentWorkspace.fulfilled, (state, action) => {
```
mgdelacroix, line 25: Why not typing the action here? (same in the other `extraReducers`)

webapp/src/store/language.ts
```diff
@@ -0,0 +1,45 @@
...
+import {RootState} from './index'
+
+export const fetchLanguage = createAsyncThunk(
+    'language/fetch',
+    async () => getCurrentLanguage(),
+)
+
+export const storeLanguage = createAsyncThunk(
+    'language/store',
+    async (lang: string) => {
```
mgdelacroix, line 17: This function doesn't seem to need `async`:

```suggestion
    (lang: string) => {
```

webapp/src/viewModel/boardTree.ts
```diff
@@ -55,57 +55,54 @@ class MutableBoardTree implements BoardTree {
...
+    usersById: {[key: string]: IUser}
 
     get allBlocks(): IBlock[] {
         return [this.board, ...this.views, ...this.allCards, ...this.cardTemplates, ...this.rawBlocks]
     }
 
-    constructor(board: MutableBoard) {
+    constructor(board: MutableBoard, usersById: {[key: string]: IUser}) {
         this.board = board
+        this.usersById = usersById
```
mgdelacroix, line 66: Should this use `setUsersById(usersById)` instead of a direct assign?

mgdelacroix (approved):
Amazing work!! Feels very idiomatic and well structured :rocket: 

Just added a couple of minor nonblocking comments

webapp/src/viewModel/boardTree.ts
jespino: Not very important, actually that new function is not used anywhere, so I can remove the function.

webapp/src/store/currentWorkspace.ts
jespino: What do you mean?

PR #649, opened 2021-06-29
Remove showView and showBoard property drilling and migrate centerPanel to functional component

jespino:


webapp/src/app.tsx
```diff
@@ -66,18 +66,18 @@ const App = React.memo((): JSX.Element => {
                             <Route path='/change_password'>
                                 <ChangePasswordPage/>
                             </Route>
-                            <Route path='/shared'>
+                            <Route path='/shared/:boardId?/:viewId?'>
```
jespino, line 69: @chenilim I'm changin this to simplify the approach and allow to move that logic into the router, like any other param. Do you see any problem on that (it is probably breaking the old links, I can try to make it backward compatible).

chenilim: Sounds good. Please add code to redirect the old format to the new one, and a note in the changelog of this breaking change.

jespino: Done.

webapp/src/pages/boardPage.tsx
```diff
@@ -41,25 +39,28 @@ class BoardPage extends React.Component<Props, State> {
...
-        let viewId = queryString.get('v') || ''
+        const queryBoardId = queryString.get('id')
+        const queryViewId = queryString.get('v')
+        if (queryBoardId) {
+            const params = {...props.match.params, boardId: queryBoardId}
+            if (queryViewId) {
+                params.viewId = queryViewId
+            }
+            const newPath = generatePath(props.match.path, params)
+            props.history.push(newPath)
```
mgdelacroix, line 52: if we're changing the URL, should we not return at this point?

webapp/src/pages/boardPage.tsx
```diff
@@ -41,25 +39,28 @@ class BoardPage extends React.Component<Props, State> {
...
-            if (boardId) {
-                Utils.replaceUrlQueryParam('id', boardId)
-            }
-            if (viewId) {
-                Utils.replaceUrlQueryParam('v', viewId)
-            }
+            const boardId = localStorage.getItem('lastBoardId') || ''
+            const viewId = localStorage.getItem('lastViewId') || ''
+            const newPath = generatePath(props.match.path, {...props.match.params, boardId, viewId})
+            props.history.push(newPath)
```
mgdelacroix, line 60: same here

webapp/src/components/sidebar/sidebarBoardItem.tsx
```diff
@@ -49,11 +60,13 @@ const SidebarBoardItem = React.memo((props: Props) => {
             intl.formatMessage({id: 'Mutator.duplicate-board', defaultMessage: 'duplicate board'}),
             false,
             async (newBoardId) => {
-                props.showBoard(newBoardId)
+                const newPath = generatePath(match.path, {...match.params, boardId: newBoardId || ''})
```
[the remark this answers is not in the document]
jespino: good point

webapp/src/pages/boardPage.tsx
jespino: I'm not 100% sure, this is the constructor of the BoardPage Component, that maybe is reused even if the route has changed, so I think the right thing is to entirely initialize the component and let the router decide if something needs to change.

webapp/src/components/sidebar/sidebarBoardItem.tsx
```diff
@@ -22,15 +23,25 @@ import './sidebarBoardItem.scss'
...
+
+    const showBoard = useCallback((boardId) => {
+        const newPath = generatePath(match.path, {...match.params, boardId: boardId || ''})
+        history.push(newPath)
+    }, [match, history])
+
+    const showView = useCallback((viewId, boardId) => {
+        const newPath = generatePath(match.path, {...match.params, boardId: boardId || '', viewId: viewId || ''})
+        history.push(newPath)
+    }, [match, history])
```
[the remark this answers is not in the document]
jespino: This is kinda an special case for path generation because there are so many ways to access a board. In normal circunstances we don't need this approach to generate the route. I think the amount of repetition is still low and I think is acceptable, we can refactor it later if we see the pattern happening again and again.

webapp/src/pages/boardPage.tsx
mgdelacroix: oks, happy to refactor this later if it actually becomes a problem

mgdelacroix (approved):
LGTM! :+1:

PR #656, opened 2021-06-30
Use cross-env instead of win-node-env

kamre:
#### Summary
These changes allow to start test server that is used for Cypress tests via npm script `runserver-test` on Windows.

#### Whats changed:
* `cross-env` is added as dev dependency and used for setting environment variables in npm scripts
* webpack production config file `webpack.js` renamed to `webpack.prod.js` so that it doesn't interfere with webpack-cli command
* all mentions of `win-node-env` removed from docs and github action scripts

#### Steps to run Cypress tests on Windows:
* `make prebuild`
* `make webapp`
* `make server-win`
* `cp bin/win/focalboard-server.exe bin`
* `cd webapp && npm run cypress:ci`


kamre:
@chenilim @jespino what do you think about these changes? Can we make developer experience on Windows a bit better?

jespino (approved):
Looks good to me. @kamre I have no windows, but looks like a good idea, I would need @chenilim to confirm that.

kamre:
@chenilim did you have a chance to look into this?

chenilim:
Sorry @kamre, I just got around to setting up my Win dev environment again. What setup did you use on Windows? E.g. git-bash or something else? I actually tried cross-env initially several months ago, and vaguely remember it has issues on a particular setup (may have been Mac, or a flavor of Windows shell?).

kamre:
@chenilim yes, I am using `git-bash` console and also have setup [make from Chocolatey](https://community.chocolatey.org/packages/make) to work with `Makefile`

chenilim (approved):
Thanks again @kamre, and sorry for the delay! I was able to build with this change successfully on Windows and Mac. Also noted that the dev-release build passes. This is much cleaner now!

Issue #650, opened 2021-06-29
Bug: Focalboard plugin server error when dragging multiple cards


chenilim:
**Steps to reproduce the behavior:**
1. Set up an on-prem Mattermost v5.36 server
  a. Or register and login on the [test server I used](http://ec2-54-209-196-240.compute-1.amazonaws.com/signup_user_complete/?id=ab3o8ftp1ibw3cy7qs3hptxc9h)
2. install the Mattermost Plugin (0.6.7+)
3. Open Focalboard and add a Personal Tasks (or similar) template board
4. Drag cards to different columns - this works
5. Drag 3 or more cards to the same column
6. Open the Chrome Dev tools console (Cmd+Shift+I)
7. Press shift, click on the first card, then click on the last card to select all cards in the column
8. Drag 3+ cards to another column
9. The drag fails
10. Notice errors in the dev console

Repeat steps 7-8 if this doesn't repro.

**Expected behavior:**
This should work.

**Platform:**
- Server: 5.36-rc2
 - Browser and version: Chrome 91 on macOS 10.15.7

**Additional context:**
The following errors are seen in the System Console log:
```
{"level":"error","ts":1624988327.5582626,"caller":"mlog/log.go:251","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"focalboard","error":"unexpected EOF"}
{"level":"error","ts":1624988327.5592828,"caller":"mlog/log.go:251","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"focalboard","error":"unexpected EOF"}
{"level":"error","ts":1624988327.5596976,"caller":"http/server.go:3093","msg":"http: response.WriteHeader on hijacked connection from github.com/mattermost/mattermost-server/v5/plugin.(*hooksRPCClient).ServeHTTP (client_rpc.go:337)","source":"httpserver"}
{"level":"error","ts":1624988327.5605214,"caller":"http/server.go:3093","msg":"http: response.Write on hijacked connection from fmt.Fprintln (print.go:265)","source":"httpserver"}
{"level":"error","ts":1624988327.5608683,"caller":"mlog/log.go:251","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"focalboard","error":"unexpected EOF"}
```

chenilim:
[...]

PR #658, opened 2021-06-30
Prevented concurrent writes to websocket

harshilsharma63:
#### Summary
Prevented concurrent writes to websockets. This fixes the the plugin crash occurring when dragging and dropping multiple cards together.

The problem is caused by gorilla WebSocket not support concurrent writes to WebSockets - https://github.com/gorilla/websocket/issues/119

#### Ticket Link
#650

### DIFF
--- a/server/ws/websockets.go
+++ b/server/ws/websockets.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"log"
+	"net"
 	"net/http"
 	"sync"
 	"time"
@@ -34,6 +35,14 @@ func (c *wsClient) WriteJSON(v interface{}) error {
 	defer c.lock.Unlock()
 	err := c.conn.WriteJSON(v)
 	return err
+}
+
+func (c *wsClient) RemoteAddr() net.Addr {
+	return c.conn.RemoteAddr()
+}
+
+func (c *wsClient) Close() error {
+	return c.conn.Close()
 }
 
 // Server is a WebSocket server.
@@ -211,7 +220,7 @@ func (ws *Server) authenticateListener(wsSession *websocketSession, workspaceID,
 	// Authenticate session
 	isValidSession := ws.isValidSessionToken(token, workspaceID)
 	if !isValidSession {
-		wsSession.client.conn.Close()
+		wsSession.client.Close()
 		return
 	}
 
@@ -262,7 +271,7 @@ func (ws *Server) addListener(wsSession *websocketSession, command *WebsocketCom
 	workspaceID, err := ws.getAuthenticatedWorkspaceID(wsSession, command)
 	if err != nil {
 		ws.logger.Error("addListener: NOT AUTHENTICATED", mlog.Err(err))
-		ws.sendError(wsSession.client.conn, "not authenticated")
+		ws.sendError(wsSession.client, "not authenticated")
 		return
 	}
 
@@ -300,7 +309,7 @@ func (ws *Server) removeListenerFromBlocks(wsSession *websocketSession, command 
 	workspaceID, err := ws.getAuthenticatedWorkspaceID(wsSession, command)
 	if err != nil {
 		ws.logger.Error("addListener: NOT AUTHENTICATED", mlog.Err(err))
-		ws.sendError(wsSession.client.conn, "not authenticated")
+		ws.sendError(wsSession.client, "not authenticated")
 		return
 	}
 
@@ -327,15 +336,15 @@ func (ws *Server) removeListenerFromBlocks(wsSession *websocketSession, command 
 	ws.mu.Unlock()
 }
 
-func (ws *Server) sendError(conn *websocket.Conn, message string) {
+func (ws *Server) sendError(wsClient *wsClient, message string) {
 	errorMsg := ErrorMsg{
 		Error: message,
 	}
 
-	err := conn.WriteJSON(errorMsg)
+	err := wsClient.WriteJSON(errorMsg)
 	if err != nil {
 		ws.logger.Error("sendError error", mlog.Err(err))
-		conn.Close()
+		wsClient.Close()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


server/ws/websockets.go
```diff
@@ -409,11 +410,14 @@ func (ws *Server) BroadcastBlockChange(workspaceID string, block model.Block) {
 				mlog.Stringer("remoteAddr", listener.RemoteAddr()),
 			)
 
+			ws.broadcastMutex.Lock()
```
jespino, line 413: This is locking the whole websocket, instead of locking the listener individually, I think this can enter like this in the repo but we need a more fine-grained locks to avoid cotention here.

jespino (approved):

server/ws/websockets.go
harshilsharma63: I see. We should prevent concurrent writes on a single connection but allow them on different connections.

Let me update that.

server/ws/websockets.go
```diff
@@ -24,10 +24,15 @@ type Hub interface {
 	SetReceiveWSMessage(func(data []byte))
 }
 
+type Listener struct {
```
jespino, line 27: Everything here can be private (lowercase).

server/ws/websockets.go
```diff
@@ -24,10 +24,15 @@ type Hub interface {
...
+	Lock *sync.RWMutex
+	Conn *websocket.Conn
```
jespino, line 29: I was thinking about embedding it, but I guess there is not a problem about using in this way, and probably it is just simpler.

server/ws/websockets.go
harshilsharma63: Done.

server/ws/websockets.go
harshilsharma63: 👍

server/ws/websockets.go
harshilsharma63: Done.

server/ws/websockets.go
```diff
@@ -258,10 +263,10 @@ func (ws *Server) addListener(wsSession *websocketSession, command *WebsocketCom
 	for _, blockID := range command.BlockIDs {
 		itemID := makeItemID(workspaceID, blockID)
 		if ws.listeners[itemID] == nil {
-			ws.listeners[itemID] = []*websocket.Conn{}
+			ws.listeners[itemID] = []*Listener{}
 		}
 
-		ws.listeners[itemID] = append(ws.listeners[itemID], wsSession.client)
+		ws.listeners[itemID] = append(ws.listeners[itemID], &Listener{&sync.RWMutex{}, wsSession.client})
```
jespino, line 269: There is a problem with this approach, you are sharing the `wsSessoin.client` and adding it to different listeners, that each one has its own Mutex, but the client is shared, so... you have still the problem of multiple writes from different places.

Probably you need to make the `websocketSession.client` a `Listener` and reuse that. (Probably the name of the type should be converted to something like `wsClient` instead of `Listener`).

server/ws/websockets.go
jespino: there is no reason to make the `Listener` struct public either.

server/ws/websockets.go
harshilsharma63: Done.

server/ws/websockets.go
harshilsharma63: Done.

jespino (changes requested):
Some small improvements and a place where the bug is not really fixed.

server/ws/websockets.go
```diff
@@ -199,7 +211,7 @@ func (ws *Server) authenticateListener(wsSession *websocketSession, workspaceID,
 	// Authenticate session
 	isValidSession := ws.isValidSessionToken(token, workspaceID)
 	if !isValidSession {
-		wsSession.client.Close()
+		wsSession.client.conn.Close()
```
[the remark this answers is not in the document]
harshilsharma63: Done.

server/ws/websockets.go
```diff
@@ -346,19 +358,19 @@ func (ws *Server) SetHub(hub Hub) {
 		}
 
 		for _, listener := range listeners {
-			log.Printf("Broadcast change, workspaceID: %s, blockID: %s, remoteAddr: %s", msg.WorkspaceID, msg.BlockID, listener.RemoteAddr())
+			log.Printf("Broadcast change, workspaceID: %s, blockID: %s, remoteAddr: %s", msg.WorkspaceID, msg.BlockID, listener.conn.RemoteAddr())
```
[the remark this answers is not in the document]
harshilsharma63: Done.

server/ws/websockets.go
```diff
@@ -288,7 +300,7 @@ func (ws *Server) removeListenerFromBlocks(wsSession *websocketSession, command
 	workspaceID, err := ws.getAuthenticatedWorkspaceID(wsSession, command)
 	if err != nil {
 		ws.logger.Error("addListener: NOT AUTHENTICATED", mlog.Err(err))
-		ws.sendError(wsSession.client, "not authenticated")
+		ws.sendError(wsSession.client.conn, "not authenticated")
```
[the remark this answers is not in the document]
harshilsharma63: Done.

server/ws/websockets.go
```diff
@@ -270,10 +291,10 @@ func (ws *Server) addListener(wsSession *websocketSession, command *WebsocketCom
 func (ws *Server) removeListener(client *websocket.Conn) {
```
jespino, line 291: nit: probably we can make this method receiving also a `wsClient` instead of `websocket.Conn`, that simplifies the comparison later, and unifies all the code.

server/ws/websockets.go
```diff
@@ -114,7 +135,7 @@ func (ws *Server) handleWebSocketOnChange(w http.ResponseWriter, r *http.Request
 	}
 
 	wsSession := websocketSession{
-		client:          client,
+		client:          &wsClient{client, &sync.RWMutex{}},
```
jespino, line 138: If we follow the other suggestion, probably we want to wrap the client ASAP, just before the defer that remove the listener and close it later.

jespino (approved):
I tiny improvement that I think would make it a bit more consistent, but just as an optional improvement. Everything looks good to me now.

mgdelacroix (approved):
LGTM, besides @jespino's comments

server/ws/websockets.go
harshilsharma63: Done. Made sure we're using WS connection via our wrapper.

server/ws/websockets.go
harshilsharma63: Done.

jespino (approved):
LGTM

harshilsharma63:
This fixes #650

PR #660, opened 2021-06-30
Implement Date ranges

sbishel:
#### Summary
Implements the ability to select a date range for a date property.

#### Ticket Link
https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=89e87813-96b3-4e1d-825a-d2d919091a0c&c=5d5f340b-1b94-4771-849a-0c008fc187eb

https://user-images.githubusercontent.com/12704875/127571325-cf3c1f74-1dbd-42c4-94fe-ab00ed31990d.mp4


webapp/src/utils.ts
```diff
@@ -121,9 +121,8 @@ class Utils {
 
     // Date and Time
 
-    static displayDate(date: Date, intl: IntlShape, dateFormat: string): string {
-        const text = intl.formatDate(date, {year: 'numeric', month: 'short', day: '2-digit', format: dateFormat})
-
+    static displayDate(date: Date, intl: IntlShape): string {
+        const text = intl.formatDate(date, {year: 'numeric', month: 'numeric', day: '2-digit'})
```
sbishel, line 125: The previous code that added `format: dateFormat` caused a javascript error here. This function will return the correct date format based on locale. This function currently is only being used for read-only views of date properties and displays the month as MMM (ie. 'Jul')

The Day Picker Input handled formatting the date as either 'MM/DD/YYYY' or 'DD/MM/YYYY' based on user setting. This function now provides all date formatting for the EditableDayPickers. It was changed to match the current date picker format and will return the correct format based on Locale. So the user setting is no longer necessary. - https://github.com/mattermost/focalboard/pull/666 

We should eventually allow user to pick date format, but not for locale, but for short vs 2-digit month, etc.

sbishel:
The user property for `preferredDateFormat` introduced [here](https://github.com/mattermost/focalboard/pull/666) should either be removed or updated to support different formats. (MMM DD, YYYY vs MM/DD/YYYY) It was implemented to fix locale issues which are now fixed in this PR.

webapp/src/widgets/editableDayPicker.tsx
```diff
@@ -1,26 +1,83 @@
...
+
+    const [dateProperty, setDateProperty] = useState<DateProperty>(createDatePropertyFromString(value as string))
+    const [showDialog, setShowDialog] = useState(false)
+
+    const dateFrom = dateProperty.from ? new Date(dateProperty.from) : undefined
+    const dateTo = dateProperty.to ? new Date(dateProperty.to) : undefined
+    const [fromInput, setFromInput] = useState<string>(getDisplayDate(dateFrom))
+    const [toInput, setToInput] = useState<string>(getDisplayDate(dateTo))
+
+    const stateRef = useRef(dateProperty)
```
jespino, line 77: I guess you are using this `stateRef` here to make it visible to `onClose` callback with the right data all the time, but makes more sense to use the `useCallback` to build `onClose` and set the `dateProperty` as a dependency to "regenerate" the `onClose` callback.

webapp/src/widgets/editableDayPicker.tsx
```diff
@@ -1,26 +1,83 @@
...
+    stateRef.current = dateProperty
+
+    const isRange = dateTo !== undefined
```
jespino, line 80: This trick work here because you are not changing the properties on the fly, but if in the future it can happen, you will end up with a wrong `isRange` inside the callbacks, so this line is perfectly fine, but you would need to generate the callbacks using `useCallback` and adding the right dependencies, en this case can be the `isRange`, or maybe the `datePropertype`.

jespino (changes requested):
Some changes needed, specially in the generation of the callbacks using hooks, and in the separation of concerns of the `EditableDayPicker` widget.

harshilsharma63:
Some suggestions - 

1. The selection circle will look a bit odd being an over instead of circle. Thoughts @asaadmahmood 
Screenshot 2021-07-12 at 2 04 36 PM

2. CUrrently, clicking on **_Today_** beings the calendar to the current month. I was confused for a while when I was already in the current month. Should it select the current date instead? This could be my confusion though.
today

asaadmahmood:
1. Agreed, should be a circle.
Also the calendar doesn't seem left aligned.

webapp/src/utils.ts
darkLord19: @sbishel is there any reason to ignore user preferred format? If user has set DD/MM/YYYY then we should show in that format right?

harshilsharma63:
I found an issue with property type conversion. If you take a select type property with a value set, and convert it to date type, the app crashes.

harshilsharma63 (changes requested):
Added some changes in comment.

webapp/src/utils.ts
sbishel: Actually, it looks like this has changed with this [PR](https://github.com/mattermost/focalboard/pull/791) I will be updating to latest main and will rework based on the latest.

webapp/src/components/properties/dateRange/dateRange.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+    let displayValue = ''
+    if (dateFrom) {
+        displayValue = getDisplayDate(dateFrom)
+    }
+    if (dateTo) {
+        displayValue += ' -> ' + getDisplayDate(dateTo)
+    }
+
+    const onClose = () => {
```
sbishel, line 135: @jespino I was not able to get this to work using `useCallback`. This [article](https://stackoverflow.com/questions/57847594/react-hooks-accessing-up-to-date-state-from-within-a-callback) discusses a couple different options. Previously the `useRef`option was used, changed this to retrieve the current state, via `setDateProperty`.

webapp/src/widgets/editableDayPicker.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+                value={''}
+                onChange={jest.fn()}
+            />,
+        )
+
+        const {container} = render(component)
+        expect(container).toMatchSnapshot()
+    })
+
+    test('returns default correctly, es local', () => {
```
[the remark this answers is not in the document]
sbishel: Renamed

webapp/src/widgets/editableDayPicker.tsx
```diff
@@ -29,31 +86,166 @@ function EditableDayPicker(props: Props): JSX.Element {
...
-                    locale,
-                    localeUtils: MomentLocaleUtils,
-                    todayButton: intl.formatMessage({id: 'EditableDayPicker.today', defaultMessage: 'Today'}),
-                }}
-                formatDate={MomentLocaleUtils.formatDate}
-                parseDate={MomentLocaleUtils.parseDate}
-                format={dateFormat}
-                placeholder={`${MomentLocaleUtils.formatDate(new Date(), dateFormat, locale)}`}
-            />
+        <div className={'EditableDayPicker '}>
```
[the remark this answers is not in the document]
sbishel: Moved to webapp/src/components/properties/dateRange/dateRange.tsx

webapp/src/utils.ts
sbishel: The preferred user format was removed with this [PR](https://github.com/mattermost/focalboard/pull/790). This PR is supporting dates in the a similar manner. However, it parses and displays dates for editing using the locale "L" Date format. Typically "MM/DD/YYYY" or "DD/MM/YYYY" depending on local. 
When empty the placeholder will show the parsable format.

sbishel:
@harshilsharma63 and @asaadmahmood 
I wasn't able to figure out the correct CSS to make the dates circles rather than ovals. I know the issue, the cells are rectangles, if they were squares we would get circles but my css isn't good enough to fix. If one of you could take a quick look. Thanks.

jespino (approved):

sbishel:
Thanks Harshil. The css has been updated to have the selectors round.

Screen Shot 2021-08-02 at 10 11 24 AM

sbishel:
@harshilsharma63, @chenilim, @asaadmahmood - Wanted to make sure you were ok with the UI before merging...we can always iterate on it.

asaadmahmood:
@sbishel Ideally it should be something like this (so that all the dates aren't circles, rather only the one at the start and the end), but we can iterate on the UI later too as you said.
Screenshot 2021-08-04 at 8 51 41 PM

chenilim:
Agree with @asaadmahmood. If the rounded rectangles are too tricky for the first iteration, perhaps try regular rectangles first. Yup, this can be iterated on.

sbishel:
I did take a stab at it again. I will also add a ticket to iterate on the day-picker look and feel.
Screen Shot 2021-08-04 at 5 13 57 PM

harshilsharma63 (approved):
LGTM!

sbishel:
I created this card for UX iteration. 

https://community-daily.mattermost.com/plugins/focalboard/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=7e38eab4-890a-4fc3-8f48-a1f9bd1ab822&c=f76ef212-053b-4abc-8cc5-160401f6a191

PR #664, opened 2021-07-01
Fixed issue where color panel git cut off in value selector

### LINKED ISSUE #636: Bug: Color Panel for Select fields is cut off
**Summary:**
The color panel for the select fields is cut off. (see image)

Screen Shot 2021-06-25 at 12 47 39 PM

1. Go to a card.
2. Click on a select field
3. Click on menu ('...') of a select item
4. See image above.

**Expected behavior:**
color panel for select field should be visible.

**Additional context:**
Appears to be a regression from this fix. 
https://github.com/mattermost/focalboard/pull/594

harshilsharma63:
#### Summary
Fixed an issue with the select colour picker being clipped out of the dropdown.

#### Ticket Link
Fixes #636

### DIFF
--- a/webapp/src/widgets/valueSelector.scss
+++ b/webapp/src/widgets/valueSelector.scss
@@ -7,6 +7,7 @@
     > .Label {
         margin: 0 10px;
         max-width: calc(100% - 10px);
+
         &.empty {
             color: rgba(var(--body-color), 0.6);
         }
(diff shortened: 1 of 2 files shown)


webapp/src/widgets/valueSelector.scss
```diff
@@ -41,21 +42,32 @@
...
                 background-color: unset;
+
                 &:hover {
                     background-color: rgba(var(--body-color), 0.1);
                 }
             }
         }
     }
+
+    div:nth-child(3) {
```
harshilsharma63, line 65: Note to reviewer-

These styles are hardcoded in the library and are not configurable from either props or the style function.

harshilsharma63:
@sbishel can you please review this?

sbishel (approved):
Tested and it works fine. However, from my testing, you don't need to set both 
`div:nth-child(3)`
and 
```
menu:
   minWidth: '260px',
```
One or the other depending on the display you desire.

harshilsharma63:
The `div:nth-child(3)` part allows you to view the entire colour panel without scrolling. This effect is easily visible with a select property that has only a few options.

Without `div:nth-child(3)`:

Jul-08-2021 09-59-22

With `div:nth-child(3)`:
with

The `minWidth` is to make the colour menu appear _inside_ the main select. It's a bit of an OCD thing :P

With:
Screenshot 2021-07-08 at 10 03 23 AM

Without:
Screenshot 2021-07-08 at 10 03 35 AM

PR #665, opened 2021-07-01
Only allowing valid color classes

jespino:
This fixes: https://mattermost.atlassian.net/browse/MM-36791

### DIFF
--- a/webapp/src/components/kanban/kanbanColumnHeader.tsx
+++ b/webapp/src/components/kanban/kanbanColumnHeader.tsx
@@ -129,12 +129,12 @@ export default function KanbanColumnHeader(props: Props): JSX.Element {
                                         onClick={() => mutator.deletePropertyOption(boardTree, boardTree.groupByProperty!, group.option)}
                                     />
                                     <Menu.Separator/>
-                                    {Constants.menuColors.map((color) => (
+                                    {Object.entries(Constants.menuColors).map(([id, color]) => (
                                         <Menu.Color
-                                            key={color.id}
-                                            id={color.id}
+                                            key={id}
+                                            id={id}
                                             name={color.name}
-                                            onClick={() => mutator.changePropertyOptionColor(boardTree.board, boardTree.groupByProperty!, group.option, color.id)}
+                                            onClick={() => mutator.changePropertyOptionColor(boardTree.board, boardTree.groupByProperty!, group.option, id)}
                                         />
                                     ))}
                                 </>}
--- a/webapp/src/components/kanban/kanbanColumnHeader.tsx
+++ b/webapp/src/components/kanban/kanbanColumnHeader.tsx
@@ -129,12 +129,12 @@ export default function KanbanColumnHeader(props: Props): JSX.Element {
                                         onClick={() => mutator.deletePropertyOption(boardTree, boardTree.groupByProperty!, group.option)}
                                     />
                                     <Menu.Separator/>
-                                    {Object.entries(Constants.menuColors).map(([id, color]) => (
+                                    {Object.entries(Constants.menuColors).map(([key, color]) => (
                                         <Menu.Color
-                                            key={id}
-                                            id={id}
-                                            name={color.name}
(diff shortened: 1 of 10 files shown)


webapp/src/components/table/tableGroupHeaderRow.tsx
```diff
@@ -127,12 +127,12 @@ const TableGroupHeaderRow = React.memo((props: Props): JSX.Element => {
                                         onClick={() => mutator.deletePropertyOption(boardTree, boardTree.groupByProperty!, group.option)}
                                     />
                                     <Menu.Separator/>
-                                    {Constants.menuColors.map((color) => (
+                                    {Object.entries(Constants.menuColors).map(([id, color]) => (
```
hahmadia, line 130: SUPER NIT!
Can we change this to `key` to make it more inline with the wording of `key / value`  cause I didn't immediately know `id` was the `key` but rather I was thinking `id` is a property of the object

webapp/src/components/kanban/kanbanColumnHeader.tsx
```diff
@@ -129,12 +129,12 @@ export default function KanbanColumnHeader(props: Props): JSX.Element {
                                         onClick={() => mutator.deletePropertyOption(boardTree, boardTree.groupByProperty!, group.option)}
                                     />
                                     <Menu.Separator/>
-                                    {Constants.menuColors.map((color) => (
+                                    {Object.entries(Constants.menuColors).map(([id, color]) => (
```
hahmadia, line 132: Same thing here

webapp/src/components/table/tableGroupHeaderRow.tsx
hahmadia: This also falls more in line with docs wording of what is returned `[key , value]` ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#parameters )

webapp/src/widgets/label.tsx
```diff
@@ -13,9 +15,13 @@ type Props = {
 
 // Switch is an on-off style switch / checkbox
 function Label(props: Props): JSX.Element {
+    let color = 'empty'
+    if (Constants.menuColors.hasOwnProperty(props.color || '')) {
```
hahmadia, line 19: Won't `Constants.menuColors.hasOwnProperty('')` always return false? You already init with `let color = 'empty'` so I don't see this necessary to check.

webapp/src/widgets/label.tsx
```diff
@@ -13,9 +15,13 @@ type Props = {
...
+        color = props.color || 'empty'
```
hahmadia, line 20: Already init `color = 'empty'` so I don't think the `||` is necessary here.

webapp/src/widgets/valueSelector.tsx
```diff
@@ -79,12 +79,12 @@ const ValueSelectorLabel = React.memo((props: LabelProps): JSX.Element => {
                         onClick={() => props.onDeleteOption(option)}
                     />
                     <Menu.Separator/>
-                    {Constants.menuColors.map((color) => (
+                    {Object.entries(Constants.menuColors).map(([id, color]: any) => (
```
hahmadia, line 82: Name to `key` instead of `id` as mentioned above?

hahmadia (approved):
LGTM! Some minor naming things but no major changes. Thanks!

webapp/src/components/table/tableGroupHeaderRow.tsx
jespino: Yes, you are right, sounds more correct. Thanks! :)

webapp/src/widgets/label.tsx
jespino: This is because `hasOwnProperty` requires an string and props.color can be undefined.

webapp/src/widgets/label.tsx
jespino: again is about `string|undefined` type.

hahmadia (approved):

Issue #550, opened 2021-06-09
Feature Idea: Different date formats


Talismanian:
**Summary:**

Ability to change how dates are displayed. The default MM/DD/YYYY format is extremely confusing for anyone outside of North America who isn't expecting it.

Alternative options an include YYYY-MM-DD or DD/MM/YYYY and could be tied with language/localisation settings.

**How important this is to me and why:**

Importance: Medium

Use cases:
1. Avoid confusion over dates involving low month numbers (e.g. 6/9/2021).
2. Better cater to users outside of North America.
3. Adoption of unambiguous date formats that are readily understood across regions.

**Additional context / similar features:**

A number of other web-based apps provide the option to change date formats. They're often tied to user settings and may not be the default.

chenilim:
Thanks @Talismanian! I've created a [card to track this](https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde/shared?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=ca1a5441-4c1c-4d0e-856e-88cc744576c8&r=762db9c6-305c-4d13-8ce6-9f11e5232c24&c=1b73e464-84cd-4d14-823f-a41a80c38ff7). Agree that we should at least support a MM/DD/YYYY or DD/MM/YYYY display format.

cc @sbishel, @michaelgamble.

darkLord19:
@chenilim can i work on this?

chenilim:
Thanks @darkLord19, you got it! Here's the basic spec:
1. Add an additional user setting to toggle the date format
  * For now, only support "M/D/YYYY" or "D/M/YYYY"
  * We can expand this in the future to support other custom formats
2. Store this preferred date format in the localStorage settings
3. When reading the format, if it is missing or invalid, default to the current behavior of "M/D/YYYY"
4. Also add this new setting to the user settings to persist in the native apps
  * See #376 for reference

That should do it for a first cut. LMK if you have any questions. Thanks!

chenilim:
[...]

PR #666, opened 2021-07-01
add user setting for preferred date format

darkLord19:
Fixes https://github.com/mattermost/focalboard/issues/550

### DIFF
--- a/webapp/src/components/propertyValueElement.tsx
+++ b/webapp/src/components/propertyValueElement.tsx
@@ -18,6 +18,8 @@ import Label from '../widgets/label'
 import EditableDayPicker from '../widgets/editableDayPicker'
 import Switch from '../widgets/switch'
 
+import {UserSettings} from '../userSettings'
+
 import UserProperty from './properties/user/user'
 import MultiSelectProperty from './properties/multiSelect'
 import URLProperty from './properties/link/link'
@@ -152,6 +154,7 @@ const PropertyValueElement = (props:Props): JSX.Element => {
             <EditableDayPicker
                 className='octo-propertyvalue'
                 value={value as string}
+                dateFormat={UserSettings.preferredDateFormat}
                 onChange={(newValue) => mutator.changePropertyValue(card, propertyTemplate.id, newValue)}
             />
         )
--- a/webapp/src/components/sidebar/sidebarSettingsMenu.tsx
+++ b/webapp/src/components/sidebar/sidebarSettingsMenu.tsx
@@ -14,6 +14,9 @@ import {
     setTheme, systemThemeName,
     Theme,
 } from '../../theme'
+import {
+    defaultDateFormat, dmyDateFormat,
+} from '../../dateFormat'
 import Menu from '../../widgets/menu'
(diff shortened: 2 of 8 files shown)


darkLord19:
@chenilim I have started working on the feature but facing a slight issue. preferredDateFormat is not being stored in localStorage. Can you have a look and help out. Also, is there a way to attach debugger into typescript code in focalboard codebase?

chenilim:
Thanks @darkLord19! I can't see at a glance what might be wrong from the code, but you can debug using the Chrome debugger - more details in the [Dev Tips & Tricks here](https://www.focalboard.com/contribute/getting-started/dev-tips/#debugging-the-web-app).

darkLord19:
@chenilim not sure why but `preferredDateFormat` is not getting called

jespino:
@darkLord19 I like the PR, but I guess the missing part is applying this format in the rest of the application, right?

darkLord19:
@jespino yes that's the plan but was facing issue with storing preference in localstorage and haven't got the chance to debug further yet

jespino (approved):
Looks good to me. Thanks @darkLord19! :confetti_ball:

webapp/src/components/sidebar/sidebarSettingsMenu.tsx
```diff
@@ -49,6 +52,11 @@ const SidebarSettingsMenu = React.memo((props: Props) => {
         setRandomIcons(!randomIcons)
     }
 
+    const [preferredDateFormat, setPreferredDateFormat] = useState(UserSettings.preferredDateFormat)
+    const updatePreferredDateFormat = (newDateFormat: string) => {
+        setPreferredDateFormat(newDateFormat)
```
jespino, line 57: I think you also need to set the setting in the local storage here. something like `UserSettings.preferredDateFormat = newDateFormat`.

jespino (changes requested):
I found a detail about storing in the value in the local storage.

webapp/src/components/sidebar/sidebarSettingsMenu.tsx
darkLord19: Oh yes. thanks. Silly me thought that state dispatch will call call the setter!

darkLord19:
@jespino I have updated date property inside card view to take format from user preference. is there any other place we are using date display? I didn't find but i haven;t explored focalboard as much so confirming

webapp/src/components/propertyValueElement.tsx
```diff
@@ -152,6 +154,7 @@ const PropertyValueElement = (props:Props): JSX.Element => {
             <EditableDayPicker
                 className='octo-propertyvalue'
                 value={value as string}
+                dateFormat={UserSettings.preferredDateFormat}
```
jespino, line 157: I tiny problem here is this is not going to be reactive, not a huge deal, but probably something to fix later.

jespino (approved):
Looks good to me.

harshilsharma63:
@darkLord19 can you make the lint and test fixes?

darkLord19:
@harshilsharma63 updated

Issue #668, opened 2021-07-02
Bug: Mattermost Plugin Title Links to Older Repository URL


N3rdP1um23:
**Summary:**
When clicking on the Focalboard option in the Mattermost Plugin Marketplace, the link that is opened references what seems to be an older repository that is no longer (https://github.com/mattermost/mattermost-plugin-focalboard).

**Steps to reproduce the behavior:**
1. Open Mattermost (desktop or browser)
2. Click on the hamburger menu beside your username
3. Click on "Marketplace"
4. Search for "Focalboard"
5. Click on plugin title
6. Opening URL will point to the older URL listed above.

**Expected behavior:**
Clicking on the plugin title should point to this repository as the current one is pointing to a repository that is non-existent.

**Screenshots:**
![image](https://user-images.githubusercontent.com/8093806/124317373-11a22f00-db45-11eb-9ddc-ff8df34c3eff.png)
![image](https://user-images.githubusercontent.com/8093806/124317595-69d93100-db45-11eb-960a-58b02229034b.png)

**Platform:**
 - Browser and version: Microsoft Edge Version 93.0.933.1 (Official build) dev (64-bit)
 - OS and version: Windows 10 21H1 

**Additional context:**
Looking into this repository a little more, it seems that there's a few files in the repository pointing to the older repository
- `mattermost-plugin/plugin.json`
- `mattermost-plugin/server/manifest.go`

jespino:
This was closed by @N3rdP1um23 a month ago.

PR #669, opened 2021-07-02
[GH-668] Upgraded older links and generalized release URL

N3rdP1um23:
#### Summary
This PR is in relation to the following issue #668 where the plugin title, along with a few other links are pointing to an older repository. This PR includes a commit to remove the older URL's and point to the current repository URL where the plugin is hosted. Also included in this commit is a generalized releases URL pointing to the releases section for this repository instead of the specific version. This will save some time when updates are published as the release notes will/should always be released to the releases tab of this repository. I've made that change as the previous release URL was pointing to v0.1.0.

#### Ticket Link
#668


jespino (approved):
Great, thanks @N3rdP1um23! :sparkles:

N3rdP1um23:
No problem at all! Always happy to help!! 😃

Issue #648, opened 2021-06-29
Bug: mattermost plugin not working 


arunzco:
Using Bitnami Mattermost stack , getting the following error 

plugin","missing_plugin_id":"focalboard","url":"/plugins/focalboard/workspace/r7qnsi6psfrdxqghx1z9pyxmih","error":"plugin not found: focalboard"}

chenilim:
Thanks @arunzco! to confirm, did you set up the [Mattermost Bitnami stack from here](https://bitnami.com/stack/mattermost/cloud)? What are the repro steps? Did this error occur immediately?

cc @ogi-m, @sbishel - we should try to repro this and determine the root cause.

arunzco:
@chenilim 
Yes I have used the official Bitnami Stack,
Did the following 
1.Configured the latest plugin
2.Tried the focalboard button 

It shows a 500 server error on browser were as in mattermost app a blank windows appears and the log says the error I mentioned. also found the following logs 

{"level":"error","ts":1625021130.5641432,"caller":"mlog/log.go:251","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"focalboard","error":"connection is shut down"}
{"level":"error","ts":1625021135.5650318,"caller":"mlog/log.go:251","msg":"Plugin failed to ServeHTTP, muxBroker couldn't accept connection","plugin_id":"focalboard","serve_http_stream_id":4,"error":"timeout waiting for accept"}
{"level":"error","ts":1625021135.565191,"caller":"mlog/log.go:251","msg":"Plugin failed to ServeHTTP, muxBroker couldn't Accept request body connection","plugin_id":"focalboard","error":"timeout waiting for accept"}

I have also tried it with official mattermost setup instructions tried with  and without proxy servers   it also produces the same error. 
DB used is mysql 
https://docs.mattermost.com/install/install-ubuntu-1804.html#installing-ubuntu-server-18-04-lts

chenilim:
Thanks @arunzco! @sbishel, do you think @harshilsharma63 or someone else could take a look at this? From the error log, it appears to be an issue between the plugin and Mattermost plugin framework. First thing would be to check if the versions on Bitnami are what we expect (Mattermost v5.36 and Focalboard 0.6.7). cc @jespino.

harshilsharma63:
@arunzco do you see Focalboard in the list of installed plugins in System Console?

arunzco:
Yes

harshilsharma63:
[...]

PR #677, opened 2021-07-06
Prevent plugin from crashing if SiteURL is not set

mgdelacroix:
#### Summary
This PR checks if the `SiteURL` property is set and uses an empty string if it's not, preventing the plugin from crash at start time due to a `nil` pointer.

#### Ticket Link
Possibly related to https://github.com/mattermost/focalboard/issues/648

### DIFF
--- a/mattermost-plugin/server/plugin.go
+++ b/mattermost-plugin/server/plugin.go
@@ -98,7 +98,7 @@ func (p *Plugin) OnActivate() error {
 		return fmt.Errorf("error initializing the DB: %v", err)
 	}
 
-	baseURL := "."
+	baseURL := ""
 	if mmconfig.ServiceSettings.SiteURL != nil {
 		baseURL = *mmconfig.ServiceSettings.SiteURL
 	}
--- a/mattermost-plugin/server/plugin.go
+++ b/mattermost-plugin/server/plugin.go
@@ -98,8 +98,13 @@ func (p *Plugin) OnActivate() error {
 		return fmt.Errorf("error initializing the DB: %v", err)
 	}
 
+	baseURL := ""
+	if mmconfig.ServiceSettings.SiteURL != nil {
+		baseURL = *mmconfig.ServiceSettings.SiteURL
+	}
+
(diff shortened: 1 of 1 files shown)


mattermost-plugin/server/plugin.go
```diff
@@ -98,8 +98,13 @@ func (p *Plugin) OnActivate() error {
 		return fmt.Errorf("error initializing the DB: %v", err)
 	}
 
+	baseURL := ""
```
jespino, line 101: I think it should be `"."` instead of `""` right?

mgdelacroix: Thought about it but if the user is using a subpath, they should have set `SiteURL`. In any case it is more correct, so happy to make the change :)

jespino: Yes, lets do the change, for correctness and we can merge it.

mgdelacroix: Fixed in cceb3801679b65bb14336e91909ea4b02f448903

mgdelacroix: Actually it looks like the `ServerRoot` doesn't work with a relative URL

harshilsharma63: @mgdelacroix we can also use this - https://developers.mattermost.com/extend/plugins/manifest-reference/#required_configuration to prevent the plugin from starting without a site URL specified. Would this make things simpler?

mgdelacroix: Thanks for the pointer @harshilsharma63!! We could use that, but if we can support the empty `SiteURL` reliably I think it is preferable.

@jespino found the issue. The URL cannot be relative as we're running the web interface in a subpath already, so it fails when filling the `index.html` template with the tag `<script defer src="./plugins/focalboard/static/main.js"></script>` and load it from the subpath, it's actually intended to be absolute.

jespino (approved):

harshilsharma63 (approved):

PR #686, opened 2021-07-07
fix remaining golangci linter warnings

wiggin77:
#### Summary
This PR fixes the remaining golangci linter warnings.  Project lint is clean once merged.

#### Ticket Link
https://focalboard-community.octo.mattermost.com/workspace/qgsck6cts3fwpqwyjiupjm5cde?id=47aa9bb4-6967-4a96-83c7-11bd6b20f1eb&v=ca1a5441-4c1c-4d0e-856e-88cc744576c8&c=c32e6473-441a-44b9-85ca-c419bd0f16ff


server/services/auth/password_test.go
```diff
@@ -135,7 +135,10 @@ func TestIsPasswordValidWithSettings(t *testing.T) {
 				assert.NoError(t, err)
 			} else {
 				require.Error(t, err)
-				assert.Equal(t, tc.ExpectedFailingCriterias, err.(*InvalidPasswordError).FailingCriterias)
+				var errFC *InvalidPasswordError
+				if assert.ErrorAs(t, err, &errFC) {
```
jespino, line 139: nit: should we assert that the error type is `InvalidPasswordError`. If `IsPasswordValid` start returning something that is not `InvalidPasswordError`, it would pass the tests, but not necesarilly means that is right.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -205,7 +212,7 @@ func (s *MattermostAuthLayer) GetWorkspace(id string) (*model.Workspace, error)
 		}
 		var name string
 		if err := rows.Scan(&name); err != nil {
-			log.Fatal(err)
```
jespino, line 208: This `log.Fatal` looks wrong here, I think this should just return an error.

server/services/config/config.go
```diff
@@ -12,7 +12,7 @@ const (
 )
 
 type AmazonS3Config struct {
-	AccessKeyId     string
+	AccessKeyId     string //nolint:stylecheck,golint
```
jespino, line 15: I think we can here use `AccessKeyID`, but we need to change that in the usage from `mattermost-plugin/server/plugin.go` and in `server/server/server.go`

jespino (changes requested):
I love this PR, just a small amount of improvements suggestions, on of them not strictly related to this PR (`log.Fatal`) but I think we can fix that right away.

server/services/auth/password_test.go
wiggin77: `assert.ErrorAs` does check that err is of type `InvalidPasswordError`.  It will fail if err does not have that type somewhere in the chain of wrapped errors.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
wiggin77: I agree.

jespino (approved):
Lgtm

PR #691, opened 2021-07-08
POC of permissions using mattermost ones

jespino:


jespino:
This POC is no longer needed, closing it.

PR #701, opened 2021-07-09
Remove plus icon from Kanban column header

maxhq:
#### Summary

There are two icons for adding a card in Kanban view:
- one in the [column header](https://github.com/mattermost/focalboard/blob/e0ec1c03e088e99d5ff2f570ed089a4aed4cfba3/webapp/src/components/kanban/kanbanColumnHeader.tsx#L144) and
- one [below the last card](https://github.com/mattermost/focalboard/blob/e0ec1c03e088e99d5ff2f570ed089a4aed4cfba3/webapp/src/components/kanban/kanban.tsx#L270).

Esp. the column headers IMHO already look cluttered (title, card no., dots, plus sign) so I suggest to only keep the plus button below the last card.

### DIFF
--- a/webapp/src/components/kanban/kanbanColumnHeader.tsx
+++ b/webapp/src/components/kanban/kanbanColumnHeader.tsx
@@ -12,7 +12,6 @@ import mutator from '../../mutator'
 import {BoardTree, BoardTreeGroup} from '../../viewModel/boardTree'
 import Button from '../../widgets/buttons/button'
 import IconButton from '../../widgets/buttons/iconButton'
-import AddIcon from '../../widgets/icons/add'
 import DeleteIcon from '../../widgets/icons/delete'
 import HideIcon from '../../widgets/icons/hide'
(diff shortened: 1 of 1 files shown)


wiggin77 (approved):
Code looks good. :+1:

chenilim:
I suggest not taking this change now. My gut reaction is the '+' icon in the header is useful when there are many cards. Otherwise, the user has to scroll to the bottom to add a new card. @asaadmahmood, any thoughts on this?

asaadmahmood:
@chenilim I agree. We may redo this, but removing the plus icon at the top especially for larger boards may cause a usability concern.

chenilim:
Thanks, @maxhq appreciate your PR, but let's think a bit more about this first. I've filed issue #702 to track this and added more comparable observations there. Feel free to add any additional input there. We might want to collect a few things to think about a broader board view design cleanup. Thanks.

Issue #113, opened 2021-03-18
Feature Request: Add possibility to delete tasks in table view (like in board view)


dataCobra:
**Summary:**
When in table view there is no option to easily delete a task. Before you can delete a task you need to open it.

**Steps to reproduce the behavior:**
1. Open a table view with tasks
2. Find option to delete task in table view
3. Need to open the task
4. Click on "3 dot" drop-down menu
5. Click delete

**Expected behavior:**
It would be nice to have the option to delete them directly in table view. Maybe with a "3 dot" drop-down menu like it
is in board view at the end of the row.

arvinDarmawan:
Hi  @chenilim @jespino   could you assign it to me? I want to try it

jespino:
Hi @arvinDarmawan, sure i can assign it to you. Let me know if you need anything from us.

arvinDarmawan:
Hi @jespino , is it the expected result ?
I just using MenuWrapper component in tableRow component (I'm not yet implement the delete functionality, just the UI)

<img width="1133" alt="Screen Shot 2021-07-05 at 09 40 27" src="https://user-images.githubusercontent.com/28821618/124410697-3846ac80-dd75-11eb-943e-68cfc2e64f6f.png">

harshilsharma63:
@dataCobra thanks for the implementation. This is certainly a useful feature.

I don't like the design though. It works fine when you only have as many columns visible as the width of the browser. When you have more, you need to perform a horizontal scroll to reach the menu, which removes all the convenience.

Let's wait for @asaadmahmood to provide a design for this.

asaadmahmood:
Yup, will have a look at this.

dataCobra:
Thanks for the mention, but I didn't implement the feature that was @arvinDarmawan. :+1:

asaadmahmood:
@jespino @harshilsharma63 @arvinDarmawan This is what I'm thinking about the delete option placement.
https://www.figma.com/proto/UPbYxPDsg1aIzozyysRPeO/Focalboard-Base-Design?page-id=11%3A3265&node-id=11%3A3273&viewport=14926%2C8782%2C5.783653736114502&scaling=min-zoom

So we show the icon buttons on the left when hovering a row.

arvinDarmawan:
> @jespino @harshilsharma63 @arvinDarmawan This is what I'm thinking about the delete option placement.
[...]

PR #707, opened 2021-07-11
[GH-113] Implement delete feature in table view

arvinDarmawan:
#### Summary
When in table view there is no option to easily delete a task. Before you can delete a task you need to open it.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/113


webapp/src/components/table/tableRow.tsx
```diff
@@ -140,6 +147,28 @@ const TableRow = React.memo((props: Props) => {
...
+            <MenuWrapper
+                className='optionsMenu'
+                stopPropagationOnToggle={true}
+            >
+                <IconButton icon={<OptionsIcon/>}/>
+                <Menu position='left'>
+                    <Menu.Text
+                        icon={<DeleteIcon/>}
+                        id='delete'
+                        name={'Delete'}
```
kamre, line 160: Should internalization be used for this `name`?

arvinDarmawan: Hi @kamre , so I need to  add "TableRow.delete" in internalization file, right ?

kamre: @arvinDarmawan yes, there are many other usages of `*.delete` as `id` for formated messages in other places.

So it looks like `TableRow.delete` is the correct name for `id` in this place.

jespino:
@asaadmahmood can you take a look to the UX/Design of this?

webapp/src/components/table/tableRow.tsx
arvinDarmawan: Hi @kamre , I already check again in en.json file, and there are  "ContentBlock.Delete" and "TableHeaderMenu.delete".
Should we use delete with initial letter d or D ?

kamre: @arvinDarmawan I don't have exact anwser, but it looks like most of other ids use lower case after the dot.

arvinDarmawan: done @kamre , just waiting for UX/design

asaadmahmood:
@jespino Added a comment on the main issue to point out how it should look like.
https://github.com/mattermost/focalboard/issues/113

hahmadia (approved):
From a code and usability perspective, I am approving because it's fine.

However, I'm not sure if UI changes will cause code changes. If they do, feel free to re-request a review from me. 

Thanks for the work either way 😄

asaadmahmood:
@hahmadia It should look like this. So this may still need some work.
image

https://www.figma.com/file/nIyq6YPnKUxurCr0AyklZj/113-Delete-Tasks?node-id=101%3A7882


We can probably just move the button to the left with css, and position the dropdown to its right.

asaadmahmood (changes requested):
Added a comment.

hahmadia (changes requested):
Changing to Request Changes so I can be re-notified when the new UI changes have been made.

Thanks @asaadmahmood for heads up :)

chenilim:
Trying to clean up stale PRs. @asaadmahmood, @hahmadia, are there still changes required? Or can they be made later? Thanks.

sbishel:
no longer relevent.

Issue #433, opened 2021-05-17
Server Unit Tests - ./services/webhook/webhook.go 


sbishel:
Write Unit Tests for ./server/services/webhook/webhook.go

PR #710, opened 2021-07-12
[GH-443] Add unit tests for ./services/webhook/webhook.go

yedamao:
#### Summary

Add unit tests for ./services/webhook/webhook.go

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/433


wiggin77 (changes requested):

jespino (approved):
Maybe we can include other test cases here, 1 is whenever the config is `nil`, another is whenever the `WebhookUpdate` is an empty list, and the other is whenever the `logger`is nil. I think at least 2 of them are going to generate problems.

yedamao:
@jespino config/logger is nil will cause panic. maybe we should check the passed parameters in NewClient, how about the below code.


```
func NewClient(config *config.Configuration, logger *mlog.Logger) (*Client, error) {
        if config == nil {
                return nil, errors.New("config is nil")
        }

        if logger == nil {
                return nil, errors.New("logger is nil")
        }

        return &Client{
                config: config,
                logger: logger,
        }, nil
}
```

jespino:
@yedamao I'm not sure about how to handle a nil loger, maybe a specific panic here is the right decision (because we don't want that to happen), other option is the approach that you just used, and the other option is to use `nil` logger as a way to deactivate logging there.

What do you think @wiggin77?

wiggin77:
> I'm not sure about how to handle a nil loger, maybe a specific panic here is the right decision (because we don't want that to happen)

We should panic on a nil logger.  Really the only way to get a nil logger is to have a syntax error in the config, which should IMO panic.

wiggin77 (approved):
LGTM :+1:

Issue #678, opened 2021-07-06
Bug: Hard to read labels in dark mode


chenilim:
**Steps to reproduce the behavior:**
1. Create a board while in dark mode
2. Add a few columns
3. Notice that the column labels are low contrast (white on gray)

**Expected behavior:**
Labels should be more readable in dark mode.

**Screenshots:**
![Low contrast in dark mode](https://user-images.githubusercontent.com/46905241/124635156-f6346e00-de3b-11eb-9141-a59c7dae7fff.png)

**Additional context:**
We should do a general review and scrub of the dark mode theme, and also align with the Mattermost dark theme.

chenilim:
Assigning to you @asaadmahmood to think about. Not a high priority, but we should incorporate this into any overall review / refactoring of the design. Thanks.

asaadmahmood:
Should be fixed by my PR.

PR #726, opened 2021-07-14
Issue 678 - Updating default dark theme label

asaadmahmood:
#### Summary
Issue 678 - Updating default dark theme label

#### Ticket Link
https://github.com/mattermost/focalboard/issues/678

#### Screenshot
Screenshot 2021-07-14 at 8 30 18 PM

### DIFF
--- a/webapp/src/theme.ts
+++ b/webapp/src/theme.ts
@@ -64,7 +64,7 @@ export const darkTheme = {
     ...defaultTheme,
 
     mainBg: '55, 53, 47',
-    mainFg: '200, 200, 200',
+    mainFg: '220, 220, 220',
     buttonBg: '80, 170, 221',
     buttonFg: '255, 255, 255',
     sidebarBg: '75, 73, 67',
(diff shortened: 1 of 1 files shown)


wiggin77 (commented):
The new text color has a yellow tint.  Is that intended?

wiggin77 (approved):

asaadmahmood:
@wiggin77 Good catch Doug, will update that, might have overlooked it.

asaadmahmood:
@wiggin77 Just checked it out, and I think rgb(220, 220, 220), shouldn't have a yellow tint to it.
The yellow tint is coming because the text has an opacity to it, and the background has a yellow tint to it.
But that's fine. So we can take this.

Screenshot 2021-07-15 at 1 33 00 AM

wiggin77:
> Just checked it out, and I think rgb(220, 220, 220), shouldn't have a yellow tint to it.

Yup - my mistake.  I typed 220 into a color picker for RGB, but the B component didn't update the color until I clicked away.  I was looking at 220/220/200

PR #736, opened 2021-07-16
Global header support

crspeller:
#### Summary
Adds support for the global header. Based off of https://github.com/mattermost/focalboard/pull/597 by @jespino 

Currently the experience when landing on the main page of boards is broken. This should be fixed before the feature is released. 

#### Screenshot
image

#### Screenshot of bugged main page
image


jespino:
@crspeller Nice!

chenilim:
@crspeller, thinking about this more, can we change it so that the App Switcher Boards item does the same thing as the Focalboard button in the channel header, i.e. open `/plugins/focalboard/workspace/CURRENT_CHANNEL_ID`? That way, the Global header will function starting today, and we can modify it when the Personal Dashboard / Boards homepage is ready (which may come later).

jespino:
This shouldn't be a draft anymore.

hahmadia (approved):

webapp/src/components/sidebar/sidebar.tsx
```diff
@@ -111,10 +101,8 @@ const Sidebar = React.memo((props: Props) => {
                 activeBoardId={props.activeBoardId}
             />
 
-            <SidebarSettingsMenu
-                setWhiteLogo={(newWhiteLogo: boolean) => setWhiteLogo(newWhiteLogo)}
-                activeTheme={getActiveThemeName()}
-            />
+            {(window as any).isFocalboardPlugin !== true &&
```
mgdelacroix, line 104: nit: this reads a bit weird, maybe is better to use something like:
```suggestion
            {!(window as any).isFocalboardPlugin &&
```

webapp/src/components/sidebar/sidebarUserMenu.tsx
```diff
@@ -20,32 +20,32 @@ import RegistrationLink from './registrationLink'
...
-}
+type Props = {}
 
 const SidebarUserMenu = React.memo((props: Props) => {
     const history = useHistory()
     const [showRegistrationLinkDialog, setShowRegistrationLinkDialog] = useState(false)
-    const {whiteLogo, showVersionBadge, showAccountActions} = props
+    const showVersionBadge = (window as any).isFocalboardPlugin === true
+    const focalboardTitle = (window as any).isFocalboardPlugin !== true
+    const showAccountActions = (window as any).isFocalboardPlugin !== true
```
mgdelacroix, line 30: Same for these, if we want to cover for the `null` cases we could either create new bools or use the `!!` trick for the first one and just negate the other two

webapp/src/components/topBar.tsx
```diff
@@ -7,6 +7,9 @@ import './topBar.scss'
 import HelpIcon from '../widgets/icons/help'
 
 const TopBar = React.memo((): JSX.Element => {
+    if ((window as any).isFocalboardPlugin === true) {
```
mgdelacroix, line 10: Same here

mgdelacroix (approved):
LGTM!! :+1: 

added some nonblocking comments for readability

webapp/src/components/sidebar/sidebarUserMenu.tsx
jespino: I think I going to move all this to a utility function that returns a boolean.

PR #745, opened 2021-07-16
Fix #739: Don't show account options as plugin

### LINKED ISSUE #739: Bug: Plugin should not show account options in the top-left menu
1. Click on the Focalboard logo to open the menu
2. It shows "Log out", "Change password", and "Invite users" as options

**Expected behavior:**
Those 3 options should not be displayed when running as a plugin

**Platform:**
 - Focalboard plugin 0.8.1, Chrome 91 on macOS

chenilim:
This hides the (non-functional) account actions in the Focalboard menu when running as a plugin.

### DIFF
--- a/webapp/src/components/sidebar/sidebar.tsx
+++ b/webapp/src/components/sidebar/sidebar.tsx
@@ -63,13 +63,15 @@ const Sidebar = React.memo((props: Props) => {
         )
     }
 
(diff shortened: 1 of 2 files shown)


jespino (approved):
LGTM.

I think this inferences about how to show the menu is ok for now, but probably we need some kind of global variable that defines if you are or not executing it as a plugin.

Issue #461, opened 2021-05-24
Feature Idea: Table column aggregations


Johennes:
**Summary:**

The table view type would be even more useful if it allowed for user-defined column aggregations. I'd be most interested in seeing a sum aggregation being implemented for number types. Other interesting numeric aggregations are minimum, maximum and average. Additionally, a count and unique count aggregation would be great to have irrespective of the data type.

**How important this is to me and why:**

Importance: Medium

Use cases:

1. I have a table of vacation days and want to compute the sum of days taken
2. I have a table of tasks and want to known the average story points / time estimate per task
3. I have a table of anything and want to know how many rows it contains

**Additional context / similar features:**

Here's a screenshot of the same feature in Notion.

![Screenshot 2021-05-24 at 19 44 48](https://user-images.githubusercontent.com/1137962/119387036-a43eeb00-bcc8-11eb-9bc3-ea2e4f0a6a64.png)

I don't think we'd need all of the aggregations Notion offers, at least not from the start.

harshilsharma63:
FYI I'm working on this.

harshilsharma63:
This was implemented in #761

PR #761, opened 2021-07-21
Table calculations

harshilsharma63:
#### Summary
Implemented table calculations

#### Ticket Link
Implements #461 

Calculations implemented -

1. count
1. countValue
1. countUniqueValue
1. sum
1. average
1. median
1. min
1. max
1. range


hahmadia:
Yo Harshil,

Before doing code review I decided to test it out. Here are some minor things I found after a quick run through. 

When the column too small, the range gets stacked on top like this ( I think it should be come ellipsis)
image

For date, the unique should be 3 here right?
image

Count values for multi select should be 7 here right?
image


For Checkbox, shouldn't false values be accounted for too?
image

Unique values here should be 2 (I assume we account for false values too if I'm not wrong)
image

Created Time 0 values? Is this right?
image


Created time unique values also 0?
image

Created by value is 0?
image


Created by unique value is also 0?
image

hahmadia (changes requested):
Found some minor bugs to address.

Note I didn't test every single property since some have overlap but you should check them all just to make sure I didn't miss something by accident. Thanks

harshilsharma63:
Thanks @hahmadia . I'm working on these.

harshilsharma63:
> When the column too small, the range gets stacked on top like this ( I think it should be come ellipsis)

Fixed.

> For date, the unique should be 3 here right?

Hmm, can you share an export of the board? Also can you verify it on the latest branch? I think it's fixed when I merged `main` in.

> Count values for multi select should be 7 here right?

Fixed

> For Checkbox, shouldn't false values be accounted for too?

Later on, we'll add checkbox specific values - `Checked` and `Unchecked`, those would make more sense. I think for now this is fine.

> Unique values here should be 2 (I assume we account for false values too if I'm not wrong)

Same as considering false values, later on, we'll add checkbox specific values - `Checked` and `Unchecked`, those would make more sense. I think for now this is fine.

> Created Time 0 values? Is this right?

Fixed

> Created time unique values also 0?

Fixed

> Created by value is 0?

Fixed

> Created by unique value is also 0?

Fixed

hahmadia:
Hi Harshil,

Regarding todays meeting, I would be in favor of having this QA review after dev review.

harshilsharma63:
@hahmadia you can review this now. I've synced this with `main`.

Agree on the QA point.

hahmadia:
Hi Harshil

Menu gets hidden on numbers. Also recommend adding "..." to range so we can know we should expand. I thought it was a bug but realized column was too small.

image

---

Have date like this

image

Delete first value

image

Change first value to '08/13/2021'
image

Unique value stays at 2

**Changing it in the card using the calendar works fine and doesn't cause this problem**

---

Shouldn't unique only be 4 for created time here? 

image

---

Last updated time unique should be 2 here right?
image



Otherwise nice job! Everything else seems to be working. I won't wait for you to fix these to not hold you back. I'll go forward with doing code review now.

webapp/src/components/calculations/calculations.ts
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Card} from '../../blocks/card'
+import {IPropertyTemplate} from '../../blocks/board'
+import {Utils} from '../../utils'
+import {Constants} from '../../constants'
+
+const roundedDecimalPlaces = 2
```
hahmadia, line 9: Probably wanna make this all caps

image

hahmadia (approved):
Looks good to me with the exception of incorrect calculations

ogi-m:
@hahmadia  Regarding the unique count on the dates:
If there's 3 dates in the column - 2x`August 13` and 1x`August 31` and `Unique` count shows 2 - isn't that correct?

hahmadia:
> @hahmadia  Regarding the unique count on the dates:
> 
> If there's 3 dates in the column - 2x`August 13` and 1x`August 31` and `Unique` count shows 2 - isn't that correct?

@ogi-m 

Yes sorry! That is correct.

ogi-m (approved):

chenilim:
@metanerd (I think you still own mattermod?), do you know if slash cherry-pick should work on this repo? If so, is there something that needs to be configured? Thanks.

Issue #431, opened 2021-05-17
Server Unit Tests - ./store/sqlstore/workspaces.go


sbishel:
Write Unit Tests for ./server/store/sqlstore/workspaces.go

see ./server/store/storetests for samples.

PR #768, opened 2021-07-22
[GH-431] Add unit tests for ./store/sqlstore/workspaces.go

yedamao:
#### Summary

Add unit tests for ./store/sqlstore/workspaces.go

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/431


yedamao:
insert workspace by store.UpsertWorkspaceSettings will failed, because signup_token NOT NULL constraint.
I'm not sure it's an expected error or a bug.

server/services/store/storetests/workspaces.go
```diff
@@ -0,0 +1,125 @@
...
+	t.Run("Insert and update workspace with signup token", func(t *testing.T) {
+		workspaceID := "0"
+		workspace := &model.Workspace{
+			ID:          workspaceID,
+			SignupToken: utils.CreateGUID(),
+		}
+
+		// insert
+		err := store.UpsertWorkspaceSignupToken(*workspace)
+		require.Nil(t, err)
```
wiggin77, line 50: Minor thing... could you use `require.NoError` for these cases.  It adds some extra type checking for the test.

yedamao:
@wiggin77 PTAL

server/services/store/storetests/workspaces.go
```diff
@@ -0,0 +1,125 @@
...
+		workspace := &model.Workspace{
+			ID: workspaceID,
+			Settings: map[string]interface{}{
+				"field1": "A",
+			},
+		}
+
+		// insert
+		err := store.UpsertWorkspaceSettings(*workspace)
+		// FIXME: insert workspace by store.UpsertWorkspaceSettings will failed
```
wiggin77, line 81: @jespino is this expected?

wiggin77 (approved):
LGTM :+1:

server/services/store/storetests/workspaces.go
jespino: No, it is not expected, you should be able to insert the workspace with the signup_token or make the signup_token not mandatory, clearly this is a bug.

jespino: I think this should create a random signup_token on create, but keep it untouched on update.

jespino:
I fixed the problem and merged the main branch. Thanks for finding this bug @yedamao! and for the new tests!

jespino (approved):

Issue #719, opened 2021-07-13
Bug: Dragging multiple cards does not work


chenilim:
**Steps to reproduce the behavior:**
1. Install Focalboard plugin v0.8.0 on Mattermost 5.37 (or macOS app, or Personal Server)
2. Or repro this on [mysql.test](https://mysql.test.mattermost.com/ogis-test-team/channels/town-square)
2. Create or open a board with 3 or more cards
3. In a board view, drag all cards to one column
4. Hold Shift, click on the top card
5. Hold Shift+Cmd, and click on the bottom card to select all 3
6. Drag the 3 cards to another column
7. Only one card is moved

**Expected behavior:**
All cards should move

**Screenshots:**
https://user-images.githubusercontent.com/46905241/125519445-80c2dc22-f755-4c56-8614-3c56b686729f.mov

**Platform:**
 - Focalboard plugin v0.8.0 on Mattermost v5.37 RC2, macOS app, and Personal Server

**Additional context:**
This appears to be a regression. Repros on c0.8.0 macOS app, plugin, and Personal Server. This worked on v0.7.3.

chenilim:
Updated: I'm able to repro this on v0.8.0 macOS App and on a fresh install of Personal Server, so this is not specific to the plugin or HA. Updated the steps. Reassigning to you @hahmadia for the initial investigation. Please reassign if appropriate. My first guess is this is a regression due to a front-end change.

hahmadia:
Not able to reproduce in Main

PR #771, opened 2021-07-22
Fix multiple cards not being able to drag into other columns in kanban view

hahmadia:
#### Summary

Multiple Cards not being able to be dragged into other columns in kanban view

#### Ticket Link

Fixes GH-719

image

image


hahmadia:
The problem here was that `selectedCardIds` was coming in as `[]` all the time (due to functional component problem). 

Thanks to Harshil for helping me debug it quick and solving the problem via the use of Ref's.

kamre:
@hahmadia can you also check `useDrop` dependencies for this issue?

webapp/src/components/kanban/kanban.tsx
```diff
@@ -103,19 +105,18 @@ const Kanban = (props: Props) => {
     }
 
     const onDropToColumn = async (option: IPropertyOption, card?: Card, dstOption?: IPropertyOption) => {
```
jespino, line 107: I would solve this with a `useCallback`, fix it like this for now.

jespino (approved):
LGTM

hahmadia:
@jespino @harshilsharma63 

Closing this PR as I was unable to reproduce this problem in Main so it seems to have been resolved.

Issue #626, opened 2021-06-24
Bug: Sometimes you can't move the cards between groups


JtheBAB:
**Summary:**
Sometimes you can't move the cards between groups. It looks like it is only(?) within empty boards. Sort is on Name

**Steps to reproduce the behavior:**
![Animation2](https://user-images.githubusercontent.com/42135332/123229171-56d7b880-d4d6-11eb-88d2-3432f59583bb.gif)

**Expected behavior:**
It should always work

**Platform:**
- Focalboard 0.6.7 within Mattermost 5.36
- Desktop App 4.6

chenilim:
Thanks @JtheBAB! Could you also provide details on the server setup (manual, docker, or MM Omnibus install? OS). If possible, also check the network and console tabs in the Chrome debugger. Are there any API errors? This might be related to #627.

JtheBAB:
Its self hosted manual setup. Network tab looks ok.

That is in the console when it happen. 

[1624631203.44] onDropToCard: bla copy copy copy copy
main.js:389 [1624631203.44] draggedCard: bla copy copy copy copy copy copy copy copy copy copy, column: undefined
main.js:389 [1624631203.44] insertBlocks: 1 blocks(s)
main.js:389 [1624631203.44] 	 view, fbfac3d2-7798-4b83-89e0-2efb655fbf5d, Board view
main.js:389 [1624631203.57] OctoListener flush update block: fbfac3d2-7798-4b83-89e0-2efb655fbf5d
main.js:389 [1624631203.57] workspaceListener.onChanged: 1
main.js:389 [1624631203.58] Sort by title
main.js:389 [1624631203.59] this.view.groupById card property not found: b43fa2a7-6adf-4eff-825f-dcb90beb0e5c
main.js:389 [1624631203.59] BoardPage.render (workspace p7ty6qhb6bbhtj9wykj5oktpjw) 
main.js:389 [1624631203.59] 5 propertyValues
main.js:389 [1624631203.60] componentDidUpdate

chenilim:
Thanks @JtheBAB! I notice the console log `OctoListener websocket onclose, code: 1006, reason`. Could you check that the [web proxy is configured](https://www.focalboard.com/download/mattermost/#with-nginx). That could explain these errors - if the websocket is intermittently dropping, then updates don't happen on changes.

JtheBAB:
I copied it to my nginx config before i upgraded to 5.36. And it is not always. When it would be a config error it should be always right?

[...]

Issue #708, opened 2021-07-11
Bug: can't drag card to column


kamre:
**Summary:**
Some cards can't be dragged to column in kanban view

**Steps to reproduce the behavior:**
1. Add Board => Roadmap
2. Select view By Sprint
3. Drag "Icons don't display" card to "SPRINT 1" column
4. Try to drag same card back to "SPRINT 2" column

**Expected behavior:**
Card should be in column "SPRINT 2" after dragging

**Platform:**
 - Browser and version: Chrome 91
 - OS and version: Windows 10

**Additional context:**
Looks like a regression in #583
@harshilsharma63 FYI

chenilim:
Thanks @kamre. I am not able to repro this (on v0.8.0 macOS App and plugin). Could you try this on the [latest release](https://github.com/mattermost/focalboard/releases/tag/v0.8.0) and attach a video?

Also, looks like #583 is a PR for the CreatedBy prop, was that what you intended to link to?

kamre:
@chenilim here is the video from latest release:
![GH-708-Can'tDragCardToColumn](https://user-images.githubusercontent.com/5587620/125560952-779644ac-e308-4725-b3e2-83a01bd0d161.gif)
This is windows application from `focalboard-win.zip` and fresh start (database removed in Documents folder).

Same behaviour can be seen for `focalboard-server-linux-amd64.tar.gz` when it is downloaded and started in WSL2.

Regarding #583: if we download windows application for revision 16708e93 before this PR was merged from [here](https://github.com/mattermost/focalboard/runs/3016336430) the issue is not reproduced.

kamre:
@chenilim @harshilsharma63 Regression happened during the transition from class to function component for `Kanban`.

In the class component there was an instance method `onDropToColumn` that was always the same and now with function component it is always a new closure function that is later passed into `KanbanColumn` and inside `useDrop` hook.

There are no dependencies for this `useDrop` hook and this leads to calls to the first version of closure for `onDropToColumn` which keeps the same data for `boardTree`.

Propose to fix this by adding dependecies to `useDrop` like this:
```ts
const KanbanColumn = React.memo((props: Props) => {
    const [{isOver}, drop] = useDrop(() => ({
        accept: 'card',
        collect: (monitor) => ({
            isOver: monitor.isOver(),
        }),
        drop: (item: Card, monitor) => {
            if (monitor.isOver({shallow: true})) {
                props.onDrop(item)
            }
        },
[...]

PR #772, opened 2021-07-22
Fix Cards not Dragging to different Columns (GH-626)

hahmadia:
#### Summary

If you create a new card and then try to drag to a new group, it won't work because boardTree is out of sync. 
Sometimes boardTree is not in sync and it causes cards not being able to be dragged to different columns hence we need to useRef to fix this.

#### Ticket Link

Fixes GH-626

### DIFF
--- a/webapp/src/components/kanban/kanban.tsx
+++ b/webapp/src/components/kanban/kanban.tsx
@@ -34,7 +34,11 @@ type Props = {
 
 const Kanban = (props: Props) => {
     const {boardTree} = props
-    const {cards, groupByProperty} = boardTree
+
+    const boardTreeRef = useRef<BoardTree>(boardTree)
+    boardTreeRef.current = boardTree
+
+    const {cards, groupByProperty} = boardTreeRef.current
 
     if (!groupByProperty) {
         Utils.assertFailure('Board views must have groupByProperty set')
@@ -44,7 +48,7 @@ const Kanban = (props: Props) => {
     const propertyValues = groupByProperty.options || []
     Utils.log(`${propertyValues.length} propertyValues`)
 
-    const {board, activeView, visibleGroups, hiddenGroups} = boardTree
+    const {board, activeView, visibleGroups, hiddenGroups} = boardTreeRef.current
     const visiblePropertyTemplates = board.cardProperties.filter((template) => activeView.visiblePropertyIds.includes(template.id))
     const isManualSort = activeView.sortOptions.length === 0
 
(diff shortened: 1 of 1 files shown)


kamre:
@hahmadia what is the reason for `boardTree` to be out of sync? Just switching to `useRef` looks like a workaround and not a fix for the real problem.

It seems that real problem is the same that I have described [here](https://github.com/mattermost/focalboard/issues/708#issuecomment-880497103). So we might just need to pass dependencies for `useDrop` inside `kanbanColumnHeader.tsx`

harshilsharma63 (approved):
LGTM!

kamre:
@harshilsharma63 @hahmadia The reason for `boardTree` to be out of sync is the following:
* everytime `const Kanban = (props: Props) => {` is called to render the component new closure is created and assigned to `const onDropToColumn = async (...) => {`
* this new closure for `onDropToColumn` is passed as callback prop to `KanbanColumnHeader`, `KanbanColumn` and `KanbanHiddenColumnItem`
* inside all of these components `useDrop` hook is used and `onDropToColumn` callback is used inside the object created for this hook
* `useDrop` has [additional argument for dependencies](https://react-dnd.github.io/react-dnd/docs/api/use-drop) with "_default value is an empty array for function spec_"
* so for the first render of `Kanban` the closure for `onDropToColumn` is created, passed to `useDrop` and is never updated inside it though new closures are created again and again that capture acutal values for `boardTree` and other variables in the scope

When `useRef` is used for the `boardTree` there is still the same first closure for `onDropToColumn` inside the `useDrop` hook. And this first closure actually **captures other variables** that **will not be in sync** and we will encounter the same problem as we have seen for `boardTree`.

Proper solution can be updating of the closure used for `useDrop` hook like this so that it doesn't cache the closure with outdated captured values:
```ts
export default function KanbanColumnHeader(props: Props): JSX.Element {
    ...
    const [{isOver}, drop] = useDrop(() => ({
        accept: 'column',
        collect: (monitor) => ({
            isOver: monitor.isOver(),
        }),
        drop: (item: IPropertyOption) => {
            props.onDropToColumn(item, undefined, group.option)
        },
    }), [props.onDropToColumn])  // <== add dependency
    ...
}

const KanbanColumn = React.memo((props: Props) => {
    const [{isOver}, drop] = useDrop(() => ({
        accept: 'card',
        collect: (monitor) => ({
            isOver: monitor.isOver(),
        }),
        drop: (item: Card, monitor) => {
            if (monitor.isOver({shallow: true})) {
                props.onDrop(item)
            }
        },
    }), [props.onDrop])   // <== add dependency
    ...
}

export default function KanbanHiddenColumnItem(props: Props): JSX.Element {
    const {boardTree, intl, group} = props
    const {activeView} = boardTree
    const [{isOver}, drop] = useDrop(() => ({
        accept: 'card',
        collect: (monitor) => ({
            isOver: monitor.isOver(),
        }),
        drop: (item: Card) => {
            props.onDrop(item)
        },
    }), [props.onDrop])  // <== add dependency
    ...
}
```

harshilsharma63:
@kamre you're right. The ref solves it for one variable, but what you're proposing will avoid issues with anything used in `onDrop` and `onDropToColumn`. @hahmadia can we make this change, please?

PR #777, opened 2021-07-24
InsertBlock per workspace logic

chenilim:
This reverts the InsertBlock logic to insert per workspace. This was changed in PR #583, and due to that change, blocks with the same ID in different workspaces would fail to update. It also exposes a security hole where the workspace of a block can be changed by an insert.

Currently, the primary key of a block is WorkspaceID+ID. If / when we change that, we will need to change it across the DB store system.


hahmadia (approved):

chenilim:
FYI @harshilsharma63, @sbishel, @jespino. I found this as an issue on the community server, and will cherry-pick this fix to `release-0.8.0`. In the mid-to-long term, we plan to refactor the block ID keys as part of the permission model changes, but this PR will maintain the existing consistency.

PR #800, opened 2021-07-29
Fix card loading on open card

jespino:


harshilsharma63 (approved):

webapp/src/components/cardDialog.tsx
```diff
@@ -27,6 +27,13 @@ const CardDialog = (props: Props) => {
     const [syncComplete, setSyncComplete] = useState(false)
     const [cardTree, setCardTree] = useState<CardTree>()
     const intl = useIntl()
+    useEffect(() => {
+        MutableCardTree.sync(props.cardId).then((ct) => {
+            setCardTree(ct)
+            setSyncComplete(true)
+        })
+    }, [props.boardTree])
```
kamre, line 35: @jespino why `props.boardTree` and not `props.cardId` or both as dependencies?

webapp/src/components/markdownEditor.tsx
```diff
@@ -52,6 +52,23 @@ const MarkdownEditor = (props: Props): JSX. Element => {
...
+        minHeight: '10px',
+        shortcuts: {
+            toggleStrikethrough: 'Cmd-.',
+            togglePreview: null,
+            drawImage: null,
+            drawLink: null,
+            toggleSideBySide: null,
+            toggleFullScreen: null,
+        },
+    }), [])
```
kamre, line 70: Why `useMemo` with closure and not just object literal?

PR #821, opened 2021-08-02
Improve cache invalidation on upgrades

jespino:


jespino:
It was already done by @harshilsharma63 in other way.

PR #822, opened 2021-08-02
[GH-817] Fix for auto-size column in Firefox

### LINKED ISSUE #817: Bug: Auto Size Column Doesn't Work In Firefox
**Summary:**
Auto sizing of columns doesn't work in Firefox. Also worth checking Safari.

**Steps to reproduce the behaviour:**
1. Open Focalboard in Firefox.
1. Create a new Roadmap board.
2. The initial board should display a board with a Priority column header truncated to Priorit...
3. Double click the resize handle on the right side of the Priority column.
4. The column size would still be the same/

**Expected behaviour:**
Double-clicking column resize handler should auto-resize column in Firefox and Safari.

**Platform:**
 - Browser and version: Firefox 90
 - OS and version: MacOS Big Sur

**Additional context:**
The feature works fine in Chrome.

kamre:
#### Summary
Fix usages of element's computed style in Firefox:
  - don't use shorthands like `borderLeft` or `font` because they are empty strings
  - construct font string that is required for setting into canvas context from
    individual font properties if `font` property is an empty string

#### Ticket Link
Fixes #817


jespino (approved):
Looks good to me. Thank @kamre! :confetti_ball:

PR #877, opened 2021-08-03
Adding support for patch blocks

jespino:
This is added because I want to improve the mutator before I migrate it to redux actions.

Also this improves the undo mechanism, because it doesn't undo the whole thing, only the field that you modified.


wiggin77 (approved):
Great work! :+1:

server/api/api.go
```diff
@@ -549,6 +550,83 @@ func (a *API) handleDeleteBlock(w http.ResponseWriter, r *http.Request) {
...
+	//   '200':
+	//     description: success
+	//   default:
+	//     description: internal error
+	//     schema:
+	//       "$ref": "#/definitions/ErrorResponse"
+
+	ctx := r.Context()
+	session := ctx.Value(sessionContextKey).(*model.Session)
+	userID := session.UserID
```
mgdelacroix, line 590: nit and non-specific to this PR: I think we could abstract this in a couple of helpers like `getSessionFromRequest(r)` and `getUserIDFromRequest(r)`, as I'm seeing this three lines repeated all over `api`

server/model/block.go
```diff
@@ -84,3 +116,34 @@ func (b Block) LogClone() interface{} {
...
+func (p *BlockPatch) Patch(block *Block) *Block {
+	if p.ParentID != nil {
+		block.ParentID = *p.ParentID
+	}
+	if p.RootID != nil {
+		block.RootID = *p.RootID
+	}
+	if p.Schema != nil {
+		block.Schema = *p.Schema
+	}
```
mgdelacroix, line 130: nit for consistency:

```suggestion
	if p.ParentID != nil {
		block.ParentID = *p.ParentID
	}
	
	if p.RootID != nil {
		block.RootID = *p.RootID
	}
	
	if p.Schema != nil {
		block.Schema = *p.Schema
	}
```

server/swagger/swagger.yml
```diff
@@ -292,7 +335,9 @@ info:
   license:
     name: Custom
     url: https://github.com/mattermost/focalboard/blob/main/LICENSE.txt
-  title: Focalboard Server
+  title: |-
+    line /home/jespino/Projects/mattermost/focalboard/server/main/main.go:1:1
```
mgdelacroix, line 339: :eyes:

mgdelacroix (approved):
Awesome work!! :+1:

Left a couple of nonblocking nits

server/swagger/swagger.yml
jespino: Yes, something weird there. I'll check it

server/api/api.go
jespino: I like the explicit better than implicit. Anyway, we can decide to move in that direction but I would prefer to do the migration all at once in another PR.

server/model/block.go
jespino: I don't see the difference there, what have you changed?

server/api/api.go
mgdelacroix: Agree on postponing this to another PR, but I don't think we're hiding any complexity here but simplifying something that we're doing almost everywhere. We already have helpers that go in that direction, like [`GetContextConn`](https://github.com/mattermost/focalboard/blob/main/server/api/context.go#L22), so I'm thinking in something close to that

Issue #879, opened 2021-08-03
Update Shared Board text to be more clear


chenilim:
**Summary:**
Update the Share Board text to read: "Anyone with the link can view this board and all cards in it".

Existing dialog for reference:
<img width="436" alt="Shrare board dialog" src="https://user-images.githubusercontent.com/46905241/128045152-ba38f619-4cff-4d28-baf8-3f67fc9cd9ab.png">

chenilim:
@sbishel, please assign to Dev for v0.9.
FYI @asaadmahmood - we could also enhance this with a small "warning" icon to the left of the text if you think that's appropriate, but that can be a P2.

PR #880, opened 2021-08-03
update text for shared board

sbishel:
#### Summary
Updates Text when sharing a board.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/879

Screen Shot 2021-08-03 at 10 53 18 AM


chenilim (approved):
👍

asaadmahmood (approved):
Looks good, don't think we need an alert here, I think this should work for now.

Issue #684, opened 2021-07-07
Bug: Double-clicking on a multi-select table column resizes too large


chenilim:
**Steps to reproduce the behavior:**
1. Open a table with a multi-select column
2. Set one of the multi-select values to at least 2 or 3 items
3. Double click on the column border to auto-size it
4. The column autosizes too large (by over double)

**Expected behavior:**
The auto-size fits the maximum number of displayed items with a reasonable margin.

**Screenshots:**
https://user-images.githubusercontent.com/46905241/124798667-7922fa80-df08-11eb-92af-5de4f12f88f9.mov

**Platform:**
- Focalboard plugin v0.7.3
- Chrome 91 on macOS 11.4

**Additional context:**
Having additional selected items may cause the size error % to be even greater.

chenilim:
This appears to be a bug in the table column resize logic. Setting it as up for grabs. cc @sbishel.

sbishel:
Yes, multiSelect is not implemented [here](https://github.com/mattermost/focalboard/blob/main/webapp/src/octoUtils.tsx#L19)

Shahzayb:
Hey @sbishel, I've done some debugging and found this. Hope that'll be useful.

`columnRefs` only contains refs of the cells of last card/row in a table, which is making it difficult to actually pick the multiSelect cell with max number of options. We need a multiSelect cell with max selected options to properly calculate the `padding`.
https://github.com/mattermost/focalboard/blob/caedc771b48ddf42ba7a738ccb88f24250a25748/webapp/src/components/table/table.tsx#L85-L89

And the `multiSelect` case in the `OctoUtils.propertyDisplayValue` can be implemented like this.

```
        case 'multiSelect': {
            if (propertyValue?.length) {
                const options = propertyTemplate.options.filter((o) => propertyValue.includes(o.id))
                if (!options?.length) {
                    Utils.assertFailure(`Invalid multiSelect option IDs ${propertyValue}, block.title: ${block.title}`)
                }
                const optionValues = options?.map((o) => o.value)
                displayValue = optionValues?.length ? optionValues.join(' ') : '(Unknown)'
            }
            break
        }
```
Also need to handle a `multiSelect` case in here:
https://github.com/mattermost/focalboard/blob/caedc771b48ddf42ba7a738ccb88f24250a25748/webapp/src/components/table/table.tsx#L91

sbishel:
@Shahzayb 
[...]

PR #882, opened 2021-08-03
[Gh 684] Autosize fix for multi-select columns

sbishel:
#### Summary
Fixes an issue where the autosize is not calculated correctly for multi-select fields.

#### Ticket Link
https://github.com/mattermost/focalboard/issues/684


webapp/src/components/properties/multiSelect.tsx
```diff
@@ -44,6 +42,9 @@ const MultiSelectProperty = (props: Props): JSX.Element => {
                         {v.value}
                     </Label>
                 ))}
+                {values.length === 0 && (
+                    <Label color='empty'>{''}</Label>
+                )}
```
sbishel, line 47: This is added so that if the selected row contains no values, the padding can still be retrieved. Otherwise, there is no way of determining the padding.

webapp/src/components/properties/multiSelect.tsx
```diff
@@ -25,9 +25,7 @@ const MultiSelectProperty = (props: Props): JSX.Element => {
     const {propertyTemplate, emptyValue, propertyValue, isEditable, onChange, onChangeColor, onDeleteOption, onCreate, onDeleteValue} = props
     const [open, setOpen] = useState(false)
 
-    const values = Array.isArray(propertyValue) ?
-        propertyValue.map((v) => propertyTemplate.options.find((o) => o!.id === v)).filter((v): v is IPropertyOption => Boolean(v)) :
-        []
+    const values = Array.isArray(propertyValue) && propertyValue.length > 0 ? propertyValue.map((v) => propertyTemplate.options.find((o) => o!.id === v)).filter((v): v is IPropertyOption => Boolean(v)) : []
```
kamre, line 28: Is there a limit for the line length for the project code style?

webapp/src/components/table/table.tsx
```diff
@@ -86,21 +86,63 @@ const Table = (props: Props): JSX.Element => {
...
-                if (!template) {
-                    return
+            let thisLen = 0
+            if (columnID === Constants.titleColumnId) {
+                thisLen = Utils.getTextWidth(card.title, columnFontPadding.fontDescriptor) + columnFontPadding.padding
+            } else if (template) {
+                const displayValue = (OctoUtils.propertyDisplayValue(card, card.fields.properties[columnID], template as IPropertyTemplate, intl) || '')
+                switch (template.type) {
+                case 'select': {
+                    thisLen = Utils.getTextWidth(displayValue.toString().toUpperCase(), columnFontPadding.fontDescriptor) + perItemPadding
```
kamre, line 127: Why `perItemPadding` is used for `select` type? It is not computed above for `select` type and has zero value.

webapp/src/components/properties/multiSelect.tsx
sbishel: Not exactly sure what you're asking. We currently don't have a limit on either the line length nor the number of property options, which this line is dealing with.

webapp/src/components/table/table.tsx
sbishel: Originally this was going to be a shared line. However, when that didn't work, I left in because it wasn't getting set above, so didn't matter. But since it has caused confusion, I removed here and in default case. I only left it where necessary.

webapp/src/components/properties/multiSelect.tsx
kamre: Was asking about the line length for the source code.

kamre (approved):

harshilsharma63 (approved):

webapp/src/components/table/table.tsx
```diff
@@ -86,21 +86,63 @@ const Table = (props: Props): JSX.Element => {
...
+            template = visibleProperties.find((t: IPropertyTemplate) => t.id === columnID)
+            if (!template) {
+                return
+            }
+            if (template.type === 'multiSelect') {
+                // For multiselect, the padding calculated above depends on the number selected when calculating the padding.
+                // Need to calculate it manually here.
+                // DOM Object hierarchy should be {cell -> property -> [value1, value2, etc]}
+                let valueCount = 0
+                if (columnRef.current && columnRef.current.childElementCount > 0) {
```
hahmadia, line 103: nit:
```suggestion
                if (columnRef?.current?.childElementCount > 0) {
```

hahmadia (approved):
Nice! Minor nits but the auto resizing actually works better with your PR than it did originally even with single select when I tried it. 👍

PR #883, opened 2021-08-04
Making theming compatible with mattermost on plugin mode

jespino:


jespino:
@harshilsharma63 @ogi-m @asaadmahmood. This is the themes support for plugin mode. This allows to re-use the existing mattermost themes in focalboard. This will allow to finish global header PR removing the set theme menu. Probably we should do the same with the language, but we can do it later.

Please give this certain priority because this is in the critical path.

harshilsharma63 (approved):

asaadmahmood (approved):

ogi-m:
@jespino In addition to the issue we discussed yesterday where the different themes per team are not applied, I'm also seeing this:
1. Install and enable the plugin
2. Open Team A - current theme is default (white/blue)
3. Open focalboard - default theme also applied
4. In Team A, change the theme to Mattermost Dark
5. Open focalboard - theme not updated, still shows the default light theme.

jespino:
@ogi-m this is not handling the theme change events, but if you refresh should work.

I think the 2 problems here are:
1. We have to support per team theming.
2. We have to support websocket events to update the theme on the fly.

Both problems are, in my opinion, minor problems that can be addressed after merge. This is going to affect a subset of the users or can be solved with a refresh.

What do you think @ogi-m and @chenilim? Are you ok merging it and iterating later?

ogi-m:
I tried refreshing but it doesn't fix it @jespino

chenilim:
I agree with merging this and fixing those 2 issues later, but by 0.9 / Sep release. If so, please file GitHub issues to track them and add a v0.9 milestone.

ogi-m (approved):
@chenilim @jespino I've filled the issues here:
- https://github.com/mattermost/focalboard/issues/901
- https://github.com/mattermost/focalboard/issues/902

jespino:
Thanks @ogi-m!!

PR #884, opened 2021-08-05
Write Test for cardDetailContents

### LINKED ISSUE #842: WebApp Unit Tests for ./webapp/src/components/cardDetail/cardDetailContents.tsx
Write Unit Tests for `./webapp/src/components/cardDetail/cardDetailContents.tsx`

see `./webapp/src/components/table/*` or`./webapp/src/components/properties/dateRange/dateRange.test.tsx` for sample tests

hahmadia:
#### Summary

Write tests for cardDetailContents.tsx

#### Ticket Link

Fixes #842


webapp/src/components/cardDetail/cardDetailContentsUtility.ts
```diff
@@ -0,0 +1,71 @@
...
+    dstBlockId: string,
+    moveTo: Position,
+}
+
+export const dragAndDropRearrange =
+    (data: DragAndDropRearrangeData): Array<string|string[]> => {
+        const {contentOrder, srcBlockX, srcBlockY, dstBlockX, dstBlockY, moveTo, srcBlockId, dstBlockId} = data
+        const newContentOrder: Array<string|string[]> = JSON.parse(JSON.stringify(contentOrder))
+
+        const copySrcBlockX = srcBlockX
```
harshilsharma63, line 21: This won't be a deep copy and only a reference. Will it work?

harshilsharma63 (dismissed):

webapp/src/components/cardDetail/cardDetailContents.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+const wrapProviders = (children: any) => {
+    return (
+        <DndProvider backend={HTML5Backend}>
+            <IntlProvider locale='en'>{children}</IntlProvider>
+        </DndProvider>
+    )
+}
+
+global.fetch = jest.fn()
```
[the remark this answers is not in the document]
hahmadia: The reason I added it is because Jest complains about not having it.
image

webapp/src/components/cardDetail/cardDetailContentsUtility.ts
hahmadia: Yep these are primitives (numbers) so it's okay.

sbishel (approved):
Nice changes to make it more testable.

webapp/src/components/cardDetail/cardDetailContents.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+    range.getClientRects = () => {
+        return {
+            item: () => null,
+            length: 0,
+            [Symbol.iterator]: jest.fn(),
+        }
+    }
+    return range
+}
+window.focus = jest.fn()
```
kamre, line 41: Should this code be replaced via calling [recently added mockDOM](https://github.com/mattermost/focalboard/blob/main/webapp/src/testUtils.tsx#L8-L25)?

kamre, line 41: Should this code be replaced with calling [recently added mockDOM](https://github.com/mattermost/focalboard/blob/main/webapp/src/testUtils.tsx#L8-L25)?

webapp/src/components/cardDetail/__snapshots__/cardDetailContents.test.tsx.snap
```diff
@@ -0,0 +1,979 @@
...
+        </div>
+      </div>
+      <div
+        class="addToRow "
+        style="width: 94%; height: 10px; margin-left: 48px;"
+      />
+    </div>
+  </div>
+</div>
+`;
```
kamre, line 979: @sbishel @hahmadia does anyone really review these 1k of lines inside the snapshot?

webapp/src/components/cardDetail/cardDetailContents.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+        ))
+
+        let container: Element | undefined
+        await act(async () => {
+            const result = render(component)
+            container = result.container
+        })
+        const markdownEditorField = container!.querySelectorAll('.octo-editor-preview.octo-placeholder')
+        expect(markdownEditorField).toBeDefined()
+        fireEvent.click(markdownEditorField[markdownEditorField.length - 1]!)
```
kamre, line 126: Why the last element of `markdownEditorField` is used? Should we expect to have only one element in this NodeList?

webapp/src/components/cardDetail/cardDetailContents.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+        })
+
+        const onBlurEvent = new FocusEvent('blur', {
+            view: window,
+            bubbles: true,
+            cancelable: true,
+        })
+
+        const textareaContainer = container!.querySelectorAll('.CodeMirror.cm-s-easymde.CodeMirror-wrap')
+        const textarea = textareaContainer[textareaContainer.length - 1].querySelector('textarea')
```
kamre, line 141: Same question about `textareaContainer`: do we expect to have single element in this NodeList?

webapp/src/components/cardDetail/__snapshots__/cardDetailContents.test.tsx.snap
hahmadia: I usually do a quick scan for keywords in each snapshot test to ensure what I am looking for is there but I don't look through ever line just keyword classes / text I'd expect to be present in the snapshot.

webapp/src/components/cardDetail/cardDetailContents.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+                contents={contents}
+                readonly={false}
+            />
+        ))
+
+        let container: Element | undefined
+        await act(async () => {
+            const result = render(component)
+            container = result.container
+        })
```
kamre, line 168: Looks like this code is repeated in several tests. Does it make sense to extract it to function?

webapp/src/components/cardDetail/cardDetailContents.test.tsx
hahmadia: I did this on my own through web browser and tested what elements are present. There are several that come up with that class so and the last one happens to be the content space I was looking for

webapp/src/components/cardDetail/cardDetailContents.test.tsx
hahmadia: No we don't. We tried `querySelector` but we were getting the wrong element. So we had to use All and the last one happens to be the one I want via trial and error through an actual browser.

webapp/src/components/cardDetail/cardDetailContents.test.tsx
hahmadia: I don't think so. Then that function would be repeated. I don't feel it's necessary to extract to a function here.

webapp/src/components/cardDetail/cardDetailContents.test.tsx
hahmadia: Done!

webapp/src/components/cardDetail/__snapshots__/cardDetailContents.test.tsx.snap
kamre: > quick scan for keywords

Sounds like this https://testing-library.com/docs/queries/bytext which can probably be done by the test itself.

hahmadia: Thanks Kamre

webapp/src/components/cardDetail/cardDetailContents.test.tsx
kamre: This is really strange. Here is how it looks in the debugger:
image
Also in the snapshot there is only one element matching the query.

hahmadia: Oh cool! I didn't notice that in the test. Not a big deal but I switch from querySelectorAll to just querySelector. Thanks

webapp/src/components/cardDetail/cardDetailContents.test.tsx
```diff
@@ -0,0 +1,248 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import React from 'react'
+
+import {fireEvent, render} from '@testing-library/react'
+
+import {IntlProvider} from 'react-intl'
+import {act} from 'react-dom/test-utils'
```
Shahzayb, line 8: > It is recommended to use the import from @testing-library/react over react-dom/test-utils for consistency reasons.

https://testing-library.com/docs/react-testing-library/api#act

PR #898, opened 2021-08-06
Fix share boards on plugin mode

### LINKED ISSUE #887: Bug: clicking on share board option in plugin mode crashes the app
1. Open Focalboard as a Mattermost plugin
2. Click on the share board button in the view. Notice the app crash.

**Expected behavior:**
Share modal should open and the app shouldn't crash.

**Screenshots:**
See attached viewo

https://user-images.githubusercontent.com/18575143/128362353-d7972379-5423-4547-b805-9ec5b4325554.mp4

**Platform:**
 - Browser and version: Firefox
 - OS and version: MacOS, latest Mattermost (master branch) and latest Focalboard (main branch)

**Additional context:**
The issue is somewhere in the `ShareBoardComponent` component.

jespino:
This was related to the fact that we were using the mattermost-webapp react-router-dom package, and we should use our own version (because the versions are not fully compatible).

Also I added an extra change to fix the MM_DEBUG flag that was causing problems.

Fixes #887


mattermost-plugin/webapp/package.json
```diff
@@ -68,17 +68,15 @@
...
   },
   "dependencies": {
     "core-js": "3.12.1",
     "mattermost-redux": "5.33.1",
-    "react": "17.0.2",
     "react-intl": "^5.13.5",
-    "react-redux": "7.2.4",
-    "redux": "4.1.0",
-    "typescript": "4.2.4"
+    "react-router-dom": "5.2.0"
```
jespino, line 79: This line fix the problem, the other lines just clean this up to not include the not-needed dependencies (they are used from the mattermost-webapp).

mattermost-plugin/webapp/webpack.config.js
```diff
@@ -11,12 +13,17 @@ const PLUGIN_ID = require('../plugin.json').id;
 const NPM_TARGET = process.env.npm_lifecycle_event; //eslint-disable-line no-process-env
 let mode = 'production';
 let devtool;
+const plugins = [];
 if (NPM_TARGET === 'debug' || NPM_TARGET === 'debug:watch') {
     mode = 'development';
     devtool = 'source-map';
+    plugins.push(
```
jespino, line 20: This fixes the `MM_DEBUG=true`

mattermost-plugin/webapp/webpack.config.js
```diff
@@ -120,7 +127,6 @@ module.exports = {
         'react-redux': 'ReactRedux',
         'prop-types': 'PropTypes',
         'react-bootstrap': 'ReactBootstrap',
```
jespino, line 129: This removed line is also part of the fix for the share boards problem.

sbishel (approved):
I'm not the best person to review. But we should probably get on the same version of dependencies if possible.

mgdelacroix (approved):

PR #905, opened 2021-08-07
Add step to delete default at sites-available

rukku:
The default site config needs to be deleted on both sites-available and sites-enabled otherwise the config test will fail.

### DIFF
--- a/website/site/content/download/personal-edition/ubuntu.md
+++ b/website/site/content/download/personal-edition/ubuntu.md
@@ -104,6 +104,7 @@ If there is a default site, you may need to delete it
 
 ```
 sudo rm /etc/nginx/sites-enabled/default
+sudo rm /etc/nginx/sites-available/default
 ```
 
 Enable the Focalboard site, test the config, and reload NGINX:


website/site/content/download/personal-edition/ubuntu.md
```diff
@@ -104,6 +104,7 @@ If there is a default site, you may need to delete it
 
 ```
 sudo rm /etc/nginx/sites-enabled/default
+sudo rm /etc/nginx/sites-available/default
```
jespino, line 107: I don't think this is necessary. `sites-available` is only a place to put your different sites configuration and then you link them from `sites-enabled`, removing that file make it harder to revert the changes and restore the default behavior of the nginx server. It is not a big deal, but I don't see the need, and I only see the drawback.

rukku: I tried following the install instructions and not removing it in `sites-available` results in an error when when reloading nginx. At least it did for me on Ubuntu 20.04.

jespino: It shouldn't in a regular nginx configuration, nginx configuration imports `/etc/nginx/sites-enabled/*` it doesn't import anything from `/etc/nginx/sites-available/`. It can be an ubuntu nginx package problem, or a configuration problem of your nginx.

You can see more info here: https://serverfault.com/questions/527630/difference-in-sites-available-vs-sites-enabled-vs-conf-d-directories-nginx

jespino:
I going to close this. @rukku If you feel strong about keep adding it we can keep talking.

Issue #858, opened 2021-08-02
WebApp Unit Tests for ./webapp/src/components/properties/multiSelect.tsx


sbishel:
Write Unit Tests for `./webapp/src/components/properties/multiSelect.tsx`

see `./webapp/src/components/table/*` or`./webapp/src/components/properties/dateRange/dateRange.test.tsx` for sample test

Shahzayb:
I'd like to work on this.

Shahzayb:
I'm working on this.

Shahzayb:
I'm not gonna mock `react-select/creatable`. Instead, I'm gonna use this lib https://github.com/romgain/react-select-event

PR #906, opened 2021-08-08
[GH-858] Add test cases for the multi-select component

Shahzayb:
#### Summary

Added test cases for the multi-select component

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/858


webapp/src/widgets/valueSelector.tsx
```diff
@@ -136,44 +142,46 @@ const valueSelectorStyle = {
...
-            options={props.options}
-            getOptionLabel={(o: IPropertyOption) => o.value}
-            getOptionValue={(o: IPropertyOption) => o.id}
-            onChange={(value: ValueType<IPropertyOption, true | false>, action: ActionMeta<IPropertyOption>): void => {
-                if (action.action === 'select-option') {
-                    if (Array.isArray(value)) {
-                        props.onChange((value as IPropertyOption[]).map((option) => option.id))
-                    } else {
-                        props.onChange((value as IPropertyOption).id)
+        <div data-testid='creatable-selector-parent'>
```
jespino, line 145: The `data-testid` is great whenever we strictly need that, but I think we don't need that for certain cases. For example, in general, all our components have a class, based on the component type, in this case the class `ValueSelector`, I think we can relay on that, instead of using the `data-testid`. That makes a bit more "artificial" the test cases. I think is a great solution in some cases, but here I think is not needed.

webapp/src/widgets/menu/textOption.tsx
```diff
@@ -18,6 +19,7 @@ function TextOption(props:TextOptionProps): JSX.Element {
     }
     return (
         <div
+            data-testid={props['data-testid']}
```
jespino, line 22: I'm ok with this `data-testid` but probably we can use directly the `props.id` field there, instead of requiring an extra id.

webapp/src/widgets/menu/colorOption.tsx
```diff
@@ -8,12 +8,14 @@ import './colorOption.scss'
...
 type ColorOptionProps = MenuOptionProps & {
     icon?: React.ReactNode
+    'data-testid'?: string
 }
 
 function ColorOption(props: ColorOptionProps): JSX.Element {
     const {id, name, icon} = props
     return (
         <div
+            data-testid={props['data-testid']}
```
jespino, line 18: Same here, I would use `props.id` instead of a new property.

webapp/src/widgets/valueSelector.tsx
```diff
@@ -54,6 +54,7 @@ const ValueSelectorLabel = React.memo((props: LabelProps): JSX.Element => {
                 <span className='Label-text'>{option.value}</span>
                 {onDeleteValue &&
                     <IconButton
+                        data-testid={`delete-value-${option.id}`}
```
jespino, line 57: Probably you can use here the `.CloseIcon` class to test it, instead of the need of an specific.

webapp/src/widgets/valueSelector.tsx
```diff
@@ -70,9 +71,13 @@ const ValueSelectorLabel = React.memo((props: LabelProps): JSX.Element => {
                 <Label color={option.color}>{option.value}</Label>
             </div>
             <MenuWrapper stopPropagationOnToggle={true}>
-                <IconButton icon={<OptionsIcon/>}/>
+                <IconButton
+                    data-testid={`delete-option-menu-${option.id}`}
```
jespino, line 75: Same here with `.OptionsIcon`

jespino (changes requested):
I see a bit of abuse of `data-testid` usage, it is generating some unneeded code and I think we can reduce the number to the ones that are strictly needed.

webapp/src/widgets/buttons/iconButton.tsx
```diff
@@ -19,10 +19,12 @@ function IconButton(props: Props): JSX.Element {
     }
     return (
         <div
+            role='button'
```
jespino, line 22: nice addition!

webapp/src/widgets/menu/colorOption.tsx
```diff
@@ -14,6 +14,8 @@ function ColorOption(props: ColorOptionProps): JSX.Element {
     const {id, name, icon} = props
     return (
         <div
+            role='button'
+            aria-label={`Select ${name} Color`}
```
jespino, line 18: This text should be translatable.

webapp/src/widgets/valueSelector.tsx
```diff
@@ -70,7 +70,10 @@ const ValueSelectorLabel = React.memo((props: LabelProps): JSX.Element => {
                 <Label color={option.color}>{option.value}</Label>
             </div>
             <MenuWrapper stopPropagationOnToggle={true}>
-                <IconButton icon={<OptionsIcon/>}/>
+                <IconButton
+                    title='Open Menu'
```
jespino, line 74: This text should be translatable.

webapp/src/widgets/valueSelector.tsx
```diff
@@ -137,6 +140,7 @@ const valueSelectorStyle = {
 function ValueSelector(props: Props): JSX.Element {
     return (
         <CreatableSelect
+            aria-label='Value selector'
```
jespino, line 143: This text should be translatable.

webapp/src/components/properties/multiSelect.tsx
```diff
@@ -32,12 +32,13 @@ const MultiSelectProperty = (props: Props): JSX.Element => {
...
                 className='octo-propertyvalue'
                 tabIndex={0}
+                data-testid='multiselect-non-editable'
                 onClick={() => setOpen(true)}
             >
                 {values.map((v) => (
                     <Label
                         key={v.id}
-                        color={v ? v.color : 'empty'}
+                        color={v.color}
```
jespino, line 41: good catch, I didn't noticed that we were already cleaning up the `values` array before :)

jespino (changes requested):
Just a couple of tiny changes related to translations needed, but otherwise it looks great. Thanks @Shahzayb! :confetti_ball:

webapp/i18n/en.json
```diff
@@ -6,8 +6,10 @@
   "BoardComponent.new": "+ New",
   "BoardComponent.no-property": "No {property}",
   "BoardComponent.no-property-title": "Items with an empty {property} property will go here. This column cannot be removed.",
+  "BoardComponent.openMenu": "Open Menu",
+  "BoardComponent.selectColor": "Select {color} Color",
   "BoardComponent.show": "Show",
-  "BoardPage.syncFailed": "Board may be deleted or access revoked.",
```
Shahzayb, line 10: This property is removed automatically by running the following command. `npm run i18n-extract`

jespino: Thats ok :)

webapp/src/widgets/menu/colorOption.tsx
```diff
@@ -12,8 +13,11 @@ type ColorOptionProps = MenuOptionProps & {
 
 function ColorOption(props: ColorOptionProps): JSX.Element {
     const {id, name, icon} = props
+    const intl = useIntl()
     return (
         <div
+            role='button'
+            aria-label={intl.formatMessage({id: 'BoardComponent.selectColor', defaultMessage: 'Select {color} Color'}, {color: name})}
```
jespino, line 20: I think the id here should be `ColorOption.selectColor`, not `BoardComponent.selectColor`.

webapp/src/widgets/valueSelector.tsx
```diff
@@ -70,7 +70,10 @@ const ValueSelectorLabel = React.memo((props: LabelProps): JSX.Element => {
                 <Label color={option.color}>{option.value}</Label>
             </div>
             <MenuWrapper stopPropagationOnToggle={true}>
-                <IconButton icon={<OptionsIcon/>}/>
+                <IconButton
+                    title={intl.formatMessage({id: 'BoardComponent.openMenu', defaultMessage: 'Open Menu'})}
```
jespino, line 74: Same here, I prefer `ValueSelectorLabel.openMenu`

webapp/src/widgets/valueSelector.tsx
```diff
@@ -135,8 +138,10 @@ const valueSelectorStyle = {
 }
 
 function ValueSelector(props: Props): JSX.Element {
+    const intl = useIntl()
     return (
         <CreatableSelect
+            aria-label={intl.formatMessage({id: 'BoardComponent.valueSelector', defaultMessage: 'Value selector'})}
```
jespino, line 144: And here I prefer `ValueSelector.valueSelector`

jespino (changes requested):
Great work, but the namespace in the translations ids looks wrong. (Remember to regenerate the translation after the changes to update the `i18n/en.json` file).

jespino (approved):
Awesome work, thanks!

PR #922, opened 2021-08-10
Adding backward compatibility for 5.38

jespino:


mattermost-plugin/webapp/src/index.tsx
```diff
@@ -80,21 +80,33 @@ export default class Plugin {
...
+                        history.goBack()
+                    }
+                }, [])
+                return <></>
+            })
             this.registry.registerProduct('/boards', GlobalHeaderIcon, 'Boards', '/plug/focalboard/go-to-current-workspace', MainApp, HeaderComponent)
+        } else {
+            this.channelHeaderButtonId = registry.registerChannelHeaderButtonAction(<FocalboardIcon/>, () => {
+                const currentChannel = store.getState().entities.channels.currentChannelId
+                window.open(`${window.location.origin}/plug/focalboard/workspace/${currentChannel}`)
```
chenilim, line 107: I believe this should be `/plugins/focalboard/...`

chenilim (dismissed):
Looks like the url needs to be updated. Another issue is the Compass icon is not visible in the channel header on 5.38.

mattermost-plugin/webapp/src/index.tsx
jespino: The problem was with other thing, see the commit that I included.

mgdelacroix (approved):

PR #923, opened 2021-08-10
Picking the default theme from the user preferences on reload when it is a plugin

jespino:


hahmadia (approved):

webapp/src/theme.ts
```diff
@@ -169,6 +169,29 @@ export function setTheme(theme: Theme | null): Theme {
...
+    document.documentElement.style.setProperty('--center-channel-bg-rgb', color(theme.centerChannelBg).rgb().array().join(', '))
+    document.documentElement.style.setProperty('--center-channel-color-rgb', color(theme.centerChannelColor).rgb().array().join(', '))
+    document.documentElement.style.setProperty('--button-bg-rgb', color(theme.buttonBg).rgb().array().join(', '))
+    document.documentElement.style.setProperty('--button-color-rgb', color(theme.buttonColor).rgb().array().join(', '))
+    document.documentElement.style.setProperty('--sidebar-bg-rgb', color(theme.sidebarBg).rgb().array().join(', '))
+    document.documentElement.style.setProperty('--sidebar-text-rgb', color(theme.sidebarText).rgb().array().join(', '))
+    document.documentElement.style.setProperty('--link-color-rgb', color(theme.linkColor).rgb().array().join(', '))
+    document.documentElement.style.setProperty('--sidebar-text-active-border', color(theme.sidebarTextActiveBorder).rgb().array().join(', '))
+
+    return setTheme({
```
hahmadia, line 182: Why do we need to return this if we aren't using it?

hahmadia (approved):

webapp/src/theme.ts
jespino: It is for consistency, in this case, this is equivalent to `setTheme` but for mattermost theme.

ogi-m:
Tested and passed - now working as expected.

PR #927, opened 2021-08-10
Use classic icon in Personal Editions

chenilim:
This will continue to use the classic icon for non-plugin Personal Editions, to match the icon used elsewhere, such as the App Stores and app icons.


wiggin77 (approved):
LGTM :+1:

PR #932, opened 2021-08-11
Fix flaky test

jespino:


hahmadia (approved):

wiggin77:
What if 100ms is not enough in a busy server, or when run locally?

jespino:
@wiggin77 this is intentionally hacky, I know 100ms has no meaning and still can fail, but my intention was to not waste too much time. The underneath problem was codemirror using a trick changing the text to "xxxxxxx" to measure the size of the font, or something like that. that change sometimes get reflected in the snapshot, giving this 100ms, in my theory, helps to settle down the current value and avoid this possible weird case.

PR #970, opened 2021-08-13
Cleanup the store on the initial load call

### LINKED ISSUE #969: Bug: Opening boards from app switcher combines the workspace with the one from the previous channel
**Summary:**
When a user opens focalboard using the app switcher, the sidebar will show boards from the channel that boards belonging to the current channel's workspace, as well as the board belonging to the channel that focalboard was previously opened from.

See screenshot for example - the workspace for R&D Meeting channel is open, but the boards from Focalboard channel are also shown (which was the channel I previously accessed boards from).

This doesn't happen when opening via the channel header icon, only 

1. In Channels open _Channel 1_
2. Click on the app switcher and open Boards
3. Note the boards for the channel are displayed in the sidebar
4. Return to Channels
5. Open _Channel 2_
6. Click on the app switcher and open Boards

**Expected behavior:**
Boards sidebar should only show boards belonging to _Channel 2_ 
(instead it shows boards belonging to both _Channel 1_ and _Channel 2_)

**Screenshots:**

Boards
when using the app switcher.

**Platform:**
 - 0.9 RC4 plugin on master (community-daily)
 - Chrome v92

jespino:
Fixes #969

### DIFF
--- a/webapp/src/store/boards.ts
+++ b/webapp/src/store/boards.ts
@@ -37,6 +37,8 @@ const boardsSlice = createSlice({
     },
     extraReducers: (builder) => {
         builder.addCase(initialReadOnlyLoad.fulfilled, (state, action) => {
+            state.boards = {}
+            state.templates = {}
             for (const block of action.payload) {
                 if (block.type === 'board' && block.fields.isTemplate) {
                     state.templates[block.id] = block as Board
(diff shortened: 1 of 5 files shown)


hahmadia (approved):

wiggin77 (approved):
LGTM :+1:

Issue #955, opened 2021-08-12
Bug: Links are not highlighted in the card markdown content


chenilim:
**Steps to reproduce the behavior:**
1. Create a card
2. Add a link to the description, e.g. `Here's [a link](http://www.focalboard.com)`
3. Exit editing
4. The link is not highlighted (e.g. underlined and in blue)

**Expected behavior:**
Links are obvious in descriptions.

**Screenshots:**
The link "Thread in Analytics channel" is not highlighted:
<img width="462" alt="Links not highlighted" src="https://user-images.githubusercontent.com/46905241/129221542-6945a635-f7a2-47fa-8fc2-1eaed7eb8277.png">

But it is on hover:
<img width="482" alt="Links not highlighted - hover" src="https://user-images.githubusercontent.com/46905241/129221563-17b6d763-f61a-4798-b9b0-196c5925730b.png">

**Platform:**
 - v0.9.0 RC3 plugin on Community
 - Chrome 91 on macOS

chenilim:
@sbishel, assigning to you to reassign. Looks like a regression, and may be related to general issues when running as a plugin with the Mattermost webapp CSS?

sbishel:
@hahmadia - May be related to https://github.com/mattermost/focalboard/issues/936

Issue #936, opened 2021-08-11
Bug: Card text editing does not display correctly


chenilim:
**Steps to reproduce the behavior:**
1. Open a card
2. Click on "Add a description" at the bottom
3. Displays extra boxes above and below cursor, text does not appear correctly

**Expected behavior:**
WYSIWYG markdown editing, as before.

**Screenshots:**
<img width="804" alt="Text content bug" src="https://user-images.githubusercontent.com/46905241/129096198-195c0c3f-1125-49d7-82cf-439dff59e7ee.png">

https://user-images.githubusercontent.com/46905241/129096170-d5df3530-27f2-4cba-8e1b-78f3d62ee8da.mov

**Platform:**
 - Focalboard v0.9.0 RC3 on Mattermost 5.38 RC1 (Community server)

sbishel:
Only duplicates when run as a plugin.

PR #971, opened 2021-08-13
Fix GH-955 and GH-936

hahmadia:
Fixes GH-955 

Fix CSS variable name to show proper color

Fixes GH-936

Needed to import SIMPLE MDE css file as per their instructions https://github.com/RIP21/react-simplemde-editor/blob/master/README.md#usage

Also a mismatch on --link-color-rgb . Standalone received a hex value, plugin received a RGB value. 
I had to make it so the plugin also passed down a hex value so the color got correctly set.


jespino (approved):

mgdelacroix (approved):

jespino (approved):

wiggin77 (approved):
LGTM :+1:

Issue #912, opened 2021-08-10
Bug: Table column calculations > Every "Created time" property field is counted as unique


ogi-m:
**Summary:**
Every field is the "Created time" field is count as a unique property. 
See the example in the screenshot - there's 6 rows but only 2 unique values:
-  _July 29, 12:06 AM_ 
- _August 09, 9:28PM_

I would expect the count to show **2**, not **6**.

**Steps to reproduce the behavior:**
1. Go to table view
2. Add a "Created time" property (or "Last Updated Time")
3. Create several cards in the table
4. In the "Created time" column, click _Calculate_ > _Count Unique Values_

**Expected behavior:**
Unique Count should show the number of unique values, not the total number of entries.

**Screenshots:**

![created_time](https://user-images.githubusercontent.com/15314183/128863496-1c693a9a-a4c3-4ae7-8a4c-e8a6d0c3c390.jpg)

Johennes:
I think the problem is that date fields are stored in milliseconds but the UI only shows the time part down to the minute. The values of "Created time" and "Last updated time" properties are populated automatically so even though you see "12:06 AM" three times, the second / millisecond components will be different under the hood.

While the count is technically correct, the UI makes it seem misleading so I suppose we should transform these two properties to drop the non-displayed time components before applying the aggregation.

NB: Notion appears to do the counting as you'd expect based on the UI in this case:

![Screenshot 2021-08-12 at 22 17 12](https://user-images.githubusercontent.com/1137962/129263677-cb9f994f-b3e7-4846-929c-83e7f316e8a4.png)

PR #981, opened 2021-08-14
[GH-912] Discard (milliseconds) when counting create/update dates

Johennes:
#### Summary

Card creation and update dates are displayed with hours and minutes only on the UI. When the full timestamp is used for counting unique values, this leads to a discrepancy because the (milli)second parts impact the count but are not visible to the user.

This commit fixes creation and update timestamps to minutes accuracy before their values enter any calculation.

The ticket wasn't marked as _help wanted_ but the fix seemed easy enough so that I just made it. Happy to improve on it of course if somebody has thoughts. :slightly_smiling_face: 

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/912


harshilsharma63:
Thanks, @Johennes for the change. looks awesome!

Only one small change request - can you add a test in the webapp specifically for this scenario?

harshilsharma63 (approved):

Johennes:
> Only one small change request - can you add a test in the webapp specifically for this scenario?

Good call. :+1:

I added dedicated tests for two cards created / updated 1 second and 1 minute apart.

Issue #967, opened 2021-08-12
Bug: URL properties overflow table cells


chenilim:
**Steps to reproduce the behavior:**
1. Open a table with a URL property cell
2. The contents of the URL overflow the cell

**Expected behavior:**
Content should not overflow cells

**Screenshots:**
<img width="387" alt="URL cell overflow" src="https://user-images.githubusercontent.com/46905241/129271732-710868dc-7ea2-4097-a06f-349bf3bf417f.png">

**Platform:**
 - v0.9.0 RC4 Mac App

chenilim:
FYI @asaadmahmood, we can leave this as a help wanted ticket, or something to check as part of the table styles cleanup. cc @sbishel.

jufab:
@chenilim Can I try to do this one? (not cleanup but to fix this bug...and for my first PR on this project 😊 )

PR #982, opened 2021-08-14
[GH-967] fix overflow URL property in cell

jufab:
#### Summary

This PR add some css properties to fix overflow in URL property cell
And add a unit test for this component

#### Ticket Link

https://github.com/mattermost/focalboard/issues/967


webapp/src/components/properties/link/link.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+                onChange={jest.fn()}
+                onSave={jest.fn()}
+                onCancel={jest.fn()}
+                validator={jest.fn(() => {
+                    return true
+                })}
+            />
+        )
+        const {container} = render(component)
+        expect(container).toMatchSnapshot()
```
kamre, line 24: How is this test related to the changes in `link.scss`?

jufab: You right.
I used an existing css class at the beginning...
I'm going to fix that 😊

jufab: That's not possible. 😢 
Jest configured with a mock on scss

```
"moduleNameMapper": {
    "\\.(scss|css)$": "<rootDir>/__mocks__/styleMock.js"
},
```
And styleMock.js is empty export

```
module.exports = {};
```

And I don't know how to do....Any idea?

kamre: I don't think there is an easy way to check css style inside a unit test.

jufab: Maybe an other issue  to resolve this problem of css in unit tests and do not block the PR for #967?

kamre: This is definitely not blocking. And even such unit test has value. Just need approval from core contributors (look at the list of reviewers).

hahmadia:
Hi @jufab 

Thank you for your contribution. I'm going to add my colleague @asaadmahmood to review your SCSS changes since he's quite the expert to make sure these are in fact the changes we want.

wiggin77 (approved):
LGTM :+1:   The unit test is not really related to the CSS change but looks fine.

asaadmahmood (changes requested):
The hover behaviour looks weird, as it actually shifts the whole rows down. The icon should instead be positioned on the top right of the cell.

Screenshot 2021-08-18 at 5 06 47 PM

kamre:
@asaadmahmood this behaviour with the link icon appearing below the input is reproducible on `main` currently.

Do you propose to have something like this:
LinkHover
?

asaadmahmood:
@kamre Yup, something like that makes sense.

webapp/src/components/properties/link/link.scss
```diff
@@ -1,6 +1,8 @@
 .focalboard-body .property-link {
     &.url {
         padding-right: 2rem;
+        width: inherit;
```
kamre, line 4: By the way why `width: inherit;` is needed here?

asaadmahmood: Shouldn't be.

kamre:
@jufab @asaadmahmood propose to try this `link.scss` variant:
```scss
.focalboard-body .property-link {
    &.url {
        display: flex;
        overflow: hidden;
    }

    a {
        padding: 0 8px; // increases clickable area for better UX
        align-self: stretch;
        display: none;
    }

    a:hover {
        background:  unset;
    }

    &:hover {
        a {
            display: flex;
            align-items: center;
        }
    }
}
```
- `flex` for `.url` allows to keep `input` and `a` on the same line and also `input` can shrink if needed.
- Vertical padding for `a` is set to `0` so that it doesn't try to increase the table row height when `a` appears on hover.

jufab:
@kamre I prefer yours but without padding-right on `url`

with padding-right on hover : 
link-hover
without padding-right on hover :
link-hover-without-padding

I'm going to push the last one.

asaadmahmood:
@jufab If I expand the width of the column, the url thing doesn't occupy the whole space, also the link icon should appear when you hover over the cell, currently it only appears when I hover over the input.

Screenshot 2021-08-19 at 5 59 01 PM

asaadmahmood (changes requested):
See my comment above.

jufab:
> @jufab If I expand the width of the column, the url thing doesn't occupy the whole space, also the link icon should appear when you hover over the cell, currently it only appears when I hover over the input.
> 
> Screenshot 2021-08-19 at 5 59 01 PM

@asaadmahmood I fix the url to occupy the whole space with a "width: inherit". 

link-hover

But for hover, it's a "cell" problem with a 8px padding

asaadmahmood:
@jufab That's fine, I can have a look at it later.

asaadmahmood (approved):

hahmadia (approved):

hahmadia:
Hi! Could please fix the broken snapshot tests so that we can merge this in? Thanks

jufab:
Hi @hahmadia,
it's done!

PR #987, opened 2021-08-16
Add paste/drop images on cards

### LINKED ISSUE #727: Feature Idea: Paste Photos Directly into a Card
**Summary:**
Allow users the ability to paste in photos that they may have in their clipboard.

For the purposes of this example, I'll use a bug board. When users mention a bug that may occur on the product that I'm working on, I will usually take a screenshot of the bug or any relevant information pertaining to that bug. With the screenshot manager that I use, I can copy the screenshot directly to my clipboard which allows saving an unnecessary file to disk and then upload from there.

Importance: Medium

Use cases:
1. Take a screenshot
2. Screenshot is copied to clipboard
3. Paste image into a card and have it added as a new content item

Asna for example allows pasting a file directly into the message box of a card and when submitted, the pasted file is saved in the comment thread of the card. This is helpful as it saves a few steps saving the screenshot to disk and uploading, along with the unnecessary saving of a file. Another example may be a user copying an image from an online source rather than saving it to disk first and then uploading.

jespino:
Fixes #727


wiggin77 (approved):
LGTM :+1:

Issue #966, opened 2021-08-12
Bug: Should not show "Empty" label for empty table cells


chenilim:
**Steps to reproduce the behavior:**
1. Create a table with cards that have URL, text, and date properties (among others)
2. These property types (and likely others) show an "Empty" placeholder
3. Also, empty dates have an "Empty" label with a darker color and larger font

**Expected behavior:**
Remove all "Empty" labels and placeholders in tables - show empty cells instead.

**Screenshots:**
<img width="387" alt="empty label" src="https://user-images.githubusercontent.com/46905241/129270976-7e4a8395-e860-491e-b391-b78d6bb09da5.png">

**Platform:**
 - v0.9.0 RC4 Mac App

chenilim:
@asaadmahmood, I think we should simply remove the "Empty" placeholder on tables. I don't think this is a high priority, but might be part of cleaning up the table styles. FYI @sbishel.

asaadmahmood:
That makes sense to me.

PR #1008, opened 2021-08-16
GH-966 - Removing empty placeholders

asaadmahmood:
#### Summary
GH-966 - Removing empty placeholders

#### Ticket Link
https://github.com/mattermost/focalboard/issues/966

### DIFF
--- a/webapp/src/components/properties/dateRange/__snapshots__/dateRange.test.tsx.snap
+++ b/webapp/src/components/properties/dateRange/__snapshots__/dateRange.test.tsx.snap
@@ -43,9 +43,7 @@ exports[`components/properties/dateRange returns default correctly 1`] = `
       class="Button    "
       type="button"
     >
-      <span>
-        Empty
-      </span>
+      <span />
     </button>
   </div>
 </div>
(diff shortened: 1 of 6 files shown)


wiggin77 (approved):
LGTM :+1:    A lot of unrelated code formatting changes... is that intentional?

webapp/src/components/properties/dateRange/__snapshots__/dateRange.test.tsx.snap
```diff
@@ -43,9 +43,7 @@ exports[`components/properties/dateRange returns default correctly 1`] = `
       class="Button    "
       type="button"
     >
-      <span>
-        Empty
-      </span>
+      <span />
```
wiggin77, line 46: Is `span` (non-container tag) allowed to open/close in one tag?  Will this snap match the actual HTML?

wiggin77: ... also it occurred to me that adding something to the span that helps screen readers would be useful.  It would be hidden but read aloud.

asaadmahmood:
@wiggin77 Make sense to me, added it as a `title`, also yes, the css changes are intentional, we should ideally have line breaks in between css declarations.

wiggin77 (approved):
LGTM :+1:

hahmadia (approved):

Issue #994, opened 2021-08-16
Bug: App switcher text color doesn't match the theme


ogi-m:
**Summary:**
The text for Channels, Playbooks and Boards in the channel switcher shows up as purple regardless of the set theme.

**Steps to reproduce the behavior:**
1. Set a dark theme
2. Open Boards
3. Open the app switcher in the top left and observe the text color

**Expected behavior:**
The text should have the same color as in Channels

**Screenshots:**
App switcher text color in Boards:

![appswitcher_boards](https://user-images.githubusercontent.com/15314183/129593575-34a35404-1144-4e8a-8fca-6d7c3ea53f96.jpg)

App switcher text color in Channels:

![appswitcher_channels](https://user-images.githubusercontent.com/15314183/129593571-d072b6a2-f322-438b-81e0-7114215fe625.jpg)

**Platform:**
 - Focalbooard v.0.9.0-RC5 plugin on master (community-daily)

chenilim:
Could you help investigate @asaadmahmood? Not sure if this is a bug on our side, or something the Global Nav team needs to fix. cc @sbishel.

asaadmahmood:
@chenilim Global, this is an issue that I see on the channel side as well on `community-daily`.
![Screenshot 2021-08-16 at 10 10 24 PM](https://user-images.githubusercontent.com/11034289/129602611-a36b9fe9-f1d1-433e-b2e1-92bd981adeb1.png)

jespino:
Closed by #1012

ogi-m:
Verified the text color is now fixed but I've [opened anew issue](https://github.com/mattermost/focalboard/issues/1095) for icon color.

ogi-m:
Tested and passed on v0.10

PR #1012, opened 2021-08-16
GH-994 - Updating color for global header menu

asaadmahmood:
#### Summary
GH-994 - Updating color for global header menu

#### Ticket Link
https://github.com/mattermost/focalboard/issues/994

#### Screenshot
Screenshot 2021-08-16 at 11 38 14 PM

### DIFF
--- a/mattermost-plugin/webapp/src/plugin.scss
+++ b/mattermost-plugin/webapp/src/plugin.scss
@@ -7,9 +7,7 @@
 
 .focalboard-body .feature-global-header>header {
     z-index: 1000;
-}
 
-.feature-global-header {
     .Menu {
         a,
         button {
--- a/mattermost-plugin/webapp/src/plugin.scss
+++ b/mattermost-plugin/webapp/src/plugin.scss
@@ -5,6 +5,24 @@
     align-items: center;
 }
 
-.focalboard-body .feature-global-header > header {
+.focalboard-body .feature-global-header>header {
     z-index: 1000;
+}
+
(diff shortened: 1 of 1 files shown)


wiggin77 (approved):
LGTM :+1:    @asaadmahmood , I think your editor has different formatting settings from the other devs. We should all get in sync other we'll keep clobbering each other's formatting.

hahmadia (approved):

mattermost-plugin/webapp/src/plugin.scss
```diff
@@ -5,6 +5,24 @@
...
 }
 
-.focalboard-body .feature-global-header > header {
+.focalboard-body .feature-global-header>header {
     z-index: 1000;
-}
\ No newline at end of file
+}
+
+.feature-global-header {
```
jespino, line 12: This is going to affect the `.feature-global-header` of outside focalboard too. if you combine this with the previous one you can make it better. Something like:

```css
.focalboard-body .feature-global-header {
    >header {
        z-index: 1000;
    }
    .Menu {
...
```

jespino (changes requested):
Small change needed.

mattermost-plugin/webapp/src/plugin.scss
asaadmahmood: @jespino Done.

jespino (approved):

jespino:
This can be tested after merge.

PR #1018, opened 2021-08-17
[GH-314] Persist user settings in native Linux app

Johennes:
#### Summary

This PR adds persistence for the user settings in the native Linux app. The settings are written to `$XDG_CONFIG_HOME/focalboard/settings`. If `XDG_CONFIG_HOME` is unset, the app falls back to `$HOME/.config`.

The change hooks into the existing settings export already used in the native macOS app which means the settings are persisted immediately on change.

Unfortunately, the golang webview package uses a custom native binding technology and doesn't allow to define WebKit message handlers. As a result, a dedicated message handler function was added to the existing `NativeApp` object. In the native macOS app, said handler is short-circuited to `window.webkit.messagehandlers.[NAME].postMessage`. This has the benefit that the web app remains agnostic of the particular native binding mechanism.

I have virtually zero experience with Go so while this appears to be working in my testing, the code very likely is an abomination. I'm looking forward to candid feedback. 🙂 

I'd also appreciate some more testing. When running the Linux app on a mac, I _once_ got a blank screen on launch but didn't manage to reproduce again. When testing on Linux, I didn't observe any issues so far.

#### Ticket Link

Relates to https://github.com/mattermost/focalboard/issues/314.


chenilim:
Cleaning up stale issues. Sorry for the delay! @jespino, could you help review this one for the Linux app? Thanks.

Johennes:
Rebased on latest `main` and did a short re-test. :slightly_smiling_face:

Issue #989, opened 2021-08-16
Bug: Shared Boards URL is Incorrect for Focalboard as a Plugin


hahmadia:
**Summary:**
When you share a board in focalboard when focalboard is a plugin of Mattermost, the shared link is incorrect resulting in incorrect behaviour. 

**Steps to reproduce the behavior:**
1. Go to a board
2. Click `...` and `Share board`
3. Link is (for example) 
`http://localhost:8065/workspace/...`

**Expected behavior:**
`http://localhost:8065/boards/workspace/...`

There should be a `boards` before the `workspace`

**Screenshots:**

![image](https://user-images.githubusercontent.com/17804942/129592109-e44f1e93-9b9a-4543-bbdd-55d26f5d8563.png)

chenilim:
Good catch. We definitely want to fix this, which should hopefully be straight forward. @sbishel, assigning to you to reassign. Note that there is a "frontendBaseURL" property used in the plugin webapp code. We might want to review / refactor to make sure there's a consistent base URL that works across the plugin and Personal Editions. cc @jespino, @mgdelacroix, @ogi-m.

PR #1023, opened 2021-08-17
GH-989 add boards to shared component link

sbishel:
#### Summary
Updates the path (adds 'boards') for a shared board when it is a plugin.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/989


webapp/src/components/shareBoardComponent.tsx
```diff
@@ -76,7 +76,7 @@ const ShareBoardComponent = React.memo((props: Props): JSX.Element => {
     shareUrl.searchParams.set('r', readToken)
 
     if (match.params.workspaceId) {
-        const newPath = generatePath('/workspace/:workspaceId/shared/:boardId/:viewId', {
+        const newPath = generatePath('boards/workspace/:workspaceId/shared/:boardId/:viewId', {
```
wiggin77, line 79: This path is also in several other places. Shouldn't we create a const somewhere so when it changes again this won't break?

sbishel: We have part of that in the form of `Utils.getFrontendBaseURL`. I updated this to use that setting instead, which is similar to how it is done [here.](https://github.com/mattermost/focalboard/blob/main/webapp/src/app.tsx#L61)

I don't see this string being used anywhere else. It is similar to those in the above mentioned function, but different due to setting vs. retrieving. I do agree we should probably create constants for all of these in `app.tsx`, so the different variations can get set in the same place but that is beyond the scope of this bug fix.  I will create a card for that.

wiggin77 (approved):
LGTM 👍

harshilsharma63 (approved):

Issue #995, opened 2021-08-16
Bug: Cannot login via mobile


wiggin77:
**Summary:**
Using a channel header in Chat mobile app, clicking a link to Focalboard workspace launches a browser with a login screen. Chat credentials are not accepted. Creating a new account displays a "invalid registration link" error.

**Steps to reproduce the behavior:**
1. Go to chat mobile app, open a channel with a link to a focalboard workspace in the channel header
2. Click on link to focalboard workspace
3. See login screen
4. Enter valid credentials for chat app
5. See login error

**Expected behavior:**
Clicking the channel header link should bring you to the workspace

**Platform:**
 - Android 11 

** Additional **
Also tried with link in chat post to a board. Same behaviour.

chenilim:
The basic login flow should work on mobile.

hahmadia:
Hi @wiggin77 

Are you sure the focalboard link you clicked wasn't the outdated one?

Make sure that the link is `boards/focalboard/...` and not just `/focalboard/...` because if it is the latter, then you will run into the issue you described (as I did myself when I was testing sharing)

Could you try adding `/boards/` right behind the `/focalboard/` and report back if the problem is resolved for you?

Thanks

/cc @chenilim

wiggin77:
I will.  But I tried taking the URL from a board in desktop browser and posting it to myself with same results.

wiggin77:
Here is the link I pulled from the browser's URL bar while using FB.  https://community.mattermost.com/plugins/focalboard/workspace/qgsck6cts3fwpqwyjiupjm5cde/47aa9bb4-6967-4a96-83c7-11bd6b20f1eb/150d1464-6d59-4c39-810f-ee201439dbaf

Clicking that in mobile brings me to the login screen as described.

ogi-m:
This also happens on the web app when opening a link do a different server - for example when opening a URL that has community-release in it but you're logged into community-daily. It should open the Mattermost login instead.

hahmadia:
Ok yep I can reproduce it @wiggin77 @ogi-m ...

hahmadia:
This bug is fixed by Miguels web socket enhancements. The specific PR that fixes this issue is 
https://github.com/mattermost/mattermost-webapp/pull/8608

PR #1032, opened 2021-08-18
Fix GH-995: Fix Redirection Problem

hahmadia:
#### Summary

Fixes redirection problem after logging onto Mattermost to Focalboard

#### Ticket Link

Fixes GH-995

Requires
https://github.com/mattermost/mattermost-webapp/pull/8632

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -89,7 +89,7 @@ const App = React.memo((): JSX.Element => {
                                     path='/workspace/:workspaceId/:boardId?/:viewId?'
                                     render={({match, location}) => {
                                         if (loggedIn === false) {
-                                            let redirectUrl = '/' + Utils.buildURL(`/workspace/${match.params.workspaceId}${match.params.boardId ? `/${match.params.boardId}` : ''}${match.params.boardId ? `/${match.params.viewId}` : ''}${location.search}`)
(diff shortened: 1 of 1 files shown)


webapp/src/app.tsx
```diff
@@ -87,13 +87,19 @@ const App = React.memo((): JSX.Element => {
...
                                     path='/workspace/:workspaceId/:boardId?/:viewId?'
-                                    render={({match}) => {
+                                    render={({match, location}) => {
                                         if (loggedIn === false) {
-                                            let redirectUrl = '/' + Utils.buildURL(`/workspace/${match.params.workspaceId}/`)
+                                            let redirectUrl = '/' + Utils.buildURL(`/workspace/${match.params.workspaceId}${match.params.boardId ? `/${match.params.boardId}` : ''}${match.params.viewId ? `/${match.params.viewId}` : ''}${location.search}`)
                                             if (redirectUrl.indexOf('//') === 0) {
                                                 redirectUrl = redirectUrl.slice(1)
                                             }
-                                            const loginUrl = `/login?r=${encodeURIComponent(redirectUrl)}`
```
chenilim, line 96: This URL should be used when not running as a plugin, so Personal Server works.

chenilim (changes requested):
Need to change the URL for non-plugin as we discussed. Otherwise this LGTM.

webapp/src/app.tsx
jespino: Actually, this url is the same for both, it is the same url for mattermost-webapp, so I think it is fine.

jespino (approved):
LGTM.

hahmadia:
Not needed anymore due to Miguels changes on Mattermost web app code

PR #1044, opened 2021-08-19
Fix border cut on kanban and gallery views

### LINKED ISSUE #1040: Bug: The border of some kanban cards is cut
**Summary:**
As you can see in the first screenshot, the left border of all cards in the first column is cut.
A solution could be to remove [`this padding`](https://github.com/mattermost/focalboard/blob/7b5a45dc6cf53fb044b99cb9b4d3ae09a4dcc958/webapp/src/components/centerPanel.scss#L50) and add a `padding-left` set to `1px`

1. Create the first column
2. Add some cards

**Expected behavior:**
The left border is visible 

**Screenshots:**
immagine

**Platform:**
 - Browser and version: Firefox 91
 - OS and version: Windows 10

**Additional context:**
Self-hosted using the docker container
Focalboard v0.9.0

jespino:
Fixes #1040


asaadmahmood (approved):
Accepting this for now, but we ideally want to refrain from solutions like this because if the dom content changes, or if someone introduces an element in the html in the future, the css applied will no longer be applicable.

chenilim:
Not sure why or when this happened, but it's a regression from v0.8, so we should take it for v0.9. BTW, it repros on Chrome on Community as well.

PR #1049, opened 2021-08-19
Remove duplicated restart directive

chmac:
SSIA

### DIFF
--- a/docker/docker-compose-db-nginx.yml
+++ b/docker/docker-compose-db-nginx.yml
@@ -36,7 +36,6 @@ services:
     image: postgres:latest
     restart: always
     container_name: focalboard-postgres
-    restart: always
     volumes:
       - "./data:/var/lib/postgresql/data"
     environment:


chmac:
FYI, I won't sign the CLA and all that admin stuff for this change. I also don't need to be added to any kind of contributor agreement. Feel free to make the change and commit it and then close this PR if that's the easiest strategy, but I won't do any more work than submit the PR, what y'all do with it now I leave up to y'all...

hahmadia:
@jespino

jespino:
@chmac, thanks for the contribution and the honest and direct comment about the CLA, we can't merge the PR without the CLA signature for legal reasons, but of course is up to you to sign it or not, and we are really appreciate all the issues that you are reporting lately and this PR too. I hope you keep reporting bugs and feature requests.

In any case, because you are not signing the CLA we are going to address the problem in another PR.

PR #1050, opened 2021-08-19
Use card id in the url

jespino:


chenilim (approved):
Yup, this path is cleaner. If we make this change, we should take it for v0.9, as it breaks existing external URLs (currently only affecting very few customers).

sbishel (approved):
LGTM - Actually looks like it has backwards compatibility for existing URLs. (or did I miss something)

hahmadia (approved):

PR #1052, opened 2021-08-20
Workspace switcher

harshilsharma63:
#### Summary
Workspace switcher. Only for Focalboard running as a plugin. Intentionally disabled on a personal server.

#### Ticket Link
https://community-daily.mattermost.com/plugins/focalboard/workspace/qgsck6cts3fwpqwyjiupjm5cde/47aa9bb4-6967-4a96-83c7-11bd6b20f1eb/392c8889-6c8f-4cb7-8407-f5c57c1acfe8?c=36f12ec9-0705-4a85-84d1-ea1e7040f2f1

#### Functionality
1. No change in functionality for personal server mode.
2. New workspace switcher is now visible when running Focalboard as a Mattermost plugin.
3. The workspace switcher shows the Focalboard workspaces (named by their corresponding Mattermost channels" and the number of boards in each workspace, for the workspaces the user has access to.
4. If a workspace has no boards, it's not shown in the list.
5. Selecting a workspace from the dropdown takes the user to that workspace.

**Testing note -** please do test on both PostgreSQL and MySQL, and if possible, on MySQL 5.8 as well (current min supported MySQL version).


server/services/store/sqlstore/migrations/migrations_files/000001_init.up.sql
```diff
@@ -12,4 +12,4 @@ CREATE TABLE IF NOT EXISTS {{.prefix}}blocks (
 	update_at BIGINT,
 	delete_at BIGINT,
 	PRIMARY KEY (id, insert_at)
-){{if .mysql}}CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci{{end}};
```
chenilim, line 15: Did we come to a conclusion on what (if any) encoding to set? Per @jespino's last comment in [this thread](https://community.mattermost.com/core/pl/tyr59rq8wb8sifxqgka8gybeqr): "Yes, the problem is, if you not "hardcoded" it, it can use utf8 and everything looks broken, we decided to choose one collation there, the user can still alter the table later. I'm not saying that decision is correct, but not setting it can lead to another kind of problems"

server/services/store/sqlstore/workspaces.go
```diff
@@ -144,3 +145,62 @@ func (s *SQLStore) GetWorkspaceCount() (int64, error) {
...
+	if err != nil {
+		s.logger.Error("ERROR GetUserWorkspaces", mlog.Err(err))
+		return nil, err
+	}
+
+	defer s.CloseRows(rows)
+	return s.userWorkspacesFromRoes(rows)
+}
+
+func (s *SQLStore) userWorkspacesFromRoes(rows *sql.Rows) ([]model.UserWorkspace, error) {
```
sbishel, line 185: nit: should be  userWorkspacesFromRows

server/services/store/sqlstore/workspaces.go
```diff
@@ -144,3 +145,62 @@ func (s *SQLStore) GetWorkspaceCount() (int64, error) {
...
+		var userWorkspace model.UserWorkspace
+
+		err := rows.Scan(
+			&userWorkspace.ID,
+			&userWorkspace.Title,
+			&userWorkspace.BoardCount,
+		)
+
+		if err != nil {
+			s.logger.Error("ERROR userWorkspacesFromRoes", mlog.Err(err))
```
sbishel, line 198: same nit.

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
```diff
@@ -0,0 +1,107 @@
...
+import {useIntl} from 'react-intl'
+
+import {getUserWorkspaceList} from '../../store/users'
+import {UserWorkspace} from '../../user'
+import {useAppSelector} from '../../store/hooks'
+
+import './workspaceOptions.scss'
+import Search from '../../widgets/icons/search'
+
+const style = {
```
sbishel, line 16: Why style here vs css?

webapp/src/pages/boardPage.tsx
```diff
@@ -206,6 +206,8 @@ const BoardPage = (props: Props) => {
         }
     })
 
+    // }
```
sbishel, line 209: nit - remove.

sbishel (approved):
Just minor nits.

server/services/store/sqlstore/migrations/migrations_files/000001_init.up.sql
harshilsharma63: @chenilim both charset and collation can get set in DB connection strings in Focalboard config files. So one doesn't need to hardcode them in here. 

One thing we need to do is document it in Focalboard docs to use `utf8mb4_general_ci` collation or any other `utf8mb4` variant. Can you point me to docs where I can document this?

server/services/store/sqlstore/workspaces.go
harshilsharma63: Done.

server/services/store/sqlstore/workspaces.go
harshilsharma63: Done.

webapp/src/pages/boardPage.tsx
harshilsharma63: Done.

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
harshilsharma63: Missed this. They are no longer needed. Removed now.

server/services/store/sqlstore/workspaces.go
```diff
@@ -144,3 +145,62 @@ func (s *SQLStore) GetWorkspaceCount() (int64, error) {
...
+
+	if s.dbType == mysqlDBType {
+		query = s.getQueryBuilder().
+			Select("Channels.ID", "Channels.DisplayName", "COUNT(focalboard_blocks.id)").
+			From("focalboard_blocks").
+			Join("ChannelMembers ON focalboard_blocks.workspace_id = ChannelMembers.ChannelId").
+			Join("Channels ON ChannelMembers.ChannelId = Channels.Id").
+			Where("ChannelMembers.UserId = ?", userID).
+			Where("focalboard_blocks.type = 'board'").
+			Where("focalboard_blocks.fields LIKE '%\"isTemplate\":false%'").
```
jespino, line 160: This is dangerous, somebody can introduce in the board description the text `"isTemplate":false` and this is going to swallow that. Personally I prefer to filter it after the query is returned. At least for MySQL.

server/services/store/sqlstore/workspaces.go
```diff
@@ -144,3 +145,62 @@ func (s *SQLStore) GetWorkspaceCount() (int64, error) {
...
+func (s *SQLStore) GetUserWorkspaces(userID string) ([]model.UserWorkspace, error) {
+	var query sq.SelectBuilder
+
+	if s.dbType == mysqlDBType {
+		query = s.getQueryBuilder().
+			Select("Channels.ID", "Channels.DisplayName", "COUNT(focalboard_blocks.id)").
+			From("focalboard_blocks").
+			Join("ChannelMembers ON focalboard_blocks.workspace_id = ChannelMembers.ChannelId").
+			Join("Channels ON ChannelMembers.ChannelId = Channels.Id").
+			Where("ChannelMembers.UserId = ?", userID).
```
jespino, line 158: I prefer to use here the `sq.Eq{"ChannelMembers.UserId": userID}` approach, is more explicit.

jespino: same in the next line, and in the postgres queries.

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
```diff
@@ -0,0 +1,94 @@
...
+        return {
+            label: workspace.title,
+            value: workspace.id,
+            boardCount: workspace.boardCount,
+        }
+    })
+
+    return (
+        <Select
+            classNamePrefix={'WorkspaceOptions'}
```
jespino, line 35: I would expect also the `className` field with `WorkspaceOptions` here.

webapp/src/components/workspaceSwitcher/workspaceOptions.scss
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    .boardCount {
+        color: rgba(var(--center-channel-color-rgb), 0.9);
+        font-size: 10px;
+        text-transform: uppercase;
+    }
+}
+
+.WorkspaceOptions__indicators {
```
jespino, line 31: I would expect all this to be something like:
```css
.WorkspaceOptions {
    &__indicators{
        ...
    }
    &__control {
        ...
    }
    ...
}
```

webapp/src/components/workspaceSwitcher/workspaceOptions.scss
```diff
@@ -0,0 +1,67 @@
+.workspaceOption {
```
jespino, line 1: If workspaceOption is a component, shouldn't it be `.WorkspaceOption`? and, because `.WorkspaceOption` is alwais inside a `WorkspaceOptions` shouldn't it be inside `WorkspaceOptions` class?

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const Option = (props: any): JSX.Element => {
+    const {innerProps, innerRef} = props
+
+    return (
+        <div
+            ref={innerRef}
+            {...innerProps}
+            className={`workspaceOption ${props.isFocused ? 'focused' : ''}`}
```
jespino, line 82: Because this is a component, I would expect this to be uppercased class name, like `WorkspaceOption`.

webapp/src/components/workspaceSwitcher/workspaceOptions.scss
jespino: To avoid collitions on css when this is running as a plugin we add always the `.focalboard-body` class as prefix, otherwise there are some risk of css classes collition.

jespino: this `.focalboard-body` applies to most of the classes defined in the focalboard css, so it also appies to the other CSS files submitted here.

webapp/src/components/workspaceSwitcher/workspaceSwitcher.tsx
```diff
@@ -0,0 +1,59 @@
...
+            </div>
+            {
+                showMenu &&
+                <WorkspaceOptions
+                    activeWorkspaceId={props.activeWorkspace.id}
+                    onClose={() => {
+                        setSelectClosed(true)
+                    }}
+                    onChange={(workspaceId: string) => {
+                        const newPath = generatePath(match.path, {...match.params, workspaceId})
```
jespino, line 48: This is not totally true. This was this way in other cases because we wanted to preserve the current path, because it can change a lot and there is not an easy way to define what is the right path, in this case, because in shared mode shouldn't be shown, and in personal server shouldn't be shown, you can assume that this url is going to be always `/workspace/:workspaceId`. In any case, if you want to keep the `generatePath` approach, you should remove the boardId and viewId from the current `match.params`, if not you are generating a route that is not correct (boardId and viewId from another workspace).

webapp/src/octoClient.ts
```diff
@@ -381,6 +381,20 @@ class OctoClient {
         }
         return (await this.getJson(response, [])) as IUser[]
     }
+
+    async getUserWorkspaces(): Promise<UserWorkspace[]> {
+        if (!Utils.isFocalboardPlugin()) {
```
jespino, line 386: I'm not 100% sure if this is correct, the client library in theory should be usable for third party applications and looks like passing here the `isFocalboardPlugin` is an implementation detail that shouldn't belong to the library.

webapp/src/store/users.ts
```diff
@@ -19,11 +19,12 @@ type UsersStatus = {
     me: IUser|null
     workspaceUsers: {[key: string]: IUser}
     loggedIn: boolean|null
+    userWorkspaces: UserWorkspace[]
```
jespino, line 22: I would say that this belong to `workspace.ts` not to `users.ts`. `users.ts` handles the users models, in this case, return myself, and the list of users. Whenever you need workspace instances, I expect them to be in `workspace.ts`.

webapp/src/widgets/icons/chevronDown.tsx
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react'
+
+import CompassIcon from './compassIcon'
+
+export default function ChevronDown(): JSX.Element {
```
jespino, line 8: nit: I personally dislike this kind of icon names, I would prefer something more meaningful. Like the `Search` icon that you also included in this PR.

jespino (changes requested):
In general the approach looks good, I added some comments of things that I think should be change or at least discussed.

server/services/store/sqlstore/migrations/migrations_files/000001_init.up.sql
jespino: I'm not 100% comfortable with this change, but I would need to think more about it.

harshilsharma63: @jespino whats your concern?

webapp/src/store/users.ts
harshilsharma63: Done.

webapp/src/octoClient.ts
harshilsharma63: Done.

webapp/src/components/workspaceSwitcher/workspaceSwitcher.tsx
harshilsharma63: Done.

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
harshilsharma63: Done.

webapp/src/components/workspaceSwitcher/workspaceOptions.scss
harshilsharma63: Oh, nice! Didn't knew we could do this.

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
harshilsharma63: Done.

server/services/store/sqlstore/workspaces.go
harshilsharma63: Done.

server/services/store/sqlstore/workspaces.go
harshilsharma63: That's not what happens. If someone types `"isTemplate":false` in UI, it gets saved as `\"isTemplate\":false` (escaped double quotes)

Regular board-
Screenshot 2021-08-23 at 6 33 39 PM

Board with description set as `"isTemplate":false`-
Screenshot 2021-08-23 at 6 34 28 PM

webapp/src/components/workspaceSwitcher/workspaceOptions.scss
harshilsharma63: Have renamed this to just `Options`.

webapp/src/widgets/icons/chevronDown.tsx
harshilsharma63: It can be used for multiple purposes such as for dropdown indicators, scroll indicators, etc. The usage doesn't change the fact that it's just a generic, chevron down icon. I prefer to keep this `ChevronDown`.

server/services/store/sqlstore/migrations/migrations_files/000001_init.up.sql
harshilsharma63: Resolving as we're fine with this approach and adding a default collation.

harshilsharma63:
@asaadmahmood can you please make a round of UI/UX review on this?

asaadmahmood (changes requested):
Have pointed out a bunch of issues to Harshil.
1. A bunch of UI changes are needed.
2. My local instance doesn't load the workspaces.

harshilsharma63:
@asaadmahmood points 1 and 3 are fixed now. I'm working with Jesus and Scott to see if it was an issue with your system or a general issue. WIll update here once tested.

asaadmahmood (approved):
UI looks good to me!

server/services/store/sqlstore/migrations/migrations_files/000011_match_collation.up.sql
```diff
@@ -0,0 +1,45 @@
+{{if .mysql}}
+SET @targetCollation = (SELECT table_collation from information_schema.tables WHERE table_name = 'Channels') AND table_schema = (SELECT DATABASE());
+
+-- blocks
+SET @updateCollationQuery = CONCAT('ALTER TABLE focalboard_blocks COLLATE ', @targetCollation);
```
harshilsharma63, line 5: We need all this to use a variable in the query.

server/services/store/sqlstore/migrations/migrations_files/000011_match_collation.up.sql
```diff
@@ -0,0 +1,58 @@
+{{if .mysql}}
```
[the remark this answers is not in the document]
harshilsharma63: Done.

webapp/src/components/dialog.tsx
```diff
@@ -28,7 +30,7 @@ const Dialog = React.memo((props: Props) => {
     useHotkeys('esc', () => props.onClose())
 
     return (
-        <div className='Dialog dialog-back'>
+        <div className={'Dialog dialog-back ' + (props.className || '')}>
```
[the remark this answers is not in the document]
harshilsharma63: Done.

config.json
```diff
@@ -1,8 +1,8 @@
 {
 	"serverRoot": "http://localhost:8000",
 	"port": 8000,
-	"dbtype": "sqlite3",
-	"dbconfig": "./focalboard.db",
+    "dbtype": "mysql",
```
jespino, line 4: Why are you changing the defaults here to mysql?

harshilsharma63: This was left from my testing. Pushing changes with that removed.

server/services/store/sqlstore/migrate.go
```diff
@@ -145,9 +158,15 @@ func (s *SQLStore) Migrate() error {
 	if err != nil {
 		return err
 	}
+
+	fmt.Println("############################################")
```
jespino, line 162: This looks like debug block. I think you can remove this.

server/services/store/sqlstore/migrations/migrations_files/000001_init.up.sql
```diff
@@ -12,4 +12,4 @@ CREATE TABLE IF NOT EXISTS {{.prefix}}blocks (
...
+) {{if .mysql}}DEFAULT CHARACTER SET utf8mb4{{end}};
```
jespino, line 15: Nice. I'm not 100% sure if this solves the problem, but looks like an acceptable solution.

server/services/store/sqlstore/migrations/migrations_files/000011_match_collation.down.sql
```diff
@@ -0,0 +1,2 @@
+-- Nothing to be done here
+-- This page is intentionally left blank
```
jespino, line 2: You added an empty query in the `.up.sql` file because if not, it fails, I guess it should be the same here.

server/services/store/sqlstore/workspaces.go
```diff
@@ -145,3 +146,62 @@ func (s *SQLStore) GetWorkspaceCount() (int64, error) {
...
+			GroupBy("Channels.Id", "Channels.DisplayName")
+	} else {
+		query = s.getQueryBuilder().
+			Select("channels.id", "channels.displayname", "count(focalboard_blocks.id)").
+			From("focalboard_blocks").
+			Join("channelmembers ON focalboard_blocks.workspace_id = channelmembers.channelid").
+			Join("channels ON channelmembers.channelid = channels.id").
+			Where(sq.Eq{"channelmembers.userid": userID}).
+			Where(sq.Eq{"focalboard_blocks.type": "board"}).
+			Where("focalboard_blocks.fields ->> 'isTemplate' = 'false'").
```
jespino, line 171: This is not going to work on SQLite. I guess you have to use the previous block for SQLite.

webapp/src/store/workspace.ts
```diff
@@ -15,9 +17,14 @@ export const fetchWorkspace = createAsyncThunk(
     async () => client.getWorkspace(),
 )
 
+type WorkspaceState = {
+    value: IWorkspace|null
```
jespino, line 21: maybe this should be now `current` instead of `value`.

webapp/src/store/workspace.ts
```diff
@@ -36,6 +44,13 @@ const workspaceSlice = createSlice({
...
 export const {reducer} = workspaceSlice
 
+export const getUserWorkspaces = (state: RootState): UserWorkspace[] => state.workspace.userWorkspaces
+
+export const getUserWorkspaceList = createSelector(
+    getUserWorkspaces,
+    (userWorkspaces) => userWorkspaces,
+)
+
 export function getWorkspace(state: RootState): IWorkspace|null {
```
jespino, line 54: Also, I think this should be `getCurrentWorkspace`.

jespino (changes requested):
Some small changes needed, but otherwise looks good to me.

server/services/store/sqlstore/workspaces.go
```diff
@@ -145,3 +146,62 @@ func (s *SQLStore) GetWorkspaceCount() (int64, error) {
...
+			GroupBy("channels.id", "channels.displayname")
```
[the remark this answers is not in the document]
harshilsharma63: Done.

config.json
harshilsharma63: Done.

server/services/store/sqlstore/migrate.go
harshilsharma63: Done.

server/services/store/sqlstore/migrations/migrations_files/000001_init.up.sql
harshilsharma63: It did work and I tested on MySQL 8.7 and the 15.

server/services/store/sqlstore/migrations/migrations_files/000011_match_collation.down.sql
harshilsharma63: Ah, great catch! Fixed now.

server/services/store/sqlstore/workspaces.go
harshilsharma63: This is only intended to run as a plugin, which doesn't use SQLite.

webapp/src/store/workspace.ts
harshilsharma63: Done.

webapp/src/store/workspace.ts
harshilsharma63: Done.

server/services/store/sqlstore/workspaces.go
```diff
@@ -150,30 +156,26 @@ func (s *SQLStore) GetWorkspaceCount() (int64, error) {
...
+	case mysqlDBType:
+		query = query.Where(sq.Like{"focalboard_blocks.fields": "%\"isTemplate\":false%"})
+	case postgresDBType:
+		query = query.Where("focalboard_blocks.fields ->> 'isTemplate' = 'false'")
+	default:
+		return nil, fmt.Errorf("GetUserWorkspaces - %w", errUnsupportedDatabaseError)
 	}
 
-	rows, err := query.Query()
+	query = query.GroupBy("Channels.Id", "Channels.DisplayName")
```
jespino, line 176: nit: the `GroupBy` can be added to the query in the initial query builder. I don't think is needed to do it later.

jespino (approved):
Now LGTM. An small nit, but I don't think is strictly necessary.

server/services/store/sqlstore/workspaces.go
harshilsharma63: Interesting! I didn't know the clauses were order independent! Nice!

harshilsharma63:
@ogi-m this PR is not dev-approved. Can you please test this out?

ogi-m:
Thanks @harshilsharma63, LGTM 🎉 
- workspace switcher visible when running Focalboard as a Mattermost plugin.
- workspace switcher shows the Focalboard workspaces named by their corresponding Mattermost channels and the number of boards in each workspace 
- selecting a workspace from the dropdown takes the user to that workspace.
- only the workspaces the user has access to are shown
- adding/removing a user from a channel adds/removes the channel's workspace to/from the user's workspace switcher
- If a workspace has no boards, it's not shown in the list
- workspaces are added to the switcher when boards are added to the the workspace, and removed if all boards get deleted
- tested on Postgres and MySQL

Note: didn't test on MySQL 5.8, I don't think we have a test server for it but I'll check.

webapp/src/components/sidebar/sidebar.scss
```diff
@@ -127,4 +127,10 @@
         stroke: rgba(var(--sidebar-text-rgb), 0.5);
         stroke-width: 6px;
     }
+
+    @media screen and (min-width: 768px) {
+        .sidebarSwitcher {
+            display: none;
+        }
+    }
```
kamre, line 135: @harshilsharma63 why is this needed? This is a kind of regression for personal server because now it is not possible to hide the side bar on wide screens.

harshilsharma63: @kamre That's a UX decision the team took - to only allow hiding the sidebar when the window size is small.

DO you think it's something that should be supported?

kamre: Yes, in my opinion there are situations when the space is not enough even on wide screens.

Look for example at these kanban boards from default templates, they all have horizontal scroll:
BoardsScroll
This is on laptop with 150% scale in Windows settings.

harshilsharma63: cc @asaadmahmood

chenilim: FYI @harshilsharma63, sent PR #1219 to you for review to only hide the collapse-sidebar button for workspaces.

Issue #1046, opened 2021-08-19
Bug: In plugin mode, the share url should be in the form of `/plugins/focalboard/shared/whatever`


jespino:
**Summary:**
This is because the `/boards/`url now requires logged in user, so we can solve that problem using directly the plugin api for that.

chenilim:
Thanks @jespino, when did / or will this behavior (requiring a logged in user) change? Both `/boards/` and `/plugins/focalboard/` work for shared boards on community-daily right now, at least for me. In any case, if that behavior is changing, I agree we'll need to make this change on our end.

We should revisit the ability to change that url in the future (i.e. to remove "focalboard" from it), but if this let's shared boards work for now, we should do it.

Assigning to you to reassign @sbishel. I guess it's a re-fix of #989.

PR #1062, opened 2021-08-20
update share url

sbishel:
#### Summary
Updates ShareBoards Url, since `boards` requires authentication. Shared Boards may be viewed by unauthorized users.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/1046


webapp/src/components/shareBoardComponent.tsx
```diff
@@ -76,8 +76,7 @@ const ShareBoardComponent = React.memo((props: Props): JSX.Element => {
     shareUrl.searchParams.set('r', readToken)
 
     if (match.params.workspaceId) {
-        const newPath = generatePath(':basename/workspace/:workspaceId/shared/:boardId/:viewId', {
-            basename: Utils.getFrontendBaseURL(),
+        const newPath = generatePath('plugins/focalboard/workspace/:workspaceId/shared/:boardId/:viewId', {
```
chenilim, line 79: I'm not 100% sure, but do we need a / at the beginning? Also, might be a separate issue, but will this work if the base Mattermost path is not on the root?

chenilim (approved):
Had a comment about the URL, but LGTM if Jesus is ok.

webapp/src/components/shareBoardComponent.tsx
jespino: Yes, I think we need the `/` at the beginning, but I don't know if the `shareurl.pathname = newPath` is already doing it.

jespino (approved):
LGTM

webapp/src/components/shareBoardComponent.tsx
sbishel: Nope...updated.

Issue #826, opened 2021-08-02
WebApp Unit Tests - ./webapp/src/components/blockIconSelector.tsx


sbishel:
Write Unit Tests for `./webapp/src/components/blockIconSelector.tsx`

see `./webapp/src/components/table/*` or`./webapp/src/components/properties/dateRange/dateRange.test.tsx` for sample tests

PR #1063, opened 2021-08-20
[GH-826] Add unit test for blockIconSelector

jufab:
#### Summary

Add unit test for BlockIconSelector.
#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/826


webapp/src/components/blockIconSelector.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+        expect(container).toBeEmptyDOMElement()
+    })
+    test('return menu on click', () => {
+        const {container} = render(wrapIntl(
+            <BlockIconSelector
+                block={board}
+                size='l'
+            />,
+        ))
+        userEvent.click(container.getElementsByTagName('div')[1]) //0 is div "BlockIconSelector"
```
kamre, line 58: This looks a bit suspicious because it relies on the tags and their order.

May be we can start adding [ARIA roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#roles) to the elements and use them for testing purposes also. For example [this `div` inside `MenuWrapper`](https://github.com/mattermost/focalboard/blob/main/webapp/src/widgets/menuWrapper.tsx#L84-L88) can be given a `button` role.

webapp/src/components/blockIconSelector.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+        const buttonRandom = screen.queryByRole('button', {name: 'Random'})
+        expect(buttonRandom).not.toBeNull()
+        userEvent.click(buttonRandom!)
+        expect(board.fields.icon).not.toBeNull()
+        expect(board.fields.icon).not.toEqual(icon)
```
kamre, line 86: Can this new random icon be the same one with some probabilty?

webapp/src/components/blockIconSelector.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+import BlockIconSelector from './blockIconSelector'
+
+const wrapIntl = (children: ReactElement) => <IntlProvider locale='en'>{children}</IntlProvider>
+const board = TestBlockFactory.createBoard()
+const icon = BlockIcons.shared.randomIcon()
+
+jest.mock('../mutator', () => ({
+    changeIcon: jest.fn((id, oldIcon, newIcon) => {
+        board.fields.icon = newIcon
```
kamre, line 23: @jespino @wiggin77 is this a "legit"/"good" way to mock the `mutator` for the unit test?

In the real app there happens a lot behind the scenes via the interaction with the server. But essentially it returns the same data that was sent and performs incremental update of the data model.

Is this approach treated like a contract that we can rely on in the unit tests replicating it via simple updates from mocks?

kamre:
Nice unit test! I like how user actions are performed with the mock of `mutator`.

webapp/src/components/blockIconSelector.test.tsx
jufab: I agree with you for the aria role

When I wrote the test, I started with the idea of leaving the existing code alone and writing the tests around that code.
But if I can add elements in the existing code then yes, I want to avoid that 😃 

Just modify the MenuWrapper by adding this aria-label and role. Can i do it? 😃

webapp/src/components/blockIconSelector.test.tsx
jufab: In theory, yes...
Let me try something else!

webapp/src/components/blockIconSelector.test.tsx
jufab: Just one thing
I mocked the mutator: 
- to make sure I only test this component and not all of the services.
- to keep component actions in testing and get the correct test coverage.

jufab: I reviewed mocked mutator:  I don't add implementation but now, I verify it been called

```
expect(mockedMutator.changeIcon).toBeCalledTimes(1)
expect(mockedMutator.changeIcon).toBeCalledWith(board.id, board.fields.icon, '👍')

File: webapp/src/components/blockIconSelector.test.tsx (hunk 1 above)
Comment (kamre) [reply webapp/src/components/blockIconSelector.test.tsx:59]: I think that you can do such things when writing unit tests. Look for example at some recent pull request with unit test that [added data-test-id](https://github.com/mattermost/focalboard/pull/906/files#diff-2e274a77dd6b2957a2f7e0579fa989e0cbfca3656a62cc5a49c772ee09aabf89R35)

Comment (jufab) [reply webapp/src/components/blockIconSelector.test.tsx]: Ok, I add data-testid on menuwrapper ;)

File: webapp/src/widgets/menuWrapper.tsx [hunk 4]
```diff
@@ -85,6 +85,7 @@ const MenuWrapper = React.memo((props: Props) => {
             className={className}
             onClick={toggle}
             ref={node}
+            data-testid='menuwrapper'
```
harshilsharma63, line 88: Where is this attribute being used and what for?

jufab: This attribute is used in test only and don't have consequence in render
more info : https://testing-library.com/docs/queries/bytestid

sbishel: I would prefer to set the `role` and `aria-label` rather than `data-testid`. 
See [priority] (https://testing-library.com/docs/queries/about#priority)

sbishel (changes requested):

sbishel (approved):
LGTM. Thanks for the changes.

PR #1069, opened 2021-08-23
Plugin telemetry

sbishel:
#### Summary
Adds telemetry to boards client. 

#### Ticket Link
https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/47aa9bb4-6967-4a96-83c7-11bd6b20f1eb/89e87813-96b3-4e1d-825a-d2d919091a0c?c=5a19ebbf-2480-4cdd-8461-a16b24c6d1f3


sbishel:
Working on Unit tests now.

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -32,6 +36,9 @@ import {PluginRegistry} from './types/mattermost-webapp'
 
 import './plugin.scss'
 
+const TELEMETRY_RUDDER_KEY = 'placeholder_rudder_key'
```
chenilim, line 39: Right now, `placeholder_rudder_key` is replaced in the [production build workflow](https://github.com/mattermost/focalboard/blob/main/.github/workflows/prod-release.yml#L14) only in `telemetry.go`. Looks like we either need to update all the build workflows to replace it in this file as well, or expose and call [getRudderConfig](https://github.com/mattermost/focalboard/blob/main/server/services/telemetry/telemetry.go#L53) in telemetry?

chenilim (changes requested):
Added a note about the changes needed for the build process.

Also, this is scope creep, so feel free to defer, but can we also add a distinct event when the user clicks the Channel Header button (in `goToFocalboardWorkspace`)? It would be good to track how often that happens via the Channel Header vs. App Switcher. Thanks.

webapp/src/components/centerPanel.tsx
```diff
@@ -79,6 +81,7 @@ class CenterPanel extends React.Component<Props, State> {
 
     componentDidMount(): void {
         this.showCardInUrl()
+        TelemetryClient.trackEvent('boards', 'view', {viewTypeTest: this.props.activeView.fields.viewType})
```
wiggin77, line 84: Why is this prop called `viewTypeTest`?

wiggin77 (approved):
LGTM :+1:

webapp/src/components/centerPanel.tsx
sbishel: Testing...updated, thanks for catching.

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -143,6 +151,50 @@ export default class Plugin {
...
+                    title: '',
+                    url: '',
+                },
+                {
+                    context: {
+                        ip: '0.0.0.0',
+                    },
+                    anonymousId: '00000000000000000000000000',
+                })
+                TelemetryClient.setTelemetryHandler(new RudderTelemetryHandler())
```
mgdelacroix, line 195: nit: I think we should move the static object to constants, both for clarity and to avoid repetition

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -143,6 +151,50 @@ export default class Plugin {
...
+        if (config?.telemetry) {
+            let rudderKey = TELEMETRY_RUDDER_KEY
+            let rudderUrl = TELEMETRY_RUDDER_DATAPLANE_URL
+
+            if (rudderKey.startsWith('placeholder') && rudderUrl.startsWith('placeholder')) {
+                rudderKey = process.env.RUDDER_KEY as string //eslint-disable-line no-process-env
+                rudderUrl = process.env.RUDDER_DATAPLANE_URL as string //eslint-disable-line no-process-env
+            }
+
+            if (rudderKey != null && rudderKey !== '') {
```
mgdelacroix, line 165: Doesn't seem possible that `rudderKey` is `null`, as it is being set to `''` if there is no env variable in `webpack.config`
```suggestion
            if (rudderKey !== '') {
```

mgdelacroix (approved):
LGTM! :+1:

mattermost-plugin/webapp/src/index.tsx
sbishel: done

mattermost-plugin/webapp/src/index.tsx
sbishel: done

chenilim (approved):
Looks good! Thanks.

Issue #824, opened 2021-08-02
WebApp Unit Tests - ./webapp/src/components/addContentMenuItem.tsx


sbishel:
Write Unit Tests for `./webapp/src/components/addContentMenuItem.tsx`

see `./webapp/src/components/table/*` or`./webapp/src/components/properties/dateRange/dateRange.test.tsx` for sample test

PR #1073, opened 2021-08-23
 [GH-824] Add unit test for addContentMenuItem

jufab:
#### Summary

Add unit test for addContentMenuItem.
#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/824


sbishel (changes requested):
Thanks for your contribution. Rather than write individual tests for each block type. I would suggest using `contentRegistry.contentTypes`. Similar to how it is used in `contentBlock`. And loop through each valid contentType using `test.each()`, doing a `toMatchSnapshot` on each.

wiggin77 (approved):
LGTM :+1:

sbishel (approved):
Nice..thank you.

PR #1075, opened 2021-08-24
Fix Image Paste Messing with Content Block

hahmadia:
#### Summary

Pasting an image had some problems mainly with using out of date state. Also content order is multi deminsional so a simple [...contentorder] does not work and must be deep cloned.


harshilsharma63 (approved):

jespino (approved):
LGTM

PR #1076, opened 2021-08-24
Made calculation options hidden when empty:

### LINKED ISSUE #960: Only show table calculate controls on hover
**Summary:**
Currently, the table calculate row is always visible:
Table calculate row

This adds visual noise, and is mostly unnecessary as users will typically set this once or a few times, and only for a few columns.

Proposal:
* Hide the un-set (default state) calculate controls until the user hovers over them
* But always show the controls that are set

harshilsharma63:
#### Summary
Displaying table calculation options only when not empty and on hover.

#### Ticket Link
Fixes #960


harshilsharma63:
@asaadmahmood can you run this and verify the UX?

asaadmahmood:
@harshilsharma63 For this one, we want to show all the calculate dropdowns if a person is hovering over the row.

asaadmahmood (changes requested):
See comment. Also we can make the Calculate option by default a bit greyed (opacity .6) but if it has a value, we bump up the opacity to default (1).

harshilsharma63:
@asaadmahmood can you take a look at this again? I've made the required changes.

asaadmahmood (changes requested):
The hover doesn't work on the plugin.

PR #1078, opened 2021-08-24
Add stylelint to the CI

jespino:


webapp/.stylelintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "stylelint-config-sass-guidelines",
+  "rules": {
+    "indentation": 4,
+    "selector-class-pattern": "[a-zA-Z_-]+",
+    "max-nesting-depth": 4,
+    "selector-max-compound-selectors": 6,
+    "selector-max-id": 1,
+    "selector-no-qualifying-type": null,
+    "order/properties-alphabetical-order": null
```
jespino, line 10: I want to enable this rule eventually, but in first place would generate problems with the other changes and the PRs and would make maintenance of the PR very painful.

PR #1081, opened 2021-08-24
Unfurl Focalboard Link

hahmadia:
#### Summary

Unfurl FOcalboard link in Mattemrost


mattermost-plugin/webapp/package.json
```diff
@@ -76,6 +76,7 @@
     "core-js": "3.12.1",
     "mattermost-redux": "5.33.1",
     "react-intl": "^5.13.5",
+    "react-markdown": "7.0.0",
```
kamre, line 79: Is there a plan to also use `react-markdown` for card content in the main app (not plugin) instead of using `marked.js` and `dangerouslySetInnerHTML `.

hahmadia: To be honest, probably not. I was just googling a markdown library i can use for the sake of displaying content in the card. But when reviews come, I'll probably be told to switch it or something. Didn't look at what we are using and just decided first result in google haha

asaadmahmood:
I get a errors before seeing the preview.
Screenshot 2021-08-30 at 12 20 46 AM

asaadmahmood:
So spoke with @hahmadia and the issue is that preview is "Not handling all option types on display".

mattermost-plugin/webapp/src/components/FocalboardUnfurl.scss
```diff
@@ -0,0 +1,139 @@
...
+               
+            }
+
+
+
+        }
+
+    }
+
+}
```
sbishel, line 139: nit - remove new lines.
```suggestion
                }             
            }
        }
    }
}
```

mattermost-plugin/webapp/src/components/FocalboardUnfurl.tsx
```diff
@@ -0,0 +1,190 @@
...
+import './FocalboardUnfurl.scss'
+
+type Props = {
+    embed: {
+        type: string,
+        data: string,
+    }
+}
+
+export const FocalboardUnfurl = (props: Props): JSX.Element => {
```
sbishel, line 19: Should `props.embed.type` be validated?

mattermost-plugin/webapp/src/components/FocalboardUnfurl.tsx
```diff
@@ -0,0 +1,190 @@
...
+    const [content, setContent] = useState<string>('')
+    const [board, setBoard] = useState<{title?: string, fields?: { cardProperties: Array<unknown> }}>({})
+
+    if (!workspaceID || !blockID || !baseURL || !boardID || !viewID) {
+        return <></>
+    }
+
+    useEffect(() => {
+        const fetchData = async () => {
+            const response = await fetch(`${baseURL}/plugins/focalboard/api/v1/workspaces/${workspaceID}/blocks?block_id=${blockID}`, {
```
sbishel, line 36: Probably should be using a selector here, rather than calling the API endpoint directly.

mattermost-plugin/webapp/src/components/FocalboardUnfurl.tsx
```diff
@@ -0,0 +1,190 @@
...
+            }
+            setCard(blocks[0])
+
+            if (blocks[0].fields?.contentOrder.length > 0) {
+                let contentID = blocks[0].fields?.contentOrder[0]
+                if (Array.isArray(blocks[0].fields?.contentOrder[0])) {
+                    contentID = blocks[0].fields?.contentOrder[0][0]
+                }
+
+                const contentResponse = await fetch(`${baseURL}/plugins/focalboard/api/v1/workspaces/${workspaceID}/blocks?block_id=${contentID}`, {
```
sbishel, line 56: Selector?

mattermost-plugin/webapp/src/components/FocalboardUnfurl.tsx
```diff
@@ -0,0 +1,190 @@
...
+            return null
+        }
+
+        fetchData()
+    }, [])
+
+    useEffect(() => {
+        let isCancelled = false
+        const fetchData = async () => {
+            const response = await fetch(`${baseURL}/plugins/focalboard/api/v1/workspaces/${workspaceID}/blocks?block_id=${boardID}`, {
```
sbishel, line 79: Selector?

sbishel (changes requested):
Regarding file names - 
FocalboardUnfurl.tsx
FocalboardUnfurl.scss

Suggest changing names to boardsUnfurl.*
The files should be camelCase. 
Also, I would suggest adding a subdirectory for them.
/components/boardsUnfurl/

mattermost-plugin/webapp/src/components/FocalboardUnfurl.tsx
sbishel: Spoke with Hossein, not available here, in plugin.

chenilim (changes requested):
Thanks for walking through this with me today @hahmadia! I've added comments from our discussion.

mattermost-plugin/webapp/src/components/FocalboardUnfurl.tsx
hahmadia: No it shouldn't. Good call cause we don't actually care about it here.

mattermost-plugin/webapp/package.json
hmhealey: FYI, both of those differ from the web app and mobile app Markdown since they used forked versions of marked.js and commonmark.js respectively. I'm still hoping to get those all sorted out and available for use in plugins eventually though.

hahmadia:
> So spoke with @hahmadia and the issue is that preview is "Not handling all option types on display".

@asaadmahmood 

This issue has been resolved.

Issue #1035, opened 2021-08-18
Bug: Card property display styles are inconsistent


chenilim:
**Steps to reproduce the behavior:**
1. Open a card with properties of various types
2. Note that the styles are inconsistent, e.g.:
* Created Time and Last Update Time font is smaller that Dates
* Dates are bold
* User properties have a larger font than text
* Date and User property values have different padding from others

**Expected behavior:**
Styles should be consistent

**Screenshots:**
<img width="402" alt="Card properties" src="https://user-images.githubusercontent.com/46905241/129979975-d3cd93c1-c61b-4a20-a8f5-d363cd3a680e.png">

**Platform:**
 - v0.9.0 RC6 on Community
 - Chrome 91 on macOS

chenilim:
@asaadmahmood, this is not a high priority, but probably can be cleaned up with a quick pass.

asaadmahmood:
Should be fixed with this, closing it as its a dup.
https://github.com/mattermost/focalboard/pull/1875

PR #1114, opened 2021-08-26
[GH-1059] Fix empty placeholders in the card dialog

### LINKED ISSUE #1059: Bug: Number property should show "Empty" placeholder in card editor
1. Open a card with a number property that is not set
2. The field placeholder is blank

**Expected behavior:**
* The field placeholder should show the text "Empty" in the card editor for all property types
* However, the placeholder should be blank in tables

**Screenshots:**
Number empty

**Platform:**
 - Main branch build, Aug 20, 2021

**Additional context:**
This is a regression caused by #966, see [thread](https://community.mattermost.com/core/pl/aaz64xgkspri3ff8k7cxoatbmo).

The property value display logic needs to be refactored, probably with a "showEmptyPlaceholder" flag. True for the card editor, and false for tables.

kamre:
#### Summary
Show empty placeholder for values in the card dialog. At the same time empty placeholders are not used for table cell, kanban or gallery cards.

Here is how it looks in the card dialog:
image

In kanban card:
image

In the table view:
TableViewRow

#### Ticket Link
Fixes #1059


harshilsharma63:
@kamre awesome work as always 👍 

One change though. The empty placeholder looks nice for text and other text-based properties (email, URL, etc) but the styling drifts away for some special property types. See the screenshot attached below-
 
Screenshot 2021-09-06 at 5 44 15 PM

It will look really nice and consistent if the empty palceholder looks consistent across property types. DO you think this is doable?

kamre:
@harshilsharma63 ok, I will look into this. But there is also more generic issue #1035 about styles.

kamre:
@harshilsharma63 updated styles for empty placeholders:
image

webapp/src/widgets/valueSelector.tsx
```diff
@@ -114,6 +114,10 @@ const valueSelectorStyle = {
         padding: '0 8px',
         overflow: 'unset',
     }),
+    placeholder: (provided: CSSObject): CSSObject => ({
+        ...provided,
+        color: 'rgba(var(--center-channel-color-rgb), 0.4)',
+    }),
```
kamre, line 120: May be this style for placeholder should be moved to [getSelectBaseStyle](https://github.com/mattermost/focalboard/blob/main/webapp/src/theme.ts#L242)

harshilsharma63 (approved):
LGTM!

webapp/src/widgets/valueSelector.tsx
harshilsharma63: If we move it to `getSelectBaseStyle` we have to specify select styling for places other than card properties. Either way, we'll end up having something specified in base and it being overridden in uses, so I think what we have here is fine.

kamre:
@chenilim do we need any additional changes for this pull request?

chenilim:
Apologies for the delay @kamre! I've been behind on my GitHub alerts. I've smoke tested it and the fix looks good to me. Unfortunately, it looks like you need a manual merge conflict resolve.

For future reference, we just need one developer approval to merge - so whoever approves it, just go ahead and merge if there are no further changes needed.

chenilim (approved):
LGTM. Thanks!

chenilim:
Was able to manually merge. Thanks!

PR #737, opened 2021-07-16
Migrate webapp global state to redux

jespino:

PR #1120, opened 2021-08-27
[GH-1056] Fix setting focus for empty card title

### LINKED ISSUE #1056: Bug: no focus for empty card title
**Summary:**
When creating a new card or opening a dialog for existing card with empty title there should be focus on the textarea used for the card title.

1. Create empty board
2. Create new card
3. There is no focus on the title in the opened dialog

**Expected behavior:**
There should be focus on the title because it is empty.

**Platform:**
 - Browser and version: Chrome 92
 - OS and version: Windows 10

**Additional context:**
This is a regression caused by #737

@jespino FYI

kamre:
#### Summary
After #737 the order of the calls to `useEffect` callbacks for [CardDetail](https://github.com/mattermost/focalboard/blob/main/webapp/src/components/cardDetail/cardDetail.tsx#L55-L59) and [RootPortal](https://github.com/mattermost/focalboard/blob/main/webapp/src/components/rootPortal.tsx#L15-L24) is not the expected one, i.e. focus is set for the element that has not been yet appended to the DOM tree.

Switching to `useLayoutEffect` allows to append the element before setting the focus.

#### Ticket Link
Fixes #1056


harshilsharma63 (approved):
LGTM!

Issue #1131, opened 2021-08-30
Bug: Use picker in global nav is cut off


chenilim:
**Steps to reproduce the behavior:**
1. On Community, switch to Boards
2. Click on your user icon at the top right
3. The dialog is cut off on the right

**Expected behavior:**
The dialog should be fully visible

**Screenshots:**
<img width="193" alt="User dialog" src="https://user-images.githubusercontent.com/46905241/131363698-ae9fe793-0a98-4706-8995-8b489a395143.png">

**Platform:**
 - v0.9.0 RC7 on Community
 - Chrome 91 on macOS

**Additional context:**
This is a regression from the last RC

chenilim:
Per the [thread here](https://community-daily.mattermost.com/core/pl/8m43totwjjbqfdkc91j8aizmqc), this is already fixed in master. If it involves Focalboard code, let's make sure this is fixed in the 0.9 branch. Assigning to you to reassign @sbishel. cc @ogi-m.

PR #1132, opened 2021-08-30
GH-1131 - fix for displaying global nav in focalboard

sbishel:
#### Summary
Updates CSS to unset menu setting that are causing an issue with the GlobalNav Bar.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/1131

### DIFF
--- a/mattermost-plugin/webapp/src/plugin.scss
+++ b/mattermost-plugin/webapp/src/plugin.scss
@@ -9,6 +9,8 @@
     z-index: 1000;
 
     .Menu {
+        position: unset;
+        min-width: unset;
         a,
         button {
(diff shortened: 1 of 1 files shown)


chenilim (approved):
LGTM, but it does seem we may need to revisit / refactor our menu system so not to conflict with the base Mattermost one.

sbishel:
@asaadmahmood There is probably a better way to fix this, rather than "unset". But these settings are necessary for other uses of this class. Let me know if a different fix would be better.

asaadmahmood (approved):
Probably fine for now, we can revisit this later as a refractor of the Menu component itself.

PR #1147, opened 2021-09-03
@mention support

wiggin77:
#### Summary
This PR adds @mention support.  When in plugin mode, a mentioned user will receive a DM from `boards` bot which includes the username of the author, the title of the card they were mentioned in, a link to the card, and some surrounding text for the mention.

- refactor server init to handle more params
- create notification service 
- create notification backend + plugin delivery
- unit tests

#### Ticket Link
https://community.mattermost.com/plugins/focalboard/workspace/qgsck6cts3fwpqwyjiupjm5cde/47aa9bb4-6967-4a96-83c7-11bd6b20f1eb/150d1464-6d59-4c39-810f-ee201439dbaf?c=c1260333-2a8a-456a-854d-4e247a4716e7


server/services/notify/notifymentions/mentions_test.go
```diff
@@ -0,0 +1,100 @@
...
+		wantErr bool
+	}{
+		{name: "user1", uname: user1.Username, want: user1, wantErr: false},
+		{name: "user1 with period", uname: user1.Username + ".", want: user1, wantErr: false},
+		{name: "user1 with period plus more", uname: user1.Username + ". ", want: user1, wantErr: false},
+		{name: "user2 with periods", uname: user2.Username + "...", want: user2, wantErr: false},
+		{name: "user2 with underscore", uname: user2.Username + "_", want: user2, wantErr: false},
+		{name: "user2 with hyphen plus more", uname: user2.Username + "- ", want: user2, wantErr: false},
+		{name: "user2 with hyphen plus all", uname: user2.Username + ".-_ ", want: user2, wantErr: false},
+		{name: "user3 with underscore", uname: user3.Username + "_", want: user3, wantErr: false},
```
sbishel, line 50: I don't see any error tests. Should a test be added for user4? (wantErr: true)

server/services/notify/notifymentions/mentions.go
```diff
@@ -0,0 +1,84 @@
...
+
+const (
+	usernameSpecialChars = ".-_ "
+)
+
+var atMentionRegexp = regexp.MustCompile(`\B@[[:alnum:]][[:alnum:]\.\-_:]*`)
+
+// extractMentions extracts any mentions in the specified block and returns
+// a slice of usernames.
+func extractMentions(block *model.Block) map[string]struct{} {
```
sbishel, line 23: Add unit test for this.

mattermost-plugin/server/notifications.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/mattermost/mattermost-server/v6/model"
+
+	"github.com/mattermost/mattermost-server/v6/shared/mlog"
+)
+
+const (
+	botUsername    = "boards"
+	botDisplayname = "Boards"
+	botDescription = "Created by Boards plugin."
```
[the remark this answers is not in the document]
sbishel: ...once we have i18n in server...(just read other comment.)

server/services/notify/notifymentions/extract.go
```diff
@@ -0,0 +1,95 @@
...
+	lines := strings.Split(s, "\n")
+
+	// find first line with mention
+	var found int
+	for i, l := range lines {
+		if strings.Contains(l, mention) {
+			found = i
+			break
+		}
+	}
```
sbishel, line 48: Should we simply return "", if not found.

sbishel (changes requested):
Looks really good and works well. Only the link didn't work, but I assume that is due to the view id issue.

Just a few nits.

mattermost-plugin/server/notifications.go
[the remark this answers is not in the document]
jespino: This is going to be used in the `EnsureBot` call, on the first run of the server (with the plugin), that means that 99% of the time this is going to be executed in the first run, and the only locale that you can use there is the server one, because is not bound to a user. That said, it means, that almost 99% of the time it is going to be in English anyway. We will need to add translations to the notifications, but I not sure if we need them here.

mattermost-plugin/server/notifications.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	delivery := &deliveryAdapter{client: client}
+
+	backend := notifymentions.New(delivery, botID, serverRoot, logger)
+
+	return backend, nil
+}
+
+type deliveryAdapter struct {
```
jespino, line 40: I have doubts about this interface/implementation.

I think this defines how you deliver the notification, in this case through the API, but it is also leaking some implementation details related to how you resolves the destination.

My expectation would be that receives a notification to deliver, and you deliver it. For example, if I add an "email adapter" later on, why that implementation requires the `GetDirectChannel`?

Maybe our plugin based delivery adapter should receive the `GetDirect`, `CreatePost`, `GetUserByID` and `GetUserByUsername` functions at construction time, and expose `Send(changes)` method that does all the magic needed to send through the API.

jespino (changes requested):
Only one real concern around the Delivery interface and how general can it be for other delivery methods.

mattermost-plugin/server/notifications.go
wiggin77: I think what you are describing is how it is implemented.  A `backend` exposes a generic method that gets called with changes and at construction time accepts the `GetDirect`, etc, functions.  This allows us to easily create different types of backends (not just for mentions). The `Delivery` interface is just for the mentions backend... other backend types (e.g. card subscription) can have different delivery interfaces.

wiggin77: BTW, the only reason there is an adapter is because the new, refactored, plugin API cannot be passed as is and fulfil an interface.  All the methods have been grouped into structs within it a client struct.

mattermost-plugin/server/notifications.go
```diff
@@ -0,0 +1,58 @@
...
+	return da.client.Post.CreatePost(post)
+}
+
+func (da *deliveryAdapter) GetUserByID(userID string) (*model.User, error) {
+	return da.client.User.Get(userID)
+}
+
+func (da *deliveryAdapter) GetUserByUsername(name string) (*model.User, error) {
+	return da.client.User.GetByUsername(name)
+}
```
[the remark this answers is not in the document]
wiggin77: Could be, but I like the idea of this simple adapter being in the one place that it is used.

mattermost-plugin/server/notifications.go
jespino: My point is. mentions can be notified through the plugin API using a message, or through an email. I see multiple delivery backends for the mentions, not only the API delivery backend, and I have the sensation that there is no other possible backend for that.

Maybe I'm understanding this all the way around. Maybe you want to have a Delivery mechanism that is for "send DMs to somebody" that can be reuse in other backends later?

server/app/blocks.go
```diff
@@ -30,40 +33,64 @@ func (a *App) GetParentID(c store.Container, blockID string) (string, error) {
...
+		needsNotify = append(needsNotify, blocks[i])
+
 		a.wsAdapter.BroadcastBlockChange(c.WorkspaceID, blocks[i])
-		a.metrics.IncrementBlocksInserted(len(blocks))
-		go a.webhook.NotifyUpdate(blocks[i])
+		a.metrics.IncrementBlocksInserted(1)
 	}
 
+	go func() {
+		for _, b := range needsNotify {
```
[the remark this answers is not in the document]
wiggin77: I can refactor this when we have an updated spec covering this.  cc @chenilim

mattermost-plugin/server/notifications.go
wiggin77: Ok, I've refactored to create a more generic `Delivery` interface, with a `PluginAPIDelivery` implementation.  We can add an `EmailDelivery` implementation, or whatever in the future.

server/go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.7.0
+	github.com/wiggin77/merror v1.0.3
```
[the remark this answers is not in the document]
wiggin77: Done via https://github.com/mattermost/focalboard/pull/1147/commits/1c131c9f301a8e2cc9e5258516de9473d0533e73

server/services/notify/notifymentions/extract.go
wiggin77: Done via https://github.com/mattermost/focalboard/pull/1147/commits/1c131c9f301a8e2cc9e5258516de9473d0533e73

server/services/notify/notifymentions/mentions_test.go
wiggin77: Done via https://github.com/mattermost/focalboard/pull/1147/commits/1c131c9f301a8e2cc9e5258516de9473d0533e73

server/services/notify/notifymentions/extract_test.go
```diff
@@ -0,0 +1,115 @@
...
+		mention string
+		limits  limits
+	}
+	tests := []struct {
+		name string
+		args args
+		want string
+	}{
+		{name: "good", want: join(s2, s3, s4, s5, s6), args: args{mention: "@lincoln", limits: extractLimits, s: allConcat}},
+		{name: "not found", want: join(s0, s1, s2), args: args{mention: "@bogus", limits: extractLimits, s: allConcat}},
```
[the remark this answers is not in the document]
wiggin77: I modified to return nothing if the mention is not found.  https://github.com/mattermost/focalboard/pull/1147/commits/1c131c9f301a8e2cc9e5258516de9473d0533e73

mattermost-plugin/server/notifications.go
jespino: I really like more the current result. Thanks!

jespino (approved):
LGTM

sbishel (approved):
Thanks Doug.

PR #1156, opened 2021-09-07
Added subpath support in one missed place

harshilsharma63:
#### Summary
Added subpath support in one missed place.


jespino (approved):
LGTM

PR #1162, opened 2021-09-08
Fix #1161. Add Give Feedback link.

### LINKED ISSUE #1161: Add link to Give Feedback form
**Summary:**
Add a link to give feedback.

Provide a low-friction way for our users to provide product feedback.

chenilim:
#### Summary
This add a "Give Feedback" link to the top-right of the main UI. The link will open an fwlink that points to an anonymous Google Form - one for Focalboard and another similar one for Boards.

#### UI in Focalboard:
Give Feedback Focalboard

#### UI in Boards:
Give Feedback Boards


webapp/src/components/topBar.tsx
```diff
@@ -4,16 +4,28 @@
...
             >
+                <a
+                    className='link'
+                    href={feedbackUrl}
+                >
+                    <FormattedMessage
+                        id='TopBar.give-feedback'
+                        defaultMessage='Give Feedback'
+                    />
+                </a>
```
sbishel, line 28: Need to run `npm run i18n-extract` in webapp to update language files. Unless there are already there.

sbishel (changes requested):

webapp/src/components/topBar.tsx
chenilim: Thanks, good catch. Ran and updated.

sbishel (approved):
LGTM

Issue #1136, opened 2021-08-31
Bug: "Something went wrong" error after converting multi select property to URL


ogi-m:
**Summary:**
Converting multiselect property to URL causes the board to return "Something went wrong" error with no apparent way to fix it. Reopening the board results in the same error.

**Steps to reproduce the behavior:**
1. Open boards from Mattermost
2. Create a card
3. Create a URL property and add a link to the card
4. Change the URL property to multi select
5. Change the property back from multi select to URL

**Expected behavior:**
"Something went wrong" error is displayed and there's no way to access the board.

**Screenshots:**
![SharedScreenshot](https://user-images.githubusercontent.com/15314183/131502993-286284ab-0538-4507-af83-3275bbdd1810.jpg)

**Platform:**
Focalboard plugin v0.9.0 RC8 on community-daily (master)

chenilim:
We should at least not permanently crash on load.

sbishel:
@ogi-m I am not able to reproduce, tried on main as well as community-daily. We can look at tomorrow.

ogi-m:
Tested and passed on v0.9.0 RC12 
- not seeing the error anymore
- property converted back to URL

PR #1174, opened 2021-09-09
GH-1136 - fix for when property value is an array

sbishel:
#### Summary
When a property gets converted from a "multi-select" property. The value is a string[], rather than just a string.

When passing to the PropertyValueElement using `value as string` this allows the typescript to compile, however, the underlying javascript is still passing an Array.  Because the URL component attempts to `.trim()`. This fails when an array is passed. 

Updated to pass `value.toString()` which will actually convert a string[] to a string. If multiple values are present, those are reduces to the first value when the property is updated, but still comes through as an array during redisplay.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/1136


webapp/src/components/propertyValueElement.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+        </DndProvider>
+    )
+}
+
+describe('components/propertyValueElement', () => {
+    const board = TestBlockFactory.createBoard()
+    const card = TestBlockFactory.createCard(board)
+    const comments = TestBlockFactory.createComment(card)
+
+    test('should match snapshot, select', async () => {
```
kamre, line 32: Why `async` is needed?

harshilsharma63 (approved):
LGTM!

Though I wasn't able to reproduce the issue.

webapp/src/components/propertyValueElement.test.tsx
sbishel: It's not...updated. Thank you.

PR #1176, opened 2021-09-10
Update plugin.json link to docs

justinegeffen:
Updated the help link to the live docs.

@chenilim, I suspect I may have the incorrect link formatting. FWIW, in terms of the docs, if anything changes with the URL we set up 301 redirects on our side so this link shouldn't break. That said, I'd like to follow best practice so let me know where I went wrong. :)


wiggin77 (approved):
LGTM

chenilim (approved):
Sorry for the delay. LGTM!

justinegeffen:
Looks like this has been removed in main. @chenilim, can you confirm whether this change is still needed?

chenilim:
@justinegeffen, this has been removed and the link is no longer needed. I'll close this PR.

In v6.0 and beyond, the user does not need to perform the additional setup steps (install from market place, update web proxy) they had to before, so the link is no longer necessary.

PR #1178, opened 2021-09-10
Update en.json - UI capitalization

justinegeffen:
Adjusted capitalization for UI text for consistency.

I'm not sure I got them all right - I couldn't quite figure out what was displayed and whether the punctuation is correct. Will check in prod and adjust if needed (unless we have a test server we can spin up).

There are still some inconsistencies, such as the Personal Goals board having "By status" and the other boards having "By Status". I'm 3/5 on changing them all to "By status/By priority" etc for consistency and also because we use sentence case in the Channels UI as well (in places that we don't we're working on standardizing it) if everyone is OK with that. Happy to discuss further! Also want to make sure there are no breaking changes here.


webapp/i18n/en.json
```diff
@@ -85,11 +85,11 @@
   "ShareBoard.copiedLink": "Copied!",
   "ShareBoard.copyLink": "Copy link",
   "ShareBoard.regenerateToken": "Regenerate token",
-  "ShareBoard.share": "Publish to web and share this board to anyone",
+  "ShareBoard.share": "Publish and share this board with anyone who has the link.",
```
chenilim, line 88: Overall, adding the proper punctuation is a great improvement. However, there may be a few cases we might want to revisit later. E.g. this text is the label for a switch in the Share Dialog:
Share board text

There are other places where switch labels don't have periods at the end, e.g. the Random icons switch:
Random icons text

It's not a huge deal either way, and we can iterate on it. Maybe something to consider for UX guidelines if not already? cc @asaadmahmood.

chenilim (approved):
Thanks for polishing this @justinegeffen! Agree we should be using sentence case consistently, and with proper punctuation in general.

There are a few corner cases where the text might be a label (e.g. the share switch text), and we might need to clarify guidance on this. We can iterate on that though.

The only other thing I can think of is to make similar changes in the localized text - I *believe* the updated strings are automatically flagged in Weblate? cc @jespino.

webapp/i18n/en.json
justinegeffen: Thanks, @chenilim! I actually didn't consider the buttons - any text on those should not get punctuation. And menu items don't get punctuation either. I'm giving this another review to ensure this is adhered to. 

I've added this to the UX Guidelines update list, thank you!

webapp/i18n/en.json
```diff
@@ -158,13 +158,13 @@
...
+  "Workspace.editing-board-template": "You're editing a board template.",
   "default-properties.title": "Title",
   "error.no-workspace": "Your session may have expired or you may not have access to this workspace.",
   "error.relogin": "Log in again",
   "login.log-in-button": "Log in",
   "login.log-in-title": "Log in",
-  "login.register-button": "or create an account if you don't have one",
-  "register.login-button": "or login if you already have an account",
+  "login.register-button": "or create an account if you don't have one.",
+  "register.login-button": "or log in if you already have an account.",
```
justinegeffen, line 168:
```suggestion
  "register.login-button": "or log in if you already have an account",
```

jespino (approved):
Nice. Thanks!

harshilsharma63:
Few tests are failing. I'll fix those.

harshilsharma63 (approved):

PR #1179, opened 2021-09-10
Displayed channels names for DMs in user workspace list

harshilsharma63:
#### Summary
Displayed channels names for DMs in user workspace list

#### Ticket Link
NA

Fixes the issue where `getUserWorkspaces()` API didn't return workspace names corresponding to DM channels.

Before-
before

After-
after


jespino (approved):
LGTM

PR #1184, opened 2021-09-10
Not importing compass-icons on plugin mode

### LINKED ISSUE #1153: Bug: Use the compass-icons from mattermost-webapp when we are in plugin mode
**Summary:**
When we are in plugin mode, we are using our own compass-icons version and that is overwritting the icons font from mattermost-webapp. Focalboard should share the compass-icons from mattermost-webapp instead of using its own version.

jespino:
Fixes #1153


wiggin77 (approved):
LGTM :+1:

Issue #1191, opened 2021-09-13
Bug: menu for card content item is covered with other items


kamre:
**Steps to reproduce the behavior:**
1. Create new empty board
2. Create new card
3. Add couple of checkbox items to card content
4. Activate popup menu for the first item
5. Menu is covered with other items below the current one

**Expected behavior:**
Menu should be on top of any other items

**Screenshots:**
![CardContentItemMenu](https://user-images.githubusercontent.com/5587620/133073715-1b9f8163-bd68-4c91-9242-220773c25673.gif)

**Platform:**
 - Browser and version: Chrome 92
 - OS and version: Windows 10

**Additional context:**
Reproducible on `v0.9.0`
Regression was introduced by fix in #1097

chenilim:
Thanks @kamre! Confirmed this is an issue on Chrome Mac as well. Looks like the menu z-order is incorrect? @asaadmahmood, this may be related to the other menu issues, and we should fix for v0.9. Could you take a quick look? If it's a more involved fix, let's discuss. cc @sbishel, @ogi-m.

ogi-m:
Tested and passed on v0.9.0 RC11

Issue #1189, opened 2021-09-13
Bug: kanban column header is overlapping with the scrolled content


kamre:
**Steps to reproduce the behavior:**
1. Go to [v1 Focalboard (read-only)](https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/shared/47aa9bb4-6967-4a96-83c7-11bd6b20f1eb/7e38eab4-890a-4fc3-8f48-a1f9bd1ab822?r=762db9c6-305c-4d13-8ce6-9f11e5232c24)
2. Select _By Status_ view
3. Scroll horizontally to see the last column
4. Scroll vertically so that some cards go above the column headers
5. Look at the overlap between the column header and the card

**Expected behavior:**
Card content should be below column header and should not be visible.

**Screenshots:**
![KanbanHeadersOverlap](https://user-images.githubusercontent.com/5587620/133043615-ae765088-d5f8-4e7f-bf1d-5190e25d9bde.gif)

**Platform:**
 - Browser and version: Chrome 91
 - OS and version: Windows 10

**Additional context:**
**This only happens when there is a horizontal scroll bar and some column headers are not visible initially.**
Seems to be the issue similar to #1053 but with kanban board.

ogi-m:
This is not limited to read-only view, I can repro in regular view as well
cc @chenilim

chenilim:
Thanks @kamre and @ogi-m! @asaadmahmood, could you investigate this too? It seems there might be some z-order issues in the UI. @ogi-m, what platform were you able to repro this on? I couldn't repro on Mac (Chrome, FF, Safari), or Windows Edge.

ogi-m:
@chenilim I can see it on Windows Chrome, Firefox & Edge. Try scaling down the browser width and make sure it's scrolled all the way to the right horizontally, that should help repro.

ogi-m:
Tested and passed on v0.9.0 RC11

Issue #1182, opened 2021-09-10
Bug: "Set language" menu opens over the root menu and pushes it to the left


ogi-m:
**Steps to reproduce the behavior:**
1. Open Boards from Mattermost
2. Click on the Settings button in the top right corner
3. Move the mouse over "Set Language" option

**Expected behavior:**
Menu should open to the left of the settings menu (as the arrow indicates)

**Screenshots:**
https://user-images.githubusercontent.com/15314183/132895008-9c926916-f6d8-4616-97bf-102265f19515.mp4

**Platform:**
 - Plugin v0.9.0 RC9 on master

chenilim:
Thanks @ogi-m! This is a regression from RC8, and should be fixed for v0.9. Might be related to #1172, and hopefully is just a CSS change. @asaadmahmood, could you help investigate? I'm concerned that this bug could affect other UI as well. cc @sbishel, @jespino.

moicool6:
Is it impossible to save the record of setting language, and you need to set it again every time.

chenilim:
Thanks @moicool6! What version are you running? I'm not able to repro this on v0.9.0 RC9 (running as a plugin on our community server). Are you running one of the desktop apps or Personal Server 0.9.0 RC9? Thanks.

ogi-m:
Tested and passed on v0.9.0 RC11

Issue #1173, opened 2021-09-09
Bug: Screen position judders when hovering over "Show Description" button


chenilim:
**Steps to reproduce the behavior:**
1. Open a board or table
2. Hover over the "Show Description" button at the top
3. The content below is pushed down, resulting in a visual judder

**Expected behavior:**
The other page content location should not change on hover.

**Screenshots:**
https://user-images.githubusercontent.com/46905241/132768286-e9e6cb88-9b76-4879-b6ed-6cb5699f2dad.mov

**Platform:**
 - v0.9.0 RC9 on Community
 - Chrome 91 on macOS

**Additional context:**
This is a regression from v0.8

chenilim:
@asaadmahmood, this is similar to #1024, but is a regression. I think we can live with this for v0.9, but it's a poor first impression, and if you think it's an easy and safe fix, we can take it for v0.9. cc @sbishel, @ogi-m.

kamre:
The reason is that `min-height` for the container is `28px` and button `height` [was changed](https://github.com/mattermost/focalboard/pull/1138/files#diff-3821ab810407205c0e9df9f20311ddedf382208a89ca3e90a2935573e698da46R16) to `32px`.
Possible fix:
```diff
diff --git a/webapp/src/components/viewTitle.scss b/webapp/src/components/viewTitle.scss
--- a/webapp/src/components/viewTitle.scss	(revision 09e35ae7c2a8e71cd0aaabdc2abc8644d15ebea9)
+++ b/webapp/src/components/viewTitle.scss	(date 1631252076594)
@@ -9,7 +9,7 @@
 
     > .add-buttons {
         flex-direction: row;
-        min-height: 28px;
+        min-height: 32px;
         color: rgba(var(--center-channel-color-rgb), 0.4);
         width: 100%;
         align-items: flex-start;
@@ -23,7 +23,6 @@
         }
 
         .Button {
-            min-height: 28px;
             display: none;
             line-height: 1;
         }
```

ogi-m:
Tested and passed on v0.9.0 RC11

PR #1194, opened 2021-09-13
Focalboard UI Fixes

asaadmahmood:
#### Summary
Focalboard UI Fixes

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/1189
Fixes https://github.com/mattermost/focalboard/issues/1182
Fixes https://github.com/mattermost/focalboard/issues/1173
Fixes https://github.com/mattermost/focalboard/issues/1172
Fixes https://github.com/mattermost/focalboard/issues/1191
Fixes https://github.com/mattermost/focalboard/issues/1197


mattermost-plugin/webapp/src/plugin.scss
```diff
@@ -9,9 +9,6 @@
     z-index: 1000;
 
     .Menu {
-        position: unset;
```
chenilim, line 12: @sbishel, was there a reason this was needed in the original checkin? I'm worried removing it might break something else?

chenilim (approved):
The changes look good to me.

mattermost-plugin/webapp/src/plugin.scss
sbishel: The `min-width: unset;` is not needed because the `min-width: 180px;` in `menu.scss` is being removed as well, see below.

However, the `position: unset;` removal does cause a slight difference in menu positioning between `Channels` and `Boards`.

Screen Shot 2021-09-13 at 5 23 11 PM
Screen Shot 2021-09-13 at 5 23 01 PM

sbishel (changes requested):
Other than the change in positioning, the other changes look good and everything works properly.

asaadmahmood:
@sbishel We should not override the position or the min/max width coming from the global header, otherwise a lot of things can get messed up.

That position is also getting messed up because we are overriding the global menus position based `position: absolute` we have on our `.Menu`.

I have however overriden it in my latest commit for now.
But we should remove the `.focalboard-body` appended to the scss selectors later.

Also the issue you're talking about should be fixed now.

Screenshot 2021-09-14 at 11 15 03 AM

ogi-m:
Verified the fixes for:
- https://github.com/mattermost/focalboard/issues/1189
- https://github.com/mattermost/focalboard/issues/1182
- https://github.com/mattermost/focalboard/issues/1173
- https://github.com/mattermost/focalboard/issues/1191
- https://github.com/mattermost/focalboard/issues/1197

But I'm still seeing the white space in the user modal
whitespace

Also opened [an issue](https://github.com/mattermost/focalboard/issues/1201) for the styling on the settings and custom status modals.

asaadmahmood:
@ogi-m Should be fixed now.

sbishel (approved):

ogi-m (approved):
Thanks @asaadmahmood , LGTM

PR #1204, opened 2021-09-14
Fix #1203. Require v6.0.0 for plugin

### LINKED ISSUE #1203: Update plugin.json to require Mattermost v6.0.0
**Summary:**
Update the plugin manifest to require Mattermost v6 for the v0.9 release.

chenilim:
Update the plugin.json and manifest.go.


jespino (approved):

mattermost-plugin/server/manifest.go
```diff
@@ -20,7 +20,7 @@ const manifestStr = `
   "release_notes_url": "https://github.com/mattermost/focalboard/releases",
   "icon_path": "assets/starter-template-icon.svg",
   "version": "0.9.0",
-  "min_server_version": "5.38.0",
+  "min_server_version": "6.0.0",
```
chenilim, line 23: @jespino, do we need to manually update the min_server_version in `manifest.go` as well? I just searched for "5.38" and found it here.

sbishel (approved):

mattermost-plugin/server/manifest.go
sbishel: It shouldn't be updated manually. But gets built when building the server. That said, the automatic build would result in the same changes you have.

chenilim: Thanks. I'll leave this in then. Maybe a question for later: Should `manifest.go` be checked in then?

PR #1208, opened 2021-09-14
Fix #1196. Open feedback form in new tab / window.

### LINKED ISSUE #1196: Feature Idea: Provide way to go back or cancel after clicking provide feedback button at top right
**Summary:**

Right now after clicking the feedback button at the top right of the program gets caught in a stuck state where the only way to progress is to give feedback.  If you're a curious user who doesn't want to actually give feedback that means you have to close and re-open the program to get back to the original screen.

Importance: Low

chenilim:
Open the feedback form in a new tab / window (from personal desktop app), consistent with the help link.


wiggin77 (approved):
LGTM :+1:

PR #1210, opened 2021-09-15
Updated 0 board to 0 boards (singular vs plural)

### LINKED ISSUE #1200: Bug: Workspace switcher > "0 Board" when a workspace doesn't have any boards
1. Open Boards
2. Open the workspace switcher
3. Observe a channel with no boards

**Expected behavior:**
It should say "0 Boards" instead of "0 Board"

**Screenshots:**
workspace_switcher

**Platform:**
 - latest main plugin build on mattermost master

harshilsharma63:
#### Summary
Updated 0 board to 0 boards (singular vs plural)

#### Ticket Link
Fixes #1200


webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
```diff
@@ -102,7 +102,7 @@ const Option = (props: any): JSX.Element => {
             {
                 props.data.value !== DashboardOption.value &&
                 <div className='boardCount'>
-                    {props.data.boardCount} {props.data.boardCount > 1 ? 'Boards' : 'Board'}
+                    {props.data.boardCount} {props.data.boardCount === 1 ? 'Board' : 'Boards'}
```
jespino, line 105: This is not translatable text, I my opinion this should be a FormattedMessage with pluralization.

harshilsharma63: Done.

jespino:
I was thinking about something more in the line of this: https://codesandbox.io/s/o56w8ywwoz

Also you shouldn't concat numbers with strings, not all the languages in the world puts the numbers first, some languages could express that like "Boards 1" for example. so try to include the whole phrase in the translation, including all the parts of the phrase, and avoid concatenations.

jespino (changes requested):
Still can be improved.

webapp/src/pages/dashboard/centerContent.tsx
```diff
@@ -87,7 +87,13 @@ const DashboardCenterContent = (): JSX.Element => {
                                     {workspace.title}
                                 </div>
                                 <div className='small'>
-                                    {`${workspace.boardCount} board${workspace.boardCount > 1 ? 's' : ''}`}
+                                    {
+                                        intl.formatMessage({
+                                            id: 'General.BoardCount',
+                                            defaultMessage: `${workspace.boardCount} Board${workspace.boardCount === 1 ? '' : 's'}`,
```
jespino, line 93: The defaultMessage should be the same used in the `i18n/es.json`

webapp/i18n/en.json
```diff
@@ -52,6 +52,7 @@
   "GalleryCard.delete": "Delete",
   "GalleryCard.duplicate": "Duplicate",
   "GroupBy.ungroup": "Ungroup",
+  "General.BoardCount": "{count, plural, =0 {# Boards} one {# Board} other {# Boards}}",
```
jespino, line 55: Probably you can simplify this as `"{count, plural, one {# Board} other {# Boards}}"`

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
```diff
@@ -102,7 +103,13 @@ const Option = (props: any): JSX.Element => {
             {
                 props.data.value !== DashboardOption.value &&
                 <div className='boardCount'>
-                    {props.data.boardCount} {props.data.boardCount > 1 ? 'Boards' : 'Board'}
+                    {
+                        intl.formatMessage({
+                            id: 'General.BoardCount',
+                            defaultMessage: `${props.data.boardCount} Board${props.data.boardCount === 1 ? '' : 's'}`,
```
jespino, line 109: The defaultMessage should be the same used in the `i18n/es.json`

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
```diff
@@ -102,7 +103,13 @@ const Option = (props: any): JSX.Element => {
             {
                 props.data.value !== DashboardOption.value &&
                 <div className='boardCount'>
-                    {props.data.boardCount} {props.data.boardCount > 1 ? 'Boards' : 'Board'}
+                    {
+                        intl.formatMessage({
```
jespino, line 107: Why do you use `intl.formatMessage` instead of the `FormattedMessage` component?

jespino (changes requested):
Still some improvements to be made.

webapp/src/pages/dashboard/centerContent.tsx
harshilsharma63: Done.

webapp/i18n/en.json
harshilsharma63: Done.

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
harshilsharma63: Done.

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
harshilsharma63: No specific reason. Why should I use `FormattedMessage` instead of `intl.formatMessage`? Is there any objective advantage in this particular case?

jespino: The main reason is, you are representing a block here, so... using FormattedMessage looks like the right thing, probably also it is more perfomant because you don't need to re-render every time. Also, you don't need to use the `useIntl()` there.

jespino (changes requested):
still some things need to be handled.

webapp/src/components/workspaceSwitcher/workspaceOptions.tsx
harshilsharma63: Done.

Also, this is such an insignificant change as it wasn't worth the effort to do this. Let just accept what's in place and be done with it.

webapp/i18n/en.json
harshilsharma63: done.

jespino (approved):
LGTM

PR #1212, opened 2021-09-15
Desktop App Navigation Fixes

### LINKED ISSUE #1142: Bug: Navigation with Desktop App doesn't work correctly.
**Summary:**
Focalboard uses the `react-router-dom` history module to navigate between its pages without refreshing. Unfortunately the desktop app is not aware of the changes this module makes and cannot facilitate any navigation between tabs this way.

We need to either:
a) Switch Focalboard to use the provided `browserHistory` module in the webapp (under `window.WebappUtils.browserHistory`)
b) Create a shim around the `react-router-dom/history` module that will do the same thing as the webapp to facilitate desktop navigation between tabs.

Click on the Global Header and go to Boards OR
Click on a link in Channels that links to a board

**Expected behavior:**
Boards should open on the intended board.

**Observed behavior:**
With the first step, Boards opens in the wrong tab because it is doing a redirect
With the second, the Boards tab opens but doesn't navigate to the desired link

**Platform:**
Desktop App v5.0

harshilsharma63:
#### Summary
Desktop App Navigation Fixes

#### Ticket Link
Fixes #1142


webapp/src/utils.ts
```diff
@@ -442,6 +442,59 @@ class Utils {
...
+    /**
+     * Boolean function to check if a server version is greater than another.
+     *
+     * currentVersionParam: The server version being checked
+     * compareVersionParam: The version to compare the former version against
+     *
+     * eg.  currentVersionParam = 4.16.0, compareVersionParam = 4.17.0 returns false
+     *      currentVersionParam = 4.16.1, compareVersionParam = 4.16.1 returns true
+     */
+    static isServerVersionGreaterThanOrEqualTo(currentVersionParam: string, compareVersionParam: string): boolean {
```
jespino, line 470: nit: you are only using it to compare the desktop app version, and the name here is misleading. probably something like `isVersionGreaterThanOrEqualTo` would be better. The comments also are misleading.

jespino (approved):
LGTM. Just an small improvement suggestion.

webapp/src/utils.ts
harshilsharma63: Done.

webapp/src/utils.ts
```diff
@@ -442,6 +442,59 @@ class Utils {
...
+
+    static getDesktopVersion(): string {
+        // use if the value window.desktop.version is not set yet
+        const regex = /Mattermost\/(\d+\.\d+\.\d+)/gm
+        const match = regex.exec(window.navigator.appVersion)?.[1] || ''
+        return match
+    }
+
+    /**
+     * Boolean function to check if a desktop version is greater than another.
```
jespino, line 462: nit: it is not about the desktop app either, this function is about comparing to different versions.

harshilsharma63: Done.

PR #1213, opened 2021-09-15
Propagate block changes to all of the cluster nodes in plugin mode

### LINKED ISSUE #1154: Bug: Refresh required after every action in HA
**Summary:**
Same as the [previously reported bug](https://github.com/mattermost/focalboard/issues/1125), but it only reproduces in HA environment

1. Install latest plugin build from main in an HA environment
2. Open boards
3. Click +Add Board and select a template

**Expected behavior:**
New board opens and appears in the sidebar without having to refresh

**Platform:**
latest main build on HA master

mgdelacroix:
#### Summary
This PR uses the plugin API to propagate the block changes to all the plugin nodes, making each node responsible of deciding if they need to distribute a given update to any of their subscribed websockets.

The implementation adds the `sendWorkspaceMessage` function with the intention of creating a baseline and add more functions if we need to direct messages to other entities in the future.

Fixes #1154


wiggin77 (approved):
LGTM :+1:

jespino (approved):
LGTM

PR #1218, opened 2021-09-15
Fix #1217. Save dialog changes before closing.

### LINKED ISSUE #1217: Bug: Clicking off card can lose content in FireFox
1. Create or edit a card; add text to the description
2. Click away from the card, meaning click somewhere else on the board
3. See card closes
4. Re-open card
5. See text added in step 1 is lost

**Expected behavior:**
Card content is saved when clicking away from a card

**Additional context:**
If user clicks another field before clicking away from the card, then content is saved correctly.

chenilim:
This resolves the timing issue in FireFox between the dialog background click and onBlur handing in MarkdownEditor.

Notes:
* When I added a breakpoint (in FireFox) to the onMouseDown handler in Dialog.tsx, the bug does not repro, even without this change
* Not sure which part of the onClose handler causes the update to abort.

### DIFF
--- a/webapp/src/components/dialog.tsx
+++ b/webapp/src/components/dialog.tsx
@@ -35,7 +35,9 @@ const Dialog = React.memo((props: Props) => {
                 className='wrapper'
                 onMouseDown={(e) => {
                     if (e.target === e.currentTarget) {
-                        props.onClose()
+                        setTimeout(() => {
+                            props.onClose()
+                        }, 0)
                     }
                 }}
             >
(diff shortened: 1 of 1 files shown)


jespino (approved):
I don't love this kind of solutions. I would add also a comment explaining that we want to give the opportunity of other components underneath to handle the click event before close.

wiggin77:
Why can't we refactor this to have deterministic ordering of the calls?

wiggin77 (approved):

wiggin77 (approved):

PR #1219, opened 2021-09-15
Fix #1215. Don't sidebar collapse on Personal Server.

### LINKED ISSUE #1215: Bug: its not always possible to hide the side bar
1. Create new board (for example "Meeting Notes" from template)
2. There is a horizontal scroll bar because not all kanban columns are visible
4. Its not possible to hide the side bar

**Expected behavior:**
It should be always possible to hide the side bar.

**Screenshots:**
Look for example at these kanban boards from default templates, they all have horizontal scroll:
BoardsScroll

Here is the threshold for width when the button to hide the side bar appears:
HideSideBarThreshold

**Platform:**
 - Browser and version: Chrome 92
 - OS and version: Windows 10
 - Laptop with **150% scale** in Windows settings.

**Additional context:**
This is a regression introduced in #1052
Reproducible in `v0.9.0`

@chenilim @asaadmahmood FYI

chenilim:
Updated the css selector to only hide the collapse-sidebar button when there is a workspace. This might need to be refactored in the future to be more robust.

### DIFF
--- a/webapp/src/components/sidebar/sidebar.scss
+++ b/webapp/src/components/sidebar/sidebar.scss
@@ -140,7 +140,7 @@
     }
 
     @media screen and (min-width: 768px) {
-        .sidebarSwitcher {
+        .WorkspaceTitle .sidebarSwitcher {
             display: none;
         }
     }


hahmadia (approved):

harshilsharma63 (approved):
LGTM

PR #1231, opened 2021-09-16
Start Welcome Page

### LINKED ISSUE #1302: Feature Idea: Implement Screen 1 of FTUE boards navigation
Implement Screen 1 of FTUE

Specifications can be viewed here https://docs.google.com/document/d/1Q0eujL89EU39Z3R1mer-tZMksEbyLtPGNqO-JsAe9Gg/edit?usp=sharing

hahmadia:
#### Summary

Screen 1 implementation

image

#### Ticket Link

Fixes #1302


webapp/src/pages/welcome/welcomePage.tsx
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import React from 'react'
+
+import './welcomePage.scss'
+
+const WelcomePage = React.memo(() => (
+    <div className='WelcomePage'>
+        <h1> TEST 123 123 123 </h1>
```
chenilim: Let's remove the `<h1>` element and test string - an empty div is ok.

Ideally, @asaadmahmood, could you make your changes in this PR, then merge it all at once? Or we should at least add a link to `/dashboard`, so users don't get stuck here, even if it's temporary.

chenilim (approved):
I'll approve this in case you guys want to do this in stages. But also, will there be a separate PR to actually implement the "show welcome page on first visit" logic?

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -140,15 +139,14 @@ export default class Plugin {
                 window.open(`${windowAny.frontendBaseURL}/workspace/${currentChannel}`)
             }
             this.channelHeaderButtonId = registry.registerChannelHeaderButtonAction(<FocalboardIcon/>, goToFocalboardWorkspace, '', 'Boards')
+            this.registry.registerProduct('/boards', 'product-boards', 'Boards', '/plug/focalboard/go-to-current-workspace', MainApp, HeaderComponent)
 
             this.registry.registerCustomRoute('go-to-current-workspace', () => {
-                const history = useHistory()
                 useEffect(() => {
-                    history.replace(`${windowAny.frontendBaseURL}/dashboard`)
+                    window.location.replace(`${windowAny.frontendBaseURL}/welcome`)
```
hahmadia, line 146: @jespino 

For some reason this kept redirecting me to a session expired when `lastWorkspaceViewed` in localstorage was deleted. I decided to use window.location.replace which works however it does a full refresh.

I'm not sure why. Hoping you can maybe have an input on this

jespino: Actually, all this code is no longer needed, this was a solution to redirect you to the current workspace, but now, you are no longer getting redirected to the current workspace, so... we only need to set the right url in the `registerProduct` call. Something like `/boards/welcome` instead of `/plug/focalboard/go-to-current-workspace`, and you can remove the entire `this.registry.registerCustomRoute` there.

webapp/src/app.tsx
```diff
@@ -149,6 +150,15 @@ const App = React.memo((): JSX.Element => {
                                             const loginUrl = `/login?r=${encodeURIComponent(redirectUrl)}`
                                             return <Redirect to={loginUrl}/>
                                         } else if (loggedIn === true) {
+                                            if (!localStorage.getItem('welcomePageViewed')) {
```
jespino, line 153: I don't love this. It looks like, if I go to a board/workspace/card/view... (using this route) without having that localStorage variable set, I get "something" that is not what i expect, and if I close that "welcome" I get redirected to the wrong place. I would expect something like a `<Redirect>` here to the `welcome` and even with a `?next=whatever` query paramter. Also, that should affect other urls, like `/board/:boardId?/...` and the `/:boardId?/...`

webapp/src/pages/welcome/welcomePage.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    if (localStorage.getItem('welcomePageViewed')) {
+        history.push('/dashboard')
+        return null
+    }
+
+    return (
+        <div className='WelcomePage'>
+            <div>
+                <h1 className='text-heading6'> {intl.formatMessage({id: 'WelcomePage.Heading', defaultMessage: 'Welcome To Boards'})} </h1>
```
jespino, line 39: Why are we using `intl.formatMessage` instead of `FormattedMessage` component?

webapp/src/pages/welcome/welcomePage.tsx
```diff
@@ -0,0 +1,64 @@
...
+                <div className='text-base'> {intl.formatMessage({id: 'WelcomePage.Description', defaultMessage: 'Boards is a project management tool that helps define, organize, track and manage work across teams, using a familiar kanban board view'})} </div>
+
+                <div className='WelcomePage__image'>
+                    <img
+                        src='../../../../boards-welcome.png'
```
jespino, line 44: This looks wrong, we should use something like webpack files import to solve this properly. I know now is not working fine, but this is not the solution, I think.

jespino (changes requested):
Some changes needed.

webapp/src/pages/welcome/welcomePage.tsx
hahmadia: Thanks! Done

webapp/src/app.tsx
hahmadia: Thanks. Done

mattermost-plugin/webapp/src/index.tsx
hahmadia: Thanks! Done.

webapp/src/pages/welcome/welcomePage.tsx
hahmadia: Fixed. Thanks

jespino (approved):
LGTM

webapp/src/app.tsx
```diff
@@ -162,10 +183,37 @@ const App = React.memo((): JSX.Element => {
...
+                                            return <Redirect to={`/welcome?${queryString}`}/>
+                                        }
+
+                                        if (loggedIn === true) {
+                                            return <BoardPage/>
+                                        }
+
+                                        return null
+                                    }}
+                                />
```
sbishel, line 216: In general, I think you should call a function for all of these cases. So that each of these routes does approximately the same thing. Something like  -

```
if (loggedIn === false) {
    return <Redirect to='/login'/>
}

if (sendtoWelcomePage()){
    redirectPage = {createRedirect} // may be a function or inline or returned from sendWelcomePage?
    return <Redirect to={`/welcome?r=${redirectPage}`}/>
}

// do whatever the default for the Route is.
//return <DashboardPage/>
return <BoardPage/>
```

sbishel (changes requested):
Some changes, but the big one is to re-work `webapp/src/app.tsx`. Let me know if you have questions.

webapp/src/app.tsx
```diff
@@ -131,24 +132,44 @@ const App = React.memo((): JSX.Element => {
...
-                                    {loggedIn === true && <BoardPage/>}
-                                </Route>
+                                <Route
+                                    path='/board/:boardId?/:viewId?/:cardId?'
+                                    render={({match}) => {
+                                        if (loggedIn === false) {
+                                            return <Redirect to='/login'/>
+                                        }
+
+                                        if (Utils.isFocalboardPlugin() && loggedIn === true && !localStorage.getItem('welcomePageViewed')) {
```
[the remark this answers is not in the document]
hahmadia: Necessary as discussed

webapp/src/app.tsx
```diff
@@ -162,10 +183,37 @@ const App = React.memo((): JSX.Element => {
...
+                                    path='/:boardId?/:viewId?/:cardId?'
+                                    render={({match}) => {
+                                        // Since these 3 path values are optional and they can be anything, we can pass /x/y/z and it will
+                                        // match this route however these values may not be valid so we should at the very least check
+                                        // board id for descisions made below
+                                        const boardIdIsValidUUIDV4 = (/^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/).test(match.params.boardId || '')
+
+                                        if (loggedIn === false) {
+                                            return <Redirect to='/login'/>
+                                        }
```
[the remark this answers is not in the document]
hahmadia: Necessary as discussed

webapp/src/app.tsx
```diff
@@ -162,10 +183,37 @@ const App = React.memo((): JSX.Element => {
...
+
+                                        if (Utils.isFocalboardPlugin() && loggedIn === true && (!localStorage.getItem('welcomePageViewed') || !boardIdIsValidUUIDV4)) {
+                                            const originalPath = `/${match.params.boardId || ''}/${match.params.viewId || ''}/${match.params.cardId || ''}`
+                                            const queryString = boardIdIsValidUUIDV4 ? `r=${originalPath}` : ''
+                                            return <Redirect to={`/welcome?${queryString}`}/>
+                                        }
+
+                                        if (loggedIn === true) {
```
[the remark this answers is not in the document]
hahmadia: Necessary as discussed

webapp/src/app.tsx
```diff
@@ -162,10 +183,37 @@ const App = React.memo((): JSX.Element => {
...
+                                        // Since these 3 path values are optional and they can be anything, we can pass /x/y/z and it will
+                                        // match this route however these values may not be valid so we should at the very least check
+                                        // board id for descisions made below
+                                        const boardIdIsValidUUIDV4 = (/^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/).test(match.params.boardId || '')
+
+                                        if (loggedIn === false) {
+                                            return <Redirect to='/login'/>
+                                        }
+
+                                        if (Utils.isFocalboardPlugin() && loggedIn === true && (!localStorage.getItem('welcomePageViewed') || !boardIdIsValidUUIDV4)) {
```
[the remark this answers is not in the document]
hahmadia: Necessary as discussed

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -142,7 +141,7 @@ export default class Plugin {
                 window.open(`${windowAny.frontendBaseURL}/workspace/${currentChannel}`)
             }
             this.channelHeaderButtonId = registry.registerChannelHeaderButtonAction(<FocalboardIcon/>, goToFocalboardWorkspace, '', 'Boards')
-            this.registry.registerProduct('/boards', 'product-boards', 'Boards', '/boards/dashboard', MainApp, HeaderComponent)
+            this.registry.registerProduct('/boards', 'product-boards', 'Boards', '/boards/welcome', MainApp, HeaderComponent)
```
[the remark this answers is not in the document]
hahmadia: The redirection to dashboard was causing extra problem cause as we discussed, this doesn't actually redirect to where it says it is so I left it. 

Sacrificing a extra redirection for the sake of not spending too much time figuring out router issue.

webapp/cypress/integration/createBoard.js
```diff
@@ -14,6 +14,7 @@ describe('Create and delete board / card', () => {
         localStorage.setItem('focalboardSessionId', 'TESTTOKEN');
         localStorage.setItem('language', 'en');
         cy.expect(localStorage.getItem('focalboardSessionId')).to.eq('TESTTOKEN');
+        localStorage.setItem('welcomePageViewed', 'true');
```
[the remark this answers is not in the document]
hahmadia: Sure done

webapp/src/app.tsx
hahmadia: Done

webapp/src/app.tsx
```diff
@@ -162,10 +183,37 @@ const App = React.memo((): JSX.Element => {
                                 >
                                     <DashboardPage/>
```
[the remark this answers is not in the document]
hahmadia: No we are redirecting to welcome and not dashboard due to the router problems discussed yesterday

hahmadia:
Since this affects User flow, I take this to be a major change since if broken, it can cause some problems.
Therefore, I think it's appropriate for QA to test this before merging rather than after.

sbishel (approved):

ogi-m (commented):
@hahmadia I'm seeing an issue - after dismissing the welcome page once, it will show up again after cache is cleared (or logging in on a different browser)

hahmadia:
> @hahmadia I'm seeing an issue - after dismissing the welcome page once, it will show up again after cache is cleared (or logging in on a different browser)

@ogi-m That is to be expected since we are using local storage. To use the database it was going to be a bit complicated so for now we decided to go with cache. 

This is intended behaviour.

ogi-m (approved):
Thanks @hahmadia ! Everything looks good then 👍🏻 
- User lands on welcome screen when accessing either from the global header or channel header
- Redirected as expected after clicking "Explore" in both cases
- Screen no longer opens after clicking "Explore" button (unless cache is cleared - expected)

PR #1245, opened 2021-09-18
Add /run/secrets to config paths

frederikaalund:
#### Summary
This makes it easier to use Docker secrets with Focalboard.
The change is backwards-compatible, since if there is no /run/secrets/config.json then
we use ./config.json as usual.

### DIFF
--- a/server/services/config/config.go
+++ b/server/services/config/config.go
@@ -60,9 +60,10 @@ type Configuration struct {
 
 // ReadConfigFile read the configuration from the filesystem.
 func ReadConfigFile() (*Configuration, error) {
-	viper.SetConfigName("config") // name of config file (without extension)
-	viper.SetConfigType("json")   // REQUIRED if the config file does not have the extension in the name
-	viper.AddConfigPath(".")      // optionally look for config in the working directory
+	viper.SetConfigName("config")       // name of config file (without extension)
+	viper.SetConfigType("json")         // REQUIRED if the config file does not have the extension in the name
+	viper.AddConfigPath("/run/secrets") // check secrets folder for config files first (e.g., for docker)
+	viper.AddConfigPath(".")            // optionally look for config in the working directory
 	viper.SetEnvPrefix("focalboard")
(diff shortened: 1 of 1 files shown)


jespino:
@frederikaalund I like the intention, but I don't love the solution. I would rather prefer some `--config` parameter that allows you to specify the config file to load (also I would accept that through environment variables). That would help others to integrate that with other system, not only docker, also linux distributions or other operating systems. Are you open to submit another PR with that approach?

frederikaalund:
> I like the intention, but I don't love the solution. I would rather prefer some --config parameter that allows you to specify the config file to load (also I would accept that through environment variables). That would help others to integrate that with other system, not only docker, also linux distributions or other operating systems.

I completely understand. Your proposal with a `--config` parameter and a `CONFIG_FILE` environment variable is the better approach. :+1:

> Are you open to submit another PR with that approach?

No, sorry. :) I don't have the bandwidth for that. This PR was all I could muster to quickly get Focalboard up and running in our self-hosted setup within the allotted time.

In any case, thanks for taking a look at this. I really appreciate the quick response. :+1:

jespino:
> > I like the intention, but I don't love the solution. I would rather prefer some --config parameter that allows you to specify the config file to load (also I would accept that through environment variables). That would help others to integrate that with other system, not only docker, also linux distributions or other operating systems.
> 
> I completely understand. Your proposal with a `--config` parameter and a `CONFIG_FILE` environment variable is the better approach. +1
> 
> > Are you open to submit another PR with that approach?
> 
> No, sorry. :) I don't have the bandwidth for that. This PR was all I could muster to quickly get Focalboard up and running in our self-hosted setup within the allotted time.

No problem, it is good feedback anyway, I'll create a new issue to track this and see if we can incorporate that sooner than later :)

> In any case, thanks for taking a look at this. I really appreciate the quick response. +1

Thank **you** for submitting it and taking the time to try to improve the project, I know how hard is to find the time to contribute and how easy is to "fix my own problem and keep going". So... I really appreciate your PR.

jespino:
@frederikaalund and here is the issue https://github.com/mattermost/focalboard/issues/1256

PR #1255, opened 2021-09-20
Fix elevation-1 variable

jespino:


jespino:
@ogi-m this should solve the problem of focalboard overriding the elevation-1 variable with a different value. It is used in mattermost in the :hover of the .post-image__column class, and in the `.post-menu` of the `:hover` post. I don't think testing this is really needed, but following the rule stablished about merging things in 0.9, I want your approval before merge.

asaadmahmood (approved):

ogi-m (approved):
Thanks @jespino, LGTM

PR #1257, opened 2021-09-20
Adds modd based targets that watch both server and webapp

mgdelacroix:
#### Summary
This is a QoL PR that adds `make` targets based on `modd` files that run both server and webapp in watch mode in the same target.


jespino (approved):
nice one

harshilsharma63 (approved):

PR #1296, opened 2021-09-21
Allow no workspace and send to the dashboard if that happens

harshilsharma63:
#### Summary
Allow no workspace and send to the dashboard if that happens

This resolves the issue of being displayed the no workspace error when opening the dashboard as a first-time boards user.


jespino (approved):
LGTM

Issue #1280, opened 2021-09-20
Bug: Updates from all workspaces are shown


chenilim:
**Steps to reproduce the behavior:**
1. Create two Channels in Mattermost, e.g. "channel 1" and "channel 2"
2. Open Boards from both, in two separate tabs
3. Create a board in "channel 1"
4. The board also appears in the "channel 2" tab / workspace

**Expected behavior:**
Updates only appear in the active workspace for the tab.

**Edition and Platform:**
 - Edition: Mattermost plugin
 - Version: v0.9.0
 - Browser and OS: Safari on macOS

**Additional context:**
The updated websocket backend code is broadcasting updates per user, across all clients. The front-end needs to filter out events that it (the particular instance) is not subscribed to, i.e. only handle updates to blocks in the current workspace.

chenilim:
This is a high-sev issue as it breaks the permission model.

PR #1299, opened 2021-09-21
Adds workspaceID to blocks and makes board page only reacts to the workspace updates

mgdelacroix:
#### Summary
This PR adds the `workspaceID` property to the blocks when they're retrieved from the store and makes the boards page react only to those updates that are aimed at the workspace that the user is currently in.

Fixes https://github.com/mattermost/focalboard/issues/1280


ogi-m (approved):
LGTM
- can't repro the issue on the PR build
- only the updates from the same workspace are shown

jespino (approved):
LGTM

sbishel (approved):

Issue #1270, opened 2021-09-20
Bug: Issue with WebSocket, not getting events after changing.


sbishel:
**Steps to reproduce the behavior:**
Seems if you get into the state, it reproduced everytime, but not sure how to get into that status.

1. Go to "New Card"
2. Click on Title
3. Change Title.
4. Close Card
5. Card does not show title changes.
6. Refresh, title changes now there.

Also, properties or any changes exhibit the same behavior

**Expected behavior:**
Changes should be saved and updated in the UI

**Screenshots:**

**Edition and Platform:**
 - Edition: Mattermost plugin
 - Version: v0.9.0 RC12
 - Browser and OS: [ 93.0.4577.82 (Official Build) (x86_64) on macO]

**Additional context:**
Hard to repo on demand. 

https://user-images.githubusercontent.com/12704875/134034209-d067d7b0-6592-430e-a987-edb15c9153df.mp4

chenilim:
@mgdelacroix, I ran a [dev build](https://github.com/mattermost/focalboard/actions/runs/1270631272) on the `release-0.9.0` branch, and observed the following issue:
1. Install and enable plugin on test server (`/cloud create SERVERNAME --version master`)
2. Switch to Boards via the App Switcher
3. Try to drag a card to another column
4. The card doesn't move (but does move after Cmd+R refresh)

However, this works if Boards is opened via the channel header button. In that case, both Boards tabs work.
1. Click on Boards button in channel header
2. Drag and drop cards - this works
3. In separate tab, open Boards again from the App Switcher
4. Drag and drop cards - this works now
5. Close the tab opened from the channel header
6. Drag and drop cards (in the tab opened from the App Switcher - this stops working

Video:
https://user-images.githubusercontent.com/46905241/134717143-15e39bf6-63d8-44ea-b876-72141f92f961.mov

Not sure if this will be fixed by the attached PR, but I believe this is the same issue. LMK if not, and we can create a separate issue, but this should be fixed before the next RC too.
cc @ogi-m, please also test these steps when verifying the fix. Thanks.

mgdelacroix:
@chenilim I've been investigating this a bit and it seems that is related to the URL schema and the subscription message. When navigating to a URL with the schema `/boards/{{boardId}}/{{viewId}}` the frontend sends a subscription message for the workspace id `0`. The updates are not received because the client never gets to subscribe properly.

[...]

PR #1324, opened 2021-09-23
Recovers inactive websockets connections on reconnect in plugin mode

mgdelacroix:
#### Summary
The reliable websockets infrastructure is storing the connection information on disconnect and recovering it on reconnect, so it doesn't notify the client nor provide of a mechanism of handling reconnections as its state is preserved. In the case of Focalboard, the plugin infrastructure notifies of the disconnection, the plugin reacts removing it from the server and removing its subscriptions and when the reconnection happens, no state has been preserved and the client has no way to react.

This PR adds the concept of `inactive` connections to the plugin adapter. When a disconnection happens, the plugin marks the connection as inactive and if the same `webConnID` reconnects, the state is correctly recovered and the subscriptions are still up, so the client doesn't need to do nothing. Inactive connections have a stale threshold of 5 minutes, following the server behaviour, and they're checked and cleaned every time a user connects with a new `webConnID` to avoid them staying in memory forever.

Fixes https://github.com/mattermost/focalboard/issues/1270


server/ws/testlib.go
```diff
@@ -0,0 +1,57 @@
+package ws
```
jespino, line 1: This file is going to be compile as part of the regular package instead of the test package. Unless you need to export this, please change it to a file helpers_test.go, or something like that.

wiggin77: Doesn't that also apply to everything in `services/store/storetests/`?  Also, why do we put store unit tests in a separate directory?

sbishel (approved):

ogi-m (approved):

jespino (approved):
LGTM

server/ws/plugin_adapter.go
```diff
@@ -103,9 +175,26 @@ func (pa *PluginAdapter) removeListener(pac *PluginAdapterClient) {
...
 	}
 
+	// user ID list
+	newUserListeners := []*PluginAdapterClient{}
+	for _, listener := range pa.listenersByUserID[pac.userID] {
+		if listener.webConnID != pac.webConnID {
+			newUserListeners = append(newUserListeners, listener)
+		}
+	}
+	pa.listenersByUserID[pac.userID] = newUserListeners
```
wiggin77, line 185: No locking needed here?

server/ws/plugin_adapter.go
```diff
@@ -115,13 +204,7 @@ func (pa *PluginAdapter) removeListenerFromWorkspace(pac *PluginAdapterClient, w
 	}
 	pa.listenersByWorkspace[workspaceID] = newWorkspaceListeners
```
wiggin77, line 205: I think the locking should go here instead of at the caller site.

wiggin77 (approved):
Excellent work :+1:

PR #1327, opened 2021-09-23
MM-38807 Fix tiny icons caused by css pollution

crspeller:
#### Summary
Fixing tiny icons caused by CSS pollution. 

Tiny icons on community:
image
Fixed version:
image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38807

### DIFF
--- a/mattermost-plugin/webapp/loaders/globalScssClassLoader.js
+++ b/mattermost-plugin/webapp/loaders/globalScssClassLoader.js
@@ -4,7 +4,7 @@
 function blockList(line) {
     return line.startsWith('.focalboard-body') ||
         line.startsWith('.GlobalHeaderComponent') ||
-        line.startsWith('.channel-header__icon .LogoIcon') ||
+        line.startsWith('.boards-rhs-icon') ||
         line.startsWith('.focalboard-plugin-root');
 }
 
--- a/webapp/src/widgets/icons/logo.tsx
+++ b/webapp/src/widgets/icons/logo.tsx
@@ -10,7 +10,7 @@ export default function LogoIcon(): JSX.Element {
     return (
         <CompassIcon
             icon='product-boards'
-            className='LogoIcon'
+            className='boards-rhs-icon'
         />
     )
 }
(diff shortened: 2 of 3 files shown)


mattermost-plugin/webapp/loaders/globalScssClassLoader.js
```diff
@@ -4,7 +4,7 @@
 function blockList(line) {
     return line.startsWith('.focalboard-body') ||
         line.startsWith('.GlobalHeaderComponent') ||
-        line.startsWith('.channel-header__icon .LogoIcon') ||
+        line.startsWith('.boards-rhs-icon') ||
```
crspeller, line 7: This wasted a bunch of my time trying to figure out what was going on with the CSS mysteriously vanishing. 
I think this would be much better handled by styled components or at least a liner.

crspeller (commented):
@jespino This probably should be included in the version you ship with 6.0. Not sure if you need a back-port or not.

crspeller:
Sorry. Not sure how @harshilsharma63 got assigned even though I already assigned @jespino

mattermost-plugin/webapp/loaders/globalScssClassLoader.js
jespino: Yes, this is to make everything scoped inside focalboard when we load that inside mattermost. It was error prone before because anything that you add could affect mattermost, this way, we only can break our side.

webapp/src/widgets/icons/logo.tsx
```diff
@@ -10,7 +10,7 @@ export default function LogoIcon(): JSX.Element {
     return (
         <CompassIcon
             icon='product-boards'
-            className='LogoIcon'
+            className='boards-rhs-icon'
```
jespino, line 13: Accordingly to our policy is always needed a class name with the component name, in this case `LogoIcon`, why do you need an specific class for this?

crspeller: @jespino Because it conflicts with the CSS in the webapp otherwise. It should be super specific here to not pollute.

cpoile (approved):

chenilim:
Trying to clean up stale PRs. @jespino, what is the next step here? Are changes still needed? Thanks.

webapp/src/widgets/icons/logo.tsx
jespino: I would prefer changing the component name a keeping the "class by component" rule. I can do it after merge, so, don't worry.

Issue #1278, opened 2021-09-20
Feature Idea: Confirm delete when deleting board


justinegeffen:
**Summary:**
Confirm that the user wants to delete the board prior to deletion.

**How important this is to me and why:**

Importance: High

Use cases:
"Delete" and "Duplicate" are close together and it's very easy to select "Delete" inadvertently. The board is deleted immediately and is not recoverable. If a user has a very detailed, long-standing board it means all the work is lost. 

**Additional context / similar features:**
1. Further down the line deleting items could also be permissions-based (e.g., only the board's team/owner can delete boards/items).
2. Here's an example of how Confluence does it:
<img width="674" alt="Screen Shot 2021-09-20 at 18 20 12" src="https://user-images.githubusercontent.com/4223530/134037735-e3cfd074-6ae4-4d46-87d2-57a9e482d9ce.png">

asaadmahmood:
Here's how we should do it:
https://www.figma.com/file/a3SJgKolniXEU2G4zG8Uc0/Issue-512-Deleting-a-property?node-id=214%3A3130

justinegeffen:
> Here's how we should do it:
> https://www.figma.com/file/a3SJgKolniXEU2G4zG8Uc0/Issue-512-Deleting-a-property?node-id=214%3A3130

Yep! That's perfect.

chenilim:
Thanks @justinegeffen and @asaadmahmood!

Putting this up for grabs. It's also related to #512 and #1140. We should use the same dialog code.

gtanczyk:
I'll take this one

sbishel:
closing as fixed

PR #1339, opened 2021-09-28
[GH-1278] Confirm delete when deleting board

gtanczyk:
#### Summary

Adding a confirmation dialog for delete board operation. 

Remarks:
- This dialog implementation might be later extracted and refactored for common usage in similar situations
- It is not perfectly aligned with the design, that's the current layout of dialog component, which is used elsewhere
- Error handling is missing - no info for the user in case of failure
- I could not find a way to bold a fragment of translated text
- Button component is missing some features (appearances, disabled state, loading state)

Mockup:
image

Current implementation:
image

#### Ticket Link

https://github.com/mattermost/focalboard/issues/1278


hahmadia:
@asaadmahmood 

Hey Asaad, 

I've tagged you in on this PR to get your input and see if we can help out @gtanczyk to get this to the finish line.

Thank you

webapp/src/components/sidebar/deleteBoardDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+                                try {
+                                    setSubmitting(true)
+                                    await props.onDelete()
+                                    setSubmitting(false)
+                                    props.onClose()
+                                } catch (e) {
+                                    setSubmitting(false)
+                                    Utils.logError(`Delete board ERROR: ${e}`)
+
+                                    // TODO: display error on screen
```
harshilsharma63, line 66: Do we want to address this?

harshilsharma63 (changes requested):

webapp/src/components/sidebar/deleteBoardDialog.tsx
gtanczyk: I'm not sure, I could not find existing component for displaying toast messages, or errors in general. If there is such implementation already existing, then I would use it, otherwise IMO it should addressed separately.

asaadmahmood:
This looks great @gtanczyk. I can have a look at improving the UI for this component and push a commit on this PR.

gtanczyk:
@asaadmahmood that would be great

asaadmahmood:
@gtanczyk Updated it.
Screenshot 2021-09-30 at 7 28 07 PM

hahmadia (changes requested):
Hello!

I decided to give your PR a go. I faced two major issues immediately 

1. The dialogue box looks like this for me
image

2. When I press `X`, it closes my entire tab (I'm running Focalboard as a plugin)

3. A minor issue but make sure you get the CI pipeline tests passing (Such as fixing all the ESlint errors). If you get stuck on any , you can always ask us to help you fix them.

hahmadia:
Oh by the way, once you've addressed our concerns, please don't forget to press the following button so that Github can notify us to re-review your PR!

image

webapp/src/components/sidebar/deleteBoardDialog.tsx
gtanczyk: @asaadmahmood Would you be able to add a variant to [this figma file](https://www.figma.com/file/a3SJgKolniXEU2G4zG8Uc0/Issue-512-Deleting-a-property?node-id=214%3A3130) for situation when delete operation causes an unexpected error?

gtanczyk:
@hahmadia This PR requires workflow approval after each of my commits, and I just pushed fixes for the problems you mentioned, thanks a lot for testing that.

webapp/src/widgets/buttons/button.scss
```diff
@@ -1,4 +1,6 @@
 .Button {
+    --danger-button-bg-rgb: 247, 67, 67;
```
gtanczyk, line 2: Previously this was added to shared variables, but it did not work well in mattermost with focalboard as a plugin.

hahmadia:
Hi @gtanczyk ,

Could we resolve the merge conflicts?

PR #1341, opened 2021-09-28
GH-1333: Implement Screen 3 and 4 of v6.0 navigation and FTUE 

### LINKED ISSUE #1333: Feature Idea: Implement Screen 3 and 4 of v6.0 navigation and FTUE
Implement The following two screens as outlined in the FTUE spec. See the Information about these screens under Screen 3 and Screen 4 heading in the following document

https://docs.google.com/document/d/1Q0eujL89EU39Z3R1mer-tZMksEbyLtPGNqO-JsAe9Gg/edit?usp=sharing

hahmadia:
Implement Screen 3 and 4 of v6.0 navigation and FTUE 

Fixes #1333 

Implementation details can be https://docs.google.com/document/d/1Q0eujL89EU39Z3R1mer-tZMksEbyLtPGNqO-JsAe9Gg/edit?usp=sharing


hahmadia:
Hi @asaadmahmood ,

I've done what I think to be the bulk of the CSS work required for this PR. I've just added you to double check my work and make any adjustments you feel are needed to be made. 

Thanks

webapp/src/components/emptyCenterPanel.tsx
```diff
@@ -1,34 +1,170 @@
...
+        <div
+            onClick={onClick}
+            className='button'
+        >
+            <i>{buttonIcon}</i>
+            <span className='button-title'>{title}</span>
+            {!readonly &&
+                <BoardTemplateMenuItemOptionsButton
+                    showBoard={showBoard!}
+                    boardTemplate={boardTemplate!}
```
sbishel, line 44: I find this to be very confusing. The "!" tells the compiler that this variable is NOT null or undefined, yet there is no other check to verify that and there are definitely instances these are undefined.  I'm actually surprised this works, I had to try it.  I guess the React simply doesn't render if a required property is null.

In any case, it may be better explicitly checking for those variables first. 

```
{!readonly && showBoard && boardTemplate &&
     <BoardTemplateMenuItemOptionsButton
     ...
```

webapp/src/components/sidebar/boardTemplateMenuItem.tsx
```diff
@@ -39,8 +32,49 @@ const addBoardFromTemplate = async (intl: IntlShape, showBoard: (id: string) =>
     }
 }
 
+type ButtonProps = {
+    showBoard: (id: string) => void
+    boardTemplate: Board
+}
+
+export const BoardTemplateMenuItemOptionsButton = React.memo((props: ButtonProps) => {
```
sbishel, line 40: Can this name be simplified. `BoardTemplateButtonMenu`

webapp/src/components/workspaceSwitcher/workspaceSwitcher.tsx
```diff
@@ -56,6 +57,14 @@ const WorkspaceSwitcher = (props: Props): JSX.Element => {
                     }}
                 />
             }
+            {activeWorkspace &&
+                <span
+                    className='add-workspace-icon'
+                    onClick={() => history.push(`/workspace/${activeWorkspace?.id}?showEmptyCenterPanel=true`)}
```
sbishel, line 63: I don't like using a query string for this. Can we just send it to the workspace? I realize we have code in `boardPage.tsx` that loads the lastBoard and lastView if not passed. I'm in favor of either removing that code or you could set `UserSettings.lastBoardId = null` here before setting the history.

However, I don't think the code in `boardPage.tsx` is necessary any more. I believe it was there because we didn't have another way to handle it...now we do.

webapp/src/components/emptyCenterPanel.tsx
```diff
@@ -1,34 +1,170 @@
...
+type ButtonProps = {
+    buttonIcon: string | React.ReactNode,
+    title: string,
+    readonly: boolean,
+    onClick: () => void,
+    showBoard?: (boardId: string) => void
+    boardTemplate?: Board
+}
+
+const EmptyCenterPanelButton = React.memo((props: ButtonProps) => {
```
sbishel, line 31: Can we just call this a `PanelButton`? There isn't anything requiring it to only be used in the `EmptyCenterPanel` is there?

sbishel (changes requested):
Hey Hossein. I would rather see the query string `showEmptyCenterPanel` not being used. As mentioned in the comment, I think there are a couple different ways we can do it and be able to remove the property that forces that action. Let me know if you have questions. Thanks.

webapp/src/components/emptyCenterPanel.tsx
```diff
@@ -1,34 +1,170 @@
...
+    let contentDisplay = (
+        <div className='Hint'>
+            <FormattedMessage
+                id='EmptyCenterPanel.no-content'
+                defaultMessage='Add or select a board from the sidebar to get started.'
+            />
+        </div>
+    )
+
+    if (Utils.isFocalboardPlugin()) {
```
harshilsharma63, line 85: It's not a nice practice to have the majority of the logic in conditional blocks.

This can be simplified as follows-

```javascript
    if (!Utils.isFocalboardPlugin()) {
        return (
            <div className='EmptyCenterPanel'>
                <div className='Hint'>
                    <FormattedMessage
                        id='EmptyCenterPanel.no-content'
                        defaultMessage='Add or select a board from the sidebar to get started.'
                    />
                </div>
            </div>
        )
    }

    return (...contents of this if block)
```

webapp/src/components/emptyCenterPanel.tsx
```diff
@@ -1,34 +1,170 @@
...
+                        defaultMessage='Create a Board in {workspaceName}'
+                        values={{workspaceName: workspace?.title}}
+                    />
+                </span>
+                <span className='description'>
+                    <FormattedMessage
+                        id='EmptyCenterPanel.plugin.no-content-description'
+                        defaultMessage='Add a board to the sidebar using any of the templates defined below or start from scratch.{lineBreak} Members of "{workspaceName}" will have access to boards created here.'
+                        values={{
+                            workspaceName: <b>{workspace?.title}</b>,
```
harshilsharma63, line 100: A template with no name shows up like this-

Screenshot 2021-10-04 at 1 06 51 PM

Can we add a default text of `Untitled Template` or something similar, same as what we do for boards-

Screenshot 2021-10-04 at 1 07 39 PM

webapp/src/components/sidebar/boardTemplateMenuItem.tsx
```diff
@@ -39,8 +32,49 @@ const addBoardFromTemplate = async (intl: IntlShape, showBoard: (id: string) =>
...
+                    onClick={() => {
+                        showBoard(boardTemplate.id || '')
+                    }}
+                />
+                <Menu.Text
+                    icon={<DeleteIcon/>}
+                    id='delete'
+                    name={intl.formatMessage({id: 'Sidebar.delete-template', defaultMessage: 'Delete'})}
+                    onClick={async () => {
+                        await mutator.deleteBlock(boardTemplate, 'delete board template')
```
harshilsharma63, line 61: Should there be a confirmation dialogue for this? I know there's another issue where someone implemented it. @asaadmahmood can confirm if we're adopting this design.

webapp/src/components/emptyCenterPanel.tsx
```diff
@@ -1,34 +1,170 @@
...
-            {workspace && workspace.id !== '0' &&
-            <div className='WorkspaceInfo'>
-                <FormattedMessage
-                    id='EmptyCenterPanel.workspace'
-                    defaultMessage='This is the workspace for:'
+        <div
+            onClick={onClick}
+            className='button'
+        >
+            <i>{buttonIcon}</i>
```
harshilsharma63, line 39: This is sheering the emojis-
Screenshot 2021-10-04 at 1 25 43 PM

We should use a span here.

harshilsharma63 (changes requested):

harshilsharma63:
Awesome feature @hahmadia . WOrks great overall and really love the UI. Some minor things I noticed-

1. The `New Template` button is a bit different from the Figma files. Is that intentional?
Screenshot 2021-10-04 at 10 58 38 AM

2. The workspace switcher select is not width-limited and is breaking for long channel names.
Screenshot 2021-10-04 at 11 00 48 AM
 
3. Long template names break the template button.
Screenshot 2021-10-04 at 11 02 48 AM

4. Some elements don't seem to be themed.
Screenshot 2021-10-04 at 11 07 32 AM

webapp/src/components/emptyCenterPanel.scss
```diff
@@ -11,4 +11,65 @@
...
+            margin-bottom: 12px;
+        }
+
+        .description {
+            text-align: center;
+            margin-bottom: 32px;
+            color: var(--sys-center-channel-color);
+        }
+
+        .bold-grey {
```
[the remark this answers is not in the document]
hahmadia: Oops! Thanks. Done

webapp/src/components/emptyCenterPanel.tsx
hahmadia: Sure! Thanks. Done

webapp/src/components/workspaceSwitcher/workspaceSwitcher.tsx
hahmadia: Nice! Good idea. Thanks. Done

webapp/src/components/emptyCenterPanel.scss
```diff
@@ -11,4 +11,65 @@
...
+            justify-content: center;
+            flex-wrap: wrap;
+        }
+
+        .button {
+            display: flex;
+            justify-items: center;
+            align-items: center;
+            padding: 16px 32px;
+            border: 1px solid rgba(61, 60, 64, 0.16);
```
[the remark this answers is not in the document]
hahmadia: Done! Thanks

webapp/src/components/emptyCenterPanel.tsx
hahmadia: Smart 💯 Thanks. Done

webapp/src/components/sidebar/boardTemplateMenuItem.tsx
hahmadia: If we want to do this, lets create a new ticket for it since it will further delay this ticket. 
I'll skip the implementation for the sake of getting this ticket out to the public sooner.

webapp/src/components/emptyCenterPanel.tsx
hahmadia: Oooooh Nice! Much better 😄

webapp/src/components/emptyCenterPanel.tsx
hahmadia: Done! Thanks

hahmadia:
@harshilsharma63 

Merci for the feedback! 💯 

1. Fixed. Sorry I forgot to do that.
2. Fixed
3. Fixed
4. Border looks better. I'll leave the other up to Asaad.

Issue #1360, opened 2021-09-28
Webapp Linter Fixes - fix test that wrap Elements with `(children: any)`


sbishel:
There are several files that use either
`wrapProviders = (children: any)` or `wrapIntl = (children: any)`.
	* Run `npx eslint --ext .tsx,.ts .` in the webapp folder to see them

These should be updated to use `ReactElement` instead

PR #1374, opened 2021-09-29
gh-1360 Fix for tests using 'any' when wrapping components

sbishel:
#### Summary
Fixes lint warnings for no-explicit-any in test files that wrap components

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/1360


webapp/src/components/cardDetail/cardDetailProperties.test.tsx
```diff
@@ -19,7 +19,7 @@ beforeEach(() => {
     FetchMock.fn.mockReset()
 })
 
-const wrapIntl = (children: any) => <IntlProvider locale='en'>{children}</IntlProvider>
+const wrapIntl = (children: ReactElement) => <IntlProvider locale='en'>{children}</IntlProvider>
```
kamre, line 22: Hi @sbishel!

This function is copy-pasted in many unit tests. Also there is [one inside testUtils.tsx](https://github.com/mattermost/focalboard/blob/main/webapp/src/testUtils.tsx#L6).

So may be it can be imported and used instead of copy-pasting.

sbishel: Thanks @kamre. Great suggestion. Done...

harshilsharma63 (approved):

Issue #1364, opened 2021-09-28
Webapp Linter Fixes - remove pgtools.d.ts


sbishel:
Instead of fixing linter issues. Remove this file after ensuring it is no longer necessary.
	* Run `npx eslint --ext .tsx,.ts .` in the webapp folder to see them

PR #1376, opened 2021-09-29
gh-1364 - remove file causing lint warnings

sbishel:
#### Summary
This pr removes the file `webapp/@custom_types/pgtools.d.ts`

I could not find any reference to this file anywhere. Not in package.json. Nor could I find any reference to the functions exported from this file. 

It is causing lint warnings, rather than attempt to fix, remove the file.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/1364


sbishel:
@chenilim, @jespino This file has been in the codebase for a while, but I couldn't find it being used anywhere. Please review and see if there is any reason to keep this file.

chenilim (approved):
I don't remember what this was for, but it looks like it was part of the initial checkin. My guess it was used by the initial DB code, which has since been refactored / removed.

jespino (approved):

Issue #1371, opened 2021-09-29
Bug: Access denied when changing views in a shared board


chenilim:
## Steps to reproduce the behavior

1. Deploy v0.9.1 RC2 and enable shared boards in the System Console
2. Share a board
3. Open a private browser window and open the shared url
4. Use the view picker to change views
5. Access denied error

## Expected behavior

No error, view is shown

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.9.1 RC2
 - Browser and OS: Chrome 93 on macOS

## Additional context

It appears that the read-only token (&r) is removed from the URL when switching views.

chenilim:
May be a dup of #1288?

PR #1382, opened 2021-09-30
gh-1371 - Add readtoken when in readonly mode

sbishel:
#### Summary
This PR fixes two issues. 

1. When we change views on shared boards, non validated users receive an Access Denied error.
This appears to work in Personal Server, however, because the token is not included in the query string, a refresh in Personal Server results in the same error.

Add the `?r=readtoken` parameter to the url path when either displaying a different view or a card view.

2. While testing, the card view was not being displayed in read-only view. 
I think this is due to the recent route changes. 

Add the `/cardId?` parameter to the "shared" routes.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/1371

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -158,7 +158,7 @@ const App = React.memo((): JSX.Element => {
                                 <Route path='/change_password'>
                                     <ChangePasswordPage/>
                                 </Route>
-                                <Route path='/shared/:boardId?/:viewId?'>
+                                <Route path='/shared/:boardId?/:viewId?/:cardId?'>
                                     <BoardPage readonly={true}/>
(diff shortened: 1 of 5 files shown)


sbishel:
@jespino, @harshilsharma63 - I know you guys have been doing some work on routes, so I wanted to make sure you see what is changing here.

harshilsharma63 (approved):

harshilsharma63:
@sbishel worked fine with 0.9 release branch.

ogi-m (approved):
LGTM
- able to change views as a non-validated user without error
- card view displayed as expected

jespino (approved):
LGTM

PR #1389, opened 2021-10-01
Live Watch make target for improve the development experience

jespino:


mattermost-plugin/webapp/src/index.tsx
```diff
@@ -141,7 +141,7 @@ export default class Plugin {
                 TelemetryClient.trackEvent(TelemetryCategory, TelemetryActions.ClickChannelHeader, {workspaceID: currentChannel})
                 window.open(`${windowAny.frontendBaseURL}/workspace/${currentChannel}`, '_blank', 'noopener')
             }
-            this.channelHeaderButtonId = registry.registerChannelHeaderButtonAction(<FocalboardIcon/>, goToFocalboardWorkspace, '', 'Boards')
+            this.channelHeaderButtonId = registry.registerChannelHeaderButtonAction(<FocalboardIcon/>, goToFocalboardWorkspace, '', 'Boards again')
```
mgdelacroix, line 144: :eyes:

mattermost-plugin/modd.conf
```diff
@@ -0,0 +1,3 @@
+server/**/*.go !server/**/*_test.go mattermost-plugin/server/**/*.go !mattermost-plugin/server/**/*_test.go {
+    prep: cd mattermost-plugin; make server; make deploy-to-mattermost-directory
```
mgdelacroix, line 2: nit:
```suggestion
    prep: cd mattermost-plugin; make server deploy-to-mattermost-directory
```

mattermost-plugin/Makefile
```diff
@@ -277,6 +285,36 @@ ifndef STARTERTEMPLATE_PATH
 endif
 	cd ${STARTERTEMPLATE_PATH} && go run ./build/sync/main.go ./build/sync/plan.yml $(PWD)
 
+## Watch webapp and server changes and redeploy locally using local filesystem (MM_SERVER_PATH)
+.PHONY: live-watch
+live-watch:
```
mgdelacroix, line 290: If this target is only planning to affect and act over plugins, I'd make reference to it both in the docline and in the name, maybe changing it to something like `make live-watch-plugin`

Why not adding this target to the root's `Makefile`? This way we could run all the watch targets from the root of the repo

mgdelacroix (commented):
Amazing job!! So eager to have this in! :rocket:

mattermost-plugin/Makefile
jespino: Sounds sensible, I was trying to keep that there because was the plugin part, I think we can just make a new target on the root makefile that delegates the work to this.

mattermost-plugin/webapp/src/index.tsx
jespino: Oops, a testing thing... I'll remove it.

mattermost-plugin/Makefile
jespino: Finally I just exposed that in the Makefile

mgdelacroix (approved):

wiggin77 (approved):
LGTM :+1:

harshilsharma63 (approved):

Issue #1288, opened 2021-09-20
Bug: Websocket connection fails on shared boards


chenilim:
**Steps to reproduce the behavior:**
1. Open a board
2. Click on the board (...) menu, and select Share Board
3. Enable shared board, and copy link
4. Open a private window and paste link
5. The board appears, but the websocket connection error banner appears after a few seconds
6. Open the developer console
7. See errors:
```
main.js?f159a7f4065888ba6939:389 WebSocket connection to 'wss://boards-bugbash.test.mattermost.cloud/plugins/focalboard/ws' failed: 
open @ main.js?f159a7f4065888ba6939:389
(anonymous) @ main.js?f159a7f4065888ba6939:389
main.js?f159a7f4065888ba6939:389 [1632161460.05] WSClient websocket onerror. data: [object Event]
logError @ main.js?f159a7f4065888ba6939:389
a.onerror @ main.js?f159a7f4065888ba6939:389
error (async)
open @ main.js?f159a7f4065888ba6939:389
(anonymous) @ main.js?f159a7f4065888ba6939:389
setTimeout (async)
a.onclose @ main.js?f159a7f4065888ba6939:389
main.js?f159a7f4065888ba6939:389 [1632161460.05] WSClient websocket onclose, code: 1006, reason: 
main.js?f159a7f4065888ba6939:389 [1632161460.05] Unexpected close, re-opening websocket
logError @ main.js?f159a7f4065888ba6939:389
a.onclose @ main.js?f159a7f4065888ba6939:389
main.js?f159a7f4065888ba6939:389
```

**Expected behavior:**
No errors

**Screenshots:**
<img width="709" alt="Websocket errors" src="https://user-images.githubusercontent.com/46905241/134053037-1a216f4f-c0ff-4eb8-99da-859409ec056d.png">

**Edition and Platform:**
 - Edition: Mattermost plugin
 - Version: v0.9.0
 - Browser and OS: Chrome 91 on macOS

**Additional context:**
May be related to #1279

chenilim:
@mgdelacroix, this looks related to the other websocket issues. Please investigate and reassign as necessary. Thanks. cc @sbishel, @ogi-m.

PR #1391, opened 2021-10-01
Disable websockets in shared boards

mgdelacroix:
#### Summary
This PR disables the websockets for all legacy plugin routes, including shared boards in plugin mode.

Fixes https://github.com/mattermost/focalboard/issues/1288


webapp/src/app.tsx
```diff
@@ -80,19 +80,24 @@ const App = React.memo((): JSX.Element => {
     const globalError = useAppSelector<string>(getGlobalError)
     const me = useAppSelector<IUser|null>(getMe)
     const dispatch = useAppDispatch()
+    // this is used to detect if we're running inside the plugin but
```
jespino, line 83: Can we explain here that this is a temporary solution until we find a better way to provide annonymous/not-websocket connected routes?

ogi-m (approved):
LGTM 
- not seeing the websocket error after several minutes or when changing views
- no errors in the console

harshilsharma63 (approved):

jespino (approved):

Issue #1395, opened 2021-10-01
Feature idea: Increase size for content on Splash screen


asaadmahmood:
Eric raised an issue regarding the content on the splash screen being quite small. We need to update it to be a big larger.

PR #1399, opened 2021-10-01
Updating splash screen sizing for image and content

asaadmahmood:
#### Summary
Updating splash screen sizing for image and content

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/1395

#### Screensho
localhost_8065_boards_welcome(1440p)
t


esethna:
Thanks Asaad, not sure how I can test this so I won't block it on my approval

esethna (approved):

asaadmahmood:
@hahmadia one problem I encounter locally is that webpack does not generate the images automatically in the pack folder. Please check that out and ensure that webpack will generate those images.

wiggin77:
@asaadmahmood there is a linter error:
```
src/pages/welcome/welcomePage.scss
 31:9  ✖  Unexpected longhand value '24px auto 32px auto' instead of '24px auto 32px'   shorthand-property-no-redundant-values
```

webapp/src/pages/welcome/welcomePage.tsx
```diff
@@ -35,25 +36,30 @@ const WelcomePage = React.memo(() => {
...
+                    src={BoardWelcomePNG}
+                    className='WelcomePage__image WelcomePage__image--large'
+                    alt='Boards Welcome Image'
+                />
+
+                <img
+                    src={BoardWelcomeSmallPNG}
+                    className='WelcomePage__image WelcomePage__image--small'
+                    alt='Boards Welcome Image'
+                />
```
hahmadia, line 62: Hey Asaad, 

I'm okay with the work done however I think we are unnecessarily rendering 2 components and hiding one when we could render one or the other conditionally based on Screen size. 

Do you think we can do something like here https://blog.logrocket.com/developing-responsive-layouts-with-react-hooks/ to render one or the other? That way it isn't confusing when inspecting the elements and noticing both are rendered but one is simply hidden.

Thoughts?

hahmadia (changes requested):
One minor question

webapp/src/pages/welcome/welcomePage.tsx
asaadmahmood: @hahmadia Not sure I want to add a resize handler for something so simple. We would be running javascript code on resize which I'm not sure is necessary.

But I'm fine if you guys think adding the resize handler and reloading the images as the user resizes the screen is a better option for this.

hahmadia: Nah I agree I think you're right. Rethinking it, it's pretty simple and adding more code could possibly complicate it more than it needs too. Can you just add a comment for each component to specify when it will render?

hahmadia (approved):
LGTM! Could you just add a comment to clarify when each image tag shows? Thanks and appreciate the clarification.

wiggin77 (approved):
LGTM :+1:

PR #1400, opened 2021-10-01
[GH-1331] Fix for URL property on kanban card

### LINKED ISSUE #1331: Bug: URL property is not visible on kanban cards
1. Create new board
2. Create new card
3. Add URL property to the card and close card dialog
4. Make URL property visible on the cards via _Properties_ menu
5. URL property is not visible on the kanban card

**Expected behavior:**
URL property should be visible on the kanban card

**Screenshots:**
URL property in the card dialog:
image

URL property on the kanban card:
image

**Edition and Platform:**
 - Edition: Personal Server
 - Version: v0.9.1
 - Browser and OS: Chrome 93 on Windows

**Additional context:**
Regression was introduced in #1160

kamre:
#### Summary
Fix displaying properties on kanban cards:
  - hide URL property only when it is empty
  - hide empty labels for Select/Multi Select properties

Here is how it looks (non-empty URL property is visible, empty property values are collapsed):
image

Also minor tweak for proper alignment of Select/Multi Select empty labels:
image

#### Ticket Link
Fixes #1331

### DIFF
--- a/webapp/src/components/kanban/kanbanCard.scss
+++ b/webapp/src/components/kanban/kanbanCard.scss
@@ -41,12 +41,14 @@
     }
 
     .octo-propertyvalue {
-        margin: 4px 0 0;
-        font-size: 12px;
-        line-height: 18px;
-
-        &:empty {
+        input[value=''] {
             display: none;
+        }
+
+        .Label.empty {
+            display: block;
+            padding: 0;
+            margin: 0;
         }
     }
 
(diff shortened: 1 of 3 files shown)


webapp/src/components/kanban/kanbanCard.scss
```diff
@@ -41,13 +41,15 @@
     }
 
     .octo-propertyvalue {
-        margin: 4px 0 0;
-        font-size: 12px;
-        line-height: 18px;
-
-        &:empty {
+        input[value=''] {
```
kamre, line 44: May be we should not use `input` for readonly rendering.

harshilsharma63 (approved):

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @kamre

PR #1408, opened 2021-10-02
[GH-705] Auto-name new property

### LINKED ISSUE #705: Feature Idea: Auto-name new properties based on type
**Summary:**
When adding a new property to a card, set the name automatically based on the property type, if it hasn't already been named.

Importance: Low

Use cases:
1. E.g. If I add a new Phone property, name the property "Phone" if not already set.

kamre:
#### Summary
Changed workflow for adding new property:
- menu with all property types is shown and allows the user to select new property type
image
- after selecting the type for new property its name is initialized from its type:
image
- pressing `Enter` closes this popup

#### Ticket Link
Fixes #705

@asaadmahmood please check this workflow from the UX perspective

### DIFF
--- a/webapp/src/components/cardDetail/__snapshots__/cardDetailProperties.test.tsx.snap
+++ b/webapp/src/components/cardDetail/__snapshots__/cardDetailProperties.test.tsx.snap
@@ -64,7 +64,7 @@ exports[`components/cardDetail/CardDetailProperties should match snapshot 1`] = 
 </div>
 `;
 
-exports[`components/cardDetail/CardDetailProperties shows menu with property types when adding new property 1`] = `
+exports[`components/cardDetail/CardDetailProperties should show property types menu 1`] = `
 <div>
   <div
     class="octo-propertylist CardDetailProperties"
--- a/webapp/src/components/cardDetail/__snapshots__/cardDetailProperties.test.tsx.snap
+++ b/webapp/src/components/cardDetail/__snapshots__/cardDetailProperties.test.tsx.snap
@@ -63,3 +63,361 @@ exports[`components/cardDetail/CardDetailProperties should match snapshot 1`] = 
   </div>
 </div>
 `;
+
+exports[`components/cardDetail/CardDetailProperties shows menu with property types when adding new property 1`] = `
+<div>
+  <div
+    class="octo-propertylist CardDetailProperties"
+  >
+    <div
+      class="octo-propertyrow"
+    >
+      <div
+        aria-label="menuwrapper"
+        class="MenuWrapper"
+        role="button"
+      >
+        <div
+          class="octo-propertyname"
+        >
+          <button
+            class="Button    "
+            type="button"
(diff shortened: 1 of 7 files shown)


asaadmahmood (commented):
Not sure if this is necessary? The name is already populated by default right? When a user clicks to create a new property?

Are we proposing we change that to something when the user selects a type.

It would be possible that the user has already named the property before selecting a new type, and we shouldn't replace that.

And I understand we can have a check that checks if the property has been renamed, but then if we change it, it may be confusing that the behaviour changes, and some uesrs may still expect it to change after they change the property type.

kamre:
> The name is already populated by default right? When a user clicks to create a new property?

By default the name is just always "New Property" and it looks like this:
AddNewPropertyDefault
So even to rename this default name it is necessary to click again.

> Are we proposing we change that to something when the user selects a type.
> 
> It would be possible that the user has already named the property before selecting a new type, and we shouldn't replace that.

After new property is added the workflow is the same as before and we never change property name when user selects new type. Only creation of new property type was changed and it requires to select the type beforehand.

asaadmahmood:
Okay so by default @kamre we would not populate the property and ask them to select the type everytime they want to add a new property?

kamre:
Yes, exactly. And if the user doesn't select the property type (clicks outside menu for example) then new property will not be even created.

jespino:
@asaadmahmood whenever you can, confirm that this is the expected UX to proceed to review the code. I like the current approach, specially for properties that are not added frequently, so... it is a bit more uncomfortable for adding a text field, but is better for everything else.

asaadmahmood:
@jespino I'm not sure about it, I would have to review this in UX, and see what people think. I may also create a test for two and see what users prefer.

hahmadia:
Hi @asaadmahmood , after you've done your research and conducted if this is the way to go, feel free to tag any 2 developers to proceed with code reviews and testing. 

Thanks

asaadmahmood:
@hahmadia Sure.

asaadmahmood:
@kamre We can take this.

asaadmahmood (approved):

kamre:
@jespino I have updated the unit test for deleting existing property because now it shows the confirmation dialog.

kamre:
@ogi-m This is strange because I can't reproduce the issue with positioning of "Add a property" button in my branch for standalone server or windows app.

For example this is how it looks for windows app downloaded from [build artifacts](https://github.com/mattermost/focalboard/actions/runs/1349651603):
image

kamre:
@ogi-m I have registered on the test server. But the version of focalboard there seems to be far from `main` branch and changes from my branch are definitely not there.

Am I doing something wrong?

I will try to setup development environment locally so that I can check the plugin locally.

ogi-m:
@kamre Looks like the plugin got overwritten - I've reuploaded the plugin from the PR and I've made you sysadmin so that you can upload it.

PR #1415, opened 2021-10-03
Typing adjustment - 1353

### LINKED ISSUE #1353: Webapp Linter Fixes - fix linter warnings for file `sortable.tsx`
Fix linter warnings for ...webapp/src/hooks/sortable.tsx
	* Run `npx eslint --ext .tsx,.ts .` in the webapp folder to see them

tiago154:
#### Summary

Hello.

Adjusted the warnings in the file
`webapp/src/hooks/sortable.tsx`

#### Ticket Link

Fixes #1353

### DIFF
--- a/webapp/src/hooks/sortable.tsx
+++ b/webapp/src/hooks/sortable.tsx
@@ -3,7 +3,7 @@
 import React, {useRef} from 'react'
 import {useDrag, useDrop, DragElementWrapper, DragSourceOptions, DragPreviewOptions} from 'react-dnd'
 
-function useSortableBase(itemType: string, item: any, enabled: boolean, handler: (src: any, st: any) => void): [boolean, boolean, DragElementWrapper<DragSourceOptions>, DragElementWrapper<any>, DragElementWrapper<DragPreviewOptions>] {
+function useSortableBase(itemType: string, item: unknown, enabled: boolean, handler: (src: unknown, st: unknown) => void): [boolean, boolean, DragElementWrapper<DragSourceOptions>, DragElementWrapper<unknown>, DragElementWrapper<DragPreviewOptions>] {
     const [{isDragging}, drag, preview] = useDrag(() => ({
         type: itemType,
         item,
@@ -17,7 +17,7 @@ function useSortableBase(itemType: string, item: any, enabled: boolean, handler:
         collect: (monitor) => ({
             isOver: monitor.isOver(),
         }),
-        drop: (dragItem: any) => {
+        drop: (dragItem) => {
             handler(dragItem, item)
         },
         canDrop: () => enabled,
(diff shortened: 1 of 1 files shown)


webapp/src/hooks/sortable.tsx
```diff
@@ -17,23 +30,23 @@ function useSortableBase(itemType: string, item: any, enabled: boolean, handler:
...
+            handler(dragItem, item as HandlerBaseItem)
         },
         canDrop: () => enabled,
     }), [item, handler, enabled])
 
     return [isDragging, isOver, drag, drop, preview]
 }
 
-export function useSortable(itemType: string, item: any, enabled: boolean, handler: (src: any, st: any) => void): [boolean, boolean, React.RefObject<HTMLDivElement>] {
+export function useSortable(itemType: string, item: SortableItem, enabled: boolean, handler: (src: HandlerItem, st: HandlerItem) => void): [boolean, boolean, React.RefObject<HTMLDivElement>] {
```
hahmadia, line 42: We can Simplify this by using Generics. Something like the following

```suggestion
export function useSortable<T>(itemType: string, item: T, enabled: boolean, handler: (src: T, st: T) => void): [boolean, boolean, React.RefObject<HTMLDivElement>] {
```

webapp/src/hooks/sortable.tsx
```diff
@@ -3,7 +3,20 @@
...
+
+type SortableItem = Card | IPropertyOption | IPropertyTemplate
+type HandlerItem = Card & IPropertyTemplate & IPropertyOption
+type HandlerBaseItem = HandlerItem & IContentBlockWithCords
+interface ISortableWithGripItem {
+    block: Block | Block[],
+    cords: {x: number, y?: number, z?: number}
+}
+
+function useSortableBase(itemType: string, item: SortableItem | ISortableWithGripItem, enabled: boolean, handler: (src: HandlerBaseItem, st: HandlerBaseItem) => void): [boolean, boolean, DragElementWrapper<DragSourceOptions>, DragElementWrapper<DragSourceOptions>, DragElementWrapper<DragPreviewOptions>] {
```
hahmadia, line 19:
```suggestion
function useSortableBase<T>(itemType: string, item: T, enabled: boolean, handler: (src: T, st: T) => void): [boolean, boolean, DragElementWrapper<DragSourceOptions>, DragElementWrapper<DragSourceOptions>, DragElementWrapper<DragPreviewOptions>] {
```

webapp/src/hooks/sortable.tsx
```diff
@@ -17,23 +30,23 @@ function useSortableBase(itemType: string, item: any, enabled: boolean, handler:
         collect: (monitor) => ({
             isOver: monitor.isOver(),
         }),
-        drop: (dragItem: any) => {
-            handler(dragItem, item)
+        drop: (dragItem: HandlerBaseItem) => {
+            handler(dragItem, item as HandlerBaseItem)
```
hahmadia, line 34:
```suggestion
        drop: (dragItem: T) => {
            handler(dragItem, item)
```

webapp/src/hooks/sortable.tsx
```diff
@@ -17,23 +30,23 @@ function useSortableBase(itemType: string, item: any, enabled: boolean, handler:
...
-export function useSortable(itemType: string, item: any, enabled: boolean, handler: (src: any, st: any) => void): [boolean, boolean, React.RefObject<HTMLDivElement>] {
+export function useSortable(itemType: string, item: SortableItem, enabled: boolean, handler: (src: HandlerItem, st: HandlerItem) => void): [boolean, boolean, React.RefObject<HTMLDivElement>] {
     const ref = useRef<HTMLDivElement>(null)
     const [isDragging, isOver, drag, drop] = useSortableBase(itemType, item, enabled, handler)
     drop(drag(ref))
     return [isDragging, isOver, ref]
 }
 
-export function useSortableWithGrip(itemType: string, item: any, enabled: boolean, handler: (src: any, st: any) => void): [boolean, boolean, React.RefObject<HTMLDivElement>, React.RefObject<HTMLDivElement>] {
+export function useSortableWithGrip(itemType: string, item: ISortableWithGripItem, enabled: boolean, handler: (src: IContentBlockWithCords, st: IContentBlockWithCords) => void): [boolean, boolean, React.RefObject<HTMLDivElement>, React.RefObject<HTMLDivElement>] {
```
hahmadia, line 49: I'm okay with leavin this the way it is.

hahmadia (changes requested):
Hello! Thank you for the changes!

I noticed that some of the work you've done can be simplified even further by the use of Generic types. That way, we don't have to restrict the types if we plan to use these function  for other types in the future. I'm okay with leaving `useSortableWithGrip` as it is but `useSortable` should be made generic where the `handler` function should handle the `item` of type `T` instead of having a type that unions all of them (In our case that doesn't make a lot of sense since we are never passing a type that has the properties of Card and IPropertyTemplate and so on and so forth). 

Hopefully this makes sense and let me know if you need any assistance.

tiago154:
> Hello! Thank you for the changes!
> 
> I noticed that some of the work you've done can be simplified even further by the use of Generic types. That way, we don't have to restrict the types if we plan to use these function for other types in the future. I'm okay with leaving `useSortableWithGrip` as it is but `useSortable` should be made generic where the `handler` function should handle the `item` of type `T` instead of having a type that unions all of them (In our case that doesn't make a lot of sense since we are never passing a type that has the properties of Card and IPropertyTemplate and so on and so forth).
> 
> Hopefully this makes sense and let me know if you need any assistance.

@hahmadia 

It makes perfect sense what you explained.
I made the changes for using Generics. Please confirm that everything is all right now.
Thanks in advance.

Issue #53, opened 2021-02-23
Extend import scripts


chenilim:
Right now, we have [scripts to import basic info](https://github.com/mattermost/focalboard/tree/main/import) from popular systems like Trello, Asana, and Notion. They can pull in the card titles, descriptions, and basic properties from a single board. Expand these further to do things like:
* Import advanced properties (e.g. dates, text, numbers)
* Handle multiple boards
* Handle Trello PowerUp properties
* Handle sub-tasks in Asana
* Use the [Notion API](https://www.notion.so/api-beta) to import rich data
* Import from other popular systems

This high-level issue can be split into smaller ones to handle targeted improvements

chenilim:
If there's another system you would like to import from, please [create a feature request](https://github.com/mattermost/focalboard/issues/new/choose) or thumbs-up an existing one.

We're also open to accepting PRs for additional importers. Please create a feature request first to track it, then use the existing ones as a template. I recommend starting with importing basic info (e.g. board and card titles, properties), then layer in additional details.

harshilsharma63:
@chenilim I'm working on adding a Todoist import script. The initial version is working well. I need to test it out more and polish out some corners. Will create a ticket and PR soon.

chenilim:
Thanks @harshilsharma63! That is awesome! LMK if you run into any issues.

harshilsharma63:
@chenilim here's the PR for Todoist - https://github.com/mattermost/focalboard/pull/285

NoLooseEnds:
This would also be nice to be able to import from Trello:
* Attachments (primarily images in my use case) (see https://bendellar.com/download-all-attachments-from-trello-card-as-a-zip-file/)
* Card Labels

chenilim:
Hi @NoLooseEnds, check out the [Trello import script here](https://github.com/mattermost/focalboard/tree/main/import/trello). It's pretty basic right now, and can use improvement.

NoLooseEnds:
@chenilim Just imported a couple of boards and have painstakingly added images to all my cards

That was what caused me to post the comment above. :)

chenilim:
Ah, sorry, I confused this with the thread for the Todoist import! :) Yes, we want to improve this, and it might be doable with the newly released [Notion API](https://developers.notion.com/).

PR #1430, opened 2021-10-04
Add the ability to import checklists from Trello

leosunmo:
#### Summary
Adds the ability to import checklists to Focalboard when importing from Trello. It's pretty simple code, but it's my first attempt at TypeScript, so appreciate any feedback. I'm not 100% sure of my `undefined` logic etc. But it does work :shrug: 

This is kind of related to https://github.com/mattermost/focalboard/issues/53 but there's no specific issue relating to this. I wrote it for my own needs and figured I should push it upstream.

I started looking at importing images as well, but they require auth now if the Trello board is private. That is a different kettle of fish, so decided to leave that out of this PR.

### DIFF
--- a/import/trello/importTrello.ts
+++ b/import/trello/importTrello.ts
@@ -9,8 +9,8 @@ import {IPropertyOption, IPropertyTemplate, createBoard} from '../../webapp/src/
 import {createBoardView} from '../../webapp/src/blocks/boardView'
 import {createCard} from '../../webapp/src/blocks/card'
 import {createTextBlock} from '../../webapp/src/blocks/textBlock'
-import {createCheckboxBlock, CheckboxBlock} from '../../webapp/src/blocks/checkboxBlock'
-import {ChecklistElement, Trello} from './trello'
+import {createCheckboxBlock} from '../../webapp/src/blocks/checkboxBlock'
+import {Trello} from './trello'
 import {Utils} from './utils'
 
 // HACKHACK: To allow Utils.CreateGuid to work
@@ -138,14 +138,14 @@ function convert(input: Trello): Block[] {
         }
 
         // Add Checklists
-        if (card.idChecklists && card.idChecklists.length) {
+        if (card.idChecklists && card.idChecklists.length > 0) {
             card.idChecklists.forEach(checklistID => {
                 const lookup = input.checklists.find(e => e.id === checklistID)
-                if (lookup !== undefined) {
+                if (lookup) {
                     lookup.checkItems.forEach(trelloCheckBox=> {
                         const checkBlock = createCheckboxBlock()
                         checkBlock.title = trelloCheckBox.name
-                        if (trelloCheckBox.state == 'complete') {
+                        if (trelloCheckBox.state === 'complete') {
                             checkBlock.fields.value = true
                         } else {
                             checkBlock.fields.value = false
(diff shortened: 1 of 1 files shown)


import/trello/importTrello.ts
```diff
@@ -135,6 +136,29 @@ function convert(input: Trello): Block[] {
...
+
+        // Add Checklists
+        if (card.idChecklists && card.idChecklists.length) {
+            card.idChecklists.forEach(checklistID => {
+                const lookup = input.checklists.find(e => e.id === checklistID)
+                if (lookup !== undefined) {
+                    lookup.checkItems.forEach(trelloCheckBox=> {
+                        const checkBlock = createCheckboxBlock()
+                        checkBlock.title = trelloCheckBox.name
+                        if (trelloCheckBox.state == 'complete') {
```
chenilim, line 148: The eslint setup for the import scripts may be out of date from the webapp, but in general we should use ===, i.e. `if (trelloCheckBox.state === 'complete') {`.

import/trello/importTrello.ts
```diff
@@ -135,6 +136,29 @@ function convert(input: Trello): Block[] {
 
             outCard.fields.contentOrder = [text.id]
         }
+
+        // Add Checklists
+        if (card.idChecklists && card.idChecklists.length) {
+            card.idChecklists.forEach(checklistID => {
+                const lookup = input.checklists.find(e => e.id === checklistID)
+                if (lookup !== undefined) {
```
chenilim, line 144: In general, we can check if a variable is "tuthy" i.e. `if (lookup) {`, which will handle undefined, null, and other cases that can lead to a null reference exception.

import/trello/importTrello.ts
```diff
@@ -135,6 +136,29 @@ function convert(input: Trello): Block[] {
 
             outCard.fields.contentOrder = [text.id]
         }
+
+        // Add Checklists
+        if (card.idChecklists && card.idChecklists.length) {
```
chenilim, line 141: This is minor, and may be stylistic, but in general lengths should be checked as > 0, i.e. ` if (card.idChecklists && card.idChecklists.length > 0) {`.

import/trello/importTrello.ts
```diff
@@ -9,7 +9,8 @@ import {IPropertyOption, IPropertyTemplate, createBoard} from '../../webapp/src/
 import {createBoardView} from '../../webapp/src/blocks/boardView'
 import {createCard} from '../../webapp/src/blocks/card'
 import {createTextBlock} from '../../webapp/src/blocks/textBlock'
-import {Trello} from './trello'
+import {createCheckboxBlock, CheckboxBlock} from '../../webapp/src/blocks/checkboxBlock'
```
chenilim, line 12: ESlint called out that `CheckboxBlock` and `ChecklistElement` aren't used, and should not be imported.

chenilim (changes requested):
Thanks @leosunmo for your contribution! This is a great addition to the import script! Logic looks good, just a few minor linter errors - the import scripts aren't part of the main CI / lint check.

import/trello/importTrello.ts
leosunmo: That was another thing I was wondering about. I originally used `> 0` (habit from writing Go) but kept seeing people doing what I ended up doing :shrug: so I wasn't sure if there was some hidden syntactic thing or just a style thing. I definitely prefer the clarity of `> 0`.

import/trello/importTrello.ts
leosunmo: Yeah my editor says the same, not sure why they're still there to be honest, must've had a language server error when I tidied and committed. Had to set up the TypeScript/JavaScript environment from scratch for this.

leosunmo:
Take a look at my pushed changed. Pretty sure I addressed all your concerns. Really appreciate the language/style pointers, as I said, I've stayed away from JS/TS :p

chenilim (approved):
Thanks @leosunmo! LGTM 👍

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @leosunmo

PR #1440, opened 2021-10-04
Adds a generator that wraps store methods with transactions and migrates implementations to use transactions

mgdelacroix:
#### Summary
This pull request migrates all store methods to be private and receive a transaction as the first parameter. Then it adds a generator to build the public store methods from its interface, wrapping the private ones and generating, committing and rolling back their inner transactions if necessary.

The intent with this change is to easily and transparently use transactions inside the store and in between methods (methods like `insertBlock` already make use of the same transaction to do multiple things) and at the same time not worry about those at the app level. The `getQueryBuilder` has been updated as well to create the builder with the transaction so the changes to the store logic are minimal.

#### Ticket Link
[Card link](https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/47aa9bb4-6967-4a96-83c7-11bd6b20f1eb/612a901f-5c6d-4dde-8e35-3c16b6024a67?c=0da320bd-d737-460e-8801-790d8dc33f6d)


server/services/store/generators/main.go
```diff
@@ -0,0 +1,305 @@
...
+				default:
+					paramsWithType = append(paramsWithType, fmt.Sprintf("%s %s", param.Name, param.Type))
+				}
+			}
+			return strings.Join(paramsWithType, ", ")
+		},
+		"renameStoreMethod": func(methodName string) string {
+			return strings.ToLower(methodName[0:1]) + methodName[1:]
+		},
+		"genErrorResultsVars": func(results []string, errName string) string {
```
wiggin77, line 280: I have some store methods for notifications that use the Channels style methods where a create/insert will return a pointer to the struct that was inserted. e.g. 
```
func (s *SQLStore) UpsertNotificationHint(hint *model.NotificationHint) (*model.NotificationHint, error) 
```
Is it correct that pattern won't work for this generator?

wiggin77: Miguel confirmed any return types can be used.

server/services/store/generators/transactional_store.go.tmpl
```diff
@@ -0,0 +1,52 @@
...
+		{{- if $element.Results | errorPresent }}
+			if {{$element.Results | errorVar}} != nil {
+                if rollbackErr := tx.Rollback(); rollbackErr != nil {
+                   s.logger.Error("transaction rollback error", mlog.Err(rollbackErr), mlog.String("methodName", "{{$index}}"))
+                }
+                return {{ genErrorResultsVars $element.Results "err"}}
+			}
+		{{end}}
+        if err := tx.Commit(); err != nil {
+           return {{ genErrorResultsVars $element.Results "err"}}
```
wiggin77, line 46: If we're logging rollback errors, should we log commit errors too?

server/services/store/sqlstore/public_methods.go
```diff
@@ -0,0 +1,820 @@
...
+	}
+
+	if err := tx.Commit(); err != nil {
+		return 0, err
+	}
+
+	return result, nil
+}
+
+func (s *SQLStore) HasWorkspaceAccess(userID string, workspaceID string) (bool, error) {
```
wiggin77, line 582: I'm still a bit confused about how to call a simple fetch without a transaction from the app layer.  Something like `HasWorkspaceAccess` (just an example) has a exported method that creates a transaction, and a non-exported method that can do the fetch without a transaction. Shouldn't we have exported version of the non-transactional methods, especially for lookups?

mgdelacroix: You're right, this approach has everything done on a transaction from the app layer, so we should change it to allow for non-transactional wrappers to be generated as well. I have an idea on how to approach this while keeping the implementation agnostic (and therefore, flexible), will send an update and re-request review

Issue #1441, opened 2021-10-04
Bug: Pressing Esc on a URL property results in a crash


chenilim:
## Steps to reproduce the behavior

1. Open a card with a URL property
2. Click on the url property
3. (Optional) Enter or paste in a URL
4. Press Esc
4. "Something went wrong error"

## Expected behavior

No error. Esc should accept the input.

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.9.1
 - Browser and OS: Chome 93 on macOS

## Additional context

Error in console:
```
react-dom.production.min.js:209 TypeError: Cannot read properties of undefined (reading 'toString')
    at t.default (focalboard_6d2ee7887ad9e516_bundle.js:2)
    at Xo (react-dom.production.min.js:153)
    at Pi (react-dom.production.min.js:175)
    at Es (react-dom.production.min.js:263)
    at mu (react-dom.production.min.js:246)
    at du (react-dom.production.min.js:246)
    at ru (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
    at Gn (react-dom.production.min.js:122)
```

chenilim:
FYI @sbishel, @ogi-m. This is a regression from 0.8.2. I think we should investigate for v0.9, as it's an unhandled exception in the webapp, that might indicate a broader issue.

sbishel:
On it.

PR #1444, opened 2021-10-04
make sure setValue receives empty string

sbishel:
#### Summary

When hitting escape, the value in state gets reset to the property value from state. However, PropertyValue may be null. Value is not expected to be null. If null, set to empty string. 

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/1441


ogi-m (approved):
LGTM, can't repro the issue on the PR build 👍🏻

Issue #1448, opened 2021-10-05
Bug: Can not invite users - Invalid registration link [v0.9.1 - RC3]


jankolf:
## Setup
I installed v0.9.1 - RC3 on my [Uberspace account](https://lab.uberspace.de/guide_focalboard.html) and adapted the installation instructions to new focalboard setup (e.g. changed directory name).
Setup is using postgresql, "dbtype", "dbconfig" and "postgres_dbconfig" is set accordingly.

## Steps to reproduce the behavior

1. Click on the left upper side on "focalboard"
2. Click "invite user"
3. Paste link into browser window of another browser
4. ?t=... is stripped from URL
5. Fill out form and click "Register"
6. Result is "Invalid registration link, please contact your administrator"

## Expected behavior

The registration should be complete.

## Edition and Platform

 - Edition: Personal Server
 - Version: v0.9.1 - RC3
 - Browser and OS:  Windows, Vivaldi 4.2.2406.54 and Edge Version 94.0.992.38

## Additional context

As mentioned above, the system is running on Uberspace on a shared hosting system.

mmmint:
Just downgraded to 0.9.0 and it works, meaning bug was introduced in 0.9.1.

chenilim:
Thanks @jankolf and @mmmint! @sbishel, looks like this is a blocking issue for v0.9 Personal Edition. cc @ogi-m.

kamre:
We need Cypress tests for this functionality!

I [started working](https://github.com/kamre/focalboard/pull/1/files#diff-4b5c671a5f23aabe1811d05711075551cb555074a50d99ae91061e62e0f7f1e6) on https://github.com/mattermost/focalboard/issues/42 and wanted to add tests for inviting users also, but the infrastructure for Cypress testing was not ready.

@saturninoabril any updates for the testing infrastructure in focalboard project?

aloks98:
I think I have identified the issue. Mind if I drop a PR on this?

sbishel:
@aloks98 - I was just getting ready to investigate...I'd love to see your PR.

saturninoabril:
Hello @kamre, sorry to block the Cypress item. Feel free to submit PR for [issue-focalboard-42](https://github.com/mattermost/focalboard/issues/42) with your https://github.com/kamre/focalboard/pull/1/files but I'll let Boards team to drive on how to proceed as it's touching the basic foundation of single-server setup. cc @sbishel 

[...]

PR #1456, opened 2021-10-05
GH-1448: change window to history

aloks98:
#### Summary

I wasn't able to figure out why the query params are getting stripped from the URL. But on the `RegisterPage.tsx` we use `useHistory()` which also has a location field storing the initial URL. I just changed the signupToken to use `history.location` instead of `window.location`

#### Ticket Link

Fixes [Issue #1448](https://github.com/mattermost/focalboard/issues/1448)

### DIFF
--- a/webapp/src/pages/registerPage.tsx
+++ b/webapp/src/pages/registerPage.tsx
@@ -20,7 +20,7 @@ const RegisterPage = React.memo(() => {
     const dispatch = useAppDispatch()
 
     const handleRegister = async (): Promise<void> => {
-        const queryString = new URLSearchParams(window.location.search)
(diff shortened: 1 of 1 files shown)


sbishel:
Thanks @aloks98. This PR fixes the issue, but not really the underlying problem. For instance, if you notice with this fix, the registrationToken still disappears. This means that if you refresh the page...it won't work. 

I started looking for the root cause and thanks to you I knew it had something to do with a change to `history` and/or `window`. There was a fix a couple days ago that did just that. Basically that call should only be called if board is running as a plugin (if at all). 

Thanks for your help in resolving this matter. 

Closing in favor of... https://github.com/mattermost/focalboard/pull/1457

Issue #1373, opened 2021-09-29
Bug: Opening URL to board sometimes opens Dashboard instead


chenilim:
## Steps to reproduce the behavior

1. Go to [Focalboard community channel](https://community-daily.mattermost.com/core/channels/focalboard) (on community-daily)
2. Click on "meeting topics" in the channel header
3. Boards dashboard opens

## Expected behavior

The linked board should open

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.9.1 RC2
 - Browser and OS: Chrome 93 on macOS

## Additional context

This might be related to the new routes and dashboard code. It seems that opening a board url in the same tab (e.g. a relative URL) triggers a different code path / navigation.

chenilim:
Assigning to @sbishel to reassign. FYI @harshilsharma63, @hahmadia, @jespino - might be something strange in the routing.

harshilsharma63:
On it.

PR #1457, opened 2021-10-05
GH-1448 - Only call history replace if a plugin

sbishel:
#### Summary
Fixes an issue where the registration token is removed from the URL. This removal results in the registration failing.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/1448

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -92,7 +92,9 @@ const App = React.memo((): JSX.Element => {
         dispatch(fetchLanguage())
         dispatch(fetchMe())
         dispatch(fetchClientConfig())
-        history.replace(window.location.pathname.replace((window as any).frontendBaseURL, ''))
+        if (Utils.isFocalboardPlugin()) {
+            history.replace(window.location.pathname.replace((window as any).frontendBaseURL, ''))
+        }
     }, [])
 
     if (!inPluginLegacy) {


webapp/src/app.tsx
```diff
@@ -92,7 +92,9 @@ const App = React.memo((): JSX.Element => {
         dispatch(fetchLanguage())
         dispatch(fetchMe())
         dispatch(fetchClientConfig())
-        history.replace(window.location.pathname.replace((window as any).frontendBaseURL, ''))
+        if (Utils.isFocalboardPlugin()) {
+            history.replace(window.location.pathname.replace((window as any).frontendBaseURL, ''))
+        }
```
sbishel, line 97: @harshilsharma63 - Is this line necessary?  I retested this [issue](https://github.com/mattermost/focalboard/issues/1373) by using a focalboard link in a channel message. And this line didn't seem to matter, I tested both ways and the other changes seem to fix the issue.

ogi-m (approved):

webapp/src/app.tsx
harshilsharma63: When I tested it it was required. Right now the state of Channels and Boards routers co-existing is a bit messed up, so I'd prefer keeping this.

Making this conditional for plugins only is fine though.

PR #1464, opened 2021-10-06
Board calculations

harshilsharma63:
#### Summary
Implemented board calculations

#### Ticket Link
Implements https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/47aa9bb4-6967-4a96-83c7-11bd6b20f1eb/392c8889-6c8f-4cb7-8407-f5c57c1acfe8?c=cfbae9e8-e4b6-40e3-80fb-38857668536e


harshilsharma63:
> I ran into a couple issues while testing. It was displaying the menu at all times. Never did close.

Fixed that. A debug setting was left on.


> I may have missed it, but is there a way to tell what the number is without opening the menu. (ie. onhover) display `Count Value [Owner]`

No for board calculation there is no way as per the design.

> But probably the biggest thing I noticed (and this applies to tables as well).
> 
>     * The calculations are applied at the board level rather than the view level.
> 
> 
> This means you cannot have two different calculations for the same column on different views. In other words, if View1, I want to 'count owners', but View2, I want to 'count unique owners' I can't do that.

That makes sense. I've updated this for Kanban calculations. For table calculations, I've created a separate card this card to tackle that - https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/47aa9bb4-6967-4a96-83c7-11bd6b20f1eb/392c8889-6c8f-4cb7-8407-f5c57c1acfe8?c=dab1db86-b731-4210-88a3-8c47b6449b2l as we'll also need to migrate any existing data.

Issue #1393, opened 2021-10-01
Bug: Comment textarea changes height when in focus 


asaadmahmood:
## Expected behavior
It should remain fixed in height in the default state, though we will still allow users to type multiple lines.

## Screenshots
![Oct-01-2021 22-11-03](https://user-images.githubusercontent.com/11034289/135660548-38ca5672-118f-4391-87f8-34f5a291023b.gif)

kamre:
Its the same discrepancy between preview and markdown editor like in #1394

sahil9001:
Can I take this issue?

akshat-kumar2109:
Please assign me this issue. I would like to work on it.

asaadmahmood:
@sahil9001 Sure.
@akshat-kumar2109 Since sahil came first, assigning it to him, but if he does not take it, we can assign it to you.

akshat-kumar2109:
No problem @asaadmahmood

sbishel:
closed as fixed

PR #1465, opened 2021-10-06
BugFix: Comment textarea changes height when in focus

sahil9001:
#### Summary

My pull request removes the default `min-height` property which was causing the shifting issue and when the height was reduced, the content goes a bit up, instead of this we want it to stay at a fixed position.

Before :

135660548-38ca5672-118f-4391-87f8-34f5a291023b

After :

Peek 2021-10-06 20-51

Multiline Screenshot:
Screenshot from 2021-10-09 18-05-10

#### Ticket Link
https://github.com/mattermost/focalboard/issues/1393

### DIFF
--- a/webapp/src/components/markdownEditor.scss
+++ b/webapp/src/components/markdownEditor.scss
@@ -52,7 +52,6 @@
     .octo-editor-preview {
         p {
             margin: 0;
-            // min-height: 32px;
             word-break: break-word;
         }
     }


sahil9001:
@asaadmahmood

asaadmahmood:
@sahil9001 Can you add a screenshot of how multiple lines in the comment box look, just want to make sure that that looks fine as well.

asaadmahmood:
Also the top and bottom spacing above/below the avatar and text should be consistent. Currently its less at the bottom, and more at the top.

sahil9001:
@asaadmahmood

harshilsharma63:
> @sahil9001 The spacing at the top and the bottom isn't equal in your example. image
> 
> Which is what I was talking about.
> 
> But I'll fix it in another PR.

@asaadmahmood I've merged this. You can now update the padding as you mentioned anythime.

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @sahil9001

PR #1497, opened 2021-10-08
Add non-reliable websockets support

mgdelacroix:
#### Summary
This PR makes use of the `firstConnectCallback` to mark the Mattermost websocket as connected instead of polling its state, allowing for instances with reliable websockets unset to use Focalboard normally.


mgdelacroix:
@ogi-m this PR should be tested on instances with `ServiceSettings.EnableReliableWebSockets` set to `true` and `false` to ensure that the changes work for both

Issue #832, opened 2021-08-02
WebApp Unit Tests ./webapp/src/components/flashMessages.tsx


sbishel:
Write Unit Tests for ./webapp/src/components/flashMessages.tsx

see `./webapp/src/components/table/*` or `./webapp/src/components/properties/dateRange/dateRange.test.tsx for sample test`

Bhavin789:
Hi @sbishel , may I add a test to this? Possible to assign it to me? Thanks!

PR #1522, opened 2021-10-12
[GH-832] Add unit test for flashMessage component

bhavin-a:
#### Summary

The PR adds unit tests for the `flashMessages.tsx` component. Some initial feedbacks would be great as I am learning RTL and Jest. I tried splitting this into multiple tests but that was throwing some error.

Since even in the app, `FlasMessage` is rendered only once and multiple flash messages are fired on single render, IMO, sending multiple messages in single renders replicates more of real world use case. Still let me know if this can be refactored somehow. 

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/832

### DIFF
--- a/webapp/src/components/__snapshots__/flashMessages.test.tsx.snap
+++ b/webapp/src/components/__snapshots__/flashMessages.test.tsx.snap
@@ -9,3 +9,47 @@ exports[`components/flashMessages renders a flash message with high severity 1`]
   </div>
 </div>
 `;
+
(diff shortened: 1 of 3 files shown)


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @Bhavin789

Bhavin789:
Thank you so much @emilyacook for the gift 😃

PR #1529, opened 2021-10-12
[GH-1510] make calculation options translatable

### LINKED ISSUE #1510: Make calculation labels translatable
The table/kanban calculation options are not translatable as of now. We should allow those to be translated.

Screenshot 2021-10-11 at 1 41 58 PM

The strings are defined here - https://github.com/mattermost/focalboard/blob/9ca5cd6a877c154f528b9d5ff56e49aed33173a4/webapp/src/components/calculations/options.tsx#L19

jrester:
#### Summary

This PR makes the calculation options translatable. For every translatable string there must be a call to `intl.formatMessage` so the translations are correctly detected by the `i18n-extract` script.

#### Ticket Link

Fixes #1510


jrester:
@jespino could you please merge before more conflicts occur?

jespino:
@jrester some tests are failing, but otherwise looks good to me, do you mind to fix the tests and ping me to merge it?

ogi-m (approved):
LGTM, calculation options working as expected

emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @jrester

Issue #1003, opened 2021-08-16
Bug: Deleting a multi-select option doesn't remove it from the property


chenilim:
**Steps to reproduce the behavior:**
1. Open a card with a multi-select property
2. Edit the property
3. Add a new option, e.g. "New option"
4. Click on the option menu and delete the option
5. "New option" is still set on the property

**Expected behavior:**
The deleted option should be removed from the property on that card, and all cards in the board.

**Screenshots:**
<img width="445" alt="Option menu" src="https://user-images.githubusercontent.com/46905241/129596235-a5ade54e-8304-4f26-b829-bb62fdb12835.png">
<img width="463" alt="Option after deleting" src="https://user-images.githubusercontent.com/46905241/129596306-107f6d5b-99f2-4016-81bd-ac66aaaf9df2.png">

**Platform:**
 - v0.9.0 RC5 on Community
 - Chrome 91 on macOS

chenilim:
@sbishel and @ogi-m, I think this issue is relatively low severity, but we can investigate for v0.9 if you feel more strongly about it.

nishantwrp:
Hi, I'd like to work on this issue.

PR #1569, opened 2021-10-15
[GH-1003] Deleted option doesn't go away immediately in select property

nishantwrp:
#### Summary

`props.value` is `undefined` when no option is selected or when the current option is deleted. Passing `null` instead of `undefined` in the `react-select/creatable` component fixes the issue. Maybe, it's some internal issue in 'react-select/creatable' (Not sure). But it definitely fixes this issue.

https://user-images.githubusercontent.com/36989112/137480534-add90254-ad4e-41f9-80c1-566cdf64f6a5.mp4

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/1003


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @nishantwrp

Issue #411, opened 2021-05-13
Feature Idea: Show relative time for comments (and elsewhere?)


chenilim:
**Summary:**
The comment timestamp is currently a date-time string. It would be great to show it in relative time instead.
i.e. 1 hr ago, today, yesterday, etc.

<img width="217" alt="comment time" src="https://user-images.githubusercontent.com/46905241/118159122-7e187180-b3d1-11eb-8a4c-5de15550dc9c.png">

**How important this is to me and why:**

Importance: Medium

Use cases:
1. Comment time
2. Other date/time displays as well? E.g. date property?
3. Maybe we should have a setting to toggle relative / absolute display?

**Additional context / similar features:**
Now that moment.js is a dependency, this should be relatively easy to implement.

chenilim:
Instead of a setting, we should show the relative time (to the user's local time), and show the (current) absolute date-time stamp on hover.

haardikdharma10:
Hi @chenilim, can you help me with the file involved where the change has to be made?

chenilim:
Hi @haardikdharma10! Thanks for picking this up! Check out the [momentjs functions for relative time](https://stackoverflow.com/questions/38367038/format-relative-time-in-momentjs). There should be a function that does exactly this, but is localized. There may be additional settings needed to make this work correctly per language - e.g. passing in the UI language.

PR #1574, opened 2021-10-15
[GH-411] Use relative date/time for comments

Johennes:
#### Summary

This replaces the absolute date for comments on cards with a relative date/time format and moves the absolute value into a tooltip.

Screenshot_20211015_205059

#### Ticket Link

https://github.com/mattermost/focalboard/issues/411

Fixes: #411

### DIFF
--- a/webapp/src/components/cardDetail/__snapshots__/comment.test.tsx.snap
+++ b/webapp/src/components/cardDetail/__snapshots__/comment.test.tsx.snap
@@ -16,9 +16,14 @@ exports[`components/cardDetail/comment return comment 1`] = `
         class="comment-username"
       />
       <div
-        class="comment-date"
+        class="octo-tooltip tooltip-top"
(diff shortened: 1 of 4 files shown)


kamre:
Hi @Johennes! This is really nice improvement!

But as far as I understand a proper implementation of relative time is a bit more complex: it should update all relative time labels because _now_ is always moving forward.

For example relative times for comments in Trello and JIRA will show something like "just now", "a few seconds ago", "a minute ago" and so on.

Johennes:
> Hi @Johennes! This is really nice improvement!
> 
> But as far as I understand a proper implementation of relative time is a bit more complex: it should update all relative time labels because _now_ is always moving forward.
> 
> For example relative times for comments in Trello and JIRA will show something like "just now", "a few seconds ago", "a minute ago" and so on.

@kamre you mean the label should update _while_ you have the card open? That's interesting, I hadn't considered that. I suppose it could be added as an extension of the change made here.

Issue #438, opened 2021-05-17
Feature Idea: Web-clipper as Chrome and Safari extension


chenilim:
**Summary:**
Create a Chrome extension that lets you easily save links to websites to Focalboard

**How important this is to me and why:**

Importance: Medium

Use cases:
1. Quickly add links to websites to create a mood board
2. Quickly compile lists of resources for research

**Additional context / similar features:**
Notion has a webclipper for [Chrome](https://chrome.google.com/webstore/detail/notion-web-clipper/knheggckgoiihginacbkhaalnibhilkk) and [Safari](https://apps.apple.com/us/app/notion-web-clipper/id1559269364?mt=12).

Rough usage flow:
* After installing, configure by specifying:
  * URL to Focalboard server (log-in)
  * Board to add links to
* Click on the webclip button to create a new card with link as a property
* P3: Insert screenshot of webpage as an image
* Show link to open created card

What do you guys think? How important is a feature like this to you?

Johennes:
@chenilim I love this idea because then I could use Focalboard as a read-later service (for which I currently run a private Wallabag instance).

Storing URLs and page titles seems like the minimum viable feature set to me. Perspectively, we could maybe take it a step further and also extract the page content into the card description. I once created [this](https://github.com/Johennes/mailto-me) extension for Firefox and Chrome which uses [Readability](https://github.com/mozilla/readability) to extract a page's content in reader-friendly format and [turndown](https://github.com/domchristie/turndown) to convert it to markdown.

How are you envisioning the integration between the browser extension and the Focalboard server to work? Would the extension store the session token it received after logging in to then use it when posting the clipped data to a dedicated endpoint? Or maybe we could turn it around and add a UI in the webapp for generating tokens to be used in external apps?

PR #1582, opened 2021-10-16
[GH-438] Web-clipper POC

Johennes:
:warning: This is a draft, not ready to be merged. I created it to help discuss the way forward. :warning: 

#### Summary

This adds a proof-of-concept web clipper browser extension. I only tested it in Firefox but it's using polyfills so should work in Chrome (and potentially Safari), too. Here is a short demo:

https://user-images.githubusercontent.com/1137962/137599627-69dc65b1-ffba-4e7b-87ce-3ea6a33e02b7.mp4

I'm curious what you think of this and whether or not you'd like to move forward on it. There are a couple loose ends and open questions:

- Currently only the page title is saved. We could either save the URL into a card property or attach it in the card's description (or maybe the former and if no URL property exists, the latter?). This could also be supported with an accompanying "Reading List" board template in Focalboard.
- After saving a page, the extension only links to the board but not the card because that would require a view ID. We could just query the first view's ID to build the link. Alternatively, maybe we could extend the web app to handle card links without a view ID?
- Should the extension use TS rather than JS? The code is rather minimal so I didn't bother complicating the build setup to be able to use TS so far but I think it shouldn't be too hard.
- We need a suitable icon for the page action. I used one of the existing icons in the repository but it doesn't fit the purpose here very well.
- I didn't put much thought into the current "design". If you'd like to move forward on this idea, it'd be great if one of your designers could provide a little guidance here.

Looking forward to thoughts and feedback. :slightly_smiling_face: 

#### Ticket Link

https://github.com/mattermost/focalboard/issues/438

Relates to: #438


jespino:
@chenilim take a look to this, for sure an interesting proof of concept for something bigger. I would love to have something even cooler, something like "Create a bug" in the browser that allow you to take an screen shot of a section and add a description of the bug, and that creates directly a bug in Focalboard.

I think having a web browser extension can open a lot of interesting scenarios. This can be the seed of all that.

Johennes:
> I would love to have something even cooler, something like "Create a bug" in the browser that allow you to take an screen shot of a section and add a description of the bug, and that creates directly a bug in Focalboard.

Ah, that's another great use case! I had only thought about reading lists so far.

harshilsharma63:
This is awesome @Johennes !

Johennes:
Thanks for the input @chenilim! :slightly_smiling_face: 

I pushed an update to add support for storing the site's URL as well. It'll simply look for the first card property of type `url` in the board and write the URL into it. Functionally, it's fit to support my original use case of building a read-later list now.

I'm curious if you'd consider it to be merged (or let's say code-reviewed) like this? It's not in a state where it could be released yet but as, you said, it might serve as a basis for further development.

I thought about changing the technical base to use React + TS. However, even if we add further features, the UI and logic will probably stay rather trivial. So I'm not sure if it's worth it or not but curious what the rest of you think.

Johennes:
No, I don't fell strongly at all. I just picked the easiest solution (vanilla JS). 😅

I can see if I can convert it to using TS / React / Sass if that's the direction you want to go. Will definitely be more maintainable after the initial setup cost.

Johennes:
Aaand it's using Sass as well now. Turned out this was pretty easy to do since Parcel already supports it out of the box. Again, the styling code is nothing to be proud of but it's ready to be improved using the desired tech stack now.

jespino (approved):
This looks good to me. Clearly a very good first iteration. :)

jespino:
@chenilim @harshilsharma63 can we end this review and merge it, after that we can create a bunch of help wanted tickets to improve this integration from the community. I think the current foundation is solid enough to merge it and start iterating.

asaadmahmood (approved):
This looks pretty cool to me. I would like it to take additional data from the site, but to get started, this is awesome!

experiments/webext/src/utils/settings.ts
```diff
@@ -0,0 +1,29 @@
...
+
+interface Settings {
+  host: string | null
+  username: string | null
+  token: string | null
+  boardId: string | null
+}
+ 
+export function loadSettings(): Settings {
+  return browser.storage.sync.get(['host', 'username', 'token', 'boardId'])
```
chenilim, line 14: Do you know if the namespace for localstorage is per extension? i.e. could these property names collide with other code?

experiments/webext/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+		"jsx": "react",
+		"target": "es2019",
+		"module": "commonjs",
+		"esModuleInterop": true,
+		"noImplicitAny": true,
+		"strict": true,
+		"strictNullChecks": true,
+		"forceConsistentCasingInFileNames": true,
+		"sourceMap": true,
+		"allowJs": true,
```
chenilim, line 12: Future note for us to clean up the various tsconfig settings. E.g. not sure if it should have allowJs. cc @jespino, @sbishel.

chenilim (approved):
Thanks again @Johennes! This is wonderful! A great start to building out an extension. There's a lot of potential here.

chenilim:
Since @harshilsharma63 is out until Nov 12, I'll go ahead and merge this, and we can continue to iterate.

experiments/webext/tsconfig.json
Johennes: I just copied this one from another file in the repo, I believe [webapp/tsconfig.json](https://github.com/mattermost/focalboard/blob/main/webapp/tsconfig.json), so none of the settings were intentionally chosen for the web extension.

Issue #512, opened 2021-06-02
Feature Idea: Confirm deleting a property or property value from a card


chenilim:
**Summary:**
We should prompt the user to confirm deleting a property or property value from a card, because this affects all cards.

**How important this is to me and why:**

Importance: High

Use cases:
1. User deletes a property from a card, and is unaware this deletes it from all cards in the board
2. User deletes a property value from a card, and is unaware this deletes it from all cards in the board

**Additional context / similar features:**
We should show a confirmation dialog, with a message like "This will delete the property from all cards in this board, continue?"

sbishel:
We will need a confirmation dialog first.

prakharporwal:
Hello @chenilim . I would like to try to implement this feature . Do you guys have a design system for the UI which I can follow to create the dialog box. Also is their any doc to refer to so as to keep the UI consistent ?

chenilim:
Thanks @prakharporwal. Yes, we need to establish a component or pattern for the dialog system first. @asaadmahmood, do you know if we have anything ready to go, e.g. reusable from the mattermost-webapp code or Compass design system? If not, we may need to defer this, and build the infrastructure as a team first.

The alternative would be to pick an "off-the-shelf" dialog system in React.

prakharporwal:
@chenilim @asaadmahmood  Any updates on the Dialog System . Did you find it in Mattermost ? If not can I just use the confirm() in javascript to raise a warning to the user .

asaadmahmood:
@prakharporwal You can use the one that's visible in the mockup.
I have opened the Figma file for viewer access, so if you can reuse that design, that would be great.
https://www.figma.com/file/a3SJgKolniXEU2G4zG8Uc0/Issue-512-Deleting-a-property?node-id=101%3A3227

prakharporwal:
You want me make a react component by my own and use this design ?

kamre:
There is [this page](https://zeroheight.com/29be2c109/p/426c9b-modals) about modals in the "Compass Design System".

Also it might be worth to look at the [CardDialog](https://github.com/mattermost/focalboard/blob/main/webapp/src/components/cardDialog.tsx#L83-L116) component and how it is used in the [CenterPanel](https://github.com/mattermost/focalboard/blob/main/webapp/src/components/centerPanel.tsx#L121-L134) component.

asaadmahmood:
@prakharporwal Or you can use the ones that @kamre mentioned.

sudoManiac:
[...]

PR #1605, opened 2021-10-20
[GH-1140] Property Type or name change Confirmation Box

### LINKED ISSUE #1140: Feature Idea: Show warning when converting property type
**Summary:**
When changing the type of a property, show a warning dialog if any of the values (across all cards in the board) is set. i.e. don't show it if no values for that property are set (such as when first creating a property).

> ### Confirm property type change
> Are you sure you want to change the type of the property **"\<Property name>"** to \<New type>?
> This will affect \<number of values> value(s) across \<number of cards> card(s) in this board, and can result in data loss.
> [Cancel] [Change to \<New type>]

This is related to #512 (confirmation when deleting property).

Importance: Medium

Use cases:
1. Users may not be aware that changing a property type affects all cards in the board
2. Users may pick a different type accidentally
3. Not all type changes may be compatible

prakharporwal:
resolves #1140 

1. Made Common Component to raise Confirmation Dialog Box from issue #512 and #1140 .
2. Could be used to solve #1278 as well .


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @prakharporwal

webapp/src/components/confirmationDialogBox.tsx
```diff
@@ -1,56 +1,67 @@
...
-    propertyId: string;
-    onClose: () => void;
-    onConfirm: () => void;
-    heading: string;
-    subText?: string;
+    dialogBox: ConfirmationDialogBoxProps
 }
 
 export const ConfirmationDialogBox = (props: Props) => {
+    const handleOnClose = useCallback(props.dialogBox.onClose, [])
```
jespino, line 25: there is not reason to make the useCallback here, the `props.dialogBox.onClose` it is already a reference to a function, what makes sense is to use the `useCallback in the `carDetailProperties.tsx` wherever that function is created.

jespino: Actually, I think we can just remove the usage of `useCallback here and let it not memoize the hander in this case, at the end of the day, is not so common to have this thing rendered, so I think it is ok to not memoize it.

jespino (changes requested):
Remove the `useCallback` because is not needed here.

webapp/src/components/confirmationDialogBox.tsx
prakharporwal: Removed useCallback in next commit.

jespino (approved):
Looks good to me. Thanks @prakharporwal! :tada:

hahmadia:
hi! Can you please fix the broken test?

prakharporwal:
@hahmadia Resolved failing test you can merge the PR now .

hahmadia:
Hi @ogi-m , 

can you please test this PR before merging! Much appreciated.

Thanks

ogi-m:
Thanks @prakharporwal, looks good!

I have one question before approving - should the modal show the property names that are displayed in the front end? For example "Created time" instead of the current "createdTime" ? cc @chenilim @wuwinson
propertytypechange

hahmadia (approved):

asaadmahmood:
@prakharporwal Can you resolve the conflicts. Also can you improve the design to match this:
https://www.figma.com/file/8wGZk2mlLixrFQxebFuCkr/Issue-512---Deleting-a-property-(Copy)?node-id=101%3A3227

asaadmahmood:
We should already have the button sizes.

wuwinson:
@ogi-m @asaadmahmood Sorry for the delayed reply - agree that the front end property names would be preferable and more user friendly, but not strongly opposed to current implementation. cc @chenilim @prakharporwal

asaadmahmood:
This should also not show up when the user is editing the name or changing to a property type that's compatible with the current type.

prakharporwal:
> @prakharporwal Can you resolve the conflicts. Also can you improve the design to match this: 
https://www.figma.com/file/8wGZk2mlLixrFQxebFuCkr/Issue-512---Deleting-a-property-(Copy)?node-id=101%3A3227

I built it on my local the design looks similar.



> We should already have the button sizes.

@asaadmahmood Can you explain this in more detail ?



> This should also not show up when the user is editing the name or changing to a property type that's compatible with the current type.

@asaadmahmood  So we don't have any validation check to see if the property value is compatible , So we need to implement that first , Also for name is should work as it is now ( say "webURL" is renamed by mistake to "githubURL" ... there should be a warning ) cc - @chenilim what do you think ?



> @ogi-m @asaadmahmood Sorry for the delayed reply - agree that the front end property names would be preferable and more user friendly, but not strongly opposed to current implementation. cc @chenilim @prakharporwal

@wuwinson  Addressed this , added a new commit  to fix this.

ogi-m:
> This should also not show up when the user is editing the name or changing to a property type that's compatible with the current type.

Looking at https://github.com/mattermost/focalboard/issues/1140, it's not clear to me if this was a requirement? @chenilim 

cc @wuwinson @asaadmahmood

chenilim:
I 100% agree that we should not show any warning when renaming a property (tracked by https://github.com/mattermost/focalboard/issues/1902).

I'm on the fence about not showing a warning when "changing to a property type that's compatible with the current type". If we want to do that, we need to define the table of "compatible types". My recommendation is to track that as a separate enhancement if necessary.

Also, I realize these issues and PRs are old. The current behavior I observe in v0.12 is:
1. A warning is shown when changing the property type
2. A warning is shown when renaming a property

From the files changed in this particular PR, it looks like some minor cleanup (setting the active button, etc.) to an already implemented feature? If so, I suggest taking this, after dev review, and tracking any remaining changes separately.

chenilim:
Branch has been updated. @sbishel, it looks like this PR is cleanup to an existing feature? It's old enough that things may have changed since, but please review if we should merge it, and if there's any additional fixup required. cc @ogi-m.

sbishel:
Unnecessary as it has been fixed here - 
Duplicate of resolved issue - https://github.com/mattermost/focalboard/issues/1655

PR #1624, opened 2021-10-22
[GH 1348] add eslint comment to ignore todo for empty if block

### LINKED ISSUE #1348: Webapp Linter Fixes - Investigate/Cleanup ToDos
There are several files with ToDos left in the code. These need to be investigated, fixed and removed.
	* Run `npx eslint --ext .tsx,.ts .` in the webapp folder to see them

If small enough and easy enough to understand and fix. They can be fixed as part of this ticket. However, if after investigating the issue is large and/or complicated, a separate ticket should be created and this ticket can be closed by referencing the new ticket and removing the TODO from code.

This ticket includes the investigation of the following files.
/webapp/src/components/table/tableHeaderMenu.tsx
/webapp/src/components/table/tableRows.tsx

marcvelasco:
#### Summary
This PR adds the ESLint comment flag to the line prior to the //TODO entries in TableHeaderMenu.tsx.  

/focalboard/webapp/src/components/table/tableHeaderMenu.tsx
  42:25  warning  Unexpected 'todo' comment: 'TODO: Handle name column'   no-warning-comments
  54:25  warning  Unexpected 'todo' comment: 'TODO: Handle title column'  no-warning-comments

Note that tableRows.tsx had no lint issues reported.

These //TODO lines are the only entries in a if statement block, so removing them would result in another ESLint issue with an empty block.  Following the thread here(https://github.com/eslint/eslint/issues/2808) it seemed like the eslint comment to ignore the subsequent line would work best to preserve the intent of noting that future work needed to be done for the if statement's first block if true. Although if a diff approach is preferred let me know.

#### Ticket Link
Fixes #1348 https://github.com/mattermost/focalboard/issues/1348

#### Verification
Ran ESlint after changes made and no issues flagged for tableheadermenu.


sbishel (approved):
I've created this issue to track the work of removing or handling this code. 

https://github.com/mattermost/focalboard/issues/1717

Issue #42, opened 2021-01-27
Add Cypress tests: Login actions


chenilim:
Simulate user creation, login, and logout actions
* Register user
* Login
* Change password
* Logout

djanda97:
Hi @chenilim! I’m excited to see that Mattermost is working on an alternative to Trello. Once I get the dev environment setup, I’ll take a shot at this issue.

jespino:
Great! thanks @djanda97!

kamre:
@chenilim @jespino I have started working on this and encountered the problem with user registration when _single user_ option is enabled.

In particular [this code](https://github.com/mattermost/focalboard/blob/main/server/api/auth.go#L230-L234) doesn't allow to process user registration.

Do we really need to use _single user_ for cypress testing?

kamre:
As [discussed in focalboard channel](https://community.mattermost.com/core/pl/iy18wtin5p8hjr7eousthuxnkw) this issue is on hold until there is "base Cypress setup with convenient UI and API commands that will make user registration/logging easier"

[My branch](https://github.com/kamre/focalboard/pull/1/files#diff-4b5c671a5f23aabe1811d05711075551cb555074a50d99ae91061e62e0f7f1e6) with Cypress tests can be helpful later.

PR #1679, opened 2021-10-27
[GH-42] Cypress tests for login actions

kamre:
#### Summary
Added cypress tests for login/register actions:
- cypress tests switched to `typescript`
- cypress command for initial registration and login of the user added
- some race conditions with websocket initialization in `BoardPage` component fixed

#### Ticket Link
Fixes: #42

@saturninoabril may be you can also review this


server/server/server.go
```diff
@@ -161,6 +161,11 @@ func New(params Params) (*Server, error) {
 	}
 	webServer.AddRoutes(focalboardAPI)
 
+	// Add testing routes if testing API is enabled
+	if params.Cfg.EnableTestingAPI {
+		focalboardAPI.RegisterTestingRoutes(webServer.Router())
+	}
```
saturninoabril, line 167: 👍

webapp/cypress/integration/loginActions.ts
```diff
@@ -0,0 +1,119 @@
...
+    const password = 'password'
+
+    beforeEach(() => {
+        cy.apiResetServer()
+    })
+
+    const workspaceIsAvailable = () => {
+        cy.location('pathname').should('eq', '/')
+        cy.get('.Workspace').should('exist')
+        cy.get('.Sidebar').should('exist')
```
saturninoabril, line 16: Not asking for change but commenting for future reference.
I'm fine as is with the use of class names or element IDs since it's what we have currently. However, they are known to be brittle and typically tests more on implementation details rather than on user's perspective (on how the user access the app). In the future, we'll be using [@testing-library/cypress](https://testing-library.com/docs/cypress-testing-library/intro/) and take advantage of its APIs especially for queries - https://testing-library.com/docs/queries/about/#priority

webapp/src/pages/boardPage.tsx
```diff
@@ -220,12 +224,14 @@ const BoardPage = (props: Props): JSX.Element => {
...
-            wsClient.unsubscribeToWorkspace(match.params.workspaceId || '0')
+            if (subscribedToWorkspace) {
+                wsClient.unsubscribeToWorkspace(match.params.workspaceId || '0')
+            }
             wsClient.removeOnChange(incrementalUpdate)
             wsClient.removeOnReconnect(() => dispatch(loadAction(match.params.boardId)))
             wsClient.removeOnStateChange(updateWebsocketState)
         }
-    }, [match.params.workspaceId, props.readonly])
+    }, [match.params.workspaceId, props.readonly, match.params.boardId])
```
saturninoabril, line 234: Do the changes in this component related to this PR? Seems not but please clarify.

saturninoabril (changes requested):
Thanks @kamre! Left few comments and notes but overall looks good to me.

server/app/reset.go
```diff
@@ -0,0 +1,19 @@
...
+		return err
+	}
+
+	if err := a.store.DeleteAllBlocksPermanently(c); err != nil {
+		return err
+	}
+
+	if err := a.store.DeleteAllUsers(); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
kamre: This is only enabled if corresponding flag in server config file is set to `true`. Also initially there are no any registered users, so it is not clear how to select "specific user".

In general I agree that such testing API for server looks dangerous if it will be enabled in any production setup. So may be we can do better than just a flag in the config.

Also I was thinking about adding log message about enabled testing API and some unit test for server that will check that it is not enabled by default.

webapp/cypress/integration/createBoard.ts
```diff
@@ -0,0 +1,100 @@
...
+        cy.apiResetServer()
+        cy.apiRegisterTestUser()
+        cy.apiLoginTestUser()
+        localStorage.setItem('welcomePageViewed', 'true')
+    })
+
+    it('Can create and delete a board and a card', () => {
+        cy.visit('/')
+
+        cy.log('Create new empty board')
```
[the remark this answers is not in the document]
kamre: Yes, initially these were comments. May be we can have both if it will make it easier for non-technical reviewers.

webapp/cypress/integration/loginActions.ts
kamre: Agree, I was trying to use `testing-library/cypress` while adding more tests and it looks that we will need to change some components so that they have proper attributes.

webapp/src/pages/boardPage.tsx
kamre: Yes, this fixes some race condition with websocket initialization. Cypress clicks too fast and sometimes websocket is still not connected during this process and tests are failing.

sbishel (changes requested):
I don't like the additional apis that delete all blocks and users. Very scary. I also don't think they are necessary. There are better ways to handle either having tests build on each other, or using different user accounts. 
cc: @jespino

kamre:
> I don't like the additional apis that delete all blocks and users. Very scary. I also don't think they are necessary. There are better ways to handle either having tests build on each other, or using different user accounts.

@sbishel  ok, I have reverted all the changes to server code.

The sequence of `loginActions.ts` and `createBoard.ts` is now fixed in the Cypress config file because testing of the registration should come first. Also this leads to situation when `loginActions.ts` can be executed only once on the fresh server. But `createBoard.ts` can be executed many times and on its own, it will just try to register and login the user every time.

It looks like `mattermost-webapp` e2e tests create new team/user/channel every time and they are really isolated on the server which is not the case for `focalboard`.

@saturninoabril FYI

sbishel (approved):
Thanks for the changes...LGTM

webapp/cypress/support/commands.ts
```diff
@@ -0,0 +1,45 @@
...
+        body: {
+            username: Cypress.env('username'),
+            password: Cypress.env('password'),
+            email: Cypress.env('email'),
+            token: '',
+        },
+        headers: {
+            'X-Requested-With': 'XMLHttpRequest',
+        },
+        failOnStatusCode: false,
```
saturninoabril, line 43: Same here to be its own Cypress command as `apiRegister(user: RegistrationRequest)`

saturninoabril (commented):
@kamre Thank you for updating! I've made several comments but those were non-blocking since those were not specifically mentioned in the ticket and could be worked on in a follow up PR, but let me know how you'd like to proceed.

I made the comments in a more opinionated manner, typically based from a pattern we have in the `mattermost-webapp`, since it's crucial we have it from the start so that succeeding PRs will have some kind of basis / structure to follow on.

Overall, I'm good as is and will approve depending on how you'd like to address the comments.

webapp/cypress/support/commands.ts
```diff
@@ -0,0 +1,45 @@
...
+            type: 'normal',
+        },
+        headers: {
+            'X-Requested-With': 'XMLHttpRequest',
+        },
+    }).then((response) => {
+        expect(response.body).to.have.property('token')
+        localStorage.setItem('focalboardSessionId', response.body.token)
+    })
+}
```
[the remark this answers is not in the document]
kamre: Converted to Cypress command. Not sure if we can use exposed `OctoClient` before we visit any page and login via API should work before visiting any page in my understanding.

kamre:
@saturninoabril thanks for you suggestions! I have updated the changes in this pull request accordingly.

saturninoabril (approved):
Thanks @kamre! LGTM.

kamre:
@jespino can you review this pull request please? It changes how Cypress testing is configured and other tasks for adding more Cypress tests depend on this pull request actually.

jespino (approved):
Looks good to me. Thanks @kamre! :tada:

PR #1706, opened 2021-10-31
webapp: make 'no options' message more explicit + translatable

JulienTant:
#### Summary
I was trying focalboard before checking the available issues, and upon trying to add a `select` property type to a card, it took me about 5 minutes of looking everywhere in the application to realize that adding an option was as simple as typing where the text `No options.` appears. This PR adds a "clearer" message ("No options. Start typing to add the first one!") and make this translatable as a bonus!

image

#### Ticket Link
No ticket exists. I can create one if necessary.


harshilsharma63:
@asaadmahmood the first-time usability issue for select and multi-select has been reported by many users. Is there something we can do about it?

harshilsharma63 (approved):

asaadmahmood:
@harshilsharma63 I think this makes senes, I wouldn't even mind adding a link to the popup saying "Add new option" that just adds a numbered option like: "Option 1".

harshilsharma63:
> @harshilsharma63 I think this makes senes, I wouldn't even mind adding a link to the popup saying "Add new option" that just adds a numbered option like: "Option 1". But for that we would need the abllity to edit the option, which I don't think we have yet.

Another idea is to replace the empty state text from `Empty` to something on the lines of `Type here to add option`?

Issue #1709, opened 2021-10-31
Bug: Problem Building Application under Windows 


Genei180:
## Steps to reproduce the behavior
I followed the Steps described under: https://www.focalboard.com/contribute/getting-started/dev-tips/

After Step 5. Hit F5 and select Go: Launch Server

The following Error Occurs for me on a Windows System:
`Unable to unlink socket.  "server/server.go:378"="The system cannot find the path specified."`

The file path that should be opened under windows after the Config File is: "/var/tmp/focalboard_local.socket"

## Expected behavior
To Build the Application for Debugging on 127.0.0.1:8000

## Edition and Platform
Forked from Github Commit: 690d424dfe28d3583634e8d7d7e59e365cac71d8

## Additional context
A Windows file Path gets Accepted if you Specify it in the Config, but Only if the File "focalboard_local.socket" has been created somewhere...

Genei180:
I am already Working on a Fix, but it still needs to be Tested on Windows

CodeCaterpillar666:
I have the same problem building application under MacOS.
`error [2021-11-17 10:36:58.845 -08:00] Unable to unlink socket.                 caller="server/server.go:378" error="no such file or directory"`

chenilim:
Housekeeping old issues: The error message indicates a problem with the admin API, which is a file socket. The workaround is to set `enableLocalMode` in the `config.json` file and restart.

Closing this issue for now, but please reopen or file a new issue if anyone is still seeing this. cc @jespino, @ogi-m, @sbishel.

PR #1710, opened 2021-10-31
[GH-1709] Fix Missing Socket when Building Application under Windows

Genei180:
#### Summary

The Pull Request Fixes the Bug encountered in the below Ticket when trying to Build for Windows.
As the default Path: "/var/tmp/focalboard_local.socket" is not Available under Windows, it was moved into the Local Folder.
Additionally, the Socket File will now be created if it does no Already Exist, and gets Deleted on Stop Server.

##### Note:
Be Aware because of: https://github.com/microsoft/vscode-go/issues/2387
I could not Test the Stop Code.

Additionally, someone should confirm that it still can be Build under Linux as Normal. 

#### Ticket Link

https://github.com/mattermost/focalboard/issues/1709

### DIFF
--- a/config.json
+++ b/config.json
@@ -17,7 +17,7 @@
 	"session_refresh_time": 18000,
 	"localOnly": false,
 	"enableLocalMode": true,
-	"localModeSocketLocation": "/var/tmp/focalboard_local.socket",
+	"localModeSocketLocation": "./focalboard_local.socket",
 	"authMode": "native",
 	"logging_cfg_file": "",
 	"audit_cfg_file": "",
--- a/server/server/server.go
+++ b/server/server/server.go
@@ -373,6 +373,13 @@ func (s *Server) startLocalModeServer() error {
 		ConnContext: api.SetContextConn,
 	}
 
+	// If the socket file doesn't exist, create it, or append to the file
+	f, err := os.OpenFile(s.config.LocalModeSocketLocation, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		s.logger.Error("Unable to create Socket File:", mlog.Err(err))
+	}
+	f.Close()
+
 	// TODO: Close and delete socket file on shutdown
 	if err := syscall.Unlink(s.config.LocalModeSocketLocation); err != nil {
 		s.logger.Error("Unable to unlink socket.", mlog.Err(err))
--- a/server/server/server.go
+++ b/server/server/server.go
@@ -373,14 +373,16 @@ func (s *Server) startLocalModeServer() error {
 		ConnContext: api.SetContextConn,
 	}
 
-	// If the socket file doesn't exist, create it, or append to the file
-	f, err := os.OpenFile(s.config.LocalModeSocketLocation, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
-	if err != nil {
-		s.logger.Error("Unable to create Socket File:", mlog.Err(err))
+	// If the socket file doesn't exist
+	if _, err := os.Stat(s.config.LocalModeSocketLocation); errors.Is(err, os.ErrNotExist) {
+		//Create it
+		f, err := os.OpenFile(s.config.LocalModeSocketLocation, os.O_APPEND|os.O_CREATE, 0644)
+		if err != nil {
+			s.logger.Error("Unable to create Socket File:", mlog.Err(err))
+		}
+		f.Close()
 	}
-	f.Close()
 
-	// TODO: Close and delete socket file on shutdown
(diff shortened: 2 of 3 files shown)


harshilsharma63 (approved):
Looks good. Works fine on Mac.

harshilsharma63:
@Genei180 can you fix the lint error?

harshilsharma63:
@jespino do you have a windows system to try this out?

server/server/server.go
```diff
@@ -403,6 +412,12 @@ func (s *Server) stopLocalModeServer() {
 		_ = s.localModeServer.Close()
 		s.localModeServer = nil
 	}
+
+	//Remove Socket File on Stop
```
jespino, line 416: I don't think the comment is needed, the code is self-explanatory.

server/server/server.go
```diff
@@ -373,7 +373,16 @@ func (s *Server) startLocalModeServer() error {
 		ConnContext: api.SetContextConn,
 	}
 
-	// TODO: Close and delete socket file on shutdown
+	// If the socket file doesn't exist
```
jespino, line 376: The comment is redundant to the code, and it is not helping to understand the intentions behind this. Why are we creating the file if it doesn't exist to just delete the file right after that?

config.json
```diff
@@ -17,7 +17,7 @@
 	"session_refresh_time": 18000,
 	"localOnly": false,
 	"enableLocalMode": true,
-	"localModeSocketLocation": "/var/tmp/focalboard_local.socket",
+	"localModeSocketLocation": "./focalboard_local.socket",
```
jespino, line 20: This change is changing the behavior on all the systems, can we make it change only for windows? For example, on windows startup if this variable is `/var/tmp/focalboard_local.socket` (the default value), transform that to something else, for example `./focalboard_local.socket`, or the way Windows store temporary files.

jespino (changes requested):
Some improvement and explanations needed, but in general looks good to me. Thanks @Genei180! :tada:

server/server/server.go
Genei180: #### About your First Issue :
Yes, it is Redundant.
I kind of Prefer it with Little Comments to help me out faster.
But i have to Appreciate good Feedback and will therefore remove them :)

#### About the Second Part:
It gets Created because, if it doesn't Exist by Default, the Program throws an Error.
(Was part of the Issue because the (Linux) Path couldn't be found under Windows) 

About the Deletion: If you mean the Unlink, it is only deleted if no existing File Handlers are there to Point to the File. (E.g. the Socket is no longer Needed / The Program has Stopped)
 
I didnt find Proper Documentation in Go but i think under the Hood the same as in Linux is Achieved: [Linux Man Page Unlink](https://linux.die.net/man/2/unlink)

The File does get created, because it is Needed at Runtime for Communication between Programms. See [Wikipedia Socket File](https://en.wikipedia.org/wiki/Unix_file_types#:~:text=A%20socket%20is%20a%20special,and%20recvmsg()%20system%20calls.) For reference.

After the Process/Server Exits, the Socket File is no Longer Needed, that's why it should get deleted/cleaned up. (Apparently the Code is Redundant for Deletion, but it Still didn't get Deleted on My System)

I hope I wasn't too Over explanatory, but I wanted to make sure we are aware of the Reasoning from one another. 
Did this solve the Second Part of your Issue?
Or did I misunderstand your Problem?

server/server/server.go
Genei180: Removed it

config.json
Genei180: Well, we could do that, but that would involve adding a lot of Code specifics for Windows that only checks if the Default Config isn't Overridden. 

And What happens on the Very rare Use case when a Windows User Actually wants to use "/var/tmp/focalboard_local.socket"?
The Error could also appear on another (Linux) System where "/var/tmp" was deleted or simply doesn't exist.
With the Current Implementation it is always made Sure that the Path Location Exists.
(And it always gets Deleted with the Folder, and doesn't Stay behind if the User Decides to do so on Every System)
And if it is really wanted, the Config can Still be Changed to Every Path the User Wants.

I personally wouldn't Implement a Windows Specific Way, as this adds a lot of additional trouble that is only Relevant for one specific System.

But I am not Qualified to make that Decision :)
So please let me know

jespino: The problem with that is the "Windows" environment is not the main expected environment. Yes, for the desktop app, but for personal server the expected environment is Linux, so if we make a non-standard behavior, I would prefer to make it in Windows, instead of moving that to Linux. What do you think @chenilim?

jespino: /cc @wuwinson

Genei180:
See https://github.com/mattermost/focalboard/issues/1709 for alternative Solution.

Issue #1664, opened 2021-10-26
Bug: Sort by date doesn't work


chenilim:
## Steps to reproduce the behavior

1. Create several cards with a date property
2. Set the date property to different values
3. Create a board view, and sort by date
4. Order doesn't change
5. Sort by date descending
6. Order doesn't change
7. Create a table view and sort by date
8. Order doesn't change
9. Sort by date descending
10. Order doesn't change

## Expected behavior

Card order changes by the date. If it is a range, sort by the earlier date.

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.9.4 and main build
 - Browser and OS: Chrome 93 on macOS

sarvani1997:
I would like to work on this.

pullopen:
Still doesn't work in v0.10.0 though.

ogi-m:
Verified that it's fixed in main. Not in v0.10 though, it didn't get cherry-picked from what I can tell @chenilim

chenilim:
Thanks @ogi-m. Makes sense. Looks like this was merged after the fork but not cherry-picked. For this particular issue, we can defer it to v0.11.

Let's circle back on improving our process to avoid missing future cherry-picks. I'll add an item to the next team meeting.

PR #1719, opened 2021-11-02
Fixed sort by date

sarvani1997:
#### Summary

Date property is in JSON string. So, we should parse it and sort later.

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/1664


wiggin77 (changes requested):
Looks good @sarvani1997.  Thanks.  There is one typing issue to fix and then it's good to go.

harshilsharma63 (approved):

wiggin77 (approved):
LGTM :+1:

Issue #1316, opened 2021-09-22
Don't show an error when the Boards plugin is updated, or server connection interrupted


chenilim:
**Summary:**
Currently, users with a Boards tab open see an error when the plugin is updated:
1. Go to system console
2. Update the Boards plugin
3. All clients see an error (can press the back button to return to the board)

***Expected:***
Users see the board they were on after the update is completed.

**Additional context:**
This likely requires the websocket error handling code to be tweaked. Currently, it switches to polling after a period (5 seconds). However, if the plugin is disabled (or connection interrupted) during the poll, this results in a general error, because the route is not found.

In general, Boards should have a similar error-handling experience to Channels, i.e. to show an intermittent error banner when the connection is interrupted, but seamlessly recovering after the connection is restored.

chenilim:
@sbishel, there will be fit and finish bugs / features like this as we converge on v1. Not sure what the best way to collect them is, but I think we can file issues here, then collect the high-pri ones in a card for specific releases? We can chat more.

mgdelacroix:
@chenilim Could not reproduce this exact behaviour. With the current server `master`, what's happening is that the plugin gets restarted server side and reloaded client side. The consequence of this is that the server loses the websockets connections data, so the client needs to reconnect.

A quick way of solving this is to refresh the page on plugin update, as the websocket connection is getting a notification when the plugin changes its status. This solution has been drafted at https://github.com/mattermost/focalboard/pull/1720

A more precise solution in which we reconnect instead of reload from the client could be possible, but will require further investigation as it doesn't seem as simple as reconnecting when the plugin update message is received.

ogi-m:
Verified the banner is appearing as expected on latest v0.10 branch

PR #1720, opened 2021-11-02
Show a refresh banner on plugin update

mgdelacroix:
#### Summary
Shows a banner when the plugin status changes on update, asking the user to click to refresh the interface.

boards-refresh

image

Fixes https://github.com/mattermost/focalboard/issues/1316


webapp/src/wsclient.ts
```diff
@@ -299,6 +301,17 @@ class WSClient {
...
     }
 
+    pluginStatusesChangedHandler(data: any): void {
+        if (this.pluginId === '') {
+            return
+        }
+
+        if (data.plugin_statuses.some((s: any) => s.plugin_id === this.pluginId)) {
+            Utils.log('Boards plugin has been updated, reloading...')
+            location.reload()
```
harshilsharma63, line 311: Can this cause issues when the user has unsaved progress? For example, what would happen if the user was editing a card description or comment?

Also, would it makes sense to show a warning first and reload after some time, let's say, a minute? This can be addressed in a separate ticket.

harshilsharma63 (changes requested):

webapp/src/wsclient.ts
mgdelacroix: This was addressed as a quick solution, I agree that something like Chats' "There is a new version of Mattermost, click here to reload" would be safer in case a user is working with the tool when the update happens

jespino: I have the same concern here. If this is a temporary solution, we should include a TODO comment here, and a ticket for resolving that sooner than later.

mgdelacroix: I'm actually taking a stab at putting together a banner and a more long-term fix, will update with the new approach

mgdelacroix: Done, PTAL 64f3db0

jespino (approved):
LGTM

mgdelacroix:
Heads up that this will only work when the plugin system doesn't refresh the interface automatically.

chenilim (approved):
LGTM!

Issue #1692, opened 2021-10-28
Feature Idea: Show card preview in @mention notification DM


chenilim:
## Summary

Show the card preview in the DM notification.

## How important this is to me and why

Importance: Medium

This would make links to cards consistent.

ogi-m:
Tested and passed on v0.10

PR #1725, opened 2021-11-02
unfurl for mentions

wiggin77:
#### Summary
This PR adds unfurl meta-data when creating @mention notification posts.  This includes support for unfurling of inline links, meaning any posts with inline links will now unfurl.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/1692


mattermost-plugin/server/embed.go
```diff
@@ -0,0 +1,154 @@
...
+	firstLink := ""
+
+	markdown.Inspect(str, func(blockOrInline interface{}) bool {
+		if autoLink, ok := blockOrInline.(*markdown.Autolink); ok {
+			if link := autoLink.Destination(); firstLink == "" {
+				firstLink = link
+				return false
+			}
+		}
+		if autoLink, ok := blockOrInline.(*markdown.InlineLink); ok {
```
hahmadia, line 101:
```suggestion
		if inlineLink, ok := blockOrInline.(*markdown.InlineLink); ok {
```

mattermost-plugin/server/embed.go
```diff
@@ -0,0 +1,154 @@
...
+			if link := autoLink.Destination(); firstLink == "" {
```
hahmadia, line 102:
```suggestion
			if link := inlineLink.Destination(); firstLink == "" {
```

hahmadia (approved):
LGTM! Thanks Doug. Nice refactoring 👍

hahmadia:
Hey Doug,

Just a small FYI that when you want your PR linked to a github issue, you must do `Fixes ...` or else the issue won't be linked. I've already changed it for you but just for future reference.

mattermost-plugin/server/notifications.go
```diff
@@ -49,7 +49,8 @@ func (da *pluginAPIAdapter) GetDirectChannel(userID1, userID2 string) (*model.Ch
 }
 
 func (da *pluginAPIAdapter) CreatePost(post *model.Post) error {
```
jespino, line 51: It is ok because for now we only use this in this context, but this function name can be misleading, because you can try to create a post without wanting to get that processed. Probably would be more clear something like:

```go
func (da *pluginAPIAdapter) CreatePost(post *model.Post) error {
    return da.client.Post.CreatePost(post)
}

func (da *pluginAPIAdapter) CreatePostWithEmbededCard(post *model.Post) error {
    postEmbed, _ := postWithBoardsEmbed(post, da.client.Configuration.GetConfig().FeatureFlags.BoardsUnfurl)
    return da.CreatePost(postEmbed)
}
```

Also, extracting the link every time, even whenever we already have the link is sounds sub-optimal, and also adds certain dependency on the structure of the message. In other hand, if you optionally pass the link, you can skip that extra step (if you want) and also remove that requirement of including explicitly the link in the message.

jespino (changes requested):
A comment about improvements that I think can be done. Otherwise looks good to me.

mattermost-plugin/server/notifications.go
wiggin77: I liked that the interface used was a subset of the real plugin interface. Now it will be something else. Not a huge deal but we're creating a new API for the unlikely case that the notification system would need to create posts that don't unfurl.  

Also, unfurl was not implemented by embedding a link.  It embeds the components of a link (workspace, board, card).  I might as well just do the embed in server code and keep the plugin API as is.

jespino: I like your refactor, but you included a ton of not related changes with dependencies.

jespino (changes requested):
Added dependency changes not related to this PR.

mattermost-plugin/server/notifications.go
wiggin77: They are related.  I needed to update server v6 dependency since an unfurl struct is in model there.  Then a `go mod tidy` did the rest.

wiggin77: ... and that necessitated a `make generate`

server/services/notify/plugindelivery/plugin_delivery.go
```diff
@@ -80,6 +81,18 @@ func (pd *PluginDelivery) Deliver(mentionUsername string, extract string, evt no
 		ChannelId: channel.Id,
 		Message:   formatMessage(author.Username, extract, evt.Card.Title, link, evt.BlockChanged),
 	}
+
+	embed := cardEmbed{
+		OriginalPath: link,
```
hahmadia, line 86: This will cause issues becuase this was intended to be `u.RequestURI(),` 

Input
```Go
package main
 
import (
   "fmt"
   "net/url"
)
 
func main() {
 
   str := "http://ispycode.com/GO/URL/Get-Request-URI"
   u, _ := url.Parse(str)
   uri := u.RequestURI()
   fmt.Println("uri:", uri)
}
```

Output
```Go
$ go run example.go
uri: /GO/URL/Get-Request-URI
```

We don't want the full link, just the path.

mattermost-plugin/server/notifications.go
jespino: Ohh, you are right.

server/services/notify/plugindelivery/plugin_delivery.go
```diff
@@ -80,6 +81,18 @@ func (pd *PluginDelivery) Deliver(mentionUsername string, extract string, evt no
...
+		WorkspaceID:  evt.Workspace,
```
hahmadia, line 87: A little confusing but this is WorkspaceID right? Just want to confirm

jespino (approved):
LGTM

server/services/notify/plugindelivery/plugin_delivery.go
```diff
@@ -91,3 +104,31 @@ func (pd *PluginDelivery) getTeamID(evt notify.BlockChangeEvent) (string, error)
...
 }
+
+type cardEmbed struct {
+	OriginalPath string `json:"originalPath"`
+	WorkspaceID  string `json:"workspaceID"`
+	BoardID      string `json:"boardID"`
+	CardID       string `json:"cardID"`
+}
+
+func embedLinkInPost(post *model.Post, embed cardEmbed) (*model.Post, error) {
```
hahmadia, line 115: Are we okay with having this function duplicated?

hahmadia (commented):
I'm **ok** with the changes but I was 4/5 on what we had originally (less duplicated code and a bit more straight forward in my opinion). I found that to be a much neater approach although I guess you feel that this is better which I'm okay with. 

I'd still like to raise the question why you felt doing it this way is better than the original way we had ? I'm not a 100% clear.

The Original way being

```Go
postEmbed, _ := postWithBoardsEmbed(post, da.client.Configuration.GetConfig().FeatureFlags.BoardsUnfurl)
```

server/services/notify/plugindelivery/plugin_delivery.go
wiggin77: I tried moving this to server model but it created dependencies between plugin and server that shouldn't exist.  It's really a trade-off between the parsing and duplicating the code a bit.

server/services/notify/plugindelivery/plugin_delivery.go
wiggin77: Ok, will fix.

wiggin77:
> why you felt doing it this way is better than the original way we had

I liked the first submit better as well.  It did have a parsing step that is no longer there, but at the cost of some code duplication. Creating a new method in the plugin API would have polluted the API (in my opinion) but more importantly needed either a long arg list or some embed struct that cannot easily be shared without creating a dependency storm.  So this is where I landed.

server/services/notify/plugindelivery/plugin_delivery.go
wiggin77: ... strangely my tests still worked with the full URL.

hahmadia: It's for frontend, not backend. This has no effect on the backend (when you click on the preview to take you to the card)

wiggin77: Fixed

Issue #1729, opened 2021-11-02
Bug: shared boards are not loading correctly


mgdelacroix:
## Steps to reproduce the behavior

1. Enable shared boards
2. Share a board and copy the shared URL
3. Open the shared URL either in a logged in session or an incognito one

## Expected behavior

The shared board should appear

## Logs

```js
[1635870753.69] attachToBoard: bn5xnhuhmdj6jepis4oojqky8co main.js:397:345829
[1635870753.70] OctoClient baseURL: https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard main.js:397:345829
TypeError: r is undefined
    default https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:397
    Pu https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:304
    unstable_runWithPriority https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:397
    za https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:304
    Bu https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:304
    Lu https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:304
    A https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:397
    onmessage https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:397
main.js:304:82323
Uncaught TypeError: r is undefined
    default https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:397
    Pu https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:304
    unstable_runWithPriority https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:397
    za https://boards-ha-mysql.test.mattermost.cloud/plugins/focalboard/static/main.js?e5f1978978ab20c9043e:304
[...]

PR #1738, opened 2021-11-02
Check nulls on board and active view when tracking telemetry events

mgdelacroix:
#### Summary
This PR checks `board` and `activeView` for `null`s when tracking the view shared board event.

Fixes https://github.com/mattermost/focalboard/issues/1729

### DIFF
--- a/webapp/src/pages/boardPage.tsx
+++ b/webapp/src/pages/boardPage.tsx
@@ -163,13 +163,20 @@ const BoardPage = (props: Props): JSX.Element => {
         }
     }, [board?.title, activeView?.title])
 
+    if (props.readonly) {
+        useEffect(() => {
+            if (board?.id && activeView?.id) {
+                TelemetryClient.trackEvent(TelemetryCategory, TelemetryActions.ViewSharedBoard, {board: board?.id, view: activeView?.id})
+            }
+        }, [board?.id, activeView?.id])
+    }
+
     useEffect(() => {
         let loadAction: any = initialLoad /* eslint-disable-line @typescript-eslint/no-explicit-any */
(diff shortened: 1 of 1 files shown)


chenilim (approved):
Ack, thanks @mgdelacroix! Do you think the other similar telemetry code should be updated to use optional values as well? I will re-review as well, but wonder if you saw other places. Thanks!

mgdelacroix:
@chenilim just saw that one but only skimmed through the rest. I'll check the rest of the original PR in case I can find other places to check for safety on the telemetry code

mgdelacroix:
@chenilim done in https://github.com/mattermost/focalboard/pull/1739

webapp/src/pages/boardPage.tsx
```diff
@@ -163,13 +163,20 @@ const BoardPage = (props: Props): JSX.Element => {
         }
     }, [board?.title, activeView?.title])
 
+    if (props.readonly) {
+        useEffect(() => {
+            if (board?.id && activeView?.id) {
+                TelemetryClient.trackEvent(TelemetryCategory, TelemetryActions.ViewSharedBoard, {board: board?.id, view: activeView?.id})
+            }
+        }, [board?.id, activeView?.id])
```
wiggin77, line 171: Nice optimization.  I guess telemetry was tracked every render before this?

wiggin77 (approved):
LGTM :+1:

sbishel (approved):

webapp/src/pages/boardPage.tsx
mgdelacroix: It was actually being tracked once unless the `workspaceId` changed, which would miss view updates

PR #1739, opened 2021-11-02
Normalise and add safety checks to telemetry code

mgdelacroix:
#### Summary
This PR normalises and adds some safety checks to the telemetry track events code. Most of the changes standardize the way the props are accessed not to mix destructuring and direct access, and the `cardDialog.tsx` changes use the prop which is guaranteed to exist instead of the obtained `card`.


sbishel (approved):
Nice, thanks for fixing this.

PR #1743, opened 2021-11-03
Changed server not bind to any port to debug, as it's a plugin only expected path

mgdelacroix:


wiggin77 (approved):
Oh yes, thank you.  :+1:

PR #1746, opened 2021-11-03
Redirect to new routes when visiting a board with the legacy ones

mgdelacroix:
#### Summary
This PR redirects the user to the new plugin routes if it detects they're visiting an old legacy board URL.


jespino (approved):
LGTM

PR #1748, opened 2021-11-03
Fix GH-1690: Default card template on a board template adds a card to the template, not the board

### LINKED ISSUE #1690: Bug: Default card template on a board template adds a card to the template, not the board

1. Click on +, and select "+ New Template", name the new template "Board template"
2. Click on the dropdown off the New button
3. Select + New template (new card template)
4. Name the new card template (e.g. "Card template")
5. Click on the dropdown off the New button again
6. Click on the "..." menu for "New card template", and select "Set as default"
7. Now create a new board off this new template:
8. Click on the + button in the sidebar
9. Select "Board template"
10. Click on the New button to add a new card
11. Notice that the card was not created in the board
12. Now click on the + button in the sidebar, then click ... off the "Board template" template
13. Notice that the card was added to the template instead

## Expected behavior

The card should be added to the board, not the template

## Screenshots

https://user-images.githubusercontent.com/46905241/139296695-90e8ef19-a7a4-4127-b94c-2d8d8c5ad8da.mov

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.10.0 RC1
 - Browser and OS: Chrome 93 on macOS

hahmadia:
#### Summary

Fixes GH-1690

#### Ticket Link

Fixes #1690


wiggin77 (approved):
LGTM :+1:   I see no snapshots got updated, does that mean no unit tests for this operation? I know it would be out of scope for this PR, but can we make sure there is a ticket for it?

webapp/src/components/gallery/galleryCard.tsx
```diff
@@ -92,8 +92,8 @@ const GalleryCard = React.memo((props: Props) => {
                             id='duplicate'
                             name={intl.formatMessage({id: 'GalleryCard.duplicate', defaultMessage: 'Duplicate'})}
                             onClick={() => {
-                                TelemetryClient.trackEvent(TelemetryCategory, TelemetryActions.DuplicateCard, {board: board.id, card: card.id})
-                                mutator.duplicateCard(card.id)
+                                TelemetryClient.trackEvent(TelemetryCategory, TelemetryActions.DuplicateCard, {board: props.board.id, card: card.id})
```
sbishel, line 95: Was this change intentional? 'props' was just removed here. https://github.com/mattermost/focalboard/pull/1739/files

sbishel (changes requested):
Probably should add a unit test for mutator and any others that can be added easily.

webapp/src/components/gallery/galleryCard.tsx
hahmadia: Not intentional. Never had intention to change that. Thanks for pointing out. I've reverted.

hahmadia:
@sbishel I've updated the snapshots that are broken in main in this PR.

sbishel (approved):

PR #1752, opened 2021-11-03
Fixes issue with dates and daylight savings

sbishel:
#### Summary

Found issue because unit tests began to fail. The TimeZoneOffset was being determined based on today. If the date being saved or retrieved is on the other side of the Daylight Savings time change, the hour would be off by 1.

Does not really affect current users because we don't save time, but could in the future.


jespino (approved):
LGTM

mgdelacroix (approved):
LGTM! :+1:

mgdelacroix:
There is a test still failing but looks unrelated and is failing on `main` as well

sbishel:
I'm investigating the main failure, now.

PR #1753, opened 2021-11-03
Fix for feature flag handling

sbishel:
#### Summary
Fixes an issue where feature flags are not updated in the webapp when they are changed via split-io.

Essentially retrieving the config using `GetUnsanitizedConfig` vs `LoadPluginConfiguration` in `onConfigurationChange`.


mattermost-plugin/server/configuration.go
```diff
@@ -71,21 +67,23 @@ func (p *Plugin) setConfiguration(configuration *configuration) {
 }
 
 // OnConfigurationChange is invoked when configuration changes may have been made.
-func (p *Plugin) OnConfigurationChange() error {
+func (p *Plugin) OnConfigurationChange() {
```
jespino, line 70: Why is this no longer returning error? I think that would break the plugin interface itself.

mattermost-plugin/server/configuration.go
```diff
@@ -71,21 +67,23 @@ func (p *Plugin) setConfiguration(configuration *configuration) {
...
 	if p.wsPluginAdapter == nil {
-		return nil
+		return
 	}
 
-	configuration := &configuration{}
-	// Load the public configuration fields from the Mattermost server configuration.
-	if err := p.API.LoadPluginConfiguration(&configuration); err != nil {
-		return errors.Wrap(err, "failed to load plugin configuration")
+	mmconfig := p.API.GetUnsanitizedConfig()
```
jespino, line 76: With this changes you are violating a lot of assumtions made by the plugins system. The `configuration` struct is expected to be defined and used to store and restore the configuration of the plugin itself. Also, the plugin API has methods to set and retrieve that configuration, this changes looks like a bit misleading for me. In the future we can try to use that methods and realized that there are no longer usable. Is there any reason to not leverage the existing `configuration` struct to inform the plugin about the current status of the config, and pass to the server what is needed?

mattermost-plugin/server/configuration_test.go
```diff
@@ -1,90 +1,86 @@
...
-		assert.Panics(t, func() {
-			plugin.setConfiguration(configuration1)
-		})
-	})
+	return srv
+}
 
-	t.Run("clearing configuration", func(t *testing.T) {
+func TestConfiguration(t *testing.T) {
+	t.Run("null configuration", func(t *testing.T) {
```
jespino, line 38: This lonely `t.Run` inside a `Test` function, looks weird to me. I would just change the name of the test function to something like `TestNullConfiguration` and remove the extra nesting there.

jespino (changes requested):
A couple of suggestions.

mattermost-plugin/server/configuration.go
sbishel: Yep. I fixed this in my upstream PR, but forgot to update here. Nice catch. It did break the interface. `GetUnsanitizedConfig` doesn't return an error, but `LoadPluginConfiguration` did.

mattermost-plugin/server/configuration.go
sbishel: I added back the property and setting the `configuration` struct. However, the plugin API `LoadPluginConfiguration` only loads the plugin configuration...FeatureFlags are NOT returned. So the call to `GetConfig` is still required. I did update from `GetUnsanitizedConfig`

jespino: Yes, I agree with that, we need both inside, but my concern is about how that data is exposed everywhere else in the plugin side.

jespino (approved):
LGTM

Issue #1509, opened 2021-10-11
Feature Idea:  LaTeX / KaTeX support


pathakshashank17:
## Summary

`KaTeX` / `LaTeX` support greatly helps in creating beautiful math embedded docs.

## How important this is to me and why

Importance: High. I use Notion **only** due to its KaTeX capabilities.

## Additional context/similar features

Notion does this by wrapping the `KaTeX` syntax between `$$$$`, like this: 
```LaTeX 
$$\bar{A}\cdot B$$
```

PR #1755, opened 2021-11-04
[GH-1509] Add support for Latex rendering in markdown cells

cphyc:
This adds support for displaying inline and displayed Latex equations using Katex in markdown cells.

The conversion happens before each paragraph is converted from markdown to html.


jespino:
@cphyc We are evaluating how is the best way to handled this. sorry for the delay answering, but let me think a bit more about it and we can decide if we merge it or try to incorporate the existing code from mattermost.

jespino:
Sorry, I still have to take a look into this.

cphyc:
Hi, this has been almost a year. Is there any hope of getting this merged or should I close the PR?

Issue #1685, opened 2021-10-28
Bug: Opening a link to a private board or card shows the add board from template UI


chenilim:
## Steps to reproduce the behavior

1. Create 2 accounts on the Mattermost server (A and B)
2. Log in with account A, create a private channel
3. Create a board associated with that private channel (so that only account A has access)
4. Copy and paste the link to the private board in a public channel (e.g. Town Square)
5. Log in with account B
6. Select the link in the public channel
7. The add board from template UI appears
8. Clicking on a template doesn't do anything
9. In addition, the sidebar is empty, and the user cannot navigate to the Dashboard

## Expected behavior

An access denied error is shown (existing standard "no workspace or access denied" error) when opening a workspace, board, or card you don't have access to.

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.10 RC1
 - Browser and OS: Chrome 93 on macOS

ogi-m:
Tested and passed on v0.10 - appropriate error shown.

PR #1773, opened 2021-11-08
GH-1685: Error if workspace is undefined

sbishel:
#### Summary
When a workspace is not defined, the api call returns an `undefined` workspace, not an error. This PR updates the initial load to throw an error if the workspace is not returned. 

#### Ticket Link
https://github.com/mattermost/focalboard/issues/1685


jespino (approved):
LGTM

PR #1776, opened 2021-11-08
Adds data migrations and the unique IDs migration

mgdelacroix:
#### Summary
This PR adds data migrations to the Boards server, and adds as well the Unique IDs data migration, to remove any unique ID that may exist in the blocks table.


jespino (changes requested):
I think this is not placed in the right package.

server/app/migrations.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/mattermost/focalboard/server/utils"
+
+	"github.com/mattermost/mattermost-server/v6/shared/mlog"
+)
+
+const (
+	UniqueIDsMigrationKey = "UniqueIDsMigrationComplete"
+)
+
+func (a *App) doUniqueIDsMigration() error {
```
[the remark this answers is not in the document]
mgdelacroix: All changes but adding tests appled on a87cb9a, PTAL

server/services/store/sqlstore/blocks.go
```diff
@@ -481,3 +481,74 @@ func (s *SQLStore) getBlock(db sq.BaseRunner, c store.Container, blockID string)
...
+	if err != nil {
+		s.logger.Error(`getBlocksWithSameID ERROR`, mlog.Err(err))
+		return nil, err
+	}
+	defer s.CloseRows(rows)
+
+	return s.blocksFromRows(rows)
+}
+
+func (s *SQLStore) replaceBlockID(db sq.BaseRunner, currentID, newID, workspaceID string) error {
```
wiggin77, line 507: From what I can see, this is only called from the migration code, which should be single-threaded.  Just want to point out that if this gets called concurrently the `runUpdateForBlocksAndHistory` methods will compete and create problems.

server/services/store/store.go
```diff
@@ -62,4 +63,8 @@ type Store interface {
 	HasWorkspaceAccess(userID string, workspaceID string) (bool, error)
 	GetWorkspaceCount() (int64, error)
 	GetUserWorkspaces(userID string) ([]model.UserWorkspace, error)
+
+	GetBlocksWithSameID() ([]model.Block, error)
```
jespino, line 67: I think this methods no longer need to be exposed.

server/services/store/sqlstore/data_migrations.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	// If the migration is already completed, do not run it again.
+	if hasAlreadyRun, _ := strconv.ParseBool(setting); hasAlreadyRun {
+		return nil
+	}
+
+	s.logger.Debug("Running Unique IDs migration")
+
+	blocks, err := s.GetBlocksWithSameID()
```
jespino, line 28: I would expect here to use the private method and also to do everything in the same transaction, to make it 1 single all or nothing. Also it is going to be more performant. And because this is being executed as part of the migrations process is going to be the only thing running at that time, so I prefer it more performant and atomic.

jespino (changes requested):
A couple of changes needed, bot otherwise looks good to me.

server/services/store/store.go
mgdelacroix: Addressed in 7f1da7c

server/services/store/sqlstore/data_migrations.go
mgdelacroix: Addressed in 7f1da7c

server/services/store/sqlstore/blocks.go
mgdelacroix: Right! Data migrations don't lock the database, so multiple nodes will compete to run them. The plugin API has a mechanism to create "advisory locks" using the plugin's `KVStore`, I'll take a stab on using it to sync the nodes and update the PR

server/services/store/sqlstore/public_methods.go
```diff
@@ -229,6 +239,27 @@ func (s *SQLStore) RefreshSession(session *model.Session) error {
 
 }
 
+func (s *SQLStore) ReplaceBlockID(currentID string, newID string, workspaceID string) error {
```
jespino, line 242: This are still here, probably you need to regenerate the public API

mgdelacroix: Arg, right!

jespino (approved):
LGTM. A tiny detail to fix, but not a blocker.

server/services/store/sqlstore/public_methods.go
mgdelacroix: Addressed in 6d349c8

server/services/store/sqlstore/blocks.go
mgdelacroix: Addressed in https://github.com/mattermost/focalboard/commit/9cf867ed9e74223083431c407f2d62b91861116a, PTAL

mgdelacroix:
@jespino re-requested your review because of the locking mechanism included in https://github.com/mattermost/focalboard/commit/9cf867ed9e74223083431c407f2d62b91861116a, PTAL

server/services/store/sqlstore/sqlstore.go
```diff
@@ -20,12 +21,12 @@ type SQLStore struct {
...
 	tablePrefix      string
 	connectionString string
-	isPlugin         bool
+	pluginAPI        plugin.API
 	logger           *mlog.Logger
 }
 
 // New creates a new SQL implementation of the store.
-func New(dbType, connectionString, tablePrefix string, logger *mlog.Logger, db *sql.DB, isPlugin bool) (*SQLStore, error) {
+func New(dbType, connectionString, tablePrefix string, logger *mlog.Logger, db *sql.DB, pluginAPI plugin.API) (*SQLStore, error) {
```
jespino, line 29: I don't like this, we are passing way too much information here. Can we receive here a mutexFactory function or can we receive something that matches an interface? I don't want to pass the whole API to the store.

jespino (changes requested):
I think the current implementation of the cluster mutex is too leaky.

server/services/store/sqlstore/sqlstore.go
mgdelacroix: Addressed in https://github.com/mattermost/focalboard/pull/1776/commits/25377a965ebcd36d76acafd71ab62ab2c038989e, PTAL

server/services/store/sqlstore/migrate.go
```diff
@@ -176,9 +176,9 @@ func (s *SQLStore) Migrate() error {
 	}
 
 	var mutex *cluster.Mutex
-	if s.pluginAPI != nil {
+	if s.IsPlugin {
 		var mutexErr error
-		mutex, mutexErr = cluster.NewMutex(s.pluginAPI, "Boards_dbMutex")
+		mutex, mutexErr = s.NewMutexFn("Boards_dbMutex")
```
jespino, line 181: This is always defined in this context, but you don't need to know that. So, if you receive 2 parameters, they should be checked independently, in this case something like, `if s.IsPlugin and s.NewMutexFn != nil` and you have to modify the other calls using the mutex too.

jespino (approved):
LGTM

server/services/store/sqlstore/data_migrations.go
```diff
@@ -0,0 +1,58 @@
...
+	for _, blocks := range blocksByID {
+		for i, block := range blocks {
+			if i == 0 {
+				// do nothing for the first ID, only updating the others
+				continue
+			}
+
+			newID := utils.NewID(model.BlockType2IDType(block.Type))
+			if err := s.ReplaceBlockID(block.ID, newID, block.WorkspaceID); err != nil {
+				return fmt.Errorf("cannot replace blockID %s: %w", block.ID, err)
```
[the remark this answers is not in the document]
wiggin77: I'm wondering how safe it is to continue using the product with the migration pending on next run.

server/services/store/storetests/system.go
```diff
@@ -7,6 +7,23 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// this system settings are created when running the data migrations,
```
wiggin77, line 10:
```suggestion
// these system settings are created when running the data migrations,
```

PR #1786, opened 2021-11-09
Only allow routes without workspace prefix if you are in non-plugin mode

### LINKED ISSUE #1702: Bug: Card links don't open the card dialog when opened in the same tab

1. Post a link to a card in channels
2. Click on the link (not the card preview) - user is taken to boards (in the same tab) and the card dialog opens
3. Return to channels
4. Click on the card link again - user is taken to boards but the card dialog doesn't open

## Expected behavior

The card dialog should open every time the user opens a link to a card

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.10
 - Browser and OS: Chrome 94 on Windows

## Additional context
This happens with card links that open in the same tab. If it opens in a new tab (which is the case when clicking on a card preview), then it opens the card dialog successfully every time.
- Opening a card link in the same tab doesn't open the card
- Opening the card link in a new tab successfully opens the card

jespino:
Fixes #1702

### DIFF
--- a/webapp/src/app.tsx
+++ b/webapp/src/app.tsx
@@ -216,31 +216,32 @@ const App = React.memo((): JSX.Element => {
                                     <WelcomePage/>
                                 </Route>
 
-                                <Route
-                                    path='/:boardId?/:viewId?/:cardId?'
-                                    render={({match: {params: {boardId, viewId, cardId}}}) => {
-                                        // Since these 3 path values are optional and they can be anything, we can pass /x/y/z and it will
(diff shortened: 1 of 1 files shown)


hahmadia:
I can approve but I am in favor of testing this before merge as it is touching a brittle part of our codebase that may or may not have effect elsewhere. Just want to confirm we will get testing before I approve.

sbishel (approved):

ogi-m (approved):
LGTM

PR #1800, opened 2021-11-11
 425|Added unit tests for workspaces.go

### LINKED ISSUE #425: Server Unit Tests - ./app/workspaces.go
Write Unit Tests for ./server/app/workspaces.go

use mock for store layer
see ./server/app/auth_test.go and ./server/app/blocks_test.go for samples.

Prassud:
Summary
Added Unit test cases for server/app/worksapces.go 

Ticket Link
[Link](https://github.com/mattermost/focalboard/issues/425)

fixes #425


server/app/workspaces.go
```diff
@@ -20,12 +20,12 @@ func (a *App) GetRootWorkspace() (*model.Workspace, error) {
 		}
 		err := a.store.UpsertWorkspaceSignupToken(*workspace)
 		if err != nil {
-			a.logger.Fatal("Unable to initialize workspace", mlog.Err(err))
```
wiggin77, line 23: This does more than change the log level.  Fatal exited the app with an error code as this condition was considered unrecoverable, and this behaviour would be changed by switching to `Error`.  Is this intentional?

Prassud: Wherever GetRootWorkspace method is referred in codebase, there is logic to check error is nil, Since Fatal was terminating the application, the logic on the other modules are not useful, so made it as error. and added test cases for the same

below is the code sample from server.go
`      // Init workspace
	if _, err := app.GetRootWorkspace(); err != nil {
		params.Logger.Error("Unable to get root workspace", mlog.Err(err))
		return nil, err
	}`

So, that's why made this change and added the testcase

hahmadia: @wiggin77 Contributor has responded

wiggin77 (approved):
LGTM :+1:   Thanks @Prassud.  Just a minor formatting issue to fix to make the linter happy then it can be merged.

Prassud:
@wiggin77 @hahmadia  Fixed the lint errors.

PR #1807, opened 2021-11-12
Shows the version change banner only on plugin update and tries to reconnect when the interface is refreshed

mgdelacroix:
#### Summary
The new version banner is showing on every plugin update, as there is no way to distinguish which plugin has been changed in the websocket message. This PR compares the new Boards version with the current one and only shows the version banner in that case.

For the cases where an update happens and the interface is refreshed, the previous fix would not work because of the plugin lifecycle, so we trigger a delayed reconnect as a temporal solution so the interface doesn't become unresponsive.

A more complete solution would be to ACK the websocket subscription message so the plugin can retry until the server acknowledges the subscription and could show a warning in the meantime.

### DIFF
--- a/mattermost-plugin/webapp/src/index.tsx
+++ b/mattermost-plugin/webapp/src/index.tsx
@@ -67,7 +67,7 @@ type Props = {
 }
 
 const MainApp = (props: Props) => {
-    wsClient.initPlugin(manifest.id, props.webSocketClient)
+    wsClient.initPlugin(manifest.id, manifest.version, props.webSocketClient)
 
     useEffect(() => {
         document.body.classList.add('focalboard-body')
--- a/webapp/src/utils.test.ts
+++ b/webapp/src/utils.test.ts
@@ -132,4 +132,18 @@ describe('utils', () => {
             expect(Utils.displayDateTime(date, intl)).toBe(`July 09, ${previousYear}, 5:35 AM`)
         })
     })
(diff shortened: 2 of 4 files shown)


jespino (approved):
LGTM

chenilim:
My initial reaction is not to take this for v0.10, given how close we are to release. From the description, this issue will only repro when any plugin is updated, which is relatively rare. And even in that case, clicking the banner will hide it without other side effects. Do you see other potential issues not taking this for v0.10? cc @ogi-m, @wuwinson, @asaadmahmood, @sbishel.

Anyone want to make the case for taking it for v0.10?

mgdelacroix:
The only risk I see is the confusion it would potentially create in servers like Community, where plugins are updated constantly. I'd cherry-pick it to the `release-0.10` branch still so the fix is included in any dot versions we might release.

The other benefit of the PR, which is the reconnect on plugin update, will actually work if included in the destination plugin (I mean, if the server is updated from `v0.10.0` that doesn't have the fix to `v0.10.1` that does) because of how the plugin lifecycle works, so that'd be another benefit of cherry-picking even if we don't include it in `v0.10.0`.

chenilim:
Thanks Miguel. Agreed. Let's cherry-pick to the `release-0.10` branch, but not make another build of v0.10.0. We can ship this if there is a dot release.

PR #1836, opened 2021-11-16
GH-1835: Show Autocomplete Textbox when At Mentioning Users in Comments or Content Section

### LINKED ISSUE #1835: Feature Idea: Show Autocomplete Textbox when At Mentioning Users in Comments or Content Section
## Summary

Showing Autocomplete Box when At Mentioning a user in a comment or content

Importance: High

hahmadia:
#### Summary

image

#### Ticket Link

Fixes #1835


webapp/src/components/simpleMentionEditor/simpleMentionEditor.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+type Props = {
+    onChange?: (text: string) => void
+    onFocus?: () => void
+    onBlur?: (text: string) => void
+    initialText?: string
+    id?: string
+}
+
+const SimpleMentionEditor = (props: Props): ReactElement => {
```
jespino, line 36: I don't like this name, it is just showing an implementation detail that is not more important that others. It should be name something like `MarkdownEditorInput` or something like that.

jespino (changes requested):
I tiny change proposal related to the naming, but otherwise looks good to me.

webapp/src/components/simpleMentionEditor/simpleMentionEditor.tsx
hahmadia: Yes you are right.

webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+        return {plugins, MentionSuggestions, EmojiSuggestions}
+    }, [])
+
+    useEffect(() => {
+        if (ref) {
+            setTimeout(() => ref.current?.focus(), 200)
+        }
+    })
+
+    const customKeyBindingFn = (e: React.KeyboardEvent) => {
```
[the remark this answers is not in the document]
hahmadia: Fair enough. Addressed.

webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const handleKeyCommand = (command: string): DraftHandleValue => {
+        if (command === 'editor-blur') {
+            ref.current?.blur()
+            return 'handled'
+        }
+
+        return 'not-handled'
+    }
+
+    const onEditorStateBlur = () => {
```
jespino, line 93: I meant all the callbacks, because there are multiple of them, and you are passing a lot of them to the editor.

hahmadia: Sorry I missed that. I've used it on all callbacks. 

Please double check to make sure that I've not missed any dependency on the callbacks. Thank you

webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+        setIsEmojiPopoverOpen(true)
+    }, [])
+
+    const onEmojiPopoverClose = useCallback(() => {
+        setIsEmojiPopoverOpen(false)
+    }, [])
+
+    const onSearchChange = useCallback(({value}: { value: string }) => {
+        setSuggestions(defaultSuggestionsFilter(value, mentions))
+    }, [])
```
jespino, line 118: `mentions` should be in the dependencies.

webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const onEditorStateBlur = useCallback(() => {
+        const text = editorState.getCurrentContent().getPlainText()
+        onBlur && onBlur(text)
+    }, [editorState])
+
+    const onEditorStateChange = useCallback((newEditorState: EditorState) => {
+        const newText = newEditorState.getCurrentContent().getPlainText()
+        onChange && onChange(newText)
+        setEditorState(newEditorState)
+    }, [])
```
jespino, line 102: onChange should be in the dependencies.

webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
```diff
@@ -0,0 +1,149 @@
...
+    }, [isEmojiPopoverOpen, isMentionPopoverOpen])
+
+    const handleKeyCommand = useCallback((command: string): DraftHandleValue => {
+        if (command === 'editor-blur') {
+            ref.current?.blur()
+            return 'handled'
+        }
+
+        return 'not-handled'
+    }, [ref])
```
jespino, line 91: The ref shouldn't change ever, a reference is a reference, and is created one time, and you can change it over time, but the ref itself doesn't change. I think is not needed here.

jespino (changes requested):

webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
hahmadia: Okay!

jespino (approved):
LGTM

harshilsharma63:
@hahmadia  The plugin crashed for me in these scenerios. Interestingly, it only crashes on FIrefox and works fine on Chrome. Can you try it out on Firefox?

1. Clicking on the comment typing area and then on the summary area without typing anything crashes Boards. The same happens if you reverse the order.
2. 
https://user-images.githubusercontent.com/18575143/142969604-70ce6e6b-6799-4020-afdf-e419266c3e2b.mov

2. Clicking on the comments area, then clicking out of it (but within the car), then clicking in the comments area again crashes the plugin. The same thing happens with the summary area.

https://user-images.githubusercontent.com/18575143/142969733-cbc2e0ef-13a4-4570-ad08-a61850fd4036.mov

3. It also seems to crash when clicking in the comments or summary area the second time.

harshilsharma63 (changes requested):
Code changes look good. Only need to address the Firefox issue.

hahmadia:
Hi @jespino / @harshilsharma63 ,

It was discovered that DraftJS did not like being mounted and unmounted on Firefox. Doing so would cause the internals to crash. Therefore, the work around I've used here is to instead hide the DraftJS input field when we are not editing it and display it when we are. 

This seems to work fine and resolves our problem. Let me know if you guys see any issues with these changes.

harshilsharma63:
@hahmadia Awesome! The app crash issue is resolved and works great now.

One UI issue though, otherwise changes look good. The autocomplete list doesn't use the theme colors-
Screenshot 2021-11-25 at 4 43 21 PM

harshilsharma63 (approved):
LGTM!

hahmadia:
> @hahmadia Awesome! The app crash issue is resolved and works great now.
> 
> One UI issue though, otherwise changes look good. The autocomplete list doesn't use the theme colors- Screenshot 2021-11-25 at 4 43 21 PM

@harshilsharma63 I will create a follow up ticket to solve this separately.

ogi-m (approved):
LGTM, I've logged the outstanding issues separately

webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
```diff
@@ -0,0 +1,172 @@
...
+        const plugins = [
+            mentionPlugin,
+            emojiPlugin,
+        ]
+        return {plugins, MentionSuggestions, EmojiSuggestions}
+    }, [])
+
+    useEffect(() => {
+        setTimeout(() => ref.current?.focus(), 200)
+    })
```
kamre, line 69: @hahmadia why do we need this `useEffect`? It breaks setting focus to card title textarea.

hahmadia: @kamre Good call. I'll fix this in a PR.
We need it so that when you are in editing mode, it auto focuses on the DraftJS editor. Otherwise, it won't if this is not called.

kamre: @hahmadia I have created pull request #1952 with the proposed fix, please review it.

PR #1849, opened 2021-11-19
[GH-1823] Added E2E test for MM-T4284

venarius:
#### Summary
Hey, everyone! This is my first commit. I would be grateful for any feedback!

This commit adds a cypress E2E test for MM-T4284

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/1823


wiggin77 (approved):
LGTM :+1:

saturninoabril (changes requested):
@venarius Thank you for your contribution! Could you please update to latest main since there's a recent commit that changed basic Cypress setup?
I'll look into the details once updated. Also, please share a screenshot that the test is passing on your local dev.

venarius:
@saturninoabril 
image

webapp/cypress.json
```diff
@@ -3,7 +3,8 @@
     "baseUrl": "http://localhost:8088",
     "testFiles": [
         "**/login*.ts",
-        "**/create*.ts"
+        "**/create*.ts",
+        "**/manage*.ts"
```
venarius, line 7: This seems somewhat weird to me. Why not suffix all cypress integration files with `.spec.ts` so we could do `"**/*.spec.ts"` here instead?

hahmadia (approved):

webapp/cypress.json
saturninoabril: I agree. It's done this way to maintain order but we'll change it next time for a better approach.

saturninoabril (approved):
Thanks @venarius! LGTM.

PR #1886, opened 2021-11-29
Close migrations

agnivade:
This was the original reason behind focalboard
not working with the DB RPC API.

Because we were not closing the migrations instance,
the original connection was never being
released and therefore we ran out of connections.

The fix we did was kind of a workaround
because we would open a new DB instance and then close
it explicitly which would close the connection also,
but that meant we could not reuse the connection.

Closing the driver also closes the source driver
which was also something that wasn't being done.

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4/cxg16f8nro7bhxjdwpnqdnumj5o

### DIFF
--- a/server/services/store/sqlstore/migrate.go
+++ b/server/services/store/sqlstore/migrate.go
@@ -100,14 +100,10 @@ func appendMultipleStatementsFlag(connectionString string) (string, error) {
 // migrations in MySQL need to run with the multiStatements flag
 // enabled, so this method creates a new connection ensuring that it's
 // enabled.
-func (s *SQLStore) getMigrationConnection() (*sql.DB, error) {
-	connectionString := s.connectionString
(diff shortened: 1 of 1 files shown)


agnivade:
I do not see an easy way to add a unit test for this without mocking the database in some way. But please let me know if there is a way

agnivade:
Ah I get it now, `golang-migrate` will not only close the connection, but also the db instance at the same time. So we cannot just reuse the instance because it will be closed as well. That's unfortunate, because the API wasn't designed to reuse connections.

cc @isacikgoz on this as well as a note for morph.

Meanwhile, let me update the PR.

PR #1890, opened 2021-11-29
Don't insert nulls into blocks_history table.

wiggin77:
#### Summary
Currently when we delete a block we insert a record in the blocks_history table with the same id but the following fields are NULL:   parent_id, root_id, schema, type, fields, title, create_at, created_by. 

This leaves us with an incomplete history of the block, and also fails when querying the table since row scan is not expecting NULLs.

This PR ensures all the fields are populated with values from the time the block was deleted.  

#### Ticket Link
https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/150d1464-6d59-4c39-810f-ee201439dbaf/cofg9io7z6jdbzreo3xcot54rfw


sbishel (approved):
LGTM

wiggin77:
I added data migration for any records in blocks_history with NULLs.  It tries to populate via other versions of the blocks.

server/services/store/sqlstore/migrations/migrations_files/000015_blocks_history_no_nulls.up.sql
```diff
@@ -0,0 +1,85 @@
...
+			FROM {{.prefix}}blocks_history bh2
+			WHERE bh1.id = bh2.id AND bh2.fields IS NOT NULL 
+			ORDER BY bh2.insert_at ASC limit 1) 
+		, '')
+	WHERE fields IS NULL;
+{{end}}
+
+-- create_at
+UPDATE {{.prefix}}blocks_history bh1
+	SET create_at = update_at
```
jespino, line 64: Accordingly to the previous code, the create at is the create at of the original block, that I think is not necesarilly the create at of the history entry. I would guess this needs a migration similar to the previous one, or we need to change the code to make the `createAt`, of the history entry the date of the creation of the history entry, not the creation of the original block.

jespino (changes requested):
I tiny inconsistency around the create_at field in the database, but other than that looks good to me.

Probably the migration can be done in less queries, but I think this way is easier to understand, and I don't think the performance implications are going to be relevant at this stage of the project.

server/services/store/sqlstore/migrations/migrations_files/000015_blocks_history_no_nulls.up.sql
wiggin77: I agree.  I'll modify to get the correct createAt.

wiggin77: Done

jespino (approved):
Nice!

Issue #1889, opened 2021-11-29
Bug: Can't Acces Config File if Server runs in Docker Container


Genei180:
## Steps to reproduce the behavior

1. run the Focal board in a Docker Container as Described in README.
2. try to Access the Config File in the Docker

## Expected behavior

A Mounted Volume that Allows for Easy Access to edit the Config file.
To Edit and Changes Setting like Described in: https://www.focalboard.com/guide/admin/

## Edition and Platform

 - Edition: Personal Server (Docker)
 - Version: v0.10.0

## Additional context

Working on it...

PR #1891, opened 2021-11-29
Added Docker Config Volume

Genei180:
#### Summary
There were three Possible Choices to Implement a Mount for a Config File in Docker:

1) docker run -v:
You could specify a Volume at the Start of a Container, but it has the Problem that the Folder on the Container, gets overwritten by the Host. So the Config File would have needed to be Included Explicitly. Additional, it would have needed to be separated from the Other files, because they would have been deleted otherwise in the Process of mounting. 

2) Docker Volume:
Instead of creating it at runtime, it could be created beforehand, which would solve the Above Issue and would Allow for Sharing of Files from the Container with the Host. But this would complicate the Setup needed even further. Therefore, three was chosen. 

3) Dockerfile Volume:
As the focalboard Database already gets its own Volume Created in the Dockerfile to persist, I added a Volume Creation for the Config File as well in the Dockerfile. This requires less Setup, while still allowing for Easy Access of the Config File.

The Config File Source on the Host System can be found via the following Command:
`docker inspect *Container Name* | grep -B 1 "/opt/focalboard" | grep "Source"`
Be Aware that with a WSL 2 Docker Setup under Windows, the Folder Location differs from the one displayed.

(Note: The Container needs to be Restarted for the Updates to take Place)

#### Ticket Link
Provides Solution too https://github.com/mattermost/focalboard/issues/1889

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,6 +33,7 @@ EXPOSE 8000/tcp
 
 EXPOSE 8000/tcp 9092/tcp
 
+VOLUME /opt/focalboard
 VOLUME /data
 
 CMD ["/opt/focalboard/bin/focalboard-server"]


jespino (approved):
Looks good to me. Thanks @Genei180! :tada:

Genei180:
I am currently Working on a Update, to make it Possible to Include the COnfig via the -v Option.

The Config.json needs to be moved into its own Directory for that, in my Understanding.

Genei180:
Well as is it turns out the Problem i was trying to avoid cant be solved with the Current State of Docker.
I was trying to somehow provide a config.json file in the Bind Mound on the first Run for the User to easily edit.

It would have been Possible if some sort of Shell would be Installed in the Container, that would have allowed copying of the Config File into the Mounted Folder on Runtime.
But because the Container is Distroless this is not easily Possible. 

Instead, a Pre-written Config File now needs to be Copied into the Mounted Folder.

Nonetheless, this should now allow for easy and persisting Configuration that is Compatible with Docker Compose, not like the Previous Setup Chosen.

Genei180:
What's Up with the `sudo chown -R nobody /home/user/focalboard-data` in the Merge Conflict?

What Purpose does it have?
Additionally, it's an Linux only Command?
Should this maybe be Split up into Multiple Instructions for Certain Operating Systems?

wjjjjyourFA:
Hi,it seems that I can't use "volumes: - /path/config:/config" in docker-compose file.

PR #1895, opened 2021-11-30
Adding explict logout logic

jespino:


webapp/src/octoClient.ts
```diff
@@ -75,8 +75,18 @@ class OctoClient {
...
     }
 
-    logout(): void {
+    async logout(): Promise<boolean> {
+        const path = '/api/v1/logout'
+        const response = await fetch(this.getBaseURL() + path, {
+            method: 'POST',
+            headers: this.headers(),
+        })
         localStorage.removeItem('focalboardSessionId')
```
jespino, line 84: Probably we should delete this only if the logout API call is a success, but we would need to notify that to the user somehow. This is the initial approach, but we should make it more explicit, so if the session is not removed from the server (for any reason) we don't consider you logged out, and we just show you an error message. But that would be needed to be done as another task.

wiggin77: IMO we should warn if user was not completely logged out, but still remove the local cookie.  This could be someone stepping away from a public computer and we should not be leaving a cookie there when they explicitly asked us not to.

wiggin77 (approved):
LGTM :+1:   From what I can see it currently removes the local cookie even if deleting the session fails, which is how I think it should be.

sbishel (approved):
LGTM...Thanks.

PR #1905, opened 2021-12-01
Add unit tests for Archiver

jufab:
#### Summary

Add unit tests for Archiver.

Add [web-encoding](https://www.npmjs.com/package/web-encoding) to help test around text encoder.

#### Ticket Link

No ticket


webapp/src/lineReader.ts
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+import {TextDecoder} from 'web-encoding'
```
chenilim, line 3: TextDecoder doesn't appear to be used? Do we need to import web-encoding?

chenilim (changes requested):
Thanks @jufab! Very cool use of the input mocks. Do we need to import web-encoding though?

webapp/src/lineReader.ts
hahmadia: To add on to this, was this added by accident? Why did you install `web-encoding` in package.json?

hahmadia (changes requested):
Why install `web-encoding` and what is it? I think it's a mistake. Please remove and I will re-review your PR. Thanks

webapp/src/lineReader.ts
jufab: TextDecoder is not present in jsdom
I always have this error if I want to do something with TextDecoder.
```console
    ReferenceError: TextDecoder is not defined

      28 |         let offset = 0
      29 |         const fr = new FileReader()
    > 30 |         const decoder = new TextDecoder()
         |                             ^
      31 | 
      32 |         fr.onload = async () => {
      33 |             const chunk = new Uint8Array(fr.result as ArrayBuffer)

      at Function.readFile (src/lineReader.ts:30:29)
      at src/archiver.ts:56:24
      at Function.importBlocksFromFile (src/archiver.ts:55:16)
      at HTMLInputElement.input.onchange (src/archiver.ts:116:32)
      at HTMLInputElement.<anonymous> (node_modules/jsdom/lib/jsdom/living/helpers/create-event-accessor.js:32:32)
      at innerInvokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:318:25)
      at invokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:274:3)
      at HTMLInputElementImpl._dispatch (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:221:9)
      at HTMLInputElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:94:17)
      at HTMLInputElement.dispatchEvent (node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:231:34)
      at node_modules/@testing-library/dom/dist/events.js:25:20
      at Object.eventWrapper (node_modules/@testing-library/dom/dist/config.js:27:23)
      at fireEvent (node_modules/@testing-library/dom/dist/events.js:16:35)
      at Function.fireEvent.<computed> [as change] (node_modules/@testing-library/dom/dist/events.js:125:36)
      at Object.<anonymous> (src/archiver.test.ts:94:23)
```

So, the only way I found is to use this lib .

But, I found an another way to use it only in tests and not in code prod => see my last commit :)

_I found this unresolved issue about this problem : https://github.com/jsdom/jsdom/issues/2524_

Issue #1879, opened 2021-11-25
Bug: Users can receive a mention notification when they're no longer a member of the channel/board


ogi-m:
## Steps to reproduce the behavior

1. Have a channel with User-1 and User-2
2. As User-1, open the associated board
3. Open a card and click on the comment field
4. Type in "@" and make sure User-1 is listed
5. As User-2, leave the channel that's associated with the board
6. As User-1, type in "@" in the comment field again (without refreshing)
7. Select User-1 from the autocomplete and post the comment - User-1 receives a notification for the mention

## Expected behavior

Users shouldn't be able to receive notifications for boards they no longer belong to

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.11
 - Browser and OS: Chrome & Firefox

## Additional context

User-1 can only see the card title, the card doesn't unfurl and they can't access it but I would expect them to not get a notification at all.
Also, if the board is refreshed then the User-1 no longer shows up in autocomplete.

sbishel:
@asaadmahmood to think about design for this.

asaadmahmood:
3 things.
1. This is a bug, as it first shows a user that is no longer in the channel associated with the board.
2. And then it also sends a notification.
3. In the future, when we allow mentioning users across the system, we should show a header above the users who are not part of the board. And that should probably be fine for now.

![Boards Card Detail](https://user-images.githubusercontent.com/11034289/144441882-15e5f291-9795-418d-89d7-8fd0fc2acb50.png)
![Boards Card Detail-1](https://user-images.githubusercontent.com/11034289/144441889-437b2bce-3e0c-42a4-935f-c956537c4478.png)

asaadmahmood:
That UI is only application when we implement the role based permissions and detach channels from boards. Right now, the user directly should be removed from the autocomplete list.

ogi-m:
Do we want to split the autocomplete part into a separate issue? That only happens if the board isn't refreshed after the user is removed from the channel and before trying to at-mention. If the board is refreshed, only members will be shown in the autocomplete.
However, they can still manually type out the username of a user outside of the channel and the notification will be send. Those are two separate issues I believe.

sbishel:
@wiggin77 - assigning to you, I think there are both server and front end fixes needed. Although the server should be fixed here. https://github.com/mattermost/focalboard/pull/1216

wiggin77:
[...]

PR #1912, opened 2021-12-02
Don't notify @mentioned users who are not members of channel

wiggin77:
#### Summary
This PR ensures that users mentioned on a card that are not members of the channel do not get notified.  This is a silent fail, a if the author typed a non-existent user e.g. `@nouser`

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/1879

### DIFF
--- a/mattermost-plugin/server/notifications.go
+++ b/mattermost-plugin/server/notifications.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/mattermost/focalboard/server/services/notify"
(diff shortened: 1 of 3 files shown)


sbishel (commented):
How do these errors display to the Front-end? If reasonable, perhaps we should throw an error instead of failing silently. Especially since we know the difference in a user that is not on the team and just an @word.

wiggin77:
> How do these errors display to the Front-end? If reasonable, perhaps we should throw an error instead of failing silently. Especially since we know the difference in a user that is not on the team and just an @word.

Currently they don't.  As you mentioned above, we don't have a mechanism for that, and I don't think doing error UI specific to this is worthwhile.  We should design a generic way to display server generated errors to users.

asaadmahmood:
Yes, @sbishel I think once we show users that are outside of the system, we can show what I have in my mockup.
But for now, we don't show anything if the user just types an @ sign and then a user that doesn't even exist in the popup.

sbishel (approved):

wiggin77:
@ogi-m do you want to look at this pre or post-merge?

ogi-m:
Post-merge is fine, thanks @wiggin77

Issue #1731, opened 2021-11-02
Bug: Horizontal drag&drop scrolling is difficult


wiggin77:
## Steps to reproduce the behavior

1. Create board view with cards in more groups than can fit on screen
2. Position bottom scroll bar to the far right
3. Grab a card and drag it to the left
4. Notice as you reach the edge of screen that it scrolls a bit, but you have to move the cursor back and forth to continue scrolling

## Expected behavior
Nice smooth scrolling experience

## Screenshots

![horz_scroll](https://user-images.githubusercontent.com/5024605/139908371-98388863-ef0a-4fb2-8c6f-cf0561cc2dba.gif)

## Edition and Platform

 - Edition: Mattermost plugin
 - Version:  v0.10.0
 - Browser and OS: Linux + Firefox 93

chenilim:
Putting this up for grabs - it seems that the direct improvement would be to increase the area for which the board scrolls. What do you think @asaadmahmood?

PR #1913, opened 2021-12-03
Added better kanban edge scroll detection whilst dragging

venarius:
#### Summary
This adds better edge detection for the kanban board while dragging a card. Before that it used the default browser implementation of detecting edge-scrolling but this works only on a very narrow space and not at all on mobile.

This PR adds full edge-scrolling-support for both mobile and desktop variants.

https://user-images.githubusercontent.com/13546486/144538257-a28064ad-9a8b-4eb7-b6ae-8a9bb84674b5.mp4

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/1731


wiggin77:
This looks great @venarius.  Will you be able to address the unit test issue?  It may just need the Jest snapshot updated.

venarius:
@wiggin77 yes sure!

venarius:
Somehow the cypress on the CI doesn't like the test that I added for the scroll detection. It works locally on my machine though using the same cypress version. Does someone maybe know why or should I remove that test?

sbishel (approved):

sbishel:
Works good. @ogi-m or @saturninoabril  Any idea why cypress fails on CI but passes locally?

saturninoabril:
I'm out yesterday but will take a look today.

webapp/cypress/integration/createBoard.ts
```diff
@@ -112,4 +112,45 @@ describe('Create and delete board / card', () => {
         cy.get('.DeleteBoardDialog button.danger').click({force: true})
         cy.contains(boardTitle).should('not.exist')
     })
+
+    it('scrolls the kanban board when dragging card to edge', () => {
```
saturninoabril, line 116: @venarius Requesting change below to include new test case identifier.
```suggestion
    it('MM-T4433 Scrolls the kanban board when dragging card to edge', () => {
```

@ogi-m Please help check the test case I've created related to this feature - [MM-T4433](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testCase/MM-T4433). Thanks in advance!

webapp/cypress/integration/createBoard.ts
```diff
@@ -112,4 +112,45 @@ describe('Create and delete board / card', () => {
...
+        cy.visit('/')
+
+        // Create empty boards
+        cy.log('**Create new empty board**')
+        cy.contains('+ Add board').click({force: true})
+        cy.contains('Empty board').click({force: true})
```
saturninoabril, line 122: Please avoid the use of force when doing an action as much as possible. I understand that it's been the pattern in other places but we'll update it soon. In here, looks like we can completely remove it but in other cases, you may add extra "guards" like visibility check (`.should('be.visible')`) before chaining an action.
The succeeding step to create group intermittently failed on me, related to "detached DOM", I'd suggest to add extra wait as well, like:
```suggestion
        cy.contains('+ Add board').should('be.visible').click()
        cy.contains('Empty board').click().wait(1000)
```

webapp/cypress/integration/createBoard.ts
```diff
@@ -112,4 +112,45 @@ describe('Create and delete board / card', () => {
...
+
+        // Create 10 empty groups
+        cy.log('**Create new empty groups**')
+        for (let i = 0; i < 10; i++) {
+            cy.contains('+ Add a group').click({force: true})
```
saturninoabril, line 127:
```suggestion
            cy.contains('+ Add a group').scrollIntoView().should('be.visible').click()
```

webapp/cypress/integration/createBoard.ts
```diff
@@ -112,4 +112,45 @@ describe('Create and delete board / card', () => {
...
+
+        // Create empty card in first group
+        cy.log('**Create new empty card in first group**')
+        cy.contains('+ New').click({force: true})
+        cy.get('.Dialog').should('exist')
+        cy.get('.Dialog .wrapper').click({force: true})
+        cy.get('.KanbanCard').should('exist')
+
+        // Drag card to right corner and expect scroll to occur
+        cy.get('.Kanban').invoke('scrollLeft').should('equal', 0)
```
saturninoabril, line 139: Looks like something is re-rendering in the background that's causing the first instance of the card to be "detached". Could you add wait time and see if it will fix the failed test?
```suggestion
        cy.get('.Kanban').invoke('scrollLeft').should('equal', 0).wait(1000)
```

webapp/package.json
```diff
@@ -20,8 +20,7 @@
 		"cypress:run:firefox": "cypress run --browser firefox",
 		"cypress:run:edge": "cypress run --browser edge",
 		"cypress:run:electron": "cypress run --browser electron",
-		"cypress:open": "cypress open",
-		"updatesnapshots": "jest --updateSnapshot"
```
saturninoabril, line 24: I wonder why this was deleted.

saturninoabril (changes requested):
Thanks @venarius! Please see comments.

webapp/package.json
venarius: This line is redundant with `updatesnapshot` on line 11 so I removed it.

venarius:
Thanks for the awesome feedback, @saturninoabril - The additional waits seem to have fixed the test. Now there seems to be another unrelated test to be failing but maybe that's just that one CI run.

saturninoabril:
Thanks @venarius for updating! You're right that the failed tests were totally not related to this PR. I could submit a separate PR to fix it so that the change here is neat but I think it makes sense to have it here directly. If you don't mind, I could try to push some changes and see if it works in CI.

venarius:
@saturninoabril sure, go ahead!

webapp/cypress/integration/createBoard.ts
```diff
@@ -112,4 +108,43 @@ describe('Create and delete board / card', () => {
...
+            wait(500)
+
+        // wait necessary to trigger scroll animation for some time
+        // eslint-disable-next-line cypress/no-unnecessary-waiting
+        cy.get('.Kanban').
+            trigger('dragover', {clientX: Cypress.config().viewportWidth - 100, clientY: Cypress.config().viewportHeight / 2}).
+            wait(1500).
+            trigger('dragend')
+
+        cy.get('.Kanban').invoke('scrollLeft').should('not.equal', 0)
```
saturninoabril, line 148: @venarius Looking at https://github.com/mattermost/focalboard/issues/1731, it's about scrolling behavior from right to left. With that, could you please add another step/check that verifies such scroll direction.

saturninoabril (changes requested):
@venarius Cypress test seemed to be passing now.

The scroll behavior going to left position has improved so much. However, when I dragged the element far left (edge) or even at left sidebar, scrolling stopped unless I moved a bit back once again. Doing the same at far right edge is a smooth experience.

webapp/cypress/integration/createBoard.ts
venarius: @saturninoabril I added another check for the other direction but imho I would advise against adding this extra step as this adds 2 more waiting-seconds and this could stack up a lot on later CI runs when there are more cypress tests.

saturninoabril: Not sure if you noticed but the added step is not working/scrolling to the left. Maybe because the card is not picked up from the right?
Noted on extra overhead. If going to choose, then I'd recommend to test scroll behavior from right to left, to align with the original issue.

wiggin77 (approved):
LGTM :+1:    Just some minor merge conflicts to address.  Great work @venarius

webapp/cypress/integration/createBoard.ts
venarius: yeah, I changed the test to be from right to left 👍

venarius:
@wiggin77 just rebased - seems like now some other unrelated cypress tests are failing 😅

wiggin77:
@saturninoabril were you able to fix the unrelated cypress failures?

saturninoabril:
fyi, I'll start looking into this tomorrow.

saturninoabril (approved):
Looks good to me, thanks @venarius!

Issue #1841, opened 2021-11-16
Bug: New properties don't get added to table view by default


ogi-m:
## Steps to reproduce the behavior

1. Open Focalboard Personal Desktop (doesn't repro on plugin)
2. Create an empty board
3. Add table view
4. Add a card in table view
5. Add a text property
6. Close the card - note the new property isn't displayed in table view, it needs to be toggled manually in the Properties dropdown.

## Expected behavior

The new text property should be shown by default in table view

## Screenshots

If applicable, add screenshots to elaborate on the problem.

## Edition and Platform

 - Edition: Personal Desktop
 - Version: main
 - Browser and OS: Windows

lindy65:
Adding that this is an intermittent issue.  At times, new properties are added to table view automatically and, at others, you need to toggle the property to "on" under Properties menu.

chenilim:
Discussed with Scott and Winson - we actually expected the property not be be auto-added to the table. i.e. the user has to manually add it in the properties menu.

lindy65:
I'll keep an eye on final decision as Rainforest tests will need to be refactored if property is not auto-added to the table (as previously properties were auto-added)

sbishel:
This is being caused by the transaction processing when using a sqlite database. 

The functionality to auto add to the table has existed since the beginning, so we should probably think about whether we want to change the functionality.

lindy65:
Thanks Scott,

> The functionality to auto add to the table has existed since the beginning

I'm seeing _intermittent_ failure in Rainforest tests where property is _not_ auto added to the table so Ogi filed this issue.  For now, I've disabled the affected tests from running until a decision is made on whether to change the functionality or not.

chenilim:
I was able to repro this issue on both:
* v0.11 RC2 Personal Desktop Mac with a new DB
* Private build off release-0.11.

macOS 11.16.1 and SQLite 3.32.3.

PR #1935, opened 2021-12-06
Gh 1841 batch patches

sbishel:
#### Summary
When calling `updateBlocks` in the mutator.ts. The code was sending these to the server in separate `updateBlock` methods. `updateBlock` employed transactions and occasionally with SQLite, the second update would be called before completion of the first transaction. This was causing an error with SQLIte and a partial update. 

This PR moves `patchBlocks` to the server and places the entire thing into a transaction. In addition, it updates `insertBlocks` on the server to place the insert for all blocks in a single transaction.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/1841


ogi-m (approved):
Tested and passed
- working as expected now
- tested Personal Desktop and plugin

server/api/api.go
```diff
@@ -676,6 +677,76 @@ func (a *API) handlePatchBlock(w http.ResponseWriter, r *http.Request) {
...
+	err = json.Unmarshal(requestBody, &patches)
+	if err != nil {
+		a.errorResponse(w, r.URL.Path, http.StatusInternalServerError, "", err)
+		return
+	}
+
+	auditRec := a.makeAuditRecord(r, "patchBlocks", audit.Fail)
+	defer a.audit.LogRecord(audit.LevelModify, auditRec)
+
+	// stampModificationMetadata(r, patches.BlockIDs, auditRec)
```
wiggin77, line 736: Should this just be removed?

server/model/block.go
```diff
@@ -95,6 +95,17 @@ type BlockPatch struct {
 	DeletedFields []string `json:"deletedFields"`
 }
 
+// BlockPatchBatch is a batch of IDs and patches for modify blocks
+// swagger:model
+type BlockPatchBatch struct {
+	// The id for this block's parent block. Empty for root blocks
```
wiggin77, line 101: Are these really the parent ids?

wiggin77 (approved):
LGTM :+1:   Just two possible comments to modify.

Issue #1922, opened 2021-12-03
Bug: Creating a template from a card renders the workspace unusable


mgdelacroix:
## Steps to reproduce the behavior

1. On a board, create a card with some content
2. Create a template from the card
3. Get an error of "the card doesn't exist" on the card modal
4. Refresh
5. From that point on, accessing the workspace will redirect to an error page with `/boards/error?id=n is null` on the URL.

## Expected behavior

The template is created from the card.

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.11.0 RC
 - Browser and OS: Firefox 94 Linux

ogi-m:
Might be related to https://github.com/mattermost/focalboard/issues/1888

PR #1937, opened 2021-12-07
Prevents the application from crashing if there is a null content order

mgdelacroix:
This only takes care of not breaking on bad data, the unique IDs migration fix will avoid bad data from being created.

Fixes https://github.com/mattermost/focalboard/issues/1922


sbishel (approved):

hahmadia (approved):

ogi-m:
Tested and passed on v0.11 RC2

PR #1943, opened 2021-12-08
Fixes MM-40517

harshilsharma63:
#### Summary
Fixes MM-40517

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-40517


server/app/user.go
```diff
@@ -3,5 +3,14 @@ package app
...
 func (a *App) GetWorkspaceUsers(workspaceID string) ([]*model.User, error) {
-	return a.store.GetUsersByWorkspace(workspaceID)
+	users, err := a.store.GetUsersByWorkspace(workspaceID)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range users {
+		users[i].Email = ""
+	}
```
hahmadia, line 13: Not a huge fan of this. I would say we can make a new store method that gets users without the email and call it here. Make the users model so that email is omitempty. This can become unnecessarily expensive if the number of users is huge ( O(n) ). Better to just retrieve users in store layer without email.

hahmadia (changes requested):

server/app/user.go
harshilsharma63: Makes sense.

Issue #1907, opened 2021-12-02
Feature Idea: Improve person popup


asaadmahmood:
## Summary
The person popup looks really bad in terms of UI, it should match our Autocomplete UI that exists in Mattermost.

<img width="491" alt="Screenshot 2021-12-02 at 8 06 48 PM" src="https://user-images.githubusercontent.com/11034289/144448144-6d21deee-e1fc-4187-95a0-571a456679a6.png">

**It should look like this:**
<img width="1072" alt="Screenshot 2021-12-02 at 8 25 30 PM" src="https://user-images.githubusercontent.com/11034289/144451808-0ba48c8e-433c-4fed-a67d-2a9d05a36782.png">

## How important this is to me and why
Importance: Medium

PR #1959, opened 2021-12-09
GH-1907 - Updating person popover

asaadmahmood:
#### Summary
GH-1907 - Updating person popover

#### Ticket Link
https://github.com/mattermost/focalboard/issues/1907

#### Screenshot
Screenshot 2021-12-09 at 10 08 38 PM
Screenshot 2021-12-09 at 11 27 57 PM

### DIFF
--- a/webapp/src/components/properties/user/__snapshots__/user.test.tsx.snap
+++ b/webapp/src/components/properties/user/__snapshots__/user.test.tsx.snap
@@ -20,7 +20,11 @@ exports[`components/properties/user not readonly 1`] = `
         <div
           class="react-select__single-value css-ciz25-singleValue"
         >
-          username-1
+          <div
+            class="UserProperty-item"
+          >
+            username-1
+          </div>
         </div>
         <div
           class="css-1shkodo-Input"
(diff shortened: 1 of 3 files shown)


sbishel (approved):
@asaadmahmood Do you know why the GUIDs are being displayed in the images? I downloaded and ran locally, I'm not seeing the GUIDs only names, so I assume it has something to do with your data but wanted to ask.

asaadmahmood:
@sbishel You should have the images if you run it in plugin mode, as I think we always have images for users in Mattermost.

jespino (approved):
LGTM

wuwinson:
@asaadmahmood Is this supposed to show the profile picture if a user has one uploaded? I tried on Scott's test server and only see the letter avatar even though I have profile pic. Also to confirm, this does not change the display of username to full name correct? cc @sbishel

asaadmahmood:
@wuwinson Yes, it should work similar to Channels, channels shows the one named avatar, even if the user has not uploaded any avatar, we basically generate avatars for them.
And yes, this doesn't add the full name.

Issue #479, opened 2021-05-25
Feature Idea: Reorder comments on cards


sbishel:
**Summary:**
Display comments on cards - newest first. (or switchable)

**How important this is to me and why:**
Medium - 
Importance: High/Medium/Low

Use cases:
1. If lots of comments, scrolling to bottom would be cumbersome to see latest comments.
2.

**Additional context / similar features:**
Any examples of good implementations of this capability.
Jira allows this capability.

chenilim:
Initial reaction is to add a toggle to change sort order.

michaelgamble:
@asaadmahmood I unassigned myself off in favor of you taking this over

sbishel:
complete

PR #1985, opened 2021-12-14
GH-479 - Updating comment order and adding images

asaadmahmood:
#### Summary
GH-479 - Updating comment order and images

#### Ticket Link
https://github.com/mattermost/focalboard/issues/479

#### Screenshot
scrnli_15_12_2021_00-52-40

### DIFF
--- a/webapp/src/components/cardDetail/commentsList.tsx
+++ b/webapp/src/components/cardDetail/commentsList.tsx
@@ -48,12 +48,13 @@ const CommentsList = React.memo((props: Props) => {
 
     const {comments} = props
     const intl = useIntl()
+    const defaultImageUrl = 'data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100" style="fill: rgb(192, 192, 192);"><rect width="100" height="100" /></svg>'
 
     const newCommentComponent = (
         <div className='CommentsList__new'>
             <img
                 className='comment-avatar'
-                src={imageURLForUser(me!.id)}
+                src={imageURLForUser ? imageURLForUser(me!.id) : defaultImageUrl}
             />
             <MarkdownEditor
                 className='newcomment'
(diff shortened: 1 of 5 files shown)


wiggin77:
@asaadmahmood CI is complaining about `TypeError: imageURLForUser is not a function`

webapp/src/components/cardDetail/commentsList.tsx
```diff
@@ -42,14 +49,11 @@ const CommentsList = React.memo((props: Props) => {
...
-    // TODO: Replace this placeholder
-    const userImageUrl = 'data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100" style="fill: rgb(192, 192, 192);"><rect width="100" height="100" /></svg>'
-
     const newCommentComponent = (
-        <div className='commentrow'>
+        <div className='CommentsList__new'>
             <img
                 className='comment-avatar'
-                src={userImageUrl}
+                src={imageURLForUser(me!.id)}
```
harshilsharma63, line 56: This will crash on the personal server and desktop app as `imageURLForUser` will be undefined. You should use something like this -

```
const defaultImageURL = 'data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100" style="fill: rgb(192, 192, 192);"><rect width="100" height="100" /></svg>'

<div ...usual stuff
    src={imageURLForUser ? imageURLForUser(me!.id) : defaultImageURL}
/>
```

harshilsharma63: This needs to done in anyplace we're using `imageURLForUser`.

harshilsharma63 (changes requested):

webapp/src/components/cardDetail/commentsList.tsx
wiggin77: Might be easier to create a wrapper method that handles the difference between SKU's so that logic does not get sprinkled everywhere.

wiggin77 (approved):
LGTM 👍

harshilsharma63 (approved):
Looks perfect! LGTM!

PR #1986, opened 2021-12-15
Add workspace to teams and boards migrations

mgdelacroix:
#### Summary
This PR contains the SQL migration that goes from workspaces to teams, and then adds the boards and board_members tables and populates them with information.

The new models are included for reference, and the migration needs of a follow-up data migration to go from having floating (as in not associated with a team) templates to migrate the templates of the database and their cards to be present on each team.


server/services/store/sqlstore/migrations/migrations_files/000015_add_teams_and_boards.up.sql
```diff
@@ -0,0 +1,361 @@
...
+    scheme_viewer BOOLEAN,
+    PRIMARY KEY (board_id, user_id)
+) {{if .mysql}}DEFAULT CHARACTER SET utf8mb4{{end}};
+
+CREATE INDEX idx_boardmembers_user_id ON {{.prefix}}board_members(user_id);
+
+{{if .plugin}}
+{{- /* if we're in plugin, migrate channel memberships to the board */ -}}
+INSERT INTO {{.prefix}}board_members (
+    SELECT B.Id, CM.UserId, CM.Roles, (CM.UserId=B.created_by) OR CM.SchemeAdmin, CM.SchemeUser, FALSE, CM.SchemeGuest
```
jespino, line 350: This is assigning "Viewer" role to guests. I don't think that is wrong, but looks like something that we need to support ASAP in the interface, then.

jespino (approved):
LGTM

server/model/team.go
```diff
@@ -0,0 +1,45 @@
...
+	// required: true
+	SignupToken string `json:"signupToken"`
+
+	// Team settings
+	// required: false
+	Settings map[string]interface{} `json:"settings"`
+
+	// ID of user who last modified this
+	// required: true
+	ModifiedBy string `json:"modifiedBy"`
```
wiggin77, line 24: So far we're inconsistent naming `ModifiedBy` vs `ModifiedByID` throughout the codebase.  Looking more closely I see my code is using `MofifiedByID` the rest is `ModifiedBy`.  :grimacing:

server/services/store/sqlstore/migrations/migrations_files/000015_add_teams_and_boards.down.sql
```diff
@@ -0,0 +1 @@
+SELECT 1;
```
wiggin77, line 1: Since the migration UP deletes boards from the blocks and blocks_history table, this empty DOWN means we cannot downgrade and expect a valid database.  Is this intended?  Are we moving away from supporting backward migrations?

mgdelacroix: Both tables can be rebuilt from the boards and boards_history tables, although is not a minor effort. I've left a `SELECT 1` here in the interest of merging fast (on the feature branch) as discussed, but we can complete this before merging if we'd like to support down migrations.

Having said that, we don't really have a mechanism currently to undo a migration AFAIK, and the data migrations would fall outside of this (and the feature will have one as well for the templates) so I don't know how much it makes sense to go through the effort, as we will not be able to migrate down without implementing a system for it besides the feature itself.

server/model/team.go
mgdelacroix: I've based myself on the `blocks` table as the boards are coming from it, but happy to standardize in either direction really :)

Besides that, just saw that the `json` annotations are using camel case and not snake case, so I'll be changing that as part of this PR too :sweat_smile:

server/services/store/sqlstore/migrations/migrations_files/000015_add_teams_and_boards.down.sql
wiggin77: Ok, makes sense.  Not sure why we're bothering to maintain all those DOWN migrations.

wiggin77 (approved):
Looks great.  I've been looking forward to this change.

server/services/store/sqlstore/migrations/migrations_files/000015_add_teams_and_boards.down.sql
mgdelacroix: Chances are that as they were mandatory for `golang-migrate` we just embraced them (and the data migrations are fairly new). In any case, I've [scheduled an item](https://community-daily.mattermost.com/boards/workspace/b1ni7nprjpfwpb566k1gk8z9nw/886d2280-e88e-4065-8c85-1b62578f192e/f30dcb19-a0e8-45d2-8768-a9489f6e7fe8/crmg8ezy167yu3p1cjrkagctcao) to discuss during tomorrow's dev meeting to make a decision on this

server/model/team.go
mgdelacroix: Updated `json` annotations on e8be3ca

server/services/store/sqlstore/migrations/migrations_files/000015_add_teams_and_boards.up.sql
```diff
@@ -0,0 +1,361 @@
...
+          FROM {{.prefix}}blocks_history
+          WHERE type='board'
+  ;
+  {{end}}
+{{end}}
+
+
+{{- /* Update block references to boards*/ -}}
+{{if .sqlite}}
+  UPDATE {{.prefix}}blocks as B
```
harshilsharma63, line 194: nit: Would this be a much faster migration -

```
UPDATE {{.prefix}}blocks as B
SET board_id = parent_id
WHERE type = 'card' OR type = 'view';
```

I think the subqueries will take a lot of time.

harshilsharma63 (approved):

server/services/store/sqlstore/migrations/migrations_files/000015_add_teams_and_boards.up.sql
mgdelacroix: The goal of the query is to have all blocks belonging to a board, and that would leave the inner blocks out sadly.

PR #1990, opened 2021-12-15
Fixing and regenerating swagger documentation

jespino:


server/main/doc.go
```diff
@@ -0,0 +1,26 @@
+// Package classification Focalboard Server
```
jespino, line 1: Moving the general documentation to an independent file fixes the home path generated reference.

server/api/api.go
```diff
@@ -1635,7 +1635,7 @@ func (a *API) handleDeleteSubscription(w http.ResponseWriter, r *http.Request) {
 	//   description: Workspace ID
 	//   required: true
 	//   type: string
-	// - name: BlockID
+	// - name: blockID
```
jespino, line 1638: Not related issue, that I had to fix to generate the swagger documentation.

jespino:
Almost everything is generated code, the only changes are in the `server/api/api.go`, in `server/main/doc.go` and in `server/main/main.go` files.

wiggin77 (approved):
LGTM :+1:    Thanks for fixing `BlockID`

server/api/api.go
wiggin77: I forgot to re-generate the the swagger files when I checked that in...  otherwise I would have caught that.

jespino: Yes, maybe the CI should check that, just "generate and see if something changes or not"

sbishel (approved):
LGTM - we should probably discuss swagger at a dev meeting to make sure everyone understands how to use and rebuild.

PR #2000, opened 2021-12-16
Test all supported databases in CI

wiggin77:
#### Summary
CI is using `make server-test` to run the unit and integration tests, which only tests using the sqlite database.  This PR adds two more make commands:  `make server-test-mysql` and `make server-test-postgres`.

With this PR CI will run unit and integration tests against all three databases.  Several syntax errors were already found and fixed for mysql and postgres.

The mysql and postgres tests use docker to install and instantiate the database engines and clean up after the tests.  Docker and docker-compose must be installed for these to work.

#### Ticket Link
https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/150d1464-6d59-4c39-810f-ee201439dbaf/cobu5qzngkbbkukg3cenduzke5w


server/services/store/sqlstore/blocks.go
```diff
@@ -182,7 +182,7 @@ func (s *SQLStore) getSubTree3(db sq.BaseRunner, c store.Container, blockID stri
 		Join(s.tablePrefix + "blocks" + " as l3 on l3.parent_id = l2.id or l3.id = l2.id").
 		Where(sq.Eq{"l1.id": blockID}).
 		Where(sq.Eq{"COALESCE(l3.workspace_id, '0')": c.WorkspaceID}).
-		OrderBy("l3.insert_at")
+		OrderBy("l3.id")
```
jespino, line 185: This is not strictly the same, why are we changing this?, maybe would have fixed it with ordering by `insert_at` and `id`?

docker-testing/docker-compose-postgres.yml
```diff
@@ -0,0 +1,16 @@
+version: '2.4'
+services:
+  postgres:
+    extends:
+        file: docker-compose.common.yml
```
jespino, line 5: Is there any reason to separate this in the `common.yml` if we are already separating the files, and in the common the services are separated too? What is the added value of the common here?

jespino (approved):
A couple of not-blocker comments, but otherwise looks good to me.

docker-testing/docker-compose-postgres.yml
wiggin77: None really.  Just ended up that way because that's how it is in channels.  I'll reorganize into one file each.

wiggin77: Fixed via https://github.com/mattermost/focalboard/pull/2000/commits/b5026d8631b50055cc0c6fe01b63534d817e70f2

server/services/store/sqlstore/blocks.go
wiggin77: Order preserved via https://github.com/mattermost/focalboard/pull/2000/commits/dda97326628ba77882fbae6d8232e3a3a1c346e4

sbishel (approved):

Issue #1916, opened 2021-12-03
Bug: Deleting the created time property when the calendar view is using it to display the cards will not change to other properties


mgdelacroix:
## Steps to reproduce the behavior

1. Create a calendar view
2. Add a date property (say `deadline`) and `Created time`
3. Change the `Display by` to use `Created time`
4. Remove the `Created time` property

## Expected behavior

The view should modify the `Display by` and enable `deadline` as it's the last one that can be used, but it will keep the filter on `Created time`.

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.11.0 RC
 - Browser and OS: Firefox 94 Linux

mgdelacroix:
After some more testing, I've realised that the `Display by` will always change to `Created time`, no matter if there are any other properties of type `date` that could be used to display the cards by.

Issue #2007, opened 2021-12-21
Bug: Issues with VisiblePropertyIds


sbishel:
## Steps to reproduce the behavior

Currently, there are two property collections.
Boards - CardProperties - contains all all properties for a card.
ActiveView - VisiblePropertyIds - contains all visible property ids for the current view.

There are a couple of issues with our current implemenation.

1. Table View -
When reordering columns (fields), the CardProperties in the Boards object is reordered. This causes other views properties order to change. Also causes default GroupByProperty and DateDisplayProperty to change.
(This should be updated to change the order of the VisiblePropertyIds.)

2. All Views -
All views, including the table view, property orders are based on the CardProperties collection. This should be based on the Views Visible properties.

## Expected behavior
Each views property orders should be independent of each other. Although only the table view allows changing of the order. Others views can be ordered by turning properties on in a specific order.

PR #2008, opened 2021-12-21
GH-1916 - Set to first date field when loading

sbishel:
#### Summary
If the display is null, set to first date property.
Previously, after deleting the date property the createDate would be taken as default.

This follows the same pattern as is used for the groupByProperty.

There are some issues with this implementation, which another [issue](https://github.com/mattermost/focalboard/issues/2007) has been created to handle these issues.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/1916


wiggin77 (approved):
LGTM :+1:

harshilsharma63 (approved):
LGTM!

PR #2010, opened 2021-12-22
GH-2007 Use VisiblePropertyIds for view property order

sbishel:
#### Summary
This PR updates all views to use the `activeView.fields.VisiblePropertyIds` array to keep order of the visible properties. Currently, we use the `board.fields.CardProperties`. 

The table view is currently the only view which allows the properties to be rearranged and it actually rearranges the `board.fields.CardProperties` which currently cause the properties order to change on all views.

With this PR the other views property orders can be controlled based on the order they are added. This currently isn't possible, the order is always based on the order of `board.fields.CardProperties`

Before merging this PR, the functionality needs to be run past the rest of the team and PM to ensure this change in 
functionality is the right change to make.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/2007


wuwinson (approved):
LGTM!

harshilsharma63 (approved):
Works great! LGTM!

asaadmahmood (approved):

PR #2011, opened 2021-12-22
Permissions phase 1 - Store changes

mgdelacroix:
#### Summary
This PR contains the necessary store changes to move from workspaces to teams, as part of the permissions phase 1.


server/model/block.go
```diff
@@ -58,9 +58,13 @@ type Block struct {
 	// required: false
 	DeleteAt int64 `json:"deleteAt"`
 
-	// The workspace id that the block belongs to
+	// Deprecated. The workspace id that the block belongs to
+	// required: false
+	WorkspaceID string `json:"-"`
```
jespino, line 63: Is there any reason to keep this here?

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -198,67 +196,38 @@ func (s *MattermostAuthLayer) CleanUpSessions(expireTime int64) error {
...
+		team := model.Team{
 			ID:    id,
 			Title: "",
 		}
 
-		return &workspace, nil
+		return &team, nil
 	}
 
+	// ToDo: this is retrieving channels, modify this
```
jespino, line 209: TODO, I guess this is still something to change.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -198,67 +196,38 @@ func (s *MattermostAuthLayer) CleanUpSessions(expireTime int64) error {
...
-		}
-		sb.WriteString(name)
-	}
-	return &model.Workspace{ID: id, Title: sb.String()}, nil
+	return &model.Team{ID: id, Title: displayName}, nil
 }
 
-func (s *MattermostAuthLayer) HasWorkspaceAccess(userID string, workspaceID string) (bool, error) {
+func (s *MattermostAuthLayer) HasTeamAccess(userID string, teamID string) (bool, error) {
+	// ToDo: modify this to be "has team permission"
```
jespino, line 226: Is this for this phase or for the next phase?

server/model/block.go
mgdelacroix: Yes, we need it for the unique IDs data migration

Makefile
```diff
@@ -94,22 +96,22 @@ modd-precheck:
...
 server-test: ## Run server tests
-	cd server; go test -race -v -count=1 ./...
+	cd server; go test -tags='$(BUILD_TAGS)' -race -v -count=1 ./...
 
 webapp: ## Build webapp.
 	cd webapp; npm run pack
 
 watch-plugin: modd-precheck ## Run and upload the plugin to a development server
-	modd -f modd-watchplugin.conf
+	env FOCALBOARD_BUILD_TAGS='$(BUILD_TAGS)' modd -f modd-watchplugin.conf
```
jespino, line 114: Does need the plugin the sqlite compilation flag? In theory, the plugin shouldn't use Sqlite ever.

server/model/block.go
jespino: Shouldn't we replicate that in the migration code and leave it as is the real value now? I think we shouldn't contaminate the rest of the code with migration code details.

Makefile
mgdelacroix: It's true that the plugin doesn't need it. I propagated it in case we'd like to use other `BUILD_TAGS`, but I'm happy to remove it if preferred

server/model/block.go
mgdelacroix: You mean creating a `LegacyBlock` type that contains the old field? I'm happy to go that way

server/services/store/sqlstore/data_migrations_test.go
```diff
@@ -11,7 +11,9 @@ import (
 )
 
 //nolint:gosec
-func TestGetBlocksWithSameID(t *testing.T) {
+func tTestGetBlocksWithSameID(t *testing.T) {
```
jespino, line 14: This initial `t` is accidental, I guess.

server/services/store/sqlstore/legacy_blocks.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/mattermost/focalboard/server/utils"
+
+	sq "github.com/Masterminds/squirrel"
+	"github.com/mattermost/focalboard/server/model"
+	"github.com/mattermost/focalboard/server/services/store"
+
+	"github.com/mattermost/mattermost-server/v6/shared/mlog"
+)
+
+// legacyBlocksFromRows is the old getBlock version that still uses
```
jespino, line 16: I'm ok with this for now, but we should move this to code of the data migration and clean up the rest of the code.

jespino (approved):
LGTM, a couple of improvement comments, but nothing blocking this to get merged.

server/model/board.go
```diff
@@ -79,6 +84,54 @@ type Board struct {
...
+// BoardPatch is a patch for modify boards
+// swagger:model
+type BoardPatch struct {
+	// The type of the board
+	// required: false
+	Type *BoardType `json:"type"`
+
+	// The title of the board
+	// required: false
+	Title *string `json:"title"`
```
wiggin77, line 96: Why are these pointers to strings, whereas they are just strings in Blocks struct?

server/model/board.go
```diff
@@ -111,6 +164,79 @@ type BoardMember struct {
 	SchemeViewer bool `json:"schemeViewer"`
 }
 
+func BoardFromJSON(data io.Reader) *Board {
+	var board *Board
+	_ = json.NewDecoder(data).Decode(board)
```
wiggin77, line 169: I think we're trying to get away from ignoring JSON decode errors.

server/model/board.go
mgdelacroix: We need to distinguish on the `(BoardPatch) Patch` function if the user sent an empty title (to set it) or nothing, and then leave the title intact

server/model/board.go
```diff
@@ -111,6 +164,79 @@ type BoardMember struct {
...
+	return board
+}
+
+func BoardsFromJSON(data io.Reader) []*Board {
+	var boards []*Board
+	_ = json.NewDecoder(data).Decode(&boards)
+	return boards
+}
+
+// Patch returns an update version of the board.
```
wiggin77, line 179:
```suggestion
// Patch returns an updated version of the board.
```

server/model/board.go
mgdelacroix: Agree, but as this is following how the code currently works, I'd leave it for a refactor later as we'd need to account for errors in the client, which is the main user of these methods

server/services/store/sqlstore/team.go
```diff
@@ -0,0 +1,217 @@
...
+		Values(
+			team.ID,
+			team.SignupToken,
+			team.ModifiedBy,
+			now,
+		)
+	if s.dbType == mysqlDBType {
+		query = query.Suffix("ON DUPLICATE KEY UPDATE signup_token = ?, modified_by = ?, update_at = ?",
+			team.SignupToken, team.ModifiedBy, now)
+	} else {
```
wiggin77, line 48: I know team store methods shouldn't be run with sqlite, however I think we should explicitly look for postgres in the `else`, and error on any other db type.

server/services/store/sqlstore/team.go
```diff
@@ -0,0 +1,217 @@
...
+		Values(
+			team.ID,
+			signupToken,
+			settingsJSON,
+			team.ModifiedBy,
+			now,
+		)
+	if s.dbType == mysqlDBType {
+		query = query.Suffix("ON DUPLICATE KEY UPDATE settings = ?, modified_by = ?, update_at = ?", settingsJSON, team.ModifiedBy, now)
+	} else {
```
wiggin77, line 86: Same... check for postgres and error on any other db type.

server/services/store/sqlstore/team.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		s.logger.Error(`ERROR GetTeam settings json.Unmarshal`, mlog.Err(err))
+		return nil, err
+	}
+
+	return &team, nil
+}
+
+func (s *SQLStore) hasTeamAccess(db sq.BaseRunner, userID string, teamID string) (bool, error) {
+	return true, nil
```
wiggin77, line 134: Is this a todo?

server/services/store/sqlstore/board.go
```diff
@@ -0,0 +1,483 @@
...
+			userID,
+			now,
+			now,
+		)
+
+	if _, err := insertQuery.Exec(); err != nil {
+		return err
+	}
+
+	deleteQuery := s.getQueryBuilder(db).
```
wiggin77, line 333: I know this is basically the same behaviour as workspaces, but should we think about what happens to the orphaned records in the `blocks` table?  This can be a follow-up discussion.

wiggin77 (approved):
This is great work @mgdelacroix.  :1st_place_medal:   I don't envy the merge conflicts :scream:

server/services/store/sqlstore/team.go
mgdelacroix: In this case the method is related to the signup token, which runs on Personal Server instances, so it should run with SQLite, and the `else` syntax is the one SQLite supports. This is covered on the upsert store tests, for example

server/services/store/sqlstore/team.go
mgdelacroix: It has been removed on a later PR :sweat_smile:

PR #2025, opened 2021-12-29
[GH-1754] Feature: card badges

### LINKED ISSUE #1754: Feature Idea: badges for kanban/gallery cards
## Summary
Cards can contain different types of content inside them. It will be useful to know what is inside the card on kabban board or in the gallery without opening the dialog for it.

This can be done with help of additional information that will be shown at the bottom of the card. So called "badges" that give quick summary about what is inside. Here how it can look:
CardBadgesPOC
Proper implementation should also count checkboxes in the markdown. Can also have a badge for images for example. And probably showing of these badges should be optional because they take place.

Importance: Medium, will allow better management of the cards.

Trello has badges for the cards:
TrelloCardBadges

@chenilim @asaadmahmood what do think about such feature?

kamre:
#### Summary
Badges with information about description/comments/checkboxes for cards on kanban and gallery views.

Looks like this:
image

Not visible by default because it takes additional space on the card. Can be shown/hidden via the view menu:
image

Corresponding code changes:
* `CardBadges` component added and used in `GalleryCard` and `KanbanCard`
* optional field `cardBadgesVisible` added to `BoardViewFields`
* action for show/hide badges added to `ViewHeaderActionsMenu` for kanban and gallery views
* `MessageIcon` component for corresponding Compass icon added
* unit and Cypress tests added
* minor improvements for `MenuWrapper` and `TestBlockFactory`

#### Ticket Link
Fixes #1754 

@asaadmahmood please review from UI/UX side


webapp/src/components/cardBadges.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const contents = useAppSelector(getCardContents(card.id))
+    const comments = useAppSelector(getCardComments(card.id))
+    const badges = useMemo(() => calculateBadges(contents, comments), [contents, comments])
+    if (!hasBadges(badges)) {
+        return null
+    }
+    const intl = useIntl()
+    const {checkboxes} = badges
+    return (
+        <div className={`cardBadges ${className || ''}`}>
```
jespino, line 88: Tiny detail, by convention we use the component name for the root component className inside the component. In this case should be `CardBadges` with uppercase at the beginning.

jespino (changes requested):
I added a comment about the name of the class name, other than that looks good to me. Great work @kamre! :tada:

webapp/src/components/cardBadges.tsx
kamre: Good to know. Fixed.

webapp/src/components/properties/lastModifiedAt/__snapshots__/lastModifiedAt.test.tsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`componnets/properties/lastModifiedAt should match snapshot 1`] = `
   <div
     class="LastModifiedAt octo-propertyvalue readonly"
   >
-    June 15, 4:22 PM
+    June 15, 2021, 4:22 PM
```
kamre, line 8: Rendering of the predefined date in the test has changed with the arrival of the new year.

jespino: Interesting.

jespino (approved):
Now looks good to me. Thanks @kamre! :confetti_ball:

sbishel (approved):
Code looks good. I just wanted to make sure we have UX (@asaadmahmood ) and PM 
 (@wuwinson) review placement of the 'Show/Hide card badges' MenuItem. (Properties vs Ellipsis Menu).

I have this PR installed at 
https://sbishel.test.mattermost.cloud/
If you want to play with it.

asaadmahmood (commented):
@kamre I think this should actually exist under the Properties as a toggle.

kamre:
@asaadmahmood I agree that from the point of view on toggling visible elements on the cards it is related to `Properties` menu, but it is not related to card properties themselves currently because only card content blocks are used.

Do you think that adding `Card badges` or even just `Badges` to `Properties` menu at the bottom will be a better solution?

asaadmahmood:
@kamre Yup, those may not strictly be card properties, but we can consider them properties in a general sense.
With regards to the term Badges, or Card Badges, not sure.
I think we should just say "Comments" and "Description"

If we're doing something for the description, I honestly am not sure about the checkboxes thing, as that's only part of the description, and I'm not sure that would be isolated as a different item for just the checkboxes.

CC @wuwinson @chenilim

wuwinson:
@kamre @asaadmahmood Agreed that we should move the toggles to the properties section for now. Would keep it simple with a single toggle for all badges. Current location is not very discoverable and don't think this feature will get much utilization if it stays there.

Not sure how helpful the "Description" badge is, but think we should at least keep the "Comments" and "Checkbox" badges since it provides a bit more helpful info without requiring the user to click into the card. Okay with keeping all 3, but would cut the "Description" one if anything. Thoughts? cc @chenilim

chenilim:
Thanks @kamre! This is a great feature! I agree with moving the toggle to under the Properties menu (conceptually, it's what to display on the card preview) - I suppose this could apply to Boards, Calendar, and Gallery. For tables, it could show up as another column, but that might be a future enhancement.

One minor thing is the "Text" icon appears larger than the others (comments and checkbox). It also looks similar to a grab-handle.

Card badges

This is something we can iterate on, but perhaps there's a better icon to represent the description? cc @asaadmahmood.

In any case, I'm in favor of checking this in and iterating - now that v0.12 has branched, we have time to make updates.

wuwinson (approved):
Sounds good @chenilim!

kamre:
@chenilim @asaadmahmood I have moved show/hide for card badges to `Properties` menu:
image
Something similar was already done for the card title in the gallery view when non-property is added to this menu.

Also I have added card badges support to calendar view. Nothing for table view yet, this can be done separately.

Regarding the icons: I have just reused ones from the context menu for `Add content` in the card dialog:
image
They are currently just inline svg and not Compass icons. But for comments I have used the icon from Compass set.

So if there are better icons in the Compass set for text and checkbox we can use them.

kamre:
@sbishel I have noticed that Cypress test is failing after the merge. Something is wrong with adding comments, will look into this tomorrow.

sbishel:
Thanks @kamre, just ping me and I'll get it merged.

kamre:
@sbishel fixed Cypress test: used explicit `blur` event after typing comment text before clicking on `Send` button. Manually this issue can't be reproduced so probably it is Cypress issue with changing focus to the button and not triggering blur event.

sbishel:
Awesome...thanks @kamre

PR #2034, opened 2022-01-03
Fix #1745. Delete socket only if it exists

### LINKED ISSUE #1745: Bug: Unable to unlink socket.  caller="server/server.go:378" error="no such file or directory"

1. docker run -it -p 80:8001 mattermost/focalboard

## Expected behavior
No error.

## Screenshots
image

## Edition and Platform

 - Edition: Personal Desktop
 - Version: docker image

## Additional context

docker image gives error on the subject at startup.

chenilim:
#### Summary
Previously, we try to unlink the existing admin local socket on startup, even if the file doesn't exists. This results in a confusing error log, even though it doesn't cause a failure. The change is to only try to unlink the socket file if it exists.

#### Ticket Link
#1745

### DIFF
--- a/server/server/server.go
+++ b/server/server/server.go
@@ -374,8 +374,11 @@ func (s *Server) startLocalModeServer() error {
 	}
 
 	// TODO: Close and delete socket file on shutdown
-	if err := syscall.Unlink(s.config.LocalModeSocketLocation); err != nil {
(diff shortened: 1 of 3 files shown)


wiggin77 (approved):
LGTM 👍

chenilim:
Thanks! Had to also update the (unrelated) jest snapshots to account for previous year's dates now also showing the year (2021).

PR #2035, opened 2022-01-03
Fix #1999, fix #2002: Alt italics regex without lookbehind

### LINKED ISSUE #1999: Bug: Boards App (macOS) Clicking "New" button and "+New" to create a new card shows a blank screen
Repros on macOS (but not Windows 10) Boards app from this link: https://focalboard-builds.s3.amazonaws.com/latest/focalboard-mac.zip

1. Download the macOS app from the above link
2. Create a board
3. Within the new board, click on "New" button to create a card and also click on "+New" to create a card
4. FYI, this also reproduces when hovering over an existing card in table view and then clicking "OPEN" 

## Observed
Blank screen from either of the actions in 3)

## Expected behavior

A new card is created and is displayed on the screen

## Screenshots

image

image

## Edition and Platform

 - Edition: Personal Desktop
 - Version: 0.12.0
 - Browser and OS: macOS

### LINKED ISSUE #2002: Bug: can't open cards on Safari
## Steps to reproduce the behaviour

1. Create new board
2. Try to create new card
3. See empty screen

## Expected behaviour

Card dialog should be visible

## Edition and Platform

 - Edition: Personal Server
 - Version: `0.12.0` on `main` branch
 - Browser and OS: Safari 15.2 on macOS

## Additional context

The issue comes from the `live-markdown-plugin`, in particular the [regexp with lookbehind](https://github.com/mattermost/focalboard/blob/main/webapp/src/components/live-markdown-plugin/inline-styles/italicStyleStrategy.ts#L7) that is [not supported in Safari](https://caniuse.com/js-regexp-lookbehind)

@chenilim FYI

chenilim:
#### Summary
The RegEx compiler throws an unhandled error on Safari because it doesn't support Lookbehinds. Change is to catch the exception and use an alternative, simplified regex.

Note: The alternative doesn't support complex nested bold and italics fully (e.g. ***italic* and bold** **bold and *italic***), but even the original had some issues. Those are rare, and only affect the preview (final rendering with markedjs should be correct), so leaving that for a future improvement (contributions welcome!).

#### Ticket Link
#1999 and #2002.


wiggin77 (approved):
LGTM :+1:   Go doesn't support it either.

PR #2040, opened 2022-01-04
fix crash in DeleteBlock

wiggin77:
#### Summary
Investigating a crash reported in MM cloud was traced to a DeleteBlock API call from the client.  The block was already deleted by the time the request hit the server.  I can see several ways this can happen:

- multiple people deleted a block at the same time
- single user trying to delete a block a second time before receiving the result from the first

The main issue is that `sqlstore.getBlock` returns a nil block and nil error when the block id doesn't exist.  Some callers of this API only check the error and either crash or don't crash completely by accident.  For example, `PatchBlock` will crash (or not) based on timing.  `PatchBlocks` will silently fail patching the batch and return success to the client.

A ticket exists to have store APIs that return a single entity return a `NotFound` error as needed.  This PR simply corrects one crash instance.

Note:  a `crash` in this context means an unhandled panic and the plugin no longer functioning until it is restarted.
 
#### Ticket Link
NONE

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -1594,8 +1594,8 @@ func (a *API) handleCreateSubscription(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// check for valid block
-	_, err = a.app.GetBlockWithID(*container, sub.BlockID)
-	if err != nil {
+	block, err := a.app.GetBlockWithID(*container, sub.BlockID)
+	if err != nil || block == nil {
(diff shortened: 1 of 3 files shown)


sbishel (approved):
Thanks Doug. Looks like a unit test needs updated as well. Otherwise LGTM

jespino (approved):
LGTM

jespino:
Merging it, we can test it in community or in the test environment.

PR #2055, opened 2022-01-06
List all authors of a card diff in the notification

wiggin77:
#### Summary
This PR addresses an issue with card change notifications when multiple users are editing a card at the same time.  Before this PR, only the first user to start editing would be listed in the notification.  After this PR, all authors of a card diff are listed.

Also, when comments are added, the author of the comment is included with the individual comment.

#### Ticket Link
https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/150d1464-6d59-4c39-810f-ee201439dbaf/c1dawrhr97frq381xkh94uno9ir

### DIFF
--- a/server/services/notify/notifysubscriptions/diff.go
+++ b/server/services/notify/notifysubscriptions/diff.go
@@ -201,9 +201,16 @@ func (dg *diffGenerator) generateDiffsForCard(card *model.Block, schema model.Pr
 }
 
 func (dg *diffGenerator) generateDiffForBlock(newBlock *model.Block, schema model.PropSchema) (*Diff, error) {
+	dg.logger.Debug("generateDiffForBlock - new block",
+		mlog.String("block_id", newBlock.ID),
+		mlog.String("block_type", string(newBlock.Type)),
+		mlog.String("modified_by", newBlock.ModifiedBy),
+		mlog.Int64("update_at", newBlock.UpdateAt),
+	)
+
 	// find the version of the block as it was at the time of last notify.
 	opts := model.QueryBlockHistoryOptions{
-		BeforeUpdateAt: dg.lastNotifyAt,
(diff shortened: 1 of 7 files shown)


sbishel (approved):
LGTM

server/services/notify/notifysubscriptions/diff.go
```diff
@@ -200,13 +195,22 @@ func (dg *diffGenerator) generateDiffsForCard(card *model.Block, schema model.Pr
...
+		mlog.String("block_id", newBlock.ID),
+		mlog.String("block_type", string(newBlock.Type)),
+		mlog.String("modified_by", newBlock.ModifiedBy),
+		mlog.Int64("update_at", newBlock.UpdateAt),
+	)
+
 	// find the version of the block as it was at the time of last notify.
 	opts := model.QueryBlockHistoryOptions{
-		BeforeUpdateAt: dg.lastNotifyAt,
+		BeforeUpdateAt: dg.lastNotifyAt + 1,
```
jespino, line 213: Nit: this can be unprecise, and can lead to very strange bugs in very weird times (Not a big deal, but with enough bad luck, probably you can miss s change in the notifications.

I wouldn't fix it here, and I don't even know if I would fix that, but at least we should know it.

jespino (approved):
LGTM. Just a tiny thing that I pointed out, but I don't think we can fix it without extra rethink how we query data.

server/services/notify/notifysubscriptions/diff.go
wiggin77: This query is to find the version of the block that is older than or equal to the last time a notification was sent.  I'm not seeing how it could miss a change.  The +1 is simply to turn the `X < updateAt` to `X <= updateAt`.  Am I missing something?

PR #2060, opened 2022-01-07
Adding initial templates load and the new templates endpoint

jespino:


server/services/store/sqlstore/initialize.go
```diff
@@ -28,60 +49,88 @@ func (s *SQLStore) InitializeTemplates() error {
...
-// ToDo: as part of the change in the above method,
-// isInitializationNeeded should be something like:
-// getTeamsThatNeedInitialization() ([]string, error)
-// and return the team IDs that have no boards
-
-// isInitializationNeeded returns true if the blocks table is empty.
-func (s *SQLStore) isInitializationNeeded() (bool, error) {
+// getTeamsThatNeedInitialization returns a list of teams without templates
+func (s *SQLStore) getTeamsThatNeedInitialization(db sq.BaseRunner) ([]string, error) {
 	query := s.getQueryBuilder(s.db).
```
harshilsharma63, line 109: Should we be using the transactional `db` object received in params?

server/services/store/sqlstore/initialize.go
```diff
@@ -28,60 +49,88 @@ func (s *SQLStore) InitializeTemplates() error {
 // ToDo: needs a rewrite to loop over teams and insert boards and
```
mgdelacroix, line 49: This `ToDo` comment can be removed

server/services/store/sqlstore/initialize.go
```diff
@@ -28,60 +49,88 @@ func (s *SQLStore) InitializeTemplates() error {
...
-func (s *SQLStore) isInitializationNeeded() (bool, error) {
+// getTeamsThatNeedInitialization returns a list of teams without templates
+func (s *SQLStore) getTeamsThatNeedInitialization(db sq.BaseRunner) ([]string, error) {
 	query := s.getQueryBuilder(s.db).
-		Select("count(*)").
-		From(s.tablePrefix + "blocks").
-		Where(sq.Eq{"COALESCE(channel_id, '0')": "0"})
-
-	row := query.QueryRow()
+		Select("T.Id").
```
harshilsharma63, line 110: This won't allow us to update existing templates as we need to do for growth spike. I'm not sure if should handle it here or cherrypick it later on from `main` after the growth spike.

server/services/store/sqlstore/initialize.go
```diff
@@ -3,23 +3,44 @@ package sqlstore
...
+	"github.com/mattermost/focalboard/server/model"
+	"github.com/mattermost/focalboard/server/services/store/sqlstore/initializations"
+	"github.com/mattermost/focalboard/server/utils"
 	"github.com/mattermost/mattermost-server/v6/shared/mlog"
 )
 
-// InitializeTemplates imports default templates if the blocks table is empty.
-func (s *SQLStore) InitializeTemplates() error {
-	isNeeded, err := s.isInitializationNeeded()
+// initializeTemplates imports default templates if the blocks table is empty.
```
mgdelacroix, line 17:
```suggestion
// initializeTemplates imports default templates if the boards table is empty for a team.
```

server/services/store/sqlstore/initialize.go
```diff
@@ -28,60 +49,88 @@ func (s *SQLStore) InitializeTemplates() error {
...
+				updatedBlock.BoardID = newBoardID
+				updatedBoardBlocks = append(updatedBoardBlocks, updatedBlock)
+			}
+			board.ID = newBoardID
+			board.Type = model.BoardTypeOpen
+			board.TeamID = teamID
+			board.IsTemplate = true
+
+			for _, block := range updatedBoardBlocks {
+				err := s.insertBlock(db, &block, "system")
```
harshilsharma63, line 92: I get a `gosec` memory aliasing error here in the IDE. Inlining the `insertBlock` in error check fixes it.

Screenshot 2022-01-12 at 5 40 42 PM

server/services/store/sqlstore/initialize.go
```diff
@@ -28,60 +49,88 @@ func (s *SQLStore) InitializeTemplates() error {
...
+		return err
+	}
 
-		globalContainer := store.Container{
-			TeamID: "0",
-		}
+	for _, teamID := range teamIDs {
+		for _, board := range archive.Boards {
+			boardBlocks := []model.Block{}
+			newBoardID := utils.NewID(utils.IDTypeBoard)
```
mgdelacroix, line 65: As we're generating new board IDs, we should store the ones from the archive file in a `templateID` field for telemetry tracking purposes

server/services/store/sqlstore/initialize.go
```diff
@@ -28,60 +49,88 @@ func (s *SQLStore) InitializeTemplates() error {
...
-				mlog.String("block_type", archive.Blocks[i].Type.String()),
-				mlog.String("block_title", archive.Blocks[i].Title),
-			)
-			err := s.InsertBlock(globalContainer, &archive.Blocks[i], "system")
+			boardBlocks = model.GenerateBlockIDs(boardBlocks)
+			updatedBoardBlocks := []model.Block{}
+
+			for _, block := range boardBlocks {
+				updatedBlock := block
+				if updatedBlock.ParentID == updatedBlock.ID {
```
harshilsharma63, line 77: Should this be 
```suggestion
				if updatedBlock.ParentID == board.ID {
```

Same for the `RootID` check below.

server/services/store/sqlstore/initialize.go
harshilsharma63, line 92: Nit: we can inline this with the error check.

server/services/store/sqlstore/initialize.go
```diff
@@ -28,60 +49,88 @@ func (s *SQLStore) InitializeTemplates() error {
...
+			board.TeamID = teamID
+			board.IsTemplate = true
+
+			for _, block := range updatedBoardBlocks {
+				err := s.insertBlock(db, &block, "system")
+				if err != nil {
+					return err
+				}
+			}
+			_, err := s.insertBoard(db, &board, "system")
```
harshilsharma63, line 97: Nit: we can inline this with the error check.

harshilsharma63 (changes requested):

webapp/src/store/globalTemplates.ts
```diff
@@ -12,9 +12,9 @@ import {RootState} from './index'
 
 export const fetchGlobalTemplates = createAsyncThunk(
     'globalTemplates/fetch',
-    async () => {
-        const rootClient = new OctoClient(client.serverUrl, '0')
-        const templates = await rootClient.getTeamTemplates('0') // ToDo: pass team id?
+    async (teamId: string) => {
+        const rootClient = new OctoClient(client.serverUrl, teamId)
+        const templates = await rootClient.getTeamTemplates(teamId) // ToDo: pass team id?
```
mgdelacroix, line 17:
```suggestion
        const templates = await rootClient.getTeamTemplates(teamId)
```

mgdelacroix (changes requested):
Looking really good!!

server/services/store/sqlstore/initialize.go
jespino: good catch

server/services/store/sqlstore/initialize.go
jespino: yes

server/model/board.go
```diff
@@ -19,6 +19,10 @@ type Board struct {
 	// required: true
 	ID string `json:"id"`
 
+	// The TemplateID for the board
```
jespino, line 22: I added this new field, I'm not sure about the name, because looks like is for storing the template that created the board, but in reality is set only in built in templates for tracking it. Maybe it should be something like `BuiltinTemplateId` or `BlueprintTemplateId` or `BuildinTemplateTrackingId`. @harshilsharma63 @mgdelacroix what do you think?

wiggin77:
@jespino @mgdelacroix FYI this PR https://github.com/mattermost/focalboard/pull/2167/files refactors the template initialization.  That is in conflict with this PR.

jespino:
@wiggin77 Ok, I'll take a look, thanks for the heads up.

jespino:
This is no longer relevant, changes applied directly into the main branch.

PR #2072, opened 2022-01-07
Stayin' Alive: Log & recover from web server panics

wiggin77:
#### Summary
Currently a panic in any Rest API handler will cause the FocalBoard plugin to terminate.  This PR handles panics from all APIs, returns a 500 error to the client, and logs the panic with stack trace, and does not terminate the plugin.

#### Ticket Link
NONE


sbishel (approved):

server/api/api.go
```diff
@@ -115,6 +116,17 @@ func (a *API) RegisterAdminRoutes(r *mux.Router) {
 
 func (a *API) requireCSRFToken(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer func() {
```
jespino, line 119: The code looks good, but I think this should be an independent middleware. Makes no sense to include the recovery mechanism as part of the CSRF token validation.

jespino (changes requested):
The code looks good, but I think can be placed in an independent middleware.

server/api/api.go
```diff
@@ -115,6 +116,17 @@ func (a *API) RegisterAdminRoutes(r *mux.Router) {
...
+			if p := recover(); p != nil {
+				a.logger.Error("Http handler panic",
+					mlog.Any("panic", p),
+					mlog.String("stack", string(debug.Stack())),
```
jespino, line 123: This `debug.Stack()` and `p` variables in the logs, can lead to leaking private information into the logs?

For example, Imagine that for some reason the database fails for a second and generates a panic while the user is logging in, and the user/password data is part of the parameters in the stack trace. Can it be a problem?

wiggin77: AFAIK, Go stack traces only contain pointers to func parameters.  As for the panic text, I can't imagine a case where someone is explicitly putting secrets in there. A database driver isn't going to do that.  Definitely something to watch out for, but I think we're OK here.

server/api/api.go
wiggin77: Ok, I'll create a separate middleeare method.

wiggin77: Fixed via https://github.com/mattermost/focalboard/pull/2072/commits/f14ef7f34e15ddd01230adfe3e449d83845093fb

jespino (approved):
LGTM

Issue #2077, opened 2022-01-07
Bug: Mac App sometimes fails to update after reload


chenilim:
## Steps to reproduce the behavior

1. Open the Mac App
2. Refresh with File \ Home or Cmd+Shift+H
3. Try to drag a card to another column
4. The card does not update (this may take a few refreshes)

## Expected behavior

Updates happen

## Screenshots

https://user-images.githubusercontent.com/46905241/148616214-564c0e7d-a421-46ed-bbac-dedac23c8dbf.mov

## Edition and Platform

 - Edition: Personal Desktop
 - Version: v0.11 and v0.12 RC1
 - Browser and OS: macOS Big Sur 16.2.1 and Safari 15.2

## Additional context

Not sure if having Safari 15.2 is a factor, as the WKWebView depends on it.

chenilim:
I've submitted a PR for this, but @ogi-m, can you help check if this repros for you? I saw this on macOS 16.2.1 (Big Sur) and with Safari 15.2 installed, and with the v0.11 and v0.12 RC1 Mac App binaries from GitHub.

ogi-m:
@chenilim I don't have macOS, but I couldn't repro on Windows

PR #2078, opened 2022-01-07
Gh 2077 mac app update

chenilim:
#### Summary
This appears to be a timing issue in JS / React:
1. UseEffect in Boardpage.tsx subscribes and unsubscribes to workspace changes on the websocket (wsClient)
2. The cleanup routine returned from useEffect does the unsubscribe
3. With the Mac App (single-user), there's a pattern of useEffect, useEffect-cleanup, useEffect
4. However, in that last useEffect call, wsClient.state is 'init' (on [this line](https://github.com/mattermost/focalboard/blob/main/webapp/src/pages/boardPage.tsx#L206)), so subscribe isn't called
5. This is because wsClient is still somewhere in the [opening routine](https://github.com/mattermost/focalboard/blob/main/webapp/src/wsclient.ts#L259).

The fix is to update `wsClient.state` at the beginning of `ws.onopen`.

I've only observed this happening in the Mac App, and not in Safari against Personal Server. It's probably due to random differences in timing. There ultimately might be a more robust way to handle "re-entering" wsclient state changes, or change subscriptions.

#### Ticket Link
#2077

### DIFF
--- a/webapp/src/wsclient.ts
+++ b/webapp/src/wsclient.ts
@@ -427,7 +427,6 @@ class WSClient {
         }
 
         this.sendCommand(command)
-        console.log('unsubscribeToWorkspace')
     }
 
     subscribeToWorkspace(workspaceId: string): void {
@@ -442,7 +441,6 @@ class WSClient {
         }
 
         this.sendCommand(command)
-        console.log('subscribeToWorkspace')
     }
 
     unsubscribeFromBlocks(workspaceId: string, blockIds: string[], readToken = ''): void {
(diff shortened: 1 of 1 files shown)


wiggin77 (approved):

chenilim:
Thanks Doug. I propose cherry-picking this change for v0.12, as I think this is affecting customers (though I did not see any other reports in the feedback form or App Store).

The risk should be low, but what do you all think? cc @jespino, @sbishel, @ogi-m.

jespino (approved):
LGTM

jespino:
Looks pretty low impact, so I'm ok merging it into 0.12. But here I prefer to get the @ogi-m opinion, If he doesn't feel comfortable with that, we can defer that until 0.13.

ogi-m (approved):
I can't say I'm too comfortable with this but given it's a one line change and if we think it's affecting customers, then let's cherry-pick it.

chenilim:
Manually cherry-picked to release-0.12.

Issue #2094, opened 2022-01-11
Bug: [Change Notifications] Date, Person, and Multi-select properties not displaying properly in the notification


wuwinson:
## Steps to reproduce the behavior

1. Have someone else or use a secondary account to make changes to the date, person, and multi-select properties on a card you're following:
2. Wait 2 minutes to receive the change notification from Boards Bot.
3. See the error in Boards Bot notification

## Expected behavior

The properties display the same as the card.

## Screenshots
![Change Notifications Bug](https://user-images.githubusercontent.com/93531870/149017549-aa2d159b-6b39-4980-9e63-404dc7785c59.png)

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.12 RC1
 - Browser and OS: Chrome Version 96.0.4664.55 on macOS

wiggin77:
Fixed `Multiselect` and `Date` property types.  `Person` type remains to be fixed.

PR #2110, opened 2022-01-12
Fix prop change formatting for date & multiselect

wiggin77:
#### Summary
This PR fixes property change formatting for notifications, specifically for `Date` and `Multiselect`.  

#### Ticket Link
https://github.com/mattermost/focalboard/issues/2094

### DIFF
--- a/server/model/properties.go
+++ b/server/model/properties.go
@@ -4,13 +4,20 @@
 package model
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
+
+	"github.com/mattermost/focalboard/server/utils"
 )
 
 var ErrInvalidBoardBlock = errors.New("invalid board block")
 var ErrInvalidPropSchema = errors.New("invalid property schema")
 var ErrInvalidProperty = errors.New("invalid property")
+var ErrInvalidPropertyValue = errors.New("invalid property value")
+var ErrInvalidPropertyValueType = errors.New("invalid property value type")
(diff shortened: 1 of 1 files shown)


server/model/properties.go
```diff
@@ -45,13 +52,76 @@ type PropDef struct {
...
+		}
+		return sb.String(), nil
 	}
-	return v
+	return fmt.Sprintf("%v", v), nil
+}
+
+func (pd PropDef) ParseDate(s string) (string, error) {
+	// s is a JSON snippet of the form: {"from":1642161600000} in milliseconds UTC
+	// The UI does not yet support date ranges.
```
sbishel, line 114: The UI does support date ranges.

Screen Shot 2022-01-12 at 2 49 35 PM

You will need to support JSON - {"from":1641513600000,"to":1642161600000} for now.
If when we support time, the full object is - 
```
export type DateProperty = {
    from?: number
    to?: number
    includeTime?: boolean
    timeZone?: string
}
```

sbishel (changes requested):
@wiggin77 We do need to support date ranges.

server/model/properties.go
wiggin77: Fixed via https://github.com/mattermost/focalboard/pull/2110/commits/7190bf108beba3a1e9cdf0c2984fc4304c49de34

sbishel (approved):
LGTM

PR #2118, opened 2022-01-14
Fix #2117. Remove Beta and Alpha language labels

### LINKED ISSUE #2117: Remove "Beta" labels on languages to avoid App Store rejections
## Summary

App Store review flagged the "Beta" labels in the language menu, and rejected a submission. Proposed removing the "(Beta)" and "(Alpha)" labels in the strings

Importance: High

chenilim:
#### Summary
Remove Beta and Alpha labels in language menu.
Also removed some "(Beta)" text in localized strings (may require Weblate update).

#### Ticket Link
#2117


webapp/i18n/fr.json
```diff
@@ -38,7 +38,7 @@
   "ContentBlock.moveUp": "Déplacer vers le haut",
   "ContentBlock.text": "texte",
   "DashboardPage.CenterPanel.ChangeChannels": "Utiliser le sélecteur pour changer de canal",
-  "DashboardPage.title": "Bienvenue dans Focalboard (Beta) !",
```
jespino, line 41: We can merge it, but the source of truth for the translations is the translation platform, not the github repo, so probably is going to be reverted by the PR for new translations. Just to keep in mind this needs to be change there too.

webapp/src/constants.ts
```diff
@@ -56,7 +56,7 @@ class Constants {
         {
             code: 'ru',
             name: 'russian',
-            displayName: 'Pусский (Beta)',
```
jespino, line 59: All this (Beta) labels are about the translation to that language, not about the stage of the project, I think this labels doesn't need to be changed.

jespino (changes requested):
I need confirmation about some changes here.

webapp/src/constants.ts
chenilim: Apple doesn't like seeing "Beta" in the UI. Could we rename this, e.g. to "level 1" or "partial"?

webapp/i18n/fr.json
chenilim: Thanks, I've made the corresponding changes in Weblate. Please check if I submitted them correctly: https://translate.mattermost.com/projects/focalboard/webapp/#history, thanks.

harshilsharma63 (approved):

chenilim:
@jespino, I've updated the weblate translations as well. Were there other changes you wanted in this PR?

If we want to keep the language labels in Weblate, another approach I can think of is to write code to strip " (Beta)" and " (Alpha)" from the menu titles at run time, but that's a bit hacky.

webapp/src/constants.ts
jespino: That is an usability problem in my opinion. Even removing the Beta entirely and not replacing that is a matter of user expectations, so from the technical side I only can say that I don't love it to just change to a language and feel like is partially translated by surprise.

@asaadmahmood what do you think? should we change this text to something like `(Partial)` or `(In Progress)` or `(Incomplete)` or something like that? do you think we should just remove it entirely?

asaadmahmood: @jespino I don't understand the exact problem. Is the reason we're removing this in the first place because of Marketing?
Renaming it to level 1, partial, in progress/incomplete, doesn't make sense to me.

What exactly do we want to tell the user by this label, and where does it appear?

chenilim: The issue is we cannot show "Beta" in the final UI for the Mac App Store (it will fail App review). I can see two ways to address this:
1. Change the label in Weblate to "partial" (or similar)
2. Write code to strip out the "Beta" and "Alpha" labels at runtime

I suppose a third option is to omit all the Beta / Alpha languages.

Do you see another alternative?

asaadmahmood: @chenilim I'm fine then hiding the beta tag, I am fine using the month for the release like we talked about.

chenilim: Thanks @asaadmahmood! @jespino, do you see any other issues before I merge this? Thanks.

jespino (approved):

PR #2133, opened 2022-01-18
Persistent user config

harshilsharma63:
#### Summary
Added API for updating user config to allow for server-side, persisted user-config. This is meant to be a cross-browser alternate for browser local storage.

#### Ticket Link
https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vpb5hydjzqtr7txbxeqpjxyfz4o/cn5nib48arpyu5rog6mx5ba3feo


webapp/src/pages/welcome/welcomePage.test.tsx
```diff
@@ -8,19 +8,28 @@ import {createMemoryHistory} from 'history'
...
+
 const w = (window as any)
 const oldBaseURL = w.baseURL
 
+global.fetch = FetchMock.fn
+
 beforeEach(() => {
-    UserSettings.welcomePageViewed = null
+    FetchMock.fn.mockReset()
+    FetchMock.fn.mockReturnValue(FetchMock.jsonResponse(JSON.stringify({})))
```
kamre, line 32: May be mocking `mutator` like in many other unit tests will be a little bit better.

harshilsharma63: @kamre I'm having a hard time finding where we are mocking `mutator`. Can you share one such example?

harshilsharma63: Ah, got it! Thanks for pointing that out. That definitely looks much better than emulating fetch. I'll make the change 👍

harshilsharma63: Done.

server/api/api.go
```diff
@@ -460,6 +461,82 @@ func (a *API) handlePostBlocks(w http.ResponseWriter, r *http.Request) {
...
+
+	vars := mux.Vars(r)
+	userID := vars["userID"]
+
+	ctx := r.Context()
+	session := ctx.Value(sessionContextKey).(*model.Session)
+
+	auditRec := a.makeAuditRecord(r, "updateUserConfig", audit.Fail)
+	defer a.audit.LogRecord(audit.LevelRead, auditRec)
+	auditRec.AddMeta("performedBy", userID)
```
wiggin77, line 515: We need to keep the fields names as consistent as possible since customers query by them.  These would be `user_id` and `session_id`.

server/server/server.go
```diff
@@ -233,13 +232,6 @@ func NewStore(config *config.Configuration, logger *mlog.Logger) (store.Store, e
 	if err != nil {
 		return nil, err
 	}
-	if config.AuthMode == MattermostAuthMod {
```
wiggin77, line 236: Is this unrelated?

server/swagger/swagger.yml
```diff
@@ -109,6 +109,23 @@ definitions:
         x-go-name: UpdatedFields
     type: object
     x-go-package: github.com/mattermost/focalboard/server/model
+  UserPropPatch:
+    description: UserConfigPatch is a patch for use config
```
wiggin77, line 113:
```suggestion
    description: UserConfigPatch is a patch for user config
```

webapp/i18n/en.json
```diff
@@ -240,7 +240,8 @@
   "ViewTitle.show-description": "show description",
   "ViewTitle.untitled-board": "Untitled board",
   "WelcomePage.Description": "Boards is a project management tool that helps define, organize, track and manage work across teams, using a familiar kanban board view",
-  "WelcomePage.Explore.Button": "Explore",
+  "WelcomePage.Explore.Button": "Take a tour",
```
wiggin77, line 243: Unrelated?

webapp/src/app.tsx
```diff
@@ -128,7 +128,7 @@ const App = React.memo((): JSX.Element => {
     }
 
     const continueToWelcomeScreen = () => {
-        return Utils.isFocalboardPlugin() && loggedIn === true && !UserSettings.welcomePageViewed
+        return Utils.isFocalboardPlugin() && loggedIn === true && !me?.props[UserSettingKey.WelcomePageViewed]
```
wiggin77, line 131: Also unrelated?  Is this PR based on another PR that is unmerged?

wiggin77 (changes requested):
It looks like there are some welcome page changes mixed in here.  I suspect this PR was based on another that is not yet merged .

server/api/api.go
harshilsharma63: `user_id` and `session_id` are already being included by `LogRecord()` method as general audit metadata. The user ID and session user ID are logged here in the context of this specific operation. I know user ID is already included by `LogRecord` but I still wanna keep it here for context.

Would renaming here as `performed_by` and `performed_on` resolve your concern?

server/server/server.go
harshilsharma63: This is intentional. `NewStore` isn't used in plugin mode. This previous code was doing nothing.

server/swagger/swagger.yml
harshilsharma63: Done.

webapp/i18n/en.json
harshilsharma63: Nope, totally related. We're changing the text of buttons that trigger the change in user config.

webapp/src/app.tsx
harshilsharma63: Nope, totally related. We're moving `WelcomePageViewed` from being managed in browser localstorage to server side user props.

harshilsharma63:
> It looks like there are some welcome page changes mixed in here. I suspect this PR was based on another that is not yet merged .

Nope, nothing is mixed in. Everything is related to one single goal - adding server-side API for persisting user config and migrating one config (`WelcomePageViewed`) to use this API.

server/api/api.go
wiggin77: When customers query these audit calls they know `user_id` and `session_id` answer the "who did this".  Adding duplicate fields with the same info doesn't break anything, but potentially confuses the customer.  They wonder, "can `performed_by` ever to be different than `user_id`?  Do I need to special case this API when querying?"

wiggin77:
> > It looks like there are some welcome page changes mixed in here. I suspect this PR was based on another that is not yet merged .
> 
> Nope, nothing is mixed in. Everything is related to one single goal - adding server-side API for persisting user config and migrating one config (`WelcomePageViewed`) to use this API.

Ok, understood.  The PR description only included an API being added.  It would be helpful to have the PR description include the full scope.  Thanks.

server/server/server.go
wiggin77: Ok, intentional but unrelated.

wiggin77 (approved):
LGTM :+1:   I request that the audit log field names be kept standard (don't add redundant fields), but everything else looks great.

server/api/api.go
harshilsharma63: Okay, that makes sense. I've removed these meta fields from here now as they are already being included in `LogRecord`.

harshilsharma63:
> LGTM 👍 I request that the audit log field names be kept standard (don't add redundant fields), but everything else looks great.

Done.

PR #2146, opened 2022-01-19
Fix add, delete and update boards and add board patch generation

mgdelacroix:


webapp/src/blocks/board.ts
```diff
@@ -146,10 +144,149 @@ type BoardGroup = {
...
+    })
+    const newUpdatedColumnCalculations: Record<string, any> = {}
+    Object.keys(newBoard.columnCalculations).forEach((val) => {
+        if (oldBoard.columnCalculations[val] !== newBoard.columnCalculations[val]) {
+            newUpdatedColumnCalculations[val] = newBoard.columnCalculations[val]
+        }
+    })
+
+    const newData: Record<string, any> = {}
+    const protectedFields = ['properties', 'cardProperties', 'columnCalculations']
```
harshilsharma63, line 219: nit: unused variable.

harshilsharma63 (approved):
LGTM 👍

webapp/src/blocks/board.ts
mgdelacroix: Good catch! Updated on 7efa390

webapp/src/mutator.ts
```diff
@@ -153,44 +153,51 @@ class Mutator {
...
     }
 
     async deleteBoard(board: Board, description?: string, beforeRedo?: () => Promise<void>, afterUndo?: () => Promise<void>) {
-        // ToDo: implement
+        await undoManager.perform(
+            async () => {
+                await octoClient.deleteBoard(board.id)
+            },
+            async () => {
+                await octoClient.createBoard(board)
```
jespino, line 196: I'm not 100% sure if this would work, because if you deleting a board, probably "undeleting" means creating it again with all the data again, or maybe we should have something like "undelete", that restore the last existing version from the history, or something like that, and the same for all the underneath children.

mgdelacroix: I just based this change on the `deleteBlock` method, in which we're doing basically the same thing. Completely agree on the need of an "undelete", "unarchive" or whatever name we agree on type of method, but right now I think we can move forward with this solution as we're not changing the original behavior with it.

jespino (approved):
LGTM

Issue #435, opened 2021-05-17
API Integration Tests - ./api/api.go (workspace)


sbishel:
Add integration tests for missing workspace endpoints 

apiv1.HandleFunc("/workspaces/{workspaceID}/blocks/export", a.sessionRequired(a.handleExport)).Methods("GET")
apiv1.HandleFunc("/workspaces/{workspaceID}/blocks/import", a.sessionRequired(a.handleImport)).Methods("POST")

apiv1.HandleFunc("/workspaces/{workspaceID}", a.sessionRequired(a.handleGetWorkspace)).Methods("GET")
apiv1.HandleFunc("/workspaces/{workspaceID}/regenerate_signup_token", a.sessionRequired(a.handlePostWorkspaceRegenerateSignupToken)).Methods("POST")

see ./integrationtests/blocks_test.go and sharing_test.go for samples.

compiledsound:
I can pick this up

purush7:
@sbishel may I work on this?

PR #2159, opened 2022-01-22
GH-435 api integration tests

jpaldeano:
#### Ticket Link

Hi! I hope you don't mind I start my open source adventure with you :) I was looking for a good first issue and found this PR: https://github.com/mattermost/focalboard/pull/1578 , I saw that had some code review comments 3 months old so I thought would be a very good chance for me and an easy thing to start with. 
Note that this my very first commit and PR on an open source project, please forgive any silly mistake I may make.


jpaldeano:
I see it is broken! Apologies - I am new here. Will work to fix it 👍

server/integrationtests/blocks_test.go
```diff
@@ -385,3 +386,139 @@ func TestGetSubtree(t *testing.T) {
...
+			blockTitles[i] = b.Title
+		}
+		require.Contains(t, blockTitles, parentBlock1Title)
+		require.Contains(t, blockTitles, parentBlock2Title)
+		require.Contains(t, blockTitles, childBlock1Title)
+		require.Contains(t, blockTitles, childBlock2Title)
+	})
+
+	t.Run("export with root id", func(t *testing.T) {
+		t.Skip("TODO")
```
jpaldeano, line 462: Good evening @wiggin77 ! I have got a question here. I was trying to Insert some blocks so I could test the import  by using the new `root_id` param. But looking to the api code, it seems I can't get the blocks by using the blockIDs I have created on the test above (new blocks l407-442), and the reason is because some new IDs are being generated internally by the api: https://github.com/mattermost/focalboard/blob/main/server/api/api.go#L433 

In this scenario, how can I import a block that I have just created using the ID I used generated for them? (`utils.NewID(utils.IDTypeBlock)`) Because I won't have access to the ID generated by the API side.

Am I thinking right? 

Best regards,
Joao A.

wiggin77: I agree @jpaldeano, that makes it a challenge to test from the client.  We need to change the block ID's because someone may export from one workspace and import into another, and if we don't change the block IDs they will collide.  

I suggest you create a block with a very unique title, and when validating you'll have to fetch all the blocks and find that unique title.  Bit of a hack but we do need to change the IDs on import.

jpaldeano: Thanks for clarifying that makes sense

wiggin77 (approved):
LGMT 👍   This is great @jpaldeano.  Thanks for adding these tests.

jpaldeano:
Hi @sbishel @wiggin77 I am not sure if these tests still make sense? I have been digging into the code  and application and noticed that the concept of workspaces changed quite a lot?

PR #2161, opened 2022-01-22
Disable `max-lines` ESLint rule

wiggin77:
#### Summary
There are several webapp files that exceed the default ESLint max-lines rule.  This PR disables the rule to avoid polluting the VSCode problems view with these warnings.

#### Ticket Link
NONE

### DIFF
--- a/webapp/.eslintrc.json
+++ b/webapp/.eslintrc.json
@@ -25,6 +25,7 @@
     }
   },
   "rules": {
+		"max-lines": "off",
     "no-unused-expressions": 0,
     "babel/no-unused-expressions": [2, {"allowShortCircuit": true}],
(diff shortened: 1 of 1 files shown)


jespino:
I'm not against that, but I think is a great thing to have to avoid super long files. Should we add the comment to skip that limit in the files itself?

wiggin77:
> I'm not against that, but I think is a great thing to have to avoid super long files. Should we add the comment to skip that limit in the files itself?

Think some alternatives are to increase the setting, or break up the files that are too long.  Otherwise I don't see any point in having the setting.  I'd rather not be tasked with adding an exception to each file that is too long.

harshilsharma63 (approved):

jespino (approved):

PR #2172, opened 2022-01-25
Adding the new template selector interface

jespino:


webapp/src/components/boardTemplateSelector/boardTemplateSelector.tsx
```diff
@@ -0,0 +1,168 @@
...
+            setActiveTemplate(templates.concat(globalTemplates)[0])
+        }
+    }, [templates, globalTemplates])
+
+    if (!allTemplates) {
+        return <div/>
+    }
+
+    return (
+        <div className='BoardTemplateSelector'>
```
wiggin77, line 85: What convention have we adopted for css class names?  We have camel-case and lowercase with hyphens mixed.  Is there some rule that determines which to use?

jespino: In this specific case is because the internal convention is to use the component name as the class of the root element of the component. This way, any where in the css you can use the class `.BoardTemplateSelector` to affect the `BoardTemplateSelector` component, and also, the `boardTemplateSelector.scss` is going to start with the `.BoardTemplateSelector` class as the root of the scss to scope the styles to only affect the current component.

asaadmahmood:
Pushed my changes for the improvement.

Screenshot 2022-01-26 at 8 28 47 PM

webapp/src/widgets/icons/edit.tsx
```diff
@@ -3,20 +3,13 @@
 
 import React from 'react'
 
-import './edit.scss'
+import CompassIcon from './compassIcon'
 
-export default function EditIcon(): JSX.Element {
+export default function DeleteIcon(): JSX.Element {
```
sbishel, line 8: I don't think this rename is correct. Should be `EditIcon` ?

webapp/src/components/boardTemplateSelector/boardTemplateSelectorPreview.tsx
```diff
@@ -0,0 +1,145 @@
...
+        }
+        return getVisibleAndHiddenGroups(activeTemplateCards, activeView.fields.visibleOptionIds, activeView?.fields.hiddenOptionIds, groupByProperty)
+    }, [activeTemplateCards, activeView, groupByProperty])
+
+    if (!activeTemplate) {
+        return null
+    }
+
+    return (
+        <div className='BoardTemplateSelectorPreview'>
```
sbishel, line 67: Should this section be a separate component used both here and in `centerPanel.tsx`?

sbishel (changes requested):
One rename issue and a possible improvement, but otherwise LGTM

webapp/src/widgets/icons/edit.tsx
jespino: :facepalm: I don't know how that happened.

webapp/src/components/boardTemplateSelector/boardTemplateSelectorPreview.tsx
jespino: I guess it can, but I would prefer to defer that until another PR. This is not strictly "repeated code", it is showing a preview, and the other one is showing the real thing. There is a lot of things populated with empty data here. Anyway, yes, we can make it less repeated code.

wiggin77 (approved):
LGTM 👍

asaadmahmood (approved):
LGTM!

sbishel (approved):

webapp/src/components/boardTemplateSelector/boardTemplateSelectorPreview.tsx
sbishel: I'm ok with that.

ogi-m:
Filed [a bug](https://github.com/mattermost/focalboard/issues/2216) for the UI jumping around when moving over custom templates but it's minor and we can address later. Otherwise looks good 👍🏻

ogi-m (approved):

PR #2174, opened 2022-01-25
Update website links.

chenilim:
#### Summary
Cleaned up the homepage links to be more direct, and clearer per audience.

homepage 012522


wuwinson (approved):
Looks good, but concerned about the amount of text the user has to read right from the start. Perhaps something like the [Mattermost pricing page](https://mattermost.com/pricing/) for the 3 user groups would be more visually appealing for a landing page.

PR #2179, opened 2022-01-26
[GH-1543] Switch component style fixed

### LINKED ISSUE #1543: Bug: Switch widget style

1. Open settings pop up menu in bottom left corner
2. See the switch widget for random icons option item.

## Expected behavior

image
image

## Screenshots

image
image

## Edition and Platform

 - Edition: Personal Desktop
 - Version: Latest
 - Browser and OS: Firefox 93 on Arch Linux

kamre:
#### Summary
Switch component style fixed: selector specificity increased by adding additional class so that it can override CSS properties [set](https://github.com/mattermost/focalboard/blob/4662b035014fd0eec503e8a11752af7d38aef3f2/webapp/src/styles/main.scss#L35-L39) in `main.scss`

Before the fix on _personal server_ it looks like this:

After the fix it looks properly:

Details in [this comment](https://github.com/mattermost/focalboard/issues/1543#issuecomment-1013702982).
#### Ticket Link

Fixes #1543


webapp/src/widgets/switch.tsx
```diff
@@ -12,7 +12,7 @@ type Props = {
 
 // Switch is an on-off style switch / checkbox
 function Switch(props: Props): JSX.Element {
-    const className = props.isOn ? 'Switch on' : 'Switch'
+    const className = `Switch OverrideMain${props.isOn ? ' on' : ''}`
```
jespino, line 15: The `OverrideMain` class souldn't be CamelCase, the `Switch` is CamelCase because is the class with the Component name itself, but you can add another class there if you want, but it needs to be lowercase, and problably with dashes.

Do you mind to rename this to `override-main`?

jespino (changes requested):
Just needed a rename of the class name, otherwise looks good to me. Thanks @kamre! :tada:

webapp/src/widgets/switch.tsx
kamre: Renamed. Where can I read about these naming rules?

jespino: I don't think is written anywhere, probably it should be documented anywhere.

jespino (approved):

asaadmahmood (approved):

jespino:
@kamre I don't see anything else needed, lets pass QA and merge it :)

Issue #670, opened 2021-07-03
Bug: Error when running Docker image


Erwyn:
**Summary:**
When I try to run the Docker image I get this error:

```
docker run --rm -it -p 8080:8000  mattermost/focalboard
2021/07/03 11:41:34 readConfigFile
2021/07/03 11:41:34 {ServerRoot:http://localhost:8000 Port:8000 DBType:sqlite3 DBConfigString:/data/focalboard.db DBTablePrefix: UseSSL:false SecureCookie:false WebPath:./pack FilesDriver:local FilesS3Config:{AccessKeyId: SecretAccessKey: Bucket: PathPrefix: Region: Endpoint: SSL:false SignV2:false SSE:false Trace:false} FilesPath:/data/files Telemetry:true PrometheusAddress: WebhookUpdate:[] Secret: SessionExpireTime:2592000 SessionRefreshTime:18000 LocalOnly:false EnableLocalMode:true LocalModeSocketLocation:/var/tmp/focalboard_local.socket AuthMode:native LoggingCfgFile: LoggingCfgJSON: AuditCfgFile: AuditCfgJSON:}
info  [2021-07-03 11:41:34.693 Z] FocalBoard Server                        version=0.6.7 edition=linux build_number=dev build_date="Fri Jul  2 16:15:49 UTC 2021" build_hash=bb7867c2cb5a85b10ece36a66fcbaae8c43233ff
error [2021-07-03 11:41:34.693 Z] Database Ping failed                     error="unable to open database file: no such file or directory"
fatal [2021-07-03 11:41:34.693 Z] server.NewStore ERROR                    error="unable to open database file: no such file or directory"
  github.com/mattermost/logr/v2.Logger.Log
      /go/pkg/mod/github.com/mattermost/logr/v2@v2.0.2/logger.go"55
  github.com/mattermost/focalboard/server/services/mlog.(*Logger).Fatal
      /go/src/focalboard/server/services/mlog/mlog.go"218
  main.main
      /go/src/focalboard/server/main/main.go"165

```

My `mattermost/focalboard` image is up to date.

**Steps to reproduce the behavior:**

```
docker pull mattermost/focalboard
docker run --rm -it -p 8080:8000  mattermost/focalboard
```

**Expected behavior:**
It should spawn the container.

**Screenshots:**
not applicable

**Platform:**
 - OS and version: Docker version 20.10.7, build f0df350 on Fedora 34

**Additional context:**
None

[...]

PR #2193, opened 2022-01-27
Add `chown` for nobody in the docker run example

K3UL:
#### Summary
As mentioned in the last comment on #670, there is an error if you run the Docker image with the instruction of the current readme, mounting a directory as `/data`.

This error (`unable to open database file: no such file or directory`) arises if the host directory is not owned by the user `nobody`.

WIth this PR, I propose to add this information to the README, as well as the `chown` command in the code snippet.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -3,8 +3,10 @@
 ## Docker
 
 The Dockerfile gives a quick and easy way to pull the latest Focalboard server and deploy it locally.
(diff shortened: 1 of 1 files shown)


K3UL:
I have no idea why the cla check is red, I filled all the info on the form and then sent the comment :(

wiggin77 (approved):
LGTM :+1:

harshilsharma63 (approved):

harshilsharma63:
Looks good. Merging this now.

PR #2196, opened 2022-01-28
Add permissions modal

mgdelacroix:
#### Summary
This PR adds the permissions modal initial implementation, which includes the user list, the possibility of updating and removing users from a board and the possibility of adding users typing their username into the user autocompletion field. The sidebar should as well react to membership updates.

Next pull requests should implement the user autocompletion and the share board section.

#### Card Link
https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/v3ng1gwe7wtfaijukrhp8siwcpy/c8ypz1iwsg7brigkgogx13kgdya


server/app/boards.go
```diff
@@ -159,14 +182,35 @@ func (a *App) DeleteBoardMember(boardID, userID string) error {
...
 	if errors.Is(err, sql.ErrNoRows) {
 		return nil
 	}
 	if err != nil {
 		return err
 	}
 
+	// if we're removing an admin, we need to check that there is at
+	// least still another admin on the board
+	if oldMember.SchemeAdmin {
```
harshilsharma63, line 195: This code block is a duplicate from `UpdateBoardMember`. Do you think it makes sense to move this to a utility?

harshilsharma63 (approved):

webapp/src/mutator.ts
```diff
@@ -322,6 +322,49 @@ class Mutator {
...
+            },
+            async () => {
+                await octoClient.updateBoardMember(oldMember)
+            },
+            description,
+            this.undoGroupId,
+        )
+    }
+
+    async deleteBoardMember(member: BoardMember, description = 'delete board member'): Promise<void> {
```
jespino, line 355: @chenilim Adding and removing members is something that we want to track as part of the undo/redo?

jespino (approved):
LGTM, just a small thing from that @harshilsharma63 mentioned so I going to fix that because Miguel is in PTO.

PR #2201, opened 2022-01-29
Import/export moved to server + image support + streaming

wiggin77:
#### Summary
This PR moves archive import & export from webapp to server.  Support for images is added as well.

Archives now have a new file format. They are ZIP files containing a directory per board, with each directory containing a `board.jsonl` file full of serialized blocks and zero or more image files.

Archives are exported in a streaming fashion, meaning the server does not keep the archive or any files in memory.  Archives are streamed when importing with the exception of the `board.jsonl` file which must be loaded before processing until `model.GenerateBlockIDs` can be made to support streams. This will happen when card content order is no longer stored in the card.

Zip was chosen over Tar because Zip can be streamed via a writer, whereas Tar requires each file have its size written before the file. 

#### Ticket Link
Fixes:
https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/150d1464-6d59-4c39-810f-ee201439dbaf/ced1jjs6hbjnyfj9xgynh7dgpoa

Fixes:
https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vd89yph84ajbkufkbgts9m3sb5a/c3semnr1t7jbhz89uy3mxz7fd5c

Required by (for image support):
https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/150d1464-6d59-4c39-810f-ee201439dbaf/7dstzuz9yjnn3pz3ycxo3od8k9r


server/api/archive.go
```diff
@@ -0,0 +1,144 @@
...
+	// ---
+	// produces:
+	// - application/json
+	// parameters:
+	// - name: workspaceID
+	//   in: path
+	//   description: Workspace ID
+	//   required: true
+	//   type: string
+	// - name: root_id
```
jespino, line 27: Should we use `board_id` instead of `root_id`, are we allowing you to export something that is not a board?, eventually the tree is going to dissapear, and also the `root_id` too, so... makes sense to just call the parameter `board_id`?

server/api/archive.go
```diff
@@ -0,0 +1,144 @@
...
+	//   required: false
+	//   type: string
+	// security:
+	// - BearerAuth: []
+	// responses:
+	//   '200':
+	//     description: success
+	//     schema:
+	//       type: array
+	//       items:
```
jespino, line 39: This is not true, the expected result is a binary file.

server/api/archive.go
wiggin77: I can change that.  It's a breaking change to the API but I guess since the file format changed it can already be considered a breaking change.

server/app/export.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	var files []string
+	container := store.Container{
+		WorkspaceID: opt.WorkspaceID,
+	}
+
+	// write the content blocks
+	// TODO: paginate this
+	contentBlocks, err := a.GetBlocksWithRootID(container, board.ID)
```
jespino, line 93: nit: This is also missleading. there are blocks in board that are content (like text, comment, image), but also you have `views`

server/app/export.go
```diff
@@ -0,0 +1,208 @@
...
+		if err := a.writeArchiveBoard(zw, board, opt); err != nil {
+			merr.Append(fmt.Errorf("cannot export board %s: %w", board.ID, err))
+			return
+		}
+	}
+	return nil
+}
+
+// writeArchiveHeader writes a version file to the zip.
+func (a *App) writeArchiveVersion(zw *zip.Writer) error {
```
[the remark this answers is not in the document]
wiggin77: It's writing a file called `version.json` which I agree could have more fields added in the future, but it's primary purpose is to check if the version is supported.  I'll fix the comment to match.  If we eventually need non-version info included we can create a different top-level filed with appropriate name.

server/api/archive.go
wiggin77: Fixed via https://github.com/mattermost/focalboard/pull/2201/commits/4635eb5b6b79391af0ab7f7ecc008d3a7d7704a8

server/api/archive.go
wiggin77: Fixed via https://github.com/mattermost/focalboard/pull/2201/commits/4635eb5b6b79391af0ab7f7ecc008d3a7d7704a8

server/app/export.go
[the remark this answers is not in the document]
wiggin77: Fixed via https://github.com/mattermost/focalboard/pull/2201/commits/4635eb5b6b79391af0ab7f7ecc008d3a7d7704a8

server/app/export.go
wiggin77: Fixed via https://github.com/mattermost/focalboard/pull/2201/commits/4635eb5b6b79391af0ab7f7ecc008d3a7d7704a8

server/app/import.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+		case "board.jsonl":
+			boardID, err := a.ImportBoardJSONL(zr, opt)
+			if err != nil {
+				return fmt.Errorf("cannot import board %s: %w", dir, err)
+			}
+			boardMap[dir] = boardID
+		default:
+			// import file/image;  dir is the old board id
+			boardID, ok := boardMap[dir]
```
jespino, line 66: Can this generate problem if I manually generate this not respecting the expected order on the zip file creation?

I mean, the `boardMap` is populated on the cases where the filename is `board.jsonl` and I guess that is done by finding that file in the zip file in the right order, but If the order is not the right one?

Maybe we should create the boardID before the loop and allow to pass to the `ImportBoardJSONL` an option with the BoardID, that way we would have the new boardID set no matter if the `board.jsonl` has been processed or not.

server/app/import.go
```diff
@@ -0,0 +1,172 @@
...
+	lineReader := bufio.NewReader(r)
+
+	userID := opt.ModifiedBy
+	if userID == model.SingleUser {
+		userID = ""
+	}
+	now := utils.GetMillis()
+
+	lineNum := 1
+	for {
```
jespino, line 104: nit: This is loading everything in memory, would be great to be able to batch this changes (I know is hard because the GenerateIDs, but probably we can get rid of that function call and handle the new Ids generation directly here during the loop.

server/app/import.go
wiggin77: We are using a streaming file format so the order is guaranteed and helps us manage the memory footprint.  That fact it is zip is incidental.  That said, I can do what you suggest to allow different ordering of files in a board directory.  We cannot handle someone changing the order of contents in the `board.jsonl`.

server/model/import_export.go
```diff
@@ -0,0 +1,85 @@
...
+
+// ImportArchiveOptions provides options when importing an archive.
+type ImportArchiveOptions struct {
+	WorkspaceID string
+	ModifiedBy  string
+}
+
+// ErrUnsupportedArchiveVersion is an error returned when trying to import an
+// archive with a version that this server does not support.
+type ErrUnsupportedArchiveVersion struct {
```
jespino, line 51: nice, I like this "got/want" errors.

server/app/import.go
wiggin77: When we get rid of `contentOrder` field in a card's `fields` column then it will be easy to do this.  I have a proposal for that we are debating.  Currently I only load a single board at a time which I figured was ok for now, until we can address the content order issue.  I agree this can and should be improved to not load a whole board, and we will.  Do you feel it has to happen this release?

jespino: No, I don't think so, I something that we need to tackle eventually, especially if the size of the exports start growing and growing, but I think we are ok for now.

server/app/import.go
jespino: We have to decide what is acceptable and what is not. Maybe we can set the limit with things like "order inside boards.jsonl" or define that the zip file needs to add the files in the correct order, but we should be clear upfront with that.

wiggin77: On closer look this is tricky without a streaming version of `GenerateBlockIDs`.  That generates a new board id and I won't know what that new board id is if a file appears first.  That is needed to write out the file.

sbishel (approved):

PR #2203, opened 2022-01-31
Changing active-border to active-border-rgb in the css variables

jespino:


webapp/src/theme.ts
```diff
@@ -183,7 +183,7 @@ export function setMattermostTheme(theme: any): Theme {
     document.documentElement.style.setProperty('--sidebar-bg-rgb', color(theme.sidebarBg).rgb().array().join(', '))
     document.documentElement.style.setProperty('--sidebar-text-rgb', color(theme.sidebarText).rgb().array().join(', '))
     document.documentElement.style.setProperty('--link-color-rgb', theme.linkColor)
-    document.documentElement.style.setProperty('--sidebar-text-active-border', color(theme.sidebarTextActiveBorder).rgb().array().join(', '))
+    document.documentElement.style.setProperty('--sidebar-text-active-border-rgb', color(theme.sidebarTextActiveBorder).rgb().array().join(', '))
```
michelengelen, line 186: just a question: do you still need to call `rgb()` since you already get the rgb channel values in the theme variable, or is this for safety purposes (when hex values come in instead)?

michelengelen (approved):
just a question regarding one of the changes ... otherwise LGTM!

webapp/src/theme.ts
jespino: Yes, exactly is for safety purposes. Shouldn't be hex, but this way can be in any supported format.

harshilsharma63 (approved):

wiggin77 (approved):

PR #2217, opened 2022-02-02
Fixed a bug causing double scroll bars to appear

### LINKED ISSUE #2211: Bug: Double vertical scroll bar

1. Go to Boards
2. Observe the vertical scroll bar - there are two
3. 
## Expected behavior

There should be a single vertical scroll bar

## Screenshots

double_scrollbar

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.12.1
 - Browser and OS: Chrome, Desktop app (Windows)

harshilsharma63:
#### Summary
Fixed a bug causing double scroll bars to appear

#### Ticket Link
Fixes #2211

### DIFF
--- a/webapp/src/pages/boardPage.scss
+++ b/webapp/src/pages/boardPage.scss
@@ -37,7 +37,7 @@
         }
     }
 
-    @media screen and (max-width: 768px) { 
+    @media screen and (max-width: 768px) {
         .mobileWarning {display: flex;}
     }
 }
@@ -45,4 +45,8 @@
 .Dialog.dialog-back.workspaceSwitcherModal > div > .dialog {
     width: 600px;
     height: 468px;
+}
+
+.mainContentRow {
+    height: calc(100% - 40px);
 }


webapp/src/pages/boardPage.scss
```diff
@@ -46,3 +46,7 @@
     width: 600px;
     height: 468px;
 }
+
+.mainContentRow {
+    height: calc(100% - 40px);
```
kamre, line 51: @harshilsharma63 should this fix be applied in `mattermost/webapp` repo? This class name is not used in `focalboard` sources.

harshilsharma63: @kamre no it should not be in webapp. The issue is only with focalboard and not other products. The main cause of this issue is the sidebar but after spending about half an hour, I couldn't figure out what was causing the issue. This CSS  change fixes it, though not the 100% ideal solution.

kamre: @harshilsharma63 but if the class name `mainContentRow` will be renamed in `mattermost/webapp` then this will require to update `focalboard` repo. Also just changing [this 40px](https://github.com/mattermost/mattermost-webapp/blob/0b82a79bd7ea8b52e93fa9f84bf3ac6fecedfb1f/components/global_header/global_header.tsx#L20) will also require an update here in `focalboard` repo inside this `calc` expression.

Looks like this styling is tightly coupled to `mattermost/webapp` repo. And this file `boardPage.scss` is not even inside the plugin code. So may be the issue with styling should be resolved in `mattermost/webapp` repo.

One more notice: for Playbooks that `height: 100%` for `mainContentRow` doesn't look correct, but root component in case of Playbooks has _overflow hidden_ and there is no scroll bar.

harshilsharma63: I'm aware of this. I'll update the styling later. We're making a lot of changes to the sidebar in a separate feature branch so I don't want to spend much time here knowing the whole component is gonna change a few weeks later.

sbishel (approved):
Let's merge and fix with new permissions. This should only be in for 1 release.

PR #2222, opened 2022-02-02
Adding sever side undelete endpoint

### LINKED ISSUE #2079: Bug: Undo delete board doesn't always restore view and cards

1. Delete a board (and confirm)
2. Press Cmd+Z (Ctrl+Z on Win) to undo
3. The board re-appears, but not the views or cards

## Expected behavior

Undo delete board should restore its views and cards as well.

## Screenshots

https://user-images.githubusercontent.com/46905241/148794862-55ece06f-8e95-4986-af2e-f1da41eaa016.mov

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.12 RC1
 - Browser and OS: macOS Chrome

### LINKED ISSUE #2080: Bug: Redo delete card doesn't delete it

1. Delete a card (and confirm)
2. Press Cmd+Z (Ctrl+Z on Win) to undo, card is restored ok
3. Press Shift+Cmd+Z to redo
4. Card is not re-deleted

## Expected behavior

Card is deleted on redo delete

## Screenshots

https://user-images.githubusercontent.com/46905241/148795196-cb825769-0fc1-46e7-89ed-7d02a1c9ba08.mov

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.12 RC1
 - Browser and OS: macOS Chrome

## Additional context

May be related to #2079.

jespino:
fixes #2080
fixes #2079


jespino:
@sbishel this is adding a bit more complexity that a regular fix, because is adding a new endpoint, but because today is the window for the RC1, I think we are on time to merge this. We can find a simpler way to address one of the tickets, but the other one is going to need this anyway, let me know what do you think about adding this to our RC1 if we review it today.

wiggin77:
@jespino this unit test is failing in CI:
```
=== RUN   TestBlocksStore/BlocksStore/UndeleteBlock/exiting_id
    blocks.go:684: 
        	Error Trace:	blocks.go:684
        	Error:      	Received unexpected error:
        	            	Error 1292: Incorrect datetime value: '1643805149690' for column 'insert_at' at row 1
        	Test:       	TestBlocksStore/BlocksStore/UndeleteBlock/exiting_id
=== RUN   TestBlocksStore/BlocksStore/UndeleteBlock/exiting_id_multiple_times
    blocks.go:704: 
        	Error Trace:	blocks.go:704
        	Error:      	Received unexpected error:
        	            	Error 1292: Incorrect datetime value: '1643805149698' for column 'insert_at' at row 1
        	Test:       	TestBlocksStore/BlocksStore/UndeleteBlock/exiting_id_multiple_times
```

server/services/store/sqlstore/blocks.go
```diff
@@ -516,6 +516,76 @@ func (s *SQLStore) deleteBlock(db sq.BaseRunner, c store.Container, blockID stri
...
+		block.RootID,
+		modifiedBy,
+		block.CreateAt,
+		now,
+		0,
+		block.CreatedBy,
+	}
+	insertHistoryQuery := s.getQueryBuilder(db).Insert(s.tablePrefix + "blocks_history").
+		Columns(append(columns, "insert_at")...).
+		Values(append(values, now)...)
```
sbishel, line 574: Should this be deleted from the history table. Otherwise isn't the history only effectively one level deep. Perhaps that is the desire.

sbishel (approved):
The code looks good. The part that confuses me is if the "undeleted" blocks should be removed from blocks history table after an undelete. I keep thinking they should, but don't think I understand the entire flow.

server/services/store/sqlstore/blocks.go
jespino: I think that is the desire. Just adding a new event in the history, it was deleted, and just right after was no longer deleted. The history keep consistent. Even if you want something like "restore" old version, for me is exactly the same, you take the status that was before, and create a new history entry with that data.

harshilsharma63 (approved):
LGTM 👍

wiggin77 (approved):
LGTM :+1:

Issue #2244, opened 2022-02-03
Bug: Images not appearing for cards created from default template


wiggin77:
## Steps to reproduce the behavior

1. Create channel & click FB icon
2. Select  a default template containing an image e.g. `Welcome to Boards`
3. Open card that should have an image e.g. `@mention teammates`

## Expected behavior

Image should appear in card.

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.14 RC
 - Browser and OS: FF 93 Linux

## Additional context

This error is appearing in log:
```json
{
"timestamp": "2022-02-03 18:26:29.939 Z",
"level": "error",
"msg": "API ERROR",
"caller": "app/plugin_api.go:931",
"plugin_id": "focalboard",
"code": "500",
"error": "\"unable to copy file from 4k47wuad37rjprarhxwxdk4jbh/hq5sn6ru5bbuzertkmb74j76ey/b1dgqnsx8pj8q7n5frewmzifiiw/7ek6wbpp19jfoujs1goh6kttbby.gif to 4k47wuad37rjprarhxwxdk4jbh/hq5sn6ru5bbuzertkmb74j76ey/bnyck66mk5jngdea5iyzoaeaoyr/7ek6wbpp19jfoujs1goh6kttbby.gif: The specified key does not exist.\"",
"_msg": "",
"api": "/api/v1/workspaces/hq5sn6ru5bbuzertkmb74j76ey/blocks"
} 
```

chenilim:
Doug confirmed that the images are in the default template. Please investigate.

PR #2253, opened 2022-02-03
Copy images when duplicating card/board

wiggin77:
#### Summary
This PR fixes several issues related to images in cards:

1. previously, only boards created from templates in the same workspace had their images copied
    - now images are copied when duplicating cards, duplicating boards, creating boards from templates, and creating boards from default templates

2. a new file id is generated for copied files so that duplicating a card would not try to copy a file over itself, and leave two cards pointing to the same file.
    - if we want to reduce file copying in the future, we will need to reference count the files

3. in order to determine the workspace id of the source board, we need the `GetBlock` API to not specify workspaceID in the filter
    - now that all block ids are unique, we are going to remove container/workspace id from the store API's anyway
    - we will likely remove the workspace id from the file paths in future

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/2244

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -297,7 +297,7 @@ func (a *API) handleGetBlocks(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	case blockID != "":
-		block, err = a.app.GetBlockByID(blockID)
+		block, err = a.app.GetBlockByID(*container, blockID)
 		if err != nil {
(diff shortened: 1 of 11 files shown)


sbishel (approved):

ogi-m (approved):
Verified the images in boards created from the default templates are now appearing as expected

PR #2255, opened 2022-02-04
Fied a bug where user Cannot edit description block if only a space

### LINKED ISSUE #2063: Bug: Cannot edit description block if only a space

1. Edit card description to have only a space
2. Set focus somewhere else to commit block
3. Try editing same block
4. Cannot edit it

## Expected behavior

Edit the block that only has a space

 - Edition: Mattermost plugin
 - Version: v0.12.0   bug bash
 - Browser and OS: Linux FF 95.0.1

harshilsharma63:
#### Summary
Fixed a bug where the user cannot edit description block if only a space.

#### Ticket Link
Fixes #2063

The problem was that only a space in the text, the HTML element didn't take any height, so the user couldn't click on the element to trigger an edit.

### DIFF
--- a/webapp/src/components/markdownEditor.scss
+++ b/webapp/src/components/markdownEditor.scss
@@ -20,6 +20,8 @@
     }
 
     .octo-editor-preview {
+        min-height: 2em;
+
         a {
             &:hover {
                 text-decoration: underline;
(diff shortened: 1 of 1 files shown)


chenilim (approved):
Looks good! Not sure why the original min-height was removed, but adding this makes sense.

chenilim:
cc @asaadmahmood.

harshilsharma63:
@ogi-m added added you as the QA - reviewer.

sbishel (approved):

ogi-m (approved):
LGTM 👍🏻

Issue #2257, opened 2022-02-04
Bug: Reordering columns in table view doesn't update calculations row


ogi-m:
## Steps to reproduce the behavior

1. Create a board and add a table view
2. Create a few cards and create a number property and a text property and fill it in on a few cards
3. Hover over the "Calculate" option below the column with number property and select one of the number calculations, for example "Sum"
4. Now switch places for the two columns
5. Observe the calculations row doesn't update - the "Sum" calculation is now below the text column and shows the sum for the other column

## Expected behavior

Calculations row should update to match the table columns

## Screenshots

Original order:
![SharedScreenshot_1](https://user-images.githubusercontent.com/15314183/152545809-841be929-3d9e-4707-8652-964b3a7843f3.jpg)

After updating columns:
![SharedScreenshot_2](https://user-images.githubusercontent.com/15314183/152545814-a3d33840-2a5f-4cb0-ad09-7e0ceced503b.jpg)

## Edition and Platform

 - Edition: Personal Desktop / Personal Server / Mattermost plugin
 - Version: v0.14

ogi-m:
Regression in v0.12, it's working as expected on v0.11

ogi-m:
Regression in v0.12, it's working as expected on v0.11 

cc @chenilim @wuwinson @sbishel

sbishel:
@chenilim, @wuwinson I have a fix for this issue, I don't think there is a work-around, so we may want to take this one.

chenilim:
Thanks @sbishel! Yup agree. Took a look at the PR, and I think this is good to CP for v0.14.

lindy65:
I'm reopening this issue as it's regressed in v0.16.0

![image](https://user-images.githubusercontent.com/13020833/161034453-4573f891-b595-4101-9e8f-da721de4b5cf.png)

sbishel:
Closing...opened as new issue - https://github.com/mattermost/focalboard/issues/2695

PR #2262, opened 2022-02-04
GH-2257: Calculation column order to be based on view

sbishel:
#### Summary
A change was made in v0.12 to start using the views "visibleProperties" to define the order of properties. Previously, all views based order on the boards "cardProperties" order.

This PR updates the calculationRows to use the view's "visibleProperties".

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2257


harshilsharma63 (approved):
LGTM 👍

Issue #2258, opened 2022-02-04
Bug: Unable to activate date picker in table view


ogi-m:
## Steps to reproduce the behavior

1. Create a board with several cards
2. Add a date property
3. Add a table view
4. Click on the fields in the Date column - nothing happens

## Expected behavior

The date picker should pop up

## Screenshots

https://user-images.githubusercontent.com/15314183/152547485-c75d1467-15a0-4fb2-9363-1295cd6b9bfb.mp4

## Edition and Platform

 - Edition: Personal Desktop / Personal Server / Mattermost plugin
 - Version: v0.14

## Additional context

Users have to open a card to add the date instead of being able to fill in the property from table view.

chenilim:
Thanks @ogi-m. Hmmm... not sure if this worked before (also repros for me on v0.12 on community). @sbishel, do you know if this is a regression? In any case, this should work, but I think it can be post v0.14 if needed. Workaround is to open the card and edit the property there. We can discuss at the next triage meeting. cc @wuwinson.

sbishel:
Yep. I'm pretty sure it is a regression. But not sure when it last worked, probably CSS change somewhere.

sbishel:
I had to go back to v0.9.0 before I found a working version, so its been this way a while.

sbishel:
I found the issue and have a fix. The fix is pretty specific, but I really don't want to take a CSS change at this point. It has been this way for the past 3 releases with no complaints.

ogi-m:
This reproduces again on latest main

ogi-m:
Tested and passed on v0.15 RC2

PR #2268, opened 2022-02-05
GH-2258: allow date range to overflow

sbishel:
#### Summary
Allows DateRange to overflow so the calendar picker is displayed.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2258


harshilsharma63:
I've fixed a bug in here causing the dates to overflow into neighbouring columns.

Screenshot 2022-02-09 at 1 48 00 PM

harshilsharma63 (approved):
LGTM 👍

PR #2287, opened 2022-02-08
Onboarding Tour

harshilsharma63:
#### Summary
Implements onboarding tour for Boards.

#### Ticket Link
https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vpb5hydjzqtr7txbxeqpjxyfz4o

Testing tip -

To restart the tour, set the user props to an empty object in the database.
For example, for MySQL -

```sql
update Users set Props = '{}' where Id = '<USER_ID>';
```


webapp/src/components/centerPanel.tsx
```diff
@@ -112,14 +135,55 @@ class CenterPanel extends React.Component<Props, State> {
         return true
     }
 
+    shouldStartBoardsTour(): boolean {
```
jespino, line 138: nit: This component is already big enough, can we move this logic into something else, maybe a "behavior component", something like a component called "BoardsTourInit" that takes care of this with the lifecycle, that way you can simplify a lot that and remove from this already big component all this logic.

server/model/user.go
```diff
@@ -60,7 +60,7 @@ type User struct {
 type UserPropPatch struct {
 	// The user prop updated fields
 	// required: false
-	UpdatedFields map[string]string `json:"updatedFields"`
+	UpdatedFields map[string]interface{} `json:"updatedFields"`
```
jespino, line 63: Why do we need this changes, is no way to keep it just `map[string]string`, this kind of things tends to be more fragile. If that is needed, is ok, but if we can solve that using multiple keys instead of a nested object, would result in something more robust IMO.

webapp/src/components/centerPanel.tsx
harshilsharma63: This is not a generic logic but rather only specific for the board tour being displayed in this component. So won't make sense storing it outside.

server/model/user.go
harshilsharma63: Done.

webapp/src/components/kanban/kanbanCard.tsx
```diff
@@ -81,18 +102,53 @@ const KanbanCard = React.memo((props: Props) => {
         setShowConfirmationDialogBox(true)
     }
 
+    const isOnboardingBoard = board.title === 'Welcome to Boards!'
```
jespino, line 105: Same here, you are adding a lot of logic here in a component that maybe can be extracted anywhere else, maybe move all this logic inside the `OpenCardTourStep` and `CopyLinkTourStep` components.

webapp/src/components/pulsating_dot/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {Coords} from '../tutorial_tour_tip/tutorial_tour_tip_backdrop'
+
+type Props = {
+    targetRef?: React.RefObject<HTMLImageElement>;
+    className?: string;
+    onClick?: (e: React.MouseEvent) => void;
+    coords?: Coords;
+}
+
+export class PulsatingDot extends React.PureComponent<Props> {
```
jespino, line 15: Why is this a class component?

webapp/src/components/tutorial_tour_tip/tutorial_tour_tip.tsx
```diff
@@ -0,0 +1,330 @@
...
+        handleSkipTutorial,
+        handleSavePreferences,
+        getLastStep,
+        handleEventPropagationAndDefault,
+        handleSendToNextTour,
+    } = useTutorialTourTipManager(managerProps)
+
+    const getButtonText = (): JSX.Element => {
+        let buttonText = (
+            <>
```
jespino, line 122: Why this have a fragment for the if only has one element?

webapp/src/components/tutorial_tour_tip/useElementAvailable.ts
```diff
@@ -0,0 +1,32 @@
...
+                setAvailable(true)
+                if (checkAvailableInterval.current) {
+                    clearInterval(checkAvailableInterval.current)
+                    checkAvailableInterval.current = null
+                }
+            }
+        }, 500)
+    }, [])
+
+    return useMemo(() => available, [available])
```
jespino, line 31: I think this memo is doing absolutely nothing

webapp/src/components/viewHeader/viewHeader.tsx
```diff
@@ -59,32 +75,71 @@ const ViewHeader = React.memo((props: Props) => {
 
     const hasFilter = activeView.fields.filter && activeView.fields.filter.filters?.length > 0
 
+    const isOnboardingBoard = props.board.title === OnboardingBoardTitle
```
jespino, line 78: All this code, is again, like 50 lines of code to add to one component to show decide to show or not the tooltip, I don't think it belongs to the main responsaibility of this component, it is making the component hard to understand.

jespino (changes requested):
In general looks really good, I have some concerns, specially about the separation of concerns of some of the components.

Also I found some small weird cases like a class based component, a unneeded useMemo or an undeeded fragment.

webapp/src/components/pulsating_dot/index.tsx
harshilsharma63: It's copy-pasted from Mattermst webapp. I don't want to make any changes to it and make it a Board's component. Copy this component was required because we don't have any good way of sharing components right now. I want to leave it as it is.

Thoughts?

webapp/src/components/tutorial_tour_tip/tutorial_tour_tip.tsx
harshilsharma63: Done.

webapp/src/components/viewHeader/viewHeader.tsx
harshilsharma63: The logic is different in different places. We still need to get the tour state in the current component because it's stored in the redux store, and a utility function can't access it.

harshilsharma63: Let me try one thing. It might work.

webapp/src/components/pulsating_dot/index.tsx
jespino: Actually, I prefer to do it all the way around, we want channels to migrate more into functional components whenever makes sense, and in this cases, looks like makes a lot of sense. So I would suggest to change this to functional component and make a PR in channels to copy this over there.

webapp/src/components/viewHeader/viewHeader.tsx
harshilsharma63: Done.

webapp/src/components/tutorial_tour_tip/useElementAvailable.ts
harshilsharma63: Done.

webapp/src/components/pulsating_dot/index.tsx
harshilsharma63: Fixed in Boards for now.

webapp/src/components/kanban/kanbanCard.tsx
harshilsharma63: Done.

webapp/src/components/onboardingTour/shareBoard/shareBoard.tsx
```diff
@@ -30,23 +28,25 @@ const ShareBoardTourStep = (): JSX.Element => {
...
-            tutorialCategory={TOUR_CARD}
-            autoTour={true}
-            placement={'bottom-end'}
-            className='ShareBoardTourStep'
-            hideBackdrop={true}
-            imageURL={Utils.buildURL(shareBoard, true)}
-            skipCategoryFromBackdrop={true}
-            telemetryTag='tourPoint2b'
-        />
+        <>
```
jespino, line 33: nit: I would use something like:

```js
if (!BoardTourSteps.SHARE_BOARD) {
  return null
}
```

and then return the `TourTipRenderer` thing. That way is just cleaner and simpler.

webapp/src/components/onboardingTour/tourTipRenderer/tourTipRenderer.tsx
```diff
@@ -0,0 +1,74 @@
...
+    classname: string
+    telemetryTag: string
+    placement: Placement | undefined
+    hideBackdrop: boolean
+    imageURL?: string
+    singleTip?: boolean
+    hideNavButtons?: boolean
+}
+
+const TourTipRenderer = (props: Props): JSX.Element | null => {
```
jespino, line 31: Nice

jespino (approved):
LGTM, I just added an small comment to improve a bit the code.

webapp/src/components/onboardingTour/tourTipRenderer/tourTipRenderer.tsx
harshilsharma63: 👍

webapp/src/components/onboardingTour/shareBoard/shareBoard.tsx
harshilsharma63: Done.

ogi-m:
1. I'm seeing an issue after finishing the 1st tour point. For the second tour point, the spec says that it can be triggered from any card as long as it's the user's very fist card. I'm seeing that it has to be the "Create a new card", clicking on any other card doesn't trigger actually triggers the first tour point and it looks like this:

onboarding 

2. I can't create a new board. Nothing happens when clicking on "Use this template". This doesn't repro on main for me.
Here's the console error:

onboarding1

3. Minor text discrepancy:
Second guided tour, third point, the text says:
_Add a description to your card so **people** know what it's about._
Expected (according to the spec): _Add a description to your card so **your teammates** know what it's about._

3. I don't see an option to restart the tour in the settings menu. Has this been delayed @wuwinson ?



@harshilsharma63 @sbishel @jespino

jespino:
@ogi-m point 1 fixed, I'll work in the other points now. It should be quick.

jespino:
@ogi-m @wuwinson just to confirm, now the tour is not available in personal server, and desktop app, is that expected, or is that something that we want to fix? (IMO we should fix that)

jespino:
@ogi-m fixed 3 and 4, and 2 is not reproducing in my environment.

wuwinson:
@ogi-m @jespino Thanks for filing and addressing these bugs!

- Point 1: User should be able to load next tour point from any card, but okay if it only loads from the card with the green dot.
- Point 2: Also seeing the same issue when I tested on https://ogi-master.test.mattermost.cloud/signup_user_complete/?id=s74jbfi4p3bhufdiaft1cdzo6y
- Point 3: Expected text is the right one, nice catch!
- Point 4: Should be able to restart the tour through the "Product Tour" menu option under Settings or by loading the Welcome template from the template picker. Unable to test the Welcome template due to the issue in Point 2.
- Tour should be available in personal server and desktop app, although the UX may be a bit weird since we call out things like @mentions and copying card links, which isn't relevant for the desktop app. Will need to address this at a later time, but tour should still work there nonetheless.

wuwinson:
@jespino After discussing with @chenilim, it may be better to turn off the product tour for the Personal Desktop app since there are a few things we call out on the tour points that will not work for Personal Desktop. cc @ogi-m

ogi-m:
@wuwinson Should the green dot persist on the first card until it's opened? The current behaviour is that it gets removed after the first tour point is dismissed.

https://user-images.githubusercontent.com/15314183/155772214-4ee7eda8-badd-4bff-8e4f-1aede2d635b2.mp4

wuwinson:
> @wuwinson Should the green dot persist on the first card until it's opened? The current behaviour is that it gets removed after the first tour point is dismissed.
> 
>  5339fe4d9242a804ffc324eb26b37b42.mp4

@ogi-m Clicking on the X skips that portion of the tour, so green dot should disappear in this case, but the next tour point should still trigger. Green dot should not disappear if user did not explicitly click on the X, but clicked on somewhere else on the board to minimize the tour point.

ogi-m:
Ah ok, got it now, thanks Winson. That's working as expected then. The previously reported issues have been resolved again.


Approving 👍🏻

ogi-m (approved):

PR #2288, opened 2022-02-08
Replace `npm install` with `npm ci` to ensure dependencies are properly pinned during CI/CD

oh6hay:
Running `npm ci` instead of `npm install` essentially verifies dependencies match the intended versions and makes the build fail if they don't match, please see https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci#53325242 and the npm docs for more information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,8 +11,8 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


jespino (approved):
LGTM. Thanks @oh6hay!

sbishel (approved):

PR #2294, opened 2022-02-08
[MM-36643] Use Mattermost SiteURL to determine rudder cookie domain

JulienTant:
#### Summary
When mattermost is installed on a subdomain (e.g. `mattermost.example.com`), rudder parses the top-level domain and sets their cookies using a general `.example.com` making the cookie available in every subdomain. This PR ensures the cookie is set to the actual subdomain by parsing SiteURL from the config

#### Related PR
Same fix on the webapp https://github.com/mattermost/mattermost-webapp/pull/9764

#### Ticket Link
Related to https://mattermost.atlassian.net/browse/MM-36643

### DIFF
--- a/mattermost-plugin/webapp/src/index.tsx
+++ b/mattermost-plugin/webapp/src/index.tsx
@@ -172,7 +172,14 @@ export default class Plugin {
             }
 
             if (rudderKey !== '') {
-                rudderAnalytics.load(rudderKey, rudderUrl)
+                const rudderCfg = {} as {setCookieDomain: string}
+                if (siteURL && siteURL !== '') {
+                    try {
+                        rudderCfg.setCookieDomain = new URL(siteURL).hostname
+                        // eslint-disable-next-line no-empty
+                    } catch (_) {}
+                }
+                rudderAnalytics.load(rudderKey, rudderUrl, rudderCfg)
 
                 rudderAnalytics.identify(config?.telemetryid, {}, TELEMETRY_OPTIONS)
 


mattermost-plugin/webapp/src/index.tsx
```diff
@@ -172,7 +172,14 @@ export default class Plugin {
             }
 
             if (rudderKey !== '') {
-                rudderAnalytics.load(rudderKey, rudderUrl)
+                const rudderCfg = {} as {setCookieDomain: string}
+                if (siteURL && siteURL !== '') {
```
pablovelezvidal, line 176: maybe worth adding a simple url validation check here? WDYT?
```suggestion
                if (siteURL && siteURL !== '' && /^(http|https):\/\/[^ "]+$/.test(siteURL)) {
```

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -172,7 +172,14 @@ export default class Plugin {
             }
 
             if (rudderKey !== '') {
-                rudderAnalytics.load(rudderKey, rudderUrl)
+                const rudderCfg = {} as {setCookieDomain: string}
```
pablovelezvidal, line 175: why not  just initialize the object instead of forcing the type and let typescript to directly infer it. WDYT?
```suggestion
                const rudderCfg = {setCookieDomain: ''}
```

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -172,7 +172,14 @@ export default class Plugin {
...
+                if (siteURL && siteURL !== '') {
+                    try {
+                        rudderCfg.setCookieDomain = new URL(siteURL).hostname
+                        // eslint-disable-next-line no-empty
+                    } catch (_) {}
```
pablovelezvidal, line 180: nit: given that we have the catch section, is there maybe something better that just ignoring? not sure if in this case is worth even doing some sort of logging or even tracking the failure somehow.
as mentioned before, it is a nit, so, feel free to don't even provide an elaborate answer to this comment.

pablovelezvidal (changes requested):
good work @JulienTant , some minor questions before approving, thanks!

mattermost-plugin/webapp/src/index.tsx
harshilsharma63: SiteURL is being received from MM config. I don't think it makes much sense to add a second validation here.

harshilsharma63 (approved):

mattermost-plugin/webapp/src/index.tsx
JulienTant: That was my first version actually, but given that this is a value we pass to a third-party library, I prefer giving `undefined` rather than an empty string when we don't set it.
Passing an empty string would work today because they are doing a pretty loose test (`options && options.setCookieDomain`) but I feel better giving them `undefined`

mattermost-plugin/webapp/src/index.tsx
JulienTant: I went this way because when your reach `new URL`, siteURL cannot be empty and have been validated beforehand by the server, so there's no reason for `new URL` to throw an exception... but if it does for some reason, I don't want to interrupt anything, so we just don't set the value.

mattermost-plugin/webapp/src/index.tsx
pablovelezvidal: ok, thanks for the clarification!

mattermost-plugin/webapp/src/index.tsx
pablovelezvidal: ok, thanks!

mattermost-plugin/webapp/src/index.tsx
pablovelezvidal: I see your point, thanks for the clarification.

pablovelezvidal (approved):
LGTM! thanks @JulienTant

sbishel (approved):

Issue #2251, opened 2022-02-03
Bug: Error starting plugin - server version mismatched


sbishel:
## Steps to reproduce the behavior

The Channel Intro PR did not get committed in time for Mattermost v6.4.

Therefore, the PR for the changes in Focalboard will need to be reverted.

PR #2299, opened 2022-02-08
wrap registration block to catch error

sbishel:
#### Summary
Fixes issue when running against an older mattermost-server that does not implement `registry.registerChannelIntroButtonAction`

#### Ticket Link
https://github.com/mattermost/focalboard/issues/2251


harshilsharma63 (approved):
This works for now but can we update this to match the existing pattern? Currently, we check if `this.registry.<some hook>` is truthy, and use it if so.

For example this line https://github.com/mattermost/focalboard/blob/9d12efa2a2577336c25cfc7a99e6bcc11f860106/mattermost-plugin/webapp/src/index.tsx#L135

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -148,7 +148,12 @@ export default class Plugin {
                 UserSettings.lastViewId = null
                 window.open(`${windowAny.frontendBaseURL}/workspace/${currentChannel}`, '_blank', 'noopener')
             }
-            this.channelHeaderButtonId = registry.registerChannelIntroButtonAction(<FocalboardIcon/>, goToFocalboardTemplate, 'Boards')
+            try {
```
jespino, line 151: I think I prefer a this way:
```ts
// backward compatiblity for version < v6.5
if (registry.registerChannelIntroButtonAction) {
    this.channelHeaderButtonId = registry.registerChannelIntroButtonAction(<FocalboardIcon/>, goToFocalboardTemplate, 'Boards')
}
```

jespino (changes requested):
I think I prefer a different way that doesn't mask possible unrelated errors.

PR #2301, opened 2022-02-09
Only build binary for current architecture in dev mode

hanzei:
#### Summary
If `MM_SERVICESETTINGS_ENABLEDEVELOPER` is set, `make sever` will build the binary only for the current architecture.

This allows dev to deploy the plugin faster. See https://github.com/mattermost/mattermost-plugin-starter-template/pull/161 for the original change.

#### Ticket Link
None

### DIFF
--- a/mattermost-plugin/Makefile
+++ b/mattermost-plugin/Makefile
@@ -29,6 +29,9 @@ GO_TEST_FLAGS ?= -race
 GO_BUILD_FLAGS ?= -ldflags '$(LDFLAGS)'
 MM_UTILITIES_DIR ?= ../mattermost-utilities
 DLV_DEBUG_PORT := 2346
+DEFAULT_GOOS := $(shell go env GOOS)
+DEFAULT_GOARCH := $(shell go env GOARCH)
+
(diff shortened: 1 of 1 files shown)


wiggin77:
@hanzei this is a great idea.  Wouldn't it be better to default to building just for the current arch and doing the full build when a env var is set?

hanzei:
> Wouldn't it be better to default to building just for the current arch and doing the full build when a env var is set?

There are two reasons why I like the current approach:
1. CI can stay the same. That makes (production) build more reliable.
2. There are a couple of plugins who already use the approach suggested in this PR. e.g. https://github.com/mattermost/mattermost-plugin-starter-template/pull/161 Consistency allows devs to move between plugins and expect the built tools to work the same.

hanzei:
@wiggin77 While our suggestions make sense to me, I won't be able to apply them to this PR due to time constrains. Would you prefer merging this PR as it's (as first step in the right direction) or closing it in favor of the HW ticket linked here?

wiggin77:
I'll create a new PR.  Thanks @hanzei

chenilim:
@wiggin77, do you have a new issue or PR for this? Please link to it when available (it'll help track progress of solutions to the core issue). Thanks.

PR #2317, opened 2022-02-11
Don't use property value for key construction.

kamre:
#### Summary
Usage of property value in key construction leads to unmount/remount for property value component when the value is changed.

This can sometimes cause a [failure in Cypress test](https://github.com/mattermost/focalboard/runs/5097312160?check_suite_focus=true) `cardURLProperty.ts` because the button that was found by Cypress will be detached from DOM after remount of react component.

#### Ticket Link
No linked ticket.

### DIFF
--- a/webapp/src/components/cardDetail/cardDetailProperties.tsx
+++ b/webapp/src/components/cardDetail/cardDetailProperties.tsx
@@ -135,10 +135,9 @@ const CardDetailProperties = React.memo((props: Props) => {
     return (
         <div className='octo-propertylist CardDetailProperties'>
             {board.fields.cardProperties.map((propertyTemplate: IPropertyTemplate) => {
-                const propertyValue = card.fields.properties[propertyTemplate.id]
                 return (
                     <div
-                        key={propertyTemplate.id + '-' + propertyTemplate.type + '-' + propertyValue}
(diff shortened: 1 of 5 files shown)


harshilsharma63:
@kamre having the same card open in two tabs and updating a date property in one tab is not updating it in the other tab. It works on `main` though. Can you verify that?

harshilsharma63 (changes requested):
Date property changes not being propagated to WebSocket.

kamre:
@harshilsharma63 this is just how `DateRange` is [written currently](https://github.com/mattermost/focalboard/blob/02b50275ee5b011b9f306f07b2cb423ef83a74e7/webapp/src/components/properties/dateRange/dateRange.tsx#L71), i.e. `value` passed in props is only used as _initial value and not current_ value. 

In general this is not a good practice as you can [read here](https://overreacted.io/writing-resilient-components/#principle-1-dont-stop-the-data-flow) for example. I will look into how this can be improved.

harshilsharma63:
Ah, okay, it was working on `main` because value itself was being used element key, forcing react to update the component. I can look into making the date property work if you want. LMK.

harshilsharma63 (approved):
LGTM!

PR #2351, opened 2022-02-16
Update readme with accurate Linux standalone app build instructions

wiggin77:
#### Summary
This PR updates the readme to reflect changes to the standalone app build for Linux.  

#### Ticket Link
NONE

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,8 @@ You can build standalone apps that package the server to run locally against SQL
         * `sudo apt-get install libgtk-3-dev`
         * `sudo apt-get install libwebkit2gtk-4.0-dev`
     * `make linux-app`
-    * run `linux/dist/focalboard-app`
+    * uncompress `linux/dist/focalboard-linux.tar.gz` to a directory of your choice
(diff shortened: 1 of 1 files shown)


chenilim:
cc @jespino. Shouldn't a developer be able to run the generated binary directly? Is the issue that certain pre-reqs are not in the same folder?

wiggin77:
> cc @jespino. Shouldn't a developer be able to run the generated binary directly? Is the issue that certain pre-reqs are not in the same folder?

The only thing output to `dist` during a build is the gzipped file.  Some artefacts are put in `bin` but running from there just displays an empty window and no errors output.

wuwinson:
Adding @chenilim as a reviewer instead. Updates to the readme look fine to me, but I'm not sure if anything should be added/changed. cc @wiggin77

chenilim (approved):
Thanks @wiggin77. This is a good change, because at least it gets the app working. Might just be a quick followup with @jespino if this is the expected workflow, and if something changed from the original.

PR #2353, opened 2022-02-17
fix lint errors

sbishel:
#### Summary
Fixes current lint errors.

### DIFF
--- a/webapp/src/octoUtils.test.ts
+++ b/webapp/src/octoUtils.test.ts
@@ -2,7 +2,6 @@
 // See LICENSE.txt for license information.
 
 import {Block} from './blocks/block'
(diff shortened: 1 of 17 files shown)


webapp/src/octoUtils.test.ts
```diff
@@ -53,26 +52,26 @@ test('duplicateBlockTree: Card', async () => {
     }
 })
 
-function createBoardTree(): [Block[], Board] {
-    const blocks: Block[] = []
+// function createBoardTree(): [Block[], Board] {
```
sbishel, line 55: commented out because the test that uses it is commented out above.

harshilsharma63 (approved):
Looks fine. The SQLite test is failing thoiugh.

PR #2354, opened 2022-02-17
Shows "(Deleted User)" instead of UUID when user not found

akkivasu:
In case a user is not found, at present unreadable and long UUIDs are shown which kill the look and feel of the application. This patch replaces the UUID with a more explanatory string.

### DIFF
--- a/server/services/store/mattermostauthlayer/mattermostauthlayer.go
+++ b/server/services/store/mattermostauthlayer/mattermostauthlayer.go
@@ -434,7 +434,7 @@ func (s *MattermostAuthLayer) userWorkspacesFromRows(rows *sql.Rows) ([]model.Us
 
 			for _, userID := range userIDs {
 				user, exists := users[userID]
-				//Shows "(Deleted User)" instead of long, unreadable UUID in case the user is not found
+				// Shows "(Deleted User)" instead of long, unreadable UUID in case the user is not found
 				username := "(Deleted User)"
 				if exists {
 					username = user.Username
--- a/server/services/store/mattermostauthlayer/mattermostauthlayer.go
+++ b/server/services/store/mattermostauthlayer/mattermostauthlayer.go
@@ -434,7 +434,8 @@ func (s *MattermostAuthLayer) userWorkspacesFromRows(rows *sql.Rows) ([]model.Us
 
 			for _, userID := range userIDs {
 				user, exists := users[userID]
-				username := userID
+				//Shows "(Deleted User)" instead of long, unreadable UUID in case the user is not found
+				username := "(Deleted User)"
 				if exists {
 					username = user.Username
 				}


wiggin77:
Hi @akkivasu.  We would need to localize this string, and it may be easier for now to do that on the client side.  Also I would like our PM to weight on whether it would be better to store a username with the userID so we can display something more meaningful.  cc @wuwinson

wuwinson:
@akkivasu @wiggin77 Agreed that showing the username would be more meaningful than "(Deleted User)", but would be even better if we could display something like "username (Deleted User)", to provide more helpful information.

Simply displaying "(Deleted User)" is still better than our current experience and could be a good first step towards showing it in combination with the username, so I'm okay with what we have here as long as we can iterate on it. cc @asaadmahmood @chenilim

chenilim:
Thanks @akkivasu! Agree this is a step in the right direction.

I think we can take this as is and I propose filing a separate card or issue to track storing the user name with the user. There are important corner cases to consider for that, e.g. that the authors of cards don't get spoofed. If we were to display the "fallback / unverified" user name, it should be indicated as such.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -434,7 +434,8 @@ func (s *MattermostAuthLayer) userWorkspacesFromRows(rows *sql.Rows) ([]model.Us
 
 			for _, userID := range userIDs {
 				user, exists := users[userID]
-				username := userID
+				//Shows "(Deleted User)" instead of long, unreadable UUID in case the user is not found
```
wiggin77, line 437:
```suggestion
				// Shows "(Deleted User)" instead of long, unreadable UUID in case the user is not found
```

wiggin77 (approved):

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -434,7 +434,8 @@ func (s *MattermostAuthLayer) userWorkspacesFromRows(rows *sql.Rows) ([]model.Us
 
 			for _, userID := range userIDs {
 				user, exists := users[userID]
-				username := userID
+				// Shows "(Deleted User)" instead of long, unreadable UUID in case the user is not found
+				username := "(Deleted User)"
```
sbishel, line 438: This should use `intl.Format()` so it will be translated.

sbishel (changes requested):
This string should be translatable. Also please run `npm run i18n-format` to create the string in the english file.

sbishel:
Actually, @wiggin77 pointed out, we do not yet have translations on the server side. Accepting this PR as is for now.

sbishel (approved):

Issue #2356, opened 2022-02-18
Permissions branch: Updating a text content block fails


chenilim:
## Steps to reproduce the behavior

1. Open a card
2. Set a description (text content, e.g. to "text 1") - this works the fist time
3. Update the description (e.g. to "text 2")
4. Nothing happens, and errors in the console

## Expected behavior

Text content (description) can be updated

## Edition and Platform

 - Edition: Personal Desktop / Personal Server / Mattermost plugin
 - Version: permissions-feature-branch
 - Browser and OS: Chrome on macOS

## Additional context

This is because [textElement.tsx](https://github.com/mattermost/focalboard/blob/permissions-feature-branch/webapp/src/components/content/textElement.tsx#L29) does not pass the block.id to `changeBlockTitle`.

PR #2359, opened 2022-02-19
Fix Image and text Elements

sbishel:
#### Summary
Image Fix -
octoClient image calls have changed, update the API to match the octoClient calls for images.

Text Fix -
Pass BoardID to mutator for BlockTitleChange

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/2356
  
  And 
 
  # Sev 1: Adding an image doesn’t work


wiggin77 (approved):
LGTM

PR #2361, opened 2022-02-21
Removing transactions from sqlite backend

### LINKED ISSUE #2300: Bug: sqlite database frequently locked

While trying to use the Personal hosted Server version, I frequently get an error about the database being locked. This happens "very" frequently, like every 5 interactions with the GUI. It doesn't happen at all using the same Desktop version (0.12.1).

Log extract:
```
caller="api/api.go:1796" code=401 error="unable to get the session for the token: database is locked" msg= api=/api/v1/workspaces/0/blocks
```
A slightly larger extract here: [focal.log](https://github.com/mattermost/focalboard/files/8028435/focal.log)

Severity wise feels like Sev/1 since the GUI locks and logs you out with an error.

Screenshot from 2022-02-08 19-34-55

## Edition and Platform

 - Edition: Personal Server (Docker)
 - Version: 0.12.1 (Also had the same happen in Mar 2022 docker latest)
 - Browser and OS: Firefox 95.0.2, Server OS: Qnap

jespino:
We expect to prevent the locking problem detected by some users in the #2300
issue by removing the transactions while using the sqlite database.

Fixes #2300

### DIFF
--- a/server/services/store/generators/main.go
+++ b/server/services/store/generators/main.go
@@ -82,6 +82,7 @@ type storeMetadata struct {
 var blacklistedStoreMethodNames = map[string]bool{
 	"Shutdown":      true,
 	"IsErrNotFound": true,
+	"DBType":        true,
 }
 
 func extractMethodMetadata(method *ast.Field, src []byte) methodData {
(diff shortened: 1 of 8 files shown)


wiggin77 (approved):
LGTM 👍

sbishel (approved):
LGTM

Issue #2369, opened 2022-02-22
Permissions Branch - Duplicated custom templates


sbishel:
## Steps to reproduce the behavior

1. Log in with a user with access to multiple custom templates (e.g. sysadmin/Sys@dmin)
2. Go to the template selector
3. The custom templates are duplicated in the list

## Expected behavior
Templates only listed once.

## Screenshots

![Screen Shot 2022-02-22 at 11 30 55 AM](https://user-images.githubusercontent.com/12704875/155196150-5291027f-4d25-4e11-b656-ec51e1277f52.png)

## Edition and Platform

- permissions-feature -branch

PR #2373, opened 2022-02-22
GH-2369 - Fix duplicate templates

sbishel:
#### Summary
Custom Templates are duplicated because retrieving "Global Templates" actually retrieves the current team templates. So the team templates are duplicated.

This creates a specific call in OctoClient for retrieving Global Templates, defined here as teamID = 0.

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/2369

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -73,7 +73,6 @@ func (a *API) RegisterRoutes(r *mux.Router) {
 
 	// Board APIs
 	apiv1.HandleFunc("/teams/{teamID}/boards", a.sessionRequired(a.handleGetBoards)).Methods("GET")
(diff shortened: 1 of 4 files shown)


sbishel:
The other way to fix this would be to update the 'teamPath' function. This function returns the `lastTeamID` if `this.teamID === 0'. However, that function is called all over the place and I like the explicitness of "getGlobalTemplates".

wiggin77:
> The other way to fix this would be to update the 'teamPath' function. This function returns the `lastTeamID` if `this.teamID === 0'. However, that function is called all over the place and I like the explicitness of "getGlobalTemplates".

I agree.  Putting them in team 0 makes sense to me.

wiggin77 (approved):

sbishel:
Closed because its is fixed as part of larger issue. 

https://github.com/mattermost/focalboard/pull/2397

PR #2383, opened 2022-02-23
Added delete feature in Table view

Rajat-Dabade:
#### Summary

> The pull request is about adding the new feature, the ability to delete the card in the Table view.
#### Ticket Link

Fixes for https://github.com/mattermost/focalboard/issues/113


wiggin77:
Thanks @Rajat-Dabade.  This looks great.  Can you address the build failure?

Rajat-Dabade:
Thanks for the review @wiggin77, Sure I will address the build failure.

Rajat-Dabade:
Hello @wiggin77, I have fixed the build failure, please let me know if there are any changes that need to be made. Thank you.

Rajat-Dabade:
@wiggin77, unfortunately, the build failed again, I know the reason, will be working on it now. Thank you.

Rajat-Dabade:
Hello @wiggin77, I have fixed the build failure issue, tested it locally. And also added a test case for the new implementation. Please review it. Thank you.

wiggin77 (approved):
This looks good to me.  Great work @Rajat-Dabade.

webapp/src/components/table/table.scss
```diff
@@ -1,7 +1,7 @@
 .Table {
     margin-top: 16px;
     padding-bottom: 0 !important;
-    overflow: auto;
+    overflow: visible;
```
jespino, line 4: I have some concerns about this, I have to test it manually.

webapp/src/components/table/tableRow.tsx
```diff
@@ -80,6 +92,35 @@ const TableRow = (props: Props) => {
         columnRefs.set(Constants.titleColumnId, React.createRef())
     }
 
+    const handleDeleteCard = async () => {
```
jespino, line 95: I would like to memoize this callback with `useCallback` or at least justify why not in this cases.

webapp/src/components/table/tableRow.tsx
```diff
@@ -80,6 +92,35 @@ const TableRow = (props: Props) => {
...
+        if (!card) {
+            Utils.assertFailure()
+            return
+        }
+        TelemetryClient.trackEvent(TelemetryCategory, TelemetryActions.DeleteCard, {board: board.id, card: card.id})
+        await mutator.deleteBlock(card, 'delete card')
+    }
+
+    const confirmDialogProps: ConfirmationDialogBoxProps = {
```
jespino, line 104: same here

webapp/src/components/table/tableRow.tsx
```diff
@@ -80,6 +92,35 @@ const TableRow = (props: Props) => {
...
+        heading: intl.formatMessage({id: 'CardDialog.delete-confirmation-dialog-heading', defaultMessage: 'Confirm card delete!'}),
+        confirmButtonText: intl.formatMessage({id: 'CardDialog.delete-confirmation-dialog-button-text', defaultMessage: 'Delete'}),
+        onConfirm: handleDeleteCard,
+        onClose: () => {
+            setShowConfirmationDialogBox(false)
+        },
+    }
+
+    const handleDeleteButtonOnClick = () => {
```
jespino, line 113: Same here with `useCallback`

jespino (changes requested):
Some improvements can be done, and I still have to check the `overflow: visible` behavior on a test server.

webapp/src/components/table/table.scss
jespino: As I expected, this generate problems with the scrolling when the window is narrow or when the number of columns (or size of the columns) is high.

I would try to avoid that and find another way to show that without messing with the overflow there.

wiggin77:
@Rajat-Dabade will you be able to address jespino's requests?

Rajat-Dabade:
@jespino @wiggin77 Sure, I will look into it.  Thank you for the review.

Rajat-Dabade:
Hello @jespino, Thank you for the review, Yes, we have to use useCallback and useMemo, I just found that whenever a card is deleted, all the cards in the TableRows.tsx are reevaluated, so it will be better if we use UseCallback and useMemo for all the delete function and variable so that the function and variable will not be generated every time the components reevaluates,  I have also crossed check with Kanban view as well, I think there also we need to implement the same thing. I will make these changes in TableRow.tsx . Shall I make the changes in KanbanCard.tsx as well? Please let me know. Thank you.

webapp/src/components/table/table.scss
Rajat-Dabade: @jespino, Thank you for the review and I understand why this overflow property should be auto, the only solution I can think about it right now is to increase the left padding of the table so that we can see the hidden column of the table on hover with keeping the table scrollable as well. Please let me know your thought. Thank you.

https://user-images.githubusercontent.com/26198062/156360741-d72f9a1d-c185-4c4e-a4e1-28bb08b41ae9.mp4

Rajat-Dabade:
@sbishel, Sure I will make the changes and update the PR.

Thank you

webapp/src/components/table/table.scss
Rajat-Dabade: Thank you for the review @jespino, I understand this didn't look nice, but this is the only solution possible I can think of with the current architecture of the code we are having because as Table have the overflow property as auto, it is not allowing the overflow div to be displayed on hover outside the table. 

If there is any other solution for it, please suggest, I can work on it. Thank you.

asaadmahmood: @Rajat-Dabade I'll have a look at the code and may push a UI fix for this if I figure it out.

Rajat-Dabade: @jespino @asaadmahmood, I have tried to solve the UI issue, and as far as I'm concerned it looks better than the previous one and moreover similar to the proposed UI design. I have pushed the changes, Can you please review them? Please let me know if there are any changes to be made. Thank you.

https://user-images.githubusercontent.com/26198062/157384305-cb6ad0b8-d382-4b14-b305-a53b91649f74.mp4

webapp/src/components/table/tableRow.tsx
Rajat-Dabade: Done

webapp/src/components/table/tableRow.tsx
Rajat-Dabade: Done

sbishel:
@asaadmahmood Could you take a look at this from the UX standpoint.

asaadmahmood:
@Rajat-Dabade Looks good to me.

asaadmahmood (approved):

Rajat-Dabade:
@jespino, please review the changes, so that I can merge. Thank you

jespino (approved):
LGTM

Issue #2368, opened 2022-02-22
Permissions Branch - Template Preview UI is squashed (No Templates)


sbishel:
## Steps to reproduce the behavior

1. Open Boards
2. Click on + (add board)
3. Preview is unreadable

## Expected behavior
Templates should be visible.

## Screenshots
![Screen Shot 2022-02-22 at 11 24 32 AM](https://user-images.githubusercontent.com/12704875/155195210-f128a275-90d7-4418-879e-437d20610af2.png)

## Edition and Platform

permission-feature-branch
 - Edition: Personal Desktop / Personal Server / Mattermost plugin
 - Version: [e.g. v0.9.0]
 - Browser and OS: [e.g. Chrome 91 on macOS, Edge 93 on Windows]

## Additional context
Probably due to there are NO Templates loaded

PR #2397, opened 2022-02-25
Update migrations to create global templates.

sbishel:
#### Summary
This PR update the migrations, so that the built in templates are created in the global (TeamID = '0') team. Combines work for board and board_history migration.

It deletes all built-in templates from the blocks table.

Removes code that is no longer necessary.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2368

  Fixes https://github.com/mattermost/focalboard/issues/2369


jespino (approved):
LGTM.

PR #2414, opened 2022-02-28
Fix migration sql error

wiggin77:
#### Summary
Fixes SQL syntax error for migrations (postgres)

#### Ticket Link
NONE


sbishel (approved):
Thanks Doug

Issue #2379, opened 2022-02-23
Permissions branch: No error shown when attempting to remove last admin


chenilim:
## Steps to reproduce the behavior

1. Create a board
2. Select "Show Share Dialog"
3. Select your user's role, and select "Remove member"
4. Nothing happens, and error in the console

## Expected behavior

A dialog showing "Boards must have at least one Administrator" appears ([spec](https://docs.google.com/document/d/1HOnoB87WZojqplpB6XGbtKsLuqjG8h-8CbJpmWsRfM4/edit#bookmark=id.16ze8juuy9wl)).

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: permissions-feature-branch
 - Browser and OS: Chrome on Mac

Issue #2378, opened 2022-02-22
Permissions branch: Cannot change admin to editor


chenilim:
## Steps to reproduce the behavior

1. Open a board
2. Select Show Share Dialog
3. Search for another user (e.g. sysadmin), and add them
4. Select their role (Editor by default) and change it to Admin
5. Select their role again, and try to change it back to Editor
6. Nothing happens

## Expected behavior

Able to change a user from Admin to Editor

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: permissions-feature-branch
 - Browser and OS: Chrome on Mac

PR #2416, opened 2022-02-28
don't allow last admin change/deleted

sbishel:
#### Summary
Checks when changing/deleting user. If lastAdmin, don't allow deletion.
Uses FlashMessage as feedback mechanism to user. 

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2379

  Fixes https://github.com/mattermost/focalboard/issues/2378


wiggin77 (approved):
LGTM :+1:

PR #2419, opened 2022-03-01
Fixed a bug where team ID was not set on boards cloned from global templates

### LINKED ISSUE #2418: Bug: Permissions Branch: Cannot use global templates

1. Go to Focalboard
2. Try using a default template
3. Notice that the board created from the template doesn't show up in UI

## Expected behavior

Using a template should create a board in UI

## Screenshots

NA

## Edition and Platform

Only on the permissions branch

harshilsharma63:
#### Summary
Fixed a bug where team ID was not set on boards cloned from global templates.

#### Ticket Link
Fixes #2418


jespino (approved):
LGTM

PR #2425, opened 2022-03-01
Permissions branch: Fix unit tests and CI errors (part 1)

wiggin77:
#### Summary
Fixes some unit tests, and CI errors, and linter errors.   Not quite passing all tests but closer.  Second PR to follow with more fixes.

**Note:** this removes some code that was added with permissions branch to use sqlite files instead of memory for unit and integration tests.  This was creating many temp files that were not cleaned up and causing the test suite to run much longer.  I have reverted back to in-memory sqlite when testing.  None of the databases created were more than a few hundred KB.

#### Ticket Link
NONE


Makefile
```diff
@@ -122,6 +122,7 @@ server-test-mysql: export FB_STORE_TEST_DOCKER_PORT=44445
 
 server-test-mysql: ## Run server tests using mysql
 	@echo Starting docker container for mysql
+	docker-compose -f ./docker-testing/docker-compose-mysql.yml down -v --remove-orphans
```
wiggin77, line 125: This change is to clean up after a previous test run crashed and left the container behind.

sbishel (approved):

jespino (approved):
LGTM

PR #2434, opened 2022-03-02
Adding join logic to the board switcher

### LINKED ISSUE #2380: Permissions branch: Unable to join / view board that is team accessible

Set up two user accounts.

1. As user 1: Create a new board
2. Open the Share dialog and grant "Everyone at ___ Team" Editor access
3. Now log in as user 2
4. Select the "Find Boards" input in the sidebar
5. Type in the name of the board, and select it
6. Nothing happens

## Expected behavior

The board is added to the sidebar and opens

## Edition and Platform

 - Edition: Mattermost plugin

jespino:
Fixes #2380


wiggin77:
If I'm reading this correctly a user is added as a board member every time they select a board to view, regardless of whether they already were.  Wouldn't it create less pressure on the master db to check if a member first?  Also I don't understand the point of storing board membership if all it means if whether someone happened to view a board yet. Why not assume everyone who can be a member already is?

jespino:
@wiggin77 it is requesting join to the channel, but if the membership already exists, it doesn't create a new one. So... there is not a big deal (except for the audit part).

About belonging to the channel or not is about compliance, for example, if you belong to a board, I understand that you has access to the board from Date X to Date Y, so... that give me certain information to decide what to export in the compliance exports. Otherwise, I would need a more "granular control" over what is visible for each user and what is not, or accept that everybody in the team has access to everything in any public board, what degrades a lot the utility of the compliance exports.

Also the memberships itself give us extra info, like "Should I show the board in the sidebar?", also allow us to store there more information in the future, for example the "lastViewAt", allowing in the future to highlight the stuff that has changed somehow, or things like that.

wiggin77:
> Now I'm thinking I should do it anyway directly in the BoardPage, if the board is there and It fails to load, I should join automatically.

Do you want to add that to this PR or is that for future?

jespino:
I think I want to add that here, but the code in boardsPage.tsx is too complex, maybe I'll start simpliying it a bit.

jespino:
Added the autojoin support (now is just simpler because everything was cleaner in the BoardsPage).

wiggin77 (approved):
LGTM :+1:

PR #2435, opened 2022-03-02
Splitting BoardPage component into simpler/smaller components

jespino:


wiggin77 (approved):
Nice!  :+1:

webapp/src/pages/boardPage/backwardCompatibilityQueryParamsRedirect.tsx
```diff
@@ -0,0 +1,75 @@
...
+    //         UserSettings.setLastViewId(boardId, viewId)
+    //     }
+    //
+    //     dispatch(setCurrentBoard(boardId || ''))
+    //     dispatch(setCurrentView(viewId || ''))
+    // }, [match.params.boardId, match.params.viewId, boardViews])
+    return null
+}
+
+export default BackwardCompatibilityQueryParamsRedirect
```
sbishel, line 75: Since this is commented out, can this be deleted?

sbishel (approved):
Just wondering if file is necessary, but otherwise looks good.

webapp/src/pages/boardPage/backwardCompatibilityQueryParamsRedirect.tsx
jespino: The thing is, I don't know if we want to maintain the backward compatibility from that part, that would be easy (for older links) @chenilim do you have any preference? Preserve backward compatiblity with links that use the ?cardId=whatever, and things like that?

PR #2477, opened 2022-03-05
[GH-2396] Fix scroll jump for kanban board

### LINKED ISSUE #2396: Bug: Scrolled down board jumps to the very top whenever any card is opened

1. Create a board view
2. Create a column
3. Add enough cards to be able to scroll down
4. Scroll to the bottom of the column 
5. Open any card
6. Close that card
7. You are now scrolled to the top of the column, completely losing the context of where you were

## Expected behavior

Opening and closing a card should not move your position inside on the board. You should stay at the same place. 

## Additional context

This is terrible UX, it renders the board almost unusable with a larger number of cards in columns. The user needs to scroll down all the time just to get where they were.

## Edition and Platform

 - Edition: Personal Desktop, Mattermost plugin
 - Version: 0.14.0 (30)
 - Browser and OS: Chrome on Mac, Safari on Mac, MacOS app, MacOS Mattermost app

I cannot set labels but I firmly believe this is Sev 1.

kamre:
#### Summary
Fixed bug with scroll jumping when the card is opened:
  - avoid remounting of `ScrollingComponent` for each render of `Kanban` component
  - property `autoFocus` set to false for `CalculationOptions` because it triggers `blur` event for the button in Jest tests and closes the menu
  - snapshots for tests with `CalculationOptions` updated

#### Ticket Link
Fixes #2396

### DIFF
--- a/webapp/src/components/calculations/options.tsx
+++ b/webapp/src/components/calculations/options.tsx
@@ -191,7 +191,7 @@ const CalculationOptions = (props: BaseCalculationOptionProps): JSX.Element => {
             isSearchable={false}
             components={{DropdownIndicator, ...(props.components || {})}}
             defaultMenuIsOpen={props.menuOpen}
-            autoFocus={true}
+            autoFocus={false}
             formatOptionLabel={(option: Option, meta) => {
                 return meta.context === 'menu' ? optionLabelString(option, intl) : optionDisplayNameString(option, intl)
             }}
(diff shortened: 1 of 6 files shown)


jespino (approved):
LGTM

webapp/src/components/calculations/options.tsx
```diff
@@ -191,7 +191,7 @@ const CalculationOptions = (props: BaseCalculationOptionProps): JSX.Element => {
             isSearchable={false}
             components={{DropdownIndicator, ...(props.components || {})}}
             defaultMenuIsOpen={props.menuOpen}
-            autoFocus={true}
+            autoFocus={false}
```
jespino, line 194: I'm not 100% sure if this is affecting other expected behavior, but otherwise looks good to me. @ogi-m can you confirm that the calculation options works as expected around the focus handling?

harshilsharma63 (approved):
LGTM 👍

harshilsharma63:
Merged into main and cherrypicked into permissions branch. cc @jespino

PR #2494, opened 2022-03-08
Adding back views to the sidebar

jespino:


webapp/src/components/sidebar/__snapshots__/sidebarBoardItem.test.tsx.snap
```diff
@@ -1,140 +1,63 @@
...
-        <button
-          class="IconButton"
-          type="button"
-        >
-          <i
-            class="CompassIcon icon-dots-horizontal OptionsIcon"
-          />
-        </button>
-      </div>
+      i
```
wiggin77, line 11: Is this meant to be a tag?

wiggin77: I guess it's always been there.  Just looked strange to have a lone `i` outside a tag.

jespino: No it wasn't, it looks like something that should be part of an "vim" error, probably between the keyboard and the chair, I'll take a look.

wiggin77 (approved):
LGTM :+1:

webapp/src/components/sidebar/__snapshots__/sidebarBoardItem.test.tsx.snap
jespino: Ok, this is not an accident, this is because we use unicode emojis, so, this is just a character, and the tests uses the letter `i` for this icon. so... it is working as expected.

Issue #2457, opened 2022-03-03
Bug: Read-only view screen when access revoked


sbishel:
## Steps to reproduce the behavior

1. Go to Share 
2. Publish External Link
3. Using Incognito, go to external link.
4. Should open find.
5. Disable External Link 
- either turn off sharing
- or regenerate token
6. Refresh incognito window

## Expected behavior
Unsure...but I think just a blank page, with top banner.

You cannot create a board from here, but shouldn't display.

## Screenshots

![Screen Shot 2022-03-03 at 11 13 32 AM](https://user-images.githubusercontent.com/12704875/156627455-de15085d-409b-4c10-985f-0a9f91dd7897.png)

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: 0.15-RC1
 - Browser and OS: Chrome on Mac

wuwinson:
Should send user to error page with message: "Board may be deleted or access revoked"

PR #2499, opened 2022-03-08
GH-2257: Don't display template page if readonly and access revoked

sbishel:
#### Summary
Don't display the workspace if there is no board and it is a read-only view.

The original desired fix was to redirect to the error page. Unfortunately, that wasn't going to work without redesigning the page, as the first time through the board is not yet defined and a redirect causes no boards to display.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2457

Screen Shot 2022-03-08 at 2 53 42 PM


amyblais:
@harshilsharma63 @ogi-m A kind note to prioritize reviews on this.

ogi-m (approved):
LGTM - the page is now blank with the error banner on top

harshilsharma63 (approved):
LGTM!

Issue #2507, opened 2022-03-09
Permissions branch: make swagger is broken


chenilim:
## Steps to reproduce the behavior

1. Check out the `permissions-feature-branch`
2. `make swagger`
3. Errors

## Expected behavior

`make swagger` works

## Edition and Platform

 - Edition: All
 - Version: permissions-feature-branch

## Additional context

The Swagger comments need to updated in sync with API changes.

chenilim:
Fixed by the PR linked above.

PR #2508, opened 2022-03-09
Gh 2507 fix make swagger

chenilim:
#### Summary
Fixed broken comments and added missing ones. `make swagger` works again now. A few notes and follow-ups:
* The Swagger documentation is currently incomplete - e.g. output schema of Onboard is missing (as one example)
* The Swagger docs are likely out of date  on main
* I did not check in the large index.html change with this, as we will likely need to update again, and not sure if there's a better alternative to checking in the generated docs.
* What's a better process to keep the API docs in sync?

#### Ticket Link
#2507


jespino (approved):
LGTM

harshilsharma63 (approved):

PR #2513, opened 2022-03-10
Generate templates.boardarchive at build time

chenilim:
#### Summary
Updated the Makefile to generate `templates.boardarchive` dynamically, from source files that are checked in. This should make future editions easier to track and review.

I tested the GitHub dev-release and prod-release workflows on my own fork, and they work.


jespino:
I think we can address this in very different ways. One other option is to embed the content of the file, and do something like "zip it on the fly whenever we want to upload the archive locally", tha twould be easier, doesn't need to commit the archive + the unarchived version, and doesn't need any compilation process. What do you think @wiggin77 ?

wiggin77:
@jespino @chenilim let's talk about this at FB dev meeting.

chenilim:
@jespino, this change removes the zipped archive and replaces it with the unarchived contents in the repo (and zips them at make time). @wiggin77, missed your comment before today's dev meeting, but what do you think of this approach? Was there another alternative already in flight?

(I was investigating the console errors issue that I thought was related to the template archive data, but it turned out not to, and I thought I would submit this as a proposal). We expect to expand the templates soon, and it would be great to have the changes be easier to review in the repo.

wiggin77:
> change removes the zipped archive and replaces it with the unarchived contents in the repo

I'm not understanding the rationale for removing the ZIP, which was exported from the source, and exploding that into source control only to zip it again.  It's not like the templates will be edited in github;  they will be edited in the source board and exported again.

chenilim:
The issue is how can we review future changes to `templates.boardarchive` if it's a zip file? I suppose you could download, unzip both versions, and run FileMerge, but it would be better to see the changes in git / GitHub directly.

The other thing is checking in the unzipped versions make it easier to make minor edits to them without using export / import - e.g. changing titles, or fixing typos.

Do you see any potential issues with generating the zip file at make time? One concern is we would have to update this if the archive format changes.

wiggin77:
> how can we review future changes to templates.boardarchive if it's a zip file

That's a fair point if changes to the templates are to have a review like docs and code.

> make it easier to make minor edits to them

You have to make the same edits in the source board anyway, so exporting the archive again is not a big step.

> see any potential issues with generating the zip file at make time

The exporter makes guarantees about the order of files in the zip so it can be streamed (not read the whole thing into memory).  When zipping from the filesystem we cannot guarantee that order and the import will fail if images are loaded before the JSON.    It's probably not a big deal to read the default templates archive into memory, but that uses the same import code as the other imports and we don't want to be reading potentially gigabytes at once.

chenilim:
Thanks @wiggin77. I'll queue an item to discuss at the next dev meeting. I didn't know about the stream order being potentially important. We expect to be adding more and more built-in templates over the next few months, so worth thinking about what we want that process to be like.

chenilim:
Followup from dev discussion: @wiggin77 will investigate further. Current plan is to build and run a small archive-creator at build-time. This will generate the manifest and create the archive file in the correct order. I've also created https://github.com/mattermost/focalboard/issues/2558 to track this issue.

wiggin77:
This PR is replaced by https://github.com/mattermost/focalboard/pull/2577

PR #2528, opened 2022-03-11
Fix menu tests

sbishel:
#### Summary
Fixes unit tests caused by updates to submenus.

https://github.com/mattermost/focalboard/commit/5b309e8e25e7455e2b328f25f0a3f62638fe4fc2


jespino (approved):
Just one doubt, are the menus showing on over directly?

sbishel:
> Just one doubt, are the menus showing on over directly?

Yes, we are. We were before as well. Now however, a click on the parent closes the submenu...I'll investigate why.

Issue #2500, opened 2022-03-08
Permissions branch:  Async notifyBlockChanged causes race


wiggin77:
## Steps to reproduce the behavior

1. Run unit tests for mysql or postgres in Permissions branch
2. Tests fail with "Database closed" errors

The async notifications when boards or blocks change (e.g. `app.CreateBoardsAndBlocks`) are done 

## Expected behavior

Don't run store commands after database is closed

## Edition and Platform

 - Edition: plugin
 - Version: permissions branch
 - Browser and OS: FF 93

## Additional context

Also set the Severity using the *Sev/1*, *Sev/2* or *Sev/3* labels as best you can:
* Sev/1: Issue affects critical functionality without a workaround
* Sev/2: Issue affects major functionality with a difficult or non-obvious workaround
* Sev/3: Issue affects minor, non-critical functionality

wiggin77:
fixed via https://github.com/mattermost/focalboard/pull/2530

PR #2530, opened 2022-03-11
Permissions branch:  No sqlstore calls after app shutdown

wiggin77:
#### Summary
This PR fixes 2 issues:

- block change events for websockets, adapter, notifications are done asynchronously and can happen after the store is shutdown;  fixed by ensuring all events are processed before app shutdown
- msql unit tests were failing because `insertAt` column only has 1 second resolution

**Note:** There is now a new issue with sqlite as of when the new migration tool was added. A table lock error error happens.  Mysql and Postgres are green.  This will be addressed in a new PR.

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/2500


jespino (approved):
LGTM

PR #2550, opened 2022-03-16
Shareboard cleanup

sbishel:
#### Summary
- Removes unused property from centerPanel, workspace.
- Fixes issue where Users list is shown for non-admin.
- Some CSS cleanup.


wiggin77 (approved):
LGTM :+1:

PR #2559, opened 2022-03-17
Adding initial data generator proof of concept

jespino:


wiggin77:
The end result looks great.  I don't understand why all the archive code needs to be there.  Why can't we just call  `generateBoards` from a test helper and not involve the archive stuff?  It could also be called via rest or cli, but just populates the database with predictable data.

jespino:
@wiggin77 I agree that this is not necessarily something that we need to use the archive code there, but I used it because was part of the conversation that we have during the developers meeting. But of course, If we see it as a problem instead of as a solution, we can just use directly the store (I think that would work even better).

Also I can keep this code as part of the testing framework only, but I think can be also useful for user testing too, not only for unit or integration testing, so I would try to find a way to apply this somehow to the current instance.

wiggin77:
@jespino is this PR still in play?  Note was a have a basic data generator here:  https://github.com/wiggin77/fbsc

jespino:
@wiggin77 I think this is no longer needed having your approach.

PR #2564, opened 2022-03-18
Fixing table header ordering in table view

jespino:


harshilsharma63:
@jespino can you please tell what does this fix?

jespino:
@harshilsharma63 it fix the order of the table view headers

wiggin77 (approved):
LGTM

PR #2568, opened 2022-03-18
Permissions branch: Fix sqlite table lock (CI)

wiggin77:
#### Summary
This PR fixes the sqlite table lock affecting integration tests.  This was originally a much bigger PR that tried to work around the in-memory with shared cache locking issue.  It would have been nice to keep using the in-memory database for faster tests, but it was getting too complex.

Also includes a few linter fixes.

#### Ticket Link
NONE


chenilim:
I tested this locally and it worked for me! Interesting that the in-memory DB can cause locks. Not urgent, but could we consider deleting the test DB on teardown?

wiggin77:
> Interesting that the in-memory DB can cause locks

It's not the in-memory aspect that causes the locks, but the shared cache mode needed by the new migrations tool.

wiggin77:
> but could we consider deleting the test DB on teardown?

Done

jespino (approved):
LGTM

PR #2569, opened 2022-03-18
GetBoardMetadata API

chenilim:
#### Summary
This is the base implementation of GetBoardMetadata, per the [spec here](https://docs.google.com/document/d/1Cyu5T9Xs3kRCUnitUYCz8Lp8sEh8WIg2ngUpKtVrm9s/edit#).

Adding:
* A new API and App function
* Check for license key in API
* Store functions to return block history descendants (related to a board), and board history
* Unit tests

Also fixes a couple issues:
* boards_history table was not storing the latest update_at time
* NULL table values were causing exceptions in retrieving history

#### Ticket link
#2600


server/services/store/sqlstore/blocks.go
```diff
@@ -634,6 +634,39 @@ func (s *SQLStore) getBlockHistory(db sq.BaseRunner, blockID string, opts model.
 	return s.blocksFromRows(rows)
 }
 
+func (s *SQLStore) getBlockHistoryDescendants(db sq.BaseRunner, boardID string, opts model.QueryBlockHistoryOptions) ([]model.Block, error) {
```
chenilim, line 637: @wiggin77, I reused the QueryBlockHistoryOptions struct here, because the intent is the same. Not sure if it was better to create a copy now, or do that later should they diverge.

wiggin77: QueryBlockHistoryOptions makes sense here.

server/api/api.go
```diff
@@ -2966,6 +2967,85 @@ func (a *API) handleDuplicateBlock(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		a.errorResponse(w, r.URL.Path, http.StatusInternalServerError, "", err)
+		return
+	}
+	if board == nil || boardMetadata == nil {
+		a.errorResponse(w, r.URL.Path, http.StatusNotFound, "", nil)
+		return
+	}
+
+	if !hasValidReadToken {
```
harshilsharma63, line 3015: This check should be right after we check for a valid token at line 2999 so that we don't even attempt to fetch metadata if the user doesn't have permission to do so.

server/api/api.go
```diff
@@ -2966,6 +2967,85 @@ func (a *API) handleDuplicateBlock(w http.ResponseWriter, r *http.Request) {
...
+				return
+			}
+		}
+	}
+
+	auditRec := a.makeAuditRecord(r, "getBoardMetadata", audit.Fail)
+	defer a.audit.LogRecord(audit.LevelRead, auditRec)
+	auditRec.AddMeta("boardID", boardID)
+
+	a.logger.Debug("GetBoardMetadata",
```
harshilsharma63, line 3033: Looks like a debugging log. Should this be removed?

server/api/api.go
```diff
@@ -2966,6 +2967,85 @@ func (a *API) handleDuplicateBlock(w http.ResponseWriter, r *http.Request) {
 	auditRec.Success()
 }
 
+func (a *API) handleGetBoardMetadata(w http.ResponseWriter, r *http.Request) {
```
harshilsharma63, line 2970: This API is missing the check for Mattermost enterprise edition.

server/api/api.go
```diff
@@ -2966,6 +2967,85 @@ func (a *API) handleDuplicateBlock(w http.ResponseWriter, r *http.Request) {
...
+	boardID := mux.Vars(r)["boardID"]
+	userID := getUserID(r)
+
+	hasValidReadToken := a.hasValidReadTokenForBoard(r, boardID)
+	if userID == "" && !hasValidReadToken {
+		a.errorResponse(w, r.URL.Path, http.StatusUnauthorized, "", PermissionError{"access denied to board"})
+		return
+	}
+
+	board, boardMetadata, err := a.app.GetBoardMetadata(boardID)
```
harshilsharma63, line 3005: This doesn't work for deleted boards and I'm not sure whether this should or should not work for deleted boards. My thinking is that deleted data should still be part of compliance. Otherwise, one can create and delete data between this API call and never come on the radar.

We can compare this with Mattermost compliance as an option.

harshilsharma63 (changes requested):

server/api/api.go
chenilim: Thanks! Removed the debug log.

server/api/api.go
chenilim: I've implemented the license check - by exposing GetLicence on the Store, and overriding this in mattermostauthlayer. Also added integration tests to cover this.

server/api/api.go
```diff
@@ -89,6 +89,7 @@ func (a *API) RegisterRoutes(r *mux.Router) {
 	apiv1.HandleFunc("/boards/{boardID}/blocks/{blockID}/undelete", a.sessionRequired(a.handleUndeleteBlock)).Methods("POST")
 	apiv1.HandleFunc("/boards/{boardID}/blocks/{blockID}/subtree", a.attachSession(a.handleGetSubTree, false)).Methods("GET")
 	apiv1.HandleFunc("/boards/{boardID}/blocks/{blockID}/duplicate", a.attachSession(a.handleDuplicateBlock, false)).Methods("POST")
+	apiv1.HandleFunc("/boards/{boardID}/metadata", a.attachSession(a.handleGetBoardMetadata, false)).Methods("GET")
```
sbishel, line 92: Should this be `a.sessionRequired`? I think `attachSession` is used with readtoken.

server/api/api.go
```diff
@@ -2966,6 +2967,89 @@ func (a *API) handleDuplicateBlock(w http.ResponseWriter, r *http.Request) {
...
+	//     description: required license not found
+	//   default:
+	//     description: internal error
+	//     schema:
+	//       "$ref": "#/definitions/ErrorResponse"
+
+	boardID := mux.Vars(r)["boardID"]
+	userID := getUserID(r)
+
+	hasValidReadToken := a.hasValidReadTokenForBoard(r, boardID)
```
sbishel, line 3003: This may be a security issue. I don't think we want to accept a read token on this api. I think this would allow a valid read token for a read-only board be used for this api.

server/api/api.go
sbishel: This variable `hasValidReadToken` should be removed see above.

sbishel (changes requested):
Other than supporting read-token, looks good.

server/services/store/sqlstore/blocks.go
```diff
@@ -588,6 +589,7 @@ func (s *SQLStore) getBlock(db sq.BaseRunner, blockID string) (*model.Block, err
 		s.logger.Error(`GetBlock ERROR`, mlog.Err(err))
 		return nil, err
 	}
+	defer s.CloseRows(rows)
```
chenilim, line 592: Added this missing defer CloseRows.

chenilim:
@sbishel and @harshilsharma63 - I've updated the PR per your comments. Please LMK if you see any other issues. Also (finally) got CI to pass. Thanks.

chenilim:
Gentle reminder to please re-review. I've merged from main, and this should be good to go. The CI failure is the Cypress issue on main. Thanks.

harshilsharma63 (approved):
LGTM!

server/api/api.go
chenilim: Fixed.

server/api/api.go
chenilim: Fixed.

sbishel (approved):

PR #2571, opened 2022-03-21
Fixing delete/undelete boards

### LINKED ISSUE #2566: Permissions Branch: Undo delete board doesn't work

1. Create a board
2. Delete it from the sidebar menu
3. Press Cmd+Z to undo
4. Error page is shown ("Board not found")

## Expected behavior

Board is un-deleted and shown

## Edition and Platform

jespino:
Fixes #2566


server/api/api.go
```diff
@@ -1062,6 +1063,53 @@ func (a *API) handleUndeleteBlock(w http.ResponseWriter, r *http.Request) {
...
+	userID := session.UserID
+
+	vars := mux.Vars(r)
+	boardID := vars["boardID"]
+
+	auditRec := a.makeAuditRecord(r, "undeleteBoard", audit.Fail)
+	defer a.audit.LogRecord(audit.LevelModify, auditRec)
+	auditRec.AddMeta("boardID", boardID)
+
+	err := a.app.UndeleteBoard(boardID, userID)
```
wiggin77, line 1101: Can anyone with a valid session undelete a board if they know the ID?

jespino: Good question, we should have the permission to create the board of the right type.

jespino: Added permissions checks for `undeleteBoard` and for `undeleteBlock` and some tests.

jespino:
@wuwinson @chenilim I'm using the `delete_board` permission for undelete boards and the `manage_board_cards` for the undelete block, I hope that is correct, I'm almost 100% sure about the later, but I have some doubts about the former.

webapp/src/octoClient.ts
```diff
@@ -335,9 +335,17 @@ class OctoClient {
...
+    async undeleteBlock(boardId: string, blockId: string): Promise<Response> {
         Utils.log(`undeleteBlock: ${blockId}`)
-        return fetch(this.getBaseURL() + this.teamPath() + `/blocks/${encodeURIComponent(blockId)}/undelete`, {
+        return fetch(`${this.getBaseURL()}/api/v1/boards/${encodeURIComponent(boardId)}/blocks/${encodeURIComponent(blockId)}/undelete`, {
+            method: 'POST',
+            headers: this.headers(),
+        })
+    }
+
+    async undeleteBoard(boardId: string): Promise<Response> {
```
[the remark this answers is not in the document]
wiggin77: Sorry - that doesn't make sense here.  I guess I'm confused about how someone undeletes a board and all the cards and blocks.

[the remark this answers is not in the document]
jespino: Yes, now we don't have cascade deletion, you delete the board, is enough because the board info is not longer accesible (from the UI perspective), in the future we can decide to cascade that, or to interpret a block in a deleted board as deleted. But... For now, we only need the undelete of the board.

server/api/api.go
```diff
@@ -1103,6 +1104,58 @@ func (a *API) handleUndeleteBlock(w http.ResponseWriter, r *http.Request) {
...
+
+	vars := mux.Vars(r)
+	boardID := vars["boardID"]
+
+	auditRec := a.makeAuditRecord(r, "undeleteBoard", audit.Fail)
+	defer a.audit.LogRecord(audit.LevelModify, auditRec)
+	auditRec.AddMeta("boardID", boardID)
+
+	if !a.permissions.HasPermissionToBoard(userID, boardID, model.PermissionDeleteBoard) {
+		a.errorResponse(w, r.URL.Path, http.StatusForbidden, "", PermissionError{"access denied to create public boards"})
```
wiggin77, line 1143: I don't think this is the right error text for this condition.

server/services/store/sqlstore/board.go
```diff
@@ -691,6 +691,98 @@ func (s *SQLStore) getBoardHistory(db sq.BaseRunner, boardID string, opts model.
...
 
+func (s *SQLStore) undeleteBoard(db sq.BaseRunner, boardID string, modifiedBy string) error {
+	boards, err := s.getBoardHistory(db, boardID, model.QueryBoardHistoryOptions{Limit: 1, Descending: true})
+	if err != nil {
+		return err
+	}
+
+	if len(boards) == 0 {
+		s.logger.Warn("undeleteBlock board not found", mlog.String("board_id", boardID))
+		return nil // deleting non-exiting board is not considered an error (for now)
```
wiggin77, line 702:
```suggestion
		return nil // undeleting non-exiting board is not considered an error (for now)
```

server/api/api.go
jespino: You are totally right 🤦‍♂️

jespino: Done!

wiggin77 (approved):
LGTM :+1:

Issue #2558, opened 2022-03-17
Enhancement: Build template archive at build time


chenilim:
## Summary

Instead of checking-in the `templates.boardarchive` file, check-in its components instead, and generate the zip archive at build time.

## How important this is to me and why

Importance: Medium

Use cases:
1. This makes seeing the diffs much easier
2. We can check-in diffs rather than the whole zip archive (saves space, etc.)

## Additional context/similar features

This will improve the process of adding more templates in the future.

PR #2577, opened 2022-03-22
Generate templates archive at build time

wiggin77:
#### Summary
This PR includes support for generating a default templates archive and embedding that at build time.  The archive is generated based on a checked-in file tree containing an exported set of templates.  

This PR also ensures older archive formats can be imported by filling in the missing fields adding for permissions.

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/2558


sbishel (approved):
LGTM

jespino:
If this is generating and embedding the file in the fly, should we remove the archived from the repo?

I mean the file: `server/assets/templates.boardarchive`

jespino (approved):
LGTM

Issue #2502, opened 2022-03-09
Bug: Permissions branch: Share board not working 


jespino:
## Steps to reproduce the behavior

1. In a personal server.
2. Go to a board.
3. Share it externally.
4. Open the URL in an incognito window.
5. It fails

## Expected behavior

It should work

## Screenshots

![image](https://user-images.githubusercontent.com/290303/157440797-4fa6164a-2be3-44f5-baf3-746aa8a5c154.png)

## Edition and Platform

 - Edition: Personal Server
 - Version: permissions-feature-branch
 - Browser and OS: Firefox

PR #2581, opened 2022-03-22
fixes for shared boards

sbishel:
#### Summary
Updates necessary for Shared Boards to work with permission changes.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2502


webapp/src/pages/boardPage/boardPage.tsx
```diff
@@ -101,23 +98,28 @@ const BoardPage = (props: Props): JSX.Element => {
...
+                dispatch(setCurrentView(match.params.viewId))
+                UserSettings.setLastViewId(match.params.boardId, match.params.viewId)
+            }
+
+            if (!props.readonly && me) {
+                loadOrJoinBoard(me.id, teamId, match.params.boardId)
+            }
         }
-    }, [teamId, match.params.boardId, match.params.viewId])
+    }, [teamId, match.params.boardId, match.params.viewId, me])
```
jespino, line 122: nit: probably a super tiny optimization is to pass use the `me?.id` as dependency, that way your user changes doesn't necesarilly retrigger this effect.

webapp/src/router.tsx
```diff
@@ -159,9 +159,13 @@ const FocalboardRouter = (props: Props): JSX.Element => {
                         <ChangePasswordPage/>
                     </FBRoute>}
 
-                <FBRoute path='/shared/:boardId?/:viewId?/:cardId?'>
+                <FBRoute path='/workspace/:workspaceId/shared/:boardId?/:viewId?/:cardId?'>
```
jespino, line 162: This is here for backward compatibility, I guess, do we need the previous one (without the team or workspace) URL too?

jespino: Also, you can use an array of paths in the `path` attribute, so probably makes sense to pass the array of all the valid urls that are for shared boards in one single route.

jespino: Also, some routes later, we have a redirect from `/workspace/:workspaceId/shared/:boardId?/:viewId?/:cardId?` for this specific case, so we should fix that there, I guess.

jespino (approved):
LGTM, just a couple of suggestions.

PR #2595, opened 2022-03-23
Add Members History support

### LINKED ISSUE #2586: Create and maintain board member history table
## Summary

Create and maintain a `board_member_history` table, analogous to the `channelmemberhistory` table for Channels.

Importance: High

Use cases:
1. This is important for compliance and auditing

jespino:
Fixes  #2586


server/services/store/sqlstore/board.go
```diff
@@ -528,3 +580,30 @@ func (s *SQLStore) searchBoardsForUserAndTeam(db sq.BaseRunner, term, userID, te
 
 	return s.boardsFromRows(rows)
 }
+
+func (s *SQLStore) getBoardMemberHistory(db sq.BaseRunner, boardID, userID string, limit uint64) ([]*model.BoardMemberHistoryEntry, error) {
```
jespino, line 584: This function and the `boardMemberHistoryEntriesFromRows` are just to be able to test it properly as part of the store, also, we are going to need something like that for the compliance export sooner than later, so I'm doing it anyway.

server/services/store/sqlstore/board.go
```diff
@@ -384,6 +406,8 @@ func (s *SQLStore) saveMember(db sq.BaseRunner, bm *model.BoardMember) (*model.B
 		"scheme_viewer":    bm.SchemeViewer,
 	}
 
+	oldMember, _ := s.getMemberForBoard(db, bm.BoardID, bm.UserID)
```
jespino, line 409: This is not amazing, but is the simplest approach that I found to detect if this is an insert or an update.

server/services/store/sqlstore/board.go
```diff
@@ -413,10 +448,27 @@ func (s *SQLStore) deleteMember(db sq.BaseRunner, boardID, userID string) error
...
+	rowsAffected, err := result.RowsAffected()
+	if err != nil {
 		return err
 	}
 
+	if rowsAffected > 0 {
+		addToMembersHistory := s.getQueryBuilder(db).
+			Insert(s.tablePrefix+"board_members_history").
+			Columns("board_id", "user_id", "action", "insert_at").
+			Values(boardID, userID, "delete", model.GetMillis())
```
chenilim, line 465: Let's make this "deleted" to be consistent with "created".

chenilim (changes requested):
I've updated the spec to reflect this implementation. Please review. One requested change: "deleted" instead of "delete".

server/services/store/sqlstore/board.go
jespino: fair point

jespino: Done.

chenilim (approved):
Looks good, thanks!

PR #2599, opened 2022-03-23
Sqlite in-memory database for unit tests

wiggin77:
#### Summary
This PR restores a previous setting that uses a in-memory database for sqlite.  The database locking has been worked around via a "busy timeout" pragma that waits up to 5 seconds for a database lock to resolve.

This pragma was also applied to personal server config to prevent table lock errors there as well.  The idea being that a personal server with contention lasting more than 5 seconds should be using postgres to handle that kind of load.

#### Ticket Link
NONE


jespino (approved):
LGTM

jespino:
@wiggin77 should it be also added to the cypress tests config?

sbishel (approved):
LGTM

PR #2605, opened 2022-03-23
Run make server-test-sqlite to run only against sqlite for that part of the job

jespino:
This is for speed up the tests, I'm not 100% sure if this has other implications. @wiggin77 can you confirm this is correct?.


wiggin77:
> This is for speed up the tests,

Oh wow.  That was indeed the issue.  `make server-test` runs the tests against all databases, so sqlite test was actually sqlite+mysql+postgres.

wiggin77 (approved):
Great spotting that!  :+1:

PR #2608, opened 2022-03-23
Trying to parallize tests

jespino:


wiggin77:
I think it is running push and pull-request workflows.  Is that double the work?

jespino:
@wiggin77 I guess so, but you changed it. I don't know if you want to polish that too.

wiggin77:
> @wiggin77 I guess so, but you changed it. I don't know if you want to polish that too.

I think the matrix stuff is now causing both push and pull-request to run.  It wasn't doing that before.

wiggin77:
Looking at the timings recorded, this does appear to be faster.

wiggin77 (approved):

Issue #2591, opened 2022-03-23
Bug: manageGroups Cypress test intermittently failing


chenilim:
## Steps to reproduce the behavior

1. cd webapp
2. Repeatedly run `npm run cypress:ci`, or run manageGroups.ts manually from Cypress
3. Test fails because create board fails

## Expected behavior

Test passes consistently

## Screenshots (optional)

<img width="478" alt="Screen Shot 2022-03-22 at 9 13 08 PM" src="https://user-images.githubusercontent.com/46905241/159622587-fd9cf1f9-9c3f-4293-ac11-e14347ae7f17.png">

## Edition and Platform

 - Edition: Personal Desktop / Personal Server / Mattermost Boards (plugin)
 - Version: v0.16 (main)
 - Browser and OS: Chrome on Mac

## Additional context (optional)

Debugged a bit into it. The failure is the `HasPermissionToTeam` call from `handleCreateBoard` because teamID is empty. It's possibly a timing issue getting the team on the client, but the bigger issue is we may want to block user actions until required state is available in the client.

chenilim:
FYI @jespino, @sbishel.

jespino:
Thanks @chenilim I guess this is going to be fixed soon with the changes proposed on the SQLite driver to make it single-threaded.

chenilim:
Do we think this is solely due to threading issues in SQLite? A quick debug (see notes above) indicated that an empty TeamID was being passed to the API call. cc @wiggin77.

wiggin77:
> Do we think this is solely due to threading issues in SQLite?

We'll see if https://github.com/mattermost/focalboard/pull/2599 stabilizes CI.

sbishel:
I ran into this issue when testing SharedBoards on Personal Server. We shouldn't be checking the Team permissions when running in Personal Server. I will take and fix this.

sbishel:
All Cypress tests have now been fixes.

PR #2624, opened 2022-03-24
update APIs to not check permissions on teamID = empty or 0

sbishel:
#### Summary
Updates APIs to not check team and board permissions if TeamID is "" or "0"

#### Ticket Link
https://github.com/mattermost/focalboard/issues/2591


chenilim:
I'll defer to @jespino, but I think this could be an issue in plugin mode - essentially allowing team=0 boards to be saved without permission checks. Maybe the easier thing is to skip the team check in `localpermissions.go`?

PR #2625, opened 2022-03-24
Fix #2589. Maintain cardProperty order on patch

### LINKED ISSUE #2589: Bug: Add a property (sometimes) doesn't add it at the end of the list

(This is from the `groupByProperty` Cypress CI test failure)

1. Open a card
2. Select "+ add a property"
3. Select the "Select" property type
4. (Sometimes) the property is added to the top of the list, rather than the bottom

## Expected behavior

New properties are added to the bottom of the list

## Screenshots (optional)

This is from the Cypress test failure - Select: "New Group" is expected to be the last property:

Group board by different properties -- MM-T4291 Group by different property (failed)

## Edition and Platform

 - Edition: Personal Desktop / Personal Server / Mattermost Boards (plugin)
 - Version: v0.16 (main)

chenilim:
#### Summary
Maintain the order of existing card properties on a patch. The order is used for display.

#### Ticket Link
#2589

### DIFF
--- a/server/model/board.go
+++ b/server/model/board.go
@@ -223,7 +223,8 @@ func (p *BoardPatch) Patch(board *Board) *Board {
 	}
 
 	if len(p.UpdatedCardProperties) != 0 || len(p.DeletedCardProperties) != 0 {
-		// first we accumulate all properties indexed by ID
+		// first we accumulate all properties indexed by, and maintain their order
+		keyOrder := []string{}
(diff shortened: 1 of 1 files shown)


chenilim:
Not sure if this issue exists elsewhere during patches, but this should at least fix the CI issue. The order of some properties may not be important. Also recommend adding more unit tests for the patch methods.

sbishel (approved):

PR #2630, opened 2022-03-24
Revert in-memory unit tests for sqlite

wiggin77:
#### Summary
The `cache=shared` pragma required by in-memory sqlite database creates more opportunity for database locked errors.  Adding a 5 second wait/timeout helped but still resulted in flaky results.  Moving back to a file based unit test is safer since each test gets a completely new instance and is impervious to the timing of the previous close within the driver.

I'm hoping the memory caching that sqlite does, and the fact unit tests create very small databases, and the parallelization of the unit tests will result in CI jobs that are not much slower.  **Update: just under 2 minutes was added to CI job.**

The temporary database file is cleaned up on test teardown.

#### Ticket Link
NONE

### DIFF
--- a/server/services/permissions/localpermissions/localpermissions.go
+++ b/server/services/permissions/localpermissions/localpermissions.go
@@ -27,7 +27,9 @@ func New(store permissions.Store, logger *mlog.Logger) *Service {
 }
 
 func (s *Service) HasPermissionToTeam(userID, teamID string, permission *mmModel.Permission) bool {
-	// Locally there is no team, so return true
+	if userID == "" || teamID == "" || permission == nil {
+		return false
+	}
 	return true
 }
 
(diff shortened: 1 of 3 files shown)


server/services/permissions/localpermissions/localpermissions.go
```diff
@@ -27,7 +27,9 @@ func New(store permissions.Store, logger *mlog.Logger) *Service {
 }
 
 func (s *Service) HasPermissionToTeam(userID, teamID string, permission *mmModel.Permission) bool {
-	// Locally there is no team, so return true
```
wiggin77, line 30: Reverting the fix for flaky cypress test as it caused consistent failures for backend tests.  IMO teamID should never be empty.

sbishel (approved):

PR #2635, opened 2022-03-24
Add TestGetTemplates integration test

chenilim:
#### Summary
Add an integration test for retrieving and using built-in templates


server/integrationtests/board_test.go
```diff
@@ -1268,3 +1269,53 @@ func TestDeleteMember(t *testing.T) {
...
+		th := SetupTestHelper(t).InitBasic()
+		defer th.TearDown()
+
+		teamID := "my-team-id"
+		rBoards, resp := th.Client.GetTemplatesForTeam("0")
+		th.CheckOK(resp)
+		require.NotNil(t, rBoards)
+		require.GreaterOrEqual(t, len(rBoards), 6)
+
+		fmt.Printf("\n\n")
```
wiggin77, line 1284: Can we use `t.Log` or `t.Logf` for these?  It captures the output in order with the rest of the log output, and also allows us to selectively output it or not based on test success/failure.

chenilim: Thanks! Fixed.

harshilsharma63 (approved):

PR #2636, opened 2022-03-25
Added support for access control on templates

### LINKED ISSUE #2370: Permissions Branch - Users have access to custom templates that were previously created in private channels

After updating, custom templates that were created in workspaces tied to private channels become accessible to any user that's part of that same team to which that private channel/workspace belonged to. 

Is this expected?

## Edition and Platform

- permissions-feature-branch

### LINKED ISSUE #2377: Permissions branch: Share dialog role selector is cut off

1. Open a board
2. Select "Show Share Dialog"
3. Select the role for your user
4. The menu is cut off

## Expected behavior

The menu should be visible

## Screenshots

Screen Shot 2022-02-22 at 3 52 15 PM

## Edition and Platform

- Edition: Mattermost plugin

harshilsharma63:
#### Summary
1. Added a separate API for fetching system default templates.
2. Updated `handleGetTemplates` API to include following boards, scoped to the current team-
    * all public templates,
    * all private templates where the user is added as a member.
3. Updated share board modal to hide sharing URL sections for templates.

#### Ticket Link
Fixes #2370
Fixes #2377


server/api/api.go
```diff
@@ -2208,6 +2209,62 @@ func (a *API) handleGetTemplates(w http.ResponseWriter, r *http.Request) {
 	auditRec.Success()
 }
 
+func (a *API) handleGetDefaultTemplates(w http.ResponseWriter, r *http.Request) {
```
wiggin77, line 2212: Not sure I understand the reasoning to make this a separate API versus fetching templates from team 0.

harshilsharma63: Yeah, this is not needed. It was needed in the middle of the implementation but I changed something and now it's not needed anymore.

Removed the API.

sbishel (commented):
I think the question Doug was asking (if not, I am :) ) is why we need `GetDefaultTemplates` at all. Shouldn't calling `GetTemplateBoards({teamID: '0', user: 'abc'}) provide the same results? If not , why. I think keeping templates working as similar to boards as possible is important.

server/services/store/sqlstore/templates.go
```diff
@@ -71,3 +71,41 @@ func (s *SQLStore) getTemplateBoards(db sq.BaseRunner, teamID string) ([]*model.
 
 	return s.boardsFromRows(rows)
 }
+
+// getDefaultTemplateBoards fetches all template blocks .
```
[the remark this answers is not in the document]
harshilsharma63: Done.

harshilsharma63:
> I think the question Doug was asking (if not, I am :) ) is why we need `GetDefaultTemplates` at all. Shouldn't calling `GetTemplateBoards({teamID: '0', user: 'abc'}) provide the same results? If not , why. I think keeping templates working as similar to boards as possible is important.

Yeah, turns out `GetDefaultTemplates` is not needed, neither in the API nor anywhere else. Removed it now.

server/app/onboarding.go
```diff
@@ -46,14 +46,14 @@ func (a *App) PrepareOnboardingTour(userID string, teamID string) (string, strin
 }
 
 func (a *App) getOnboardingBoardID() (string, error) {
-	boards, err := a.store.GetTemplateBoards(globalTeamID)
+	boards, err := a.store.GetTemplateBoards("0", "")
```
sbishel, line 49: Use globalTeamID instead of "0". Here and elsewhere.

sbishel (changes requested):

server/app/onboarding.go
harshilsharma63: Done.

sbishel (approved):

wiggin77 (approved):
LGTM :+1:

PR #2642, opened 2022-03-25
Upgrading most of the webapp dependencies.

jespino:
We are upgrading most of them, but we are not upgrading react-router, react-dnd
and react-select because they have more profound changes. I'll upgrade it
later.


sbishel (approved):
LGTM

Issue #2551, opened 2022-03-17
Permissions Branch: Not a Valid Team Error on first visit to Boards


chenilim:
## Steps to reproduce the behavior

1. Create a new account on the server
2. Select Boards from the App Switcher (top left menu)
3. "Not a Valid Team" error

## Expected behavior

No error.

## Screenshots (optional)

If applicable, add screenshots or a screen recording to elaborate on the problem.

## Edition and Platform

 - Edition: Mattermost Boards (plugin) 
 - Version: permissions-feature-branch
 - Browser and OS: Chrome on Mac

## Additional context (optional)

URL is `/boards/error?id=team-undefined`
Selecting "Back to Home" opens Channels, but opening Boards again works. This seems to only repro on the first visit.

PR #2598, opened 2022-03-23
Fixed bug where match didn't receive current router's route

harshilsharma63:

PR #2644, opened 2022-03-25
GH-2551 - Update correct routes after welcome page

sbishel:
#### Summary
Fixes issue with routing when user first logs in. They are redirected incorrectly to teamID - "undefined". 

This either routes them to back to home - `/boards` if originally routed just to `/boards`
Also fixes the match so that team it correctly pulled.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/2551

### DIFF
--- a/webapp/src/route.tsx
+++ b/webapp/src/route.tsx
@@ -27,7 +27,7 @@ type RouteProps = {
 
 function FBRoute(props: RouteProps) {
     const loggedIn = useAppSelector<boolean|null>(getLoggedIn)
-    const match = useRouteMatch<any>()
+    const match = useRouteMatch<any>({path: props.path})
     const me = useAppSelector<IUser|null>(getMe)
     const clientConfig = useAppSelector<ClientConfig>(getClientConfig)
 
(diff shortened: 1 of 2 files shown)


webapp/src/route.tsx
```diff
@@ -27,7 +27,7 @@ type RouteProps = {
 
 function FBRoute(props: RouteProps) {
     const loggedIn = useAppSelector<boolean|null>(getLoggedIn)
-    const match = useRouteMatch<any>()
+    const match = useRouteMatch<any>({path: props.path})
```
sbishel, line 30: Because a <Route> hasn't been called at this point, we need to set the path when calling.

sbishel: This change may not be necessary based on this fix. 
https://github.com/mattermost/focalboard/pull/2598

wiggin77 (approved):
LGTM :+1:

webapp/src/route.tsx
jespino, line 30: If this is no longer needed, should we remove it from the PR?

jespino (approved):

PR #2660, opened 2022-03-28
joinBoard API

chenilim:
#### Summary
Implement joinBoard API to replace calling addMember for users joining a board. This simplifies the permissions checks for both APIs.


chenilim:
@jespino, this should simplify the permissions checks for joinBoard and addMember (which can then just check for Admin rights to a board). If you agree, we can merge this separately, or feel free to incorporate it into your PR.

jespino (approved):
I like it.

Issue #2667, opened 2022-03-29
Personal-Server-only APIs should return 501 when running as a plugin


chenilim:
## Summary

The following APIs should return a 501 (not implemented) error when running as a plugin:
POST	/teams/{teamID}/regenerate_signup_token
GET	/teams/{teamID}/archive/export
POST	/users/{userID}/changepassword
POST	/login
POST	/logout
POST	/register

## How important this is to me and why

Importance: Medium

Use cases:
1. This reduces the attack-surface for the APIs

PR #2684, opened 2022-03-29
GH-2667: Update apis to return 501 if not available for plugin

sbishel:
#### Summary
Returns a 501 for those apis that should not be called when running as a plugin.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2667
  Fixes https://mattermost.atlassian.net/browse/MM-40553


jespino (approved):
LGTM

Issue #2712, opened 2022-04-01
Bug: Users can't create boards from templates


ogi-m:
## Steps to reproduce the behavior

1. Log in as a non-admin user
2. Open Boards
3. Select any template on the template picker screen
4. Click "Use this template" - nothing happens (see screenshot, note the template preview isn't loaded either)

## Expected behavior

A board should be created from the template

## Screenshots (optional)

![image](https://user-images.githubusercontent.com/15314183/161321491-32c289d5-af37-4da4-8be7-cf571a8092c6.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16
 - Browser and OS: Chrome on Windows]

wuwinson:
Duplicate of #2552 @ogi-m?

PR #2716, opened 2022-04-01
GH-2712:Don't check team permission for GlobalTeamID

sbishel:
#### Summary
Permission Checks for teamID = '0' fails for non admin users. 

When retrieving GlobalTemplates, don't check team permissions.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2712


jespino (approved):
LGTM

Issue #2666, opened 2022-03-29
Bug: Don't send notifications for non-board members


chenilim:
## Steps to reproduce the behavior

(Didn't actually test this, but the notification code is missing a membership check):

1. As User1 (a board admin), add User2 to a board
2. As User2 follow a card
3. As User1, remove User2 from the board
4. Change the card
5. User2 still gets a change notification for the card

## Expected behavior

User2 should not get the change notification.

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16 (main)

## Additional context (optional)

The notification code should check that a user is (still) a member of a board before sending any notifications to them.

PR #2718, opened 2022-04-01
Don't notify non-board members of card changes

wiggin77:
#### Summary
Don't notify non-board members of card changes 

- include permissions service in notifications backends 
- use permissions service to ensure @mentions are to users on team 
- use permissions service to ensure subscribers are members of board

**Note:  PM decision pending on whether mentioned users get automatically added to board when they get subscribed to cards**

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/2666


server/services/notify/plugindelivery/plugin_delivery.go
```diff
@@ -42,45 +39,17 @@ type PluginAPI interface {
...
-func New(botID string, serverRoot string, api PluginAPI) *PluginDelivery {
+func New(botID string, serverRoot string, api PluginAPI, permissions permissions.PermissionsService) *PluginDelivery {
 	return &PluginDelivery{
-		botID:      botID,
-		serverRoot: serverRoot,
-		api:        api,
-	}
-}
-
-func (pd *PluginDelivery) Deliver(mentionUsername string, extract string, evt notify.BlockChangeEvent) error {
```
wiggin77, line 58: This method had been moved/copied during a previous refactor but left here unintentionally. It is not called.

server/services/notify/notifysubscriptions/subscriptions_backend.go
```diff
@@ -179,6 +183,10 @@ func (b *Backend) OnMention(userID string, evt notify.BlockChangeEvent) {
 		return
 	}
 
+	// TODO: Automatically add user to board?  Fail and show UI?  Waiting for PM decision.
```
jespino, line 186: @chenilim @wuwinson we should make a decision here, for consistency with Mattermost Channels, we should ask the user in the frontend for that, if the user has the right permissions to do it.

server/services/notify/plugindelivery/mention_deliver.go
```diff
@@ -40,5 +50,5 @@ func (pd *PluginDelivery) MentionDeliver(mentionUsername string, extract string,
 		ChannelId: channel.Id,
 		Message:   formatMessage(author.Username, extract, evt.Card.Title, link, evt.BlockChanged),
 	}
-	return member.UserId, pd.api.CreatePost(post)
+	return user.Id, pd.api.CreatePost(post)
```
jespino, line 53: Should I check if the user has permission to view the board to receive mentions?

server/services/notify/plugindelivery/mention_deliver.go
```diff
@@ -24,12 +29,17 @@ func (pd *PluginDelivery) MentionDeliver(mentionUsername string, extract string,
 		}
 	}
 
+	// make sure mentioned user has permissions to team.
+	if !pd.permissions.HasPermissionToTeam(user.Id, evt.TeamID, model.PermissionViewTeam) {
```
jespino, line 33: I'm not sure if this should be the ViewBoard permission, instead of the view team.

jespino (changes requested):
Some doubts around the mentions and permissions checking, otherwise looks good to me.

server/services/notify/plugindelivery/mention_deliver.go
wiggin77: That would mean you can only mention users who are already members of the board.  Doesn't seem right to me.

wiggin77:
> Some doubts around the mentions and permissions checking

Thanks @jespino.  The main purpose of this PR was to ensure permissions were correct for receiving change notifications and adding the plumbing for permissions to the notification services.  The spec for mentions wasn't ready when this PR was created so I left a placeholder for that.  With this PR the mention permissions is the same as before:  users can mention anyone on the team.

Another PR will implement the new mentions spec which will address your concerns.

server/services/notify/plugindelivery/mention_deliver.go
jespino: Yes, I think you are right.

PR #2721, opened 2022-04-04
Avoid possible problem of block history with null in the title

jespino:


jespino:
Yes, it totally makes sense, doing the change.

Pinjasaur (approved):

PR #2722, opened 2022-04-04
Fixing Category scheme in postgres and MySQL

### LINKED ISSUE #2717: Bug: CreateCategories fails on Postgres and MySQL

1. Call the createCategories API on a Postgres DB
2. Error, e.g.:
```
error  [2022-04-01 21:43:40.341 Z]  Error creating category  category name="Test category" error="pq: invalid input syntax for integer: "7nwnj6c3jotn7jm4d1hnbk9ztac""
```
## Expected behavior

API works

## Edition and Platform

 - Version: v0.16 (main)

The [table schema](https://github.com/mattermost/focalboard/blob/main/server/services/store/sqlstore/migrations/000018_populate_categories.up.sql#L3) is varchar for SQLite (which works), but Serial for Postgres and INT for MySQL.

jespino:
Fixes #2717

I'm changing directly the migration here because I expect people to not use the main branch before the release. I think we also can just create a new migration at the end. Any strong opinion @mgdelacroix @wiggin77?

### DIFF
--- a/server/services/store/sqlstore/migrations/000018_populate_categories.up.sql
+++ b/server/services/store/sqlstore/migrations/000018_populate_categories.up.sql
@@ -3,9 +3,7 @@ CREATE TABLE {{.prefix}}categories (
     name varchar(100) NOT NULL,
     user_id varchar(32) NOT NULL,
     team_id varchar(32) NOT NULL,
-    {{if not .sqlite}}
-        channel_id varchar(32) NOT NULL,
-    {{end}}
+    channel_id varchar(32),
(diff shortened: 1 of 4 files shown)


wiggin77:
> Fixes #2717
> 
> I'm changing directly the migration here because I expect people to not use the main branch before the release. I think we also can just create a new migration at the end. Any strong opinion @mgdelacroix @wiggin77?

Makes sense.

wiggin77 (approved):
LGTM :+1:

PR #2735, opened 2022-04-06
Removing not used code

jespino:


server/services/store/sqlstore/blocks.go
```diff
@@ -799,7 +745,7 @@ func (s *SQLStore) replaceBlockID(db sq.BaseRunner, currentID, newID, workspaceI
 }
 
 func (s *SQLStore) duplicateBlock(db sq.BaseRunner, boardID string, blockID string, userID string, asTemplate bool) ([]model.Block, error) {
-	blocks, err := s.getSubTree3(db, boardID, blockID, model.QuerySubtreeOptions{})
+	blocks, err := s.getSubTree2(db, boardID, blockID, model.QuerySubtreeOptions{})
```
jespino, line 748: The only possible concerning part is this one. But I think it is correct. It basically gets the subtree of 2 now that the only block containing other blocks is the cards, so... looks safe to search for only the block and his children.

server/services/store/storetests/blocks.go
```diff
@@ -849,14 +802,14 @@ func testDuplicateBlock(t *testing.T, store store.Store) {
 	t.Run("duplicate existing block as no template", func(t *testing.T) {
 		blocks, err := store.DuplicateBlock(testBoardID, "child1", testUserID, false)
 		require.NoError(t, err)
-		require.Len(t, blocks, 3)
+		require.Len(t, blocks, 2)
```
jespino, line 805: This tests changes because is no longer duplicating grandchildren.

sbishel (approved):

PR #2747, opened 2022-04-07
Don't warn when  appears in card followed by text that is not a username

wiggin77:
#### Summary
When a `@` appeared in card text followed by something other than a username, an attempt was made to subscribe the non-user to the card.  This resulted in a warning emitted to the logs.  This PR corrects that.   e.g. a default template includes the phrase `to @mention teammates`

#### Ticket Link
NONE


Pinjasaur (approved):
Nice catch, LGTM

Issue #2745, opened 2022-04-07
Code debt - improve CI run times


chenilim:
## Summary

CI tests currently take around 30m in GitHub.

One area we can improve is the template init - which takes around 8 seconds on GitHub, multiplied by 130+ times (for a total of 17 minutes). From the last dev meeting, the proposal is to reduce or skip this for most tests (except the ones that test the templates).

## How important this is to me and why

Importance: Medium

chenilim:
cc @wiggin77 - followup from the dev meeting. Go ahead and grab it if you wanted to start working on it. Thanks.

wiggin77:
@chenilim https://github.com/mattermost/focalboard/pull/2748 has brought the CI job down to 14 minutes.

PR #2748, opened 2022-04-07
GH-2745  Don't init default templates when unit testing

wiggin77:
#### Summary
This PR speeds up unit testing (and integration testing) by selectively initializing default templates only for tests that need them.  Some log noise is also removed.

CI went from ~30 minutes to 14 minutes.

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/2745


jespino (approved):
LGTM

PR #2751, opened 2022-04-08
Loading the templates on demand on integration tests

jespino:


server/server/server.go
```diff
@@ -130,14 +130,15 @@ func New(params Params) (*Server, error) {
...
-		Permissions:   params.PermissionsService,
+		Auth:             authenticator,
+		Store:            params.DBStore,
+		FilesBackend:     filesBackend,
+		Webhook:          webhookClient,
+		Metrics:          metricsService,
+		Notifications:    notificationService,
+		Logger:           params.Logger,
+		Permissions:      params.PermissionsService,
+		SkipTemplateInit: params.SkipInitTemplates,
```
jespino, line 141: This was already there, I only needed to use it.

server/server/server.go
```diff
@@ -203,11 +204,6 @@ func New(params Params) (*Server, error) {
 
 	server.initHandlers()
 
-	if err := app.InitTemplates(); err != nil {
```
jespino, line 206: This was not needed because the app creation initialize the templates if it is needed.

jespino:
Closing this in favor of #2748

PR #2752, opened 2022-04-08
Applying migration changes

jespino:


server/services/store/sqlstore/migrations/000019_populate_category_blocks.up.sql
```diff
@@ -1,16 +1,18 @@
-CREATE TABLE {{.prefix}}category_blocks (
+CREATE TABLE {{.prefix}}category_boards (
     id varchar(36) NOT NULL,
     user_id varchar(32) NOT NULL,
```
jespino, line 3: @harshilsharma63 we think this user_id is not needed because the category_id points to the already existing category, that has a user_id. So... can a category be global? looks like this user ID is needed, but I don't have a clear vision of the data model.

server/app/category_boards.go
```diff
@@ -0,0 +1,26 @@
...
+
+	go func() {
+		a.wsAdapter.BroadcastCategoryBoardChange(
+			teamID,
+			userID,
+			model.BoardCategoryWebsocketData{
+				BoardID:    boardID,
+				CategoryID: categoryID,
+			})
+	}()
```
wiggin77, line 23:
```suggestion
	a.blockChangeNotifier.Enqueue(func() error {
		a.wsAdapter.BroadcastCategoryBoardChange(
			teamID,
			userID,
			model.BoardCategoryWebsocketData{
				BoardID:    boardID,
				CategoryID: categoryID,
			})
		return nil
	})
```

server/services/store/sqlstore/migrations/000019_populate_category_blocks.up.sql
wiggin77: If it is needed, please make it varchar(36)

server/model/board.go
```diff
@@ -359,5 +360,5 @@ type BoardMemberHistoryEntry struct {
 
 	// The insertion time
 	// required: true
-	InsertAt int64 `json:"insertAt"`
+	InsertAt time.Time `json:"insertAt"`
```
wiggin77, line 363: Do we need this in the model?  Generally we're just using `insert_at` to get a consistent order from queries.  If we need it then store needs to make the conversion.

jespino: This is for the membership history, so... the whole point is to expose the join/leave date here. We can expose it in other way, but we need to expose it (only for admins, of course, and I think is going to be behind the license).

Pinjasaur (approved):
LGTM

sbishel (approved):

server/services/store/sqlstore/migrations/000019_populate_category_blocks.up.sql
harshilsharma63: Okay, so withing storing user_id here as well, handling the query of moving a board from one category to another was getting a little complex. This makes checking for an existing category-block mapping for the current users much easier. I don't remember the exact issue but I remember I intentionally decided to have the user_id here as well.

server/services/store/sqlstore/migrations/000018_populate_categories.up.sql
```diff
@@ -30,6 +32,7 @@ CREATE TABLE {{.prefix}}categories (
         {{ end }}
         COALESCE(nullif(c.DisplayName, ''), 'Direct Message') as category_name,
         cm.UserId,
+        /* TODO: Ask Harshil about it */
```
harshilsharma63, line 35: @jespino @wiggin77 any idea what the issue is here?

harshilsharma63 (changes requested):

server/services/store/sqlstore/migrations/000019_populate_category_blocks.up.sql
harshilsharma63: @wiggin77 I've update the user_id length to 36.

server/services/store/sqlstore/migrations/000018_populate_categories.up.sql
wiggin77: We're trying to understand why we have a fake team named `direct_message` and a category named `Direct Message`.    Shouldn't these be in the default category (i.e. no category)?

harshilsharma63: Yup, this was not needed. We shouldn't be creating categories for DM and private group messages. Fixed this to only process public and private channels and skip DM and GM.

wiggin77 (approved):

sbishel (approved):
Looks good.  Thanks Doug.

PR #2757, opened 2022-04-08
Add readme for Mac and Win apps

chenilim:
#### Summary
Added a readme with tips on debugging the Mac and Windows app, plus running the server in single-user mode.

Also added a `Go: Launch Single-user Server` option to VSCode's launch.json.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,6 +14,19 @@
             "cwd": "${workspaceFolder}"
         },
         {
+            "name": "Go: Launch Single-user Server",
+            "type": "go",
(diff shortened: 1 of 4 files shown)


Pinjasaur (approved):
LGTM. When I was getting everything setup initially I went down a small rabbit hole debugging why the `xcodebuild` command wasn't happy. Didn't realize initially I needed the full Xcode app and _not_ just the CLI tools I was used to installing. I added a condition in the Makefile to output slightly more helpful information if there _is_ an error.

harshilsharma63 (approved):

PR #2758, opened 2022-04-09
Implement permissions specs for mentions.

wiggin77:
#### Summary
This PR implements [permissions spec](https://docs.google.com/document/d/1HOnoB87WZojqplpB6XGbtKsLuqjG8h-8CbJpmWsRfM4/edit#bookmark=id.7k6tbu6a6odp) for mentions.

- public boards: admin, editor, commenter can mention team members and auto-add them to board; guests can mention board members
- private boards:  admin, editor, commenter, guest can mention board members
- viewers cannot mention

#### Ticket Link
https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/150d1464-6d59-4c39-810f-ee201439dbaf/ciu68wqtbp3nuxnrt9o6db86htr

### DIFF
--- a/server/app/blocks.go
+++ b/server/app/blocks.go
@@ -358,7 +358,7 @@ func (a *App) notifyBlockChanged(action notify.Action, block *model.Block, oldBl
 
 	boardMember, _ := a.GetMemberForBoard(board.ID, modifiedByID)
 	if boardMember == nil {
-		// create guest board member
+		// create temporary guest board member
 		boardMember = &model.BoardMember{
 			BoardID: board.ID,
 			UserID:  modifiedByID,
--- a/server/app/blocks.go
+++ b/server/app/blocks.go
@@ -356,6 +356,15 @@ func (a *App) notifyBlockChanged(action notify.Action, block *model.Block, oldBl
 		return
 	}
 
+	boardMember, _ := a.GetMemberForBoard(board.ID, modifiedByID)
+	if boardMember == nil {
+		// create guest board member
+		boardMember = &model.BoardMember{
+			BoardID: board.ID,
+			UserID:  modifiedByID,
+		}
+	}
+
 	evt := notify.BlockChangeEvent{
 		Action:       action,
 		TeamID:       board.TeamID,
@@ -363,7 +372,7 @@ func (a *App) notifyBlockChanged(action notify.Action, block *model.Block, oldBl
 		Card:         card,
 		BlockChanged: block,
 		BlockOld:     oldBlock,
-		ModifiedByID: modifiedByID,
+		ModifiedBy:   boardMember,
 	}
 	a.notifications.BlockChanged(evt)
 }
--- a/server/services/notify/notifymentions/mentions_backend.go
+++ b/server/services/notify/notifymentions/mentions_backend.go
@@ -4,12 +4,14 @@
 package notifymentions
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
 	"github.com/mattermost/focalboard/server/model"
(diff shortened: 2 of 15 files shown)


server/services/notify/notifymentions/mentions_backend.go
```diff
@@ -141,3 +159,79 @@ func safeCallListener(listener MentionListener, userID string, evt notify.BlockC
...
+		//    - admin, editor, commenter: can mention anyone on team (mentioned users are automatically added to board)
+		//    - guest: can mention board members
+		switch {
+		case evt.ModifiedBy.SchemeAdmin, evt.ModifiedBy.SchemeEditor, evt.ModifiedBy.SchemeCommenter:
+			if !b.permissions.HasPermissionToTeam(mentionedUser.Id, evt.TeamID, model.PermissionViewTeam) {
+				return "", fmt.Errorf("%s cannot mention non-team member %s : %w", evt.ModifiedBy.UserID, mentionedUser.Id, ErrMentionPermission)
+			}
+			// add mentioned user to board (if not already a member)
+			member, err := b.store.GetMemberForBoard(evt.Board.ID, mentionedUser.Id)
+			if member == nil || b.store.IsErrNotFound(err) {
```
sbishel, line 189: Should this be an && check? Assuming member is nil on any error, then only the member check would be necessary.

server/app/blocks.go
```diff
@@ -356,14 +356,23 @@ func (a *App) notifyBlockChanged(action notify.Action, block *model.Block, oldBl
...
 	}
 
+	boardMember, _ := a.GetMemberForBoard(board.ID, modifiedByID)
+	if boardMember == nil {
+		// create temporary guest board member
+		boardMember = &model.BoardMember{
+			BoardID: board.ID,
+			UserID:  modifiedByID,
+		}
+	}
```
sbishel, line 366: Could you explain this? I don't understand the need for this. If the `modifiedByID` isn't a board member, wouldn't that be an error? Can that even happen? A mentioned user, may not be a member but I would think the modifier must be a member.

sbishel (approved):
Just a couple questions on some areas I'm not understanding. Otherwise, LGTM.

server/services/notify/notifymentions/mentions_backend.go
wiggin77: Our store APIs are inconsistent in returning an error when a lookup has zero records found.  Sometimes an ErrNotFound is returned with nil pointer, sometimes a nil error and nil pointer.  Also some store APIs return `store.ErrNotFound` and others return `sql.ErrNoRows`. The latter doesn't match `store.IsErrNotFound` which is the problem in this case.  Once the tech debt card to clean this up is complete the code above will still work, but also works now.

server/app/blocks.go
wiggin77: My understanding is that if the user making the card edit is a guest, then `GetMemberForBoard` will return nil, since guests are not members.  However the APIs are expecting a `model.BoardMember` so I create a dummy one with no permissions.

sbishel: We aren't supporting guest accounts in v0.16. So (in theory), `boardMember` will never be `nil`, so I don't think we need to change it now.

But, just to be clear, when we add guest support, they must be explicitly added to the board according to this spec. 
https://docs.google.com/document/d/1hQAjl-4FFaEVCVYtU2BtyGxA97hcnNxCBDSnpnDT_cw/edit#bookmark=id.807067fo17yq

PR #2765, opened 2022-04-12
Various UI updates on permissions

asaadmahmood:
#### Summary
Various UI updates on permissions

#### Screenshots
Screenshot 2022-04-12 at 6 10 42 PM
Screenshot 2022-04-12 at 6 10 47 PM


Pinjasaur (approved):
LGTM. Non-blocking question: I noticed a few different `opacity` values that feel a bit like "magic numbers"—would it be worth making them variables or maybe multiples of some base amount? Not familiar enough to know if there's some larger pattern like that e.g. design system or not.

asaadmahmood:
@Pinjasaur We use a 4/8px grid system for the opacities. That's how we do it in the webapp too. We previously had all the colors defined as variables, but we shifted to this approach.

Issue #2709, opened 2022-04-01
Permissions Branch: Missing channel intro link to Boards


wuwinson:
## Steps to reproduce the behavior

1. Create a new channel or DM
2. See error

## Expected behavior

"Create a board" link should be on the channel intro

## Screenshots (optional)
![Screen Shot 2022-04-01 at 1 30 57 PM](https://user-images.githubusercontent.com/93531870/161313593-ee4f216b-93a3-4c93-8fe5-16d5155701db.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: permissions-feature-branch
 - Browser and OS: Chrome on Mac

chenilim:
Looks like we just need to update the server version.

sbishel:
Investigated, when the app bar feature flag is on this doesn't work. Updating in mattermost-webapp.

Also the link still goes to "workspace", updating to "team"

PR #2766, opened 2022-04-12
update link for teams

sbishel:
#### Summary
Updates introduction link to use teams rather than channels.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/2709

### DIFF
--- a/mattermost-plugin/webapp/src/index.tsx
+++ b/mattermost-plugin/webapp/src/index.tsx
@@ -233,13 +233,6 @@ export default class Plugin {
             }
 
             this.registry.registerPostWillRenderEmbedComponent((embed) => embed.type === 'boards', BoardsUnfurl, false)
-        } else {
-            windowAny.frontendBaseURL = subpath + '/plug/focalboard'
-            this.channelHeaderButtonId = registry.registerChannelHeaderButtonAction(<FocalboardIcon/>, () => {
-                const currentTeam = mmStore.getState().entities.teams.currentTeamId
-                window.open(`${window.location.origin}/plug/focalboard/team/${currentTeam}`)
-            }, 'Boards', 'Boards')
-            this.registry.registerCustomRoute('/', MainApp)
         }
 
         const config = await octoClient.getClientConfig()
--- a/mattermost-plugin/webapp/src/index.tsx
+++ b/mattermost-plugin/webapp/src/index.tsx
@@ -218,9 +218,9 @@ export default class Plugin {
             )
 
             const goToFocalboardTemplate = () => {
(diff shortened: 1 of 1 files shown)


mattermost-plugin/webapp/src/index.tsx
```diff
@@ -236,8 +236,8 @@ export default class Plugin {
         } else {
             windowAny.frontendBaseURL = subpath + '/plug/focalboard'
             this.channelHeaderButtonId = registry.registerChannelHeaderButtonAction(<FocalboardIcon/>, () => {
-                const currentChannel = mmStore.getState().entities.channels.currentChannelId
-                window.open(`${window.location.origin}/plug/focalboard/workspace/${currentChannel}`)
+                const currentTeam = mmStore.getState().entities.teams.currentTeamId
+                window.open(`${window.location.origin}/plug/focalboard/team/${currentTeam}`)
```
Pinjasaur, line 240: Any reason not doing the `_blank` and `noopener` like above here too?

```suggestion
                window.open(`${window.location.origin}/plug/focalboard/team/${currentTeam}`, '_blank', 'noopener')
```

sbishel: After spending some time investigating, I'm not sure the "else" section even gets hit. The path provided `/plug/focalboard` is wrong as well, leading me to believe the section is unnecessary.

Pinjasaur (approved):
LGTM

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -233,13 +233,6 @@ export default class Plugin {
             }
 
             this.registry.registerPostWillRenderEmbedComponent((embed) => embed.type === 'boards', BoardsUnfurl, false)
-        } else {
-            windowAny.frontendBaseURL = subpath + '/plug/focalboard'
-            this.channelHeaderButtonId = registry.registerChannelHeaderButtonAction(<FocalboardIcon/>, () => {
-                const currentChannel = mmStore.getState().entities.channels.currentChannelId
-                window.open(`${window.location.origin}/plug/focalboard/workspace/${currentChannel}`)
-            }, 'Boards', 'Boards')
-            this.registry.registerCustomRoute('/', MainApp)
```
sbishel, line 242: This section is never called. The initialize function is called by the web app - here 
https://github.com/mattermost/mattermost-webapp/blob/master/plugins/index.js#L172

The PluginRegistry is created with a "registerProduct" function. I'm leaving the check as a sanity check, but the else section is not needed and is wrong as well.

Pinjasaur (approved):

Issue #2593, opened 2022-03-23
Bug: AWS builds for Windows10 standalone app is broken


lindy65:
## Steps to reproduce the behavior

1. Download Windows build from https://focalboard-builds.s3.amazonaws.com/latest/focalboard-win.zip
2. Unzip
3. Open `focalboard.exe`
4. Bypass Windows Defender: Select More Info, and select `Run Anyway`
5. App doesn't open

## Expected

App runs. See note below for additional debug info.

## Edition and Platform

 - Edition: Personal Desktop
 - Version: v0.16.0
 - Browser and OS: macOS/Safari & Windows 10/Chrome

Add any other context about the problem here, and any notes about the severity:
* Sev 1: Affects critical functionality without a workaround

(Outdated info) Please see conversation in [~focalboard channel](https://community.mattermost.com/core/pl/n3wbwih3bpr9pfhm8ew9wjj79c) for more details.

lindy65:
Hi @chenilim - FYI, this bug is blocking tests from running in Rainforest

chenilim:
Thanks @lindy65. We should triage this today, cc @sbishel, @wuwinson.

lindy65:
Thanks @chenilim :)

chenilim:
@jespino, are there more fixes required for this?

lindy65:
@chenilim - the AWS build for Windows 10 does not open after extracting the files. Double-clicking on "Focalboard.exe" does nothing.

There seem to be a few bugs around Status and adding properties in the AWS build for macOS

![image](https://user-images.githubusercontent.com/13020833/161693810-8b2be2c9-53ee-479b-9ac4-424d463cf132.png)

chenilim:
Thanks @lindy65. Are **both** the Mac and Windows build broken for you? I am able to repro it for Windows. I spent some time debugging, and the error appears to be:
```
2022/04/05 18:23:19 server.NewStore ERROR{error 1 0 0 cannot read migration "000001_init.down.sql": open migrations\000001_init.down.sql: file does not exist <nil>}
```
From what I can tell, the migration system was changed to morph in the permissions branch, and that was when this issue started happening. i.e. it repros on the permissions-branch PR (https://github.com/mattermost/focalboard/pull/2578), but not on the one right before (https://github.com/mattermost/focalboard/pull/2572).

@sbishel, this appears to be a different issue from the menu behavior change, and appears to be related to migrations. Could we get someone to investigate?

[...]

PR #2767, opened 2022-04-12
GH 2593: Fix Windows migration

chenilim:
#### Summary
Fix path of resources to avoid crash during migration.

#### Ticket Link
#2593


chenilim:
Yup, the issue was indeed that Path.Join uses the wrong separator on Windows, and the resources always use "/" internally. I verified that the updated Windows App launches successfully with this change.

Pinjasaur (approved):
LGTM—is it worth reviewing the other instances of `.Join` in the codebase or is this a one-off because of the `embed` usage?

chenilim:
Thanks @Pinjasaur! Good call! I did a quick search of `filepath.Join` and all the other instances deal with actual files (in the archive system, and file uploads). I also did a smoke test of export archive on Windows, and that appears to work correctly. The learning for the future is that the embedded resources use the "/" directory separator.

Issue #2759, opened 2022-04-11
Bug: Image is not displayed / shows a broken icon when adding Content of type "image" to a card


lindy65:
## Steps to reproduce the behavior

1. Open a card
2. Hover over "Add a description..." near the bottom of the card
3. Click on "Add content", select type "image"
4. Select an image to add and click "Open" (Win10)
5. Observed: an outline of where the image is displayed is shown but no image is displayed

## Expected behavior

The image you added is displayed

## Screenshots (optional)

![image](https://user-images.githubusercontent.com/13020833/162786555-f83fe6ea-8226-4d25-92d1-19513a560ee6.png)

![image](https://user-images.githubusercontent.com/13020833/162786780-35883317-3b0c-4203-93a5-2866823d7da2.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16.0 master
 - Browser and OS: Win10/Chrome and macOS/Safari

## Additional context (optional)
* Sev 1: Affects critical functionality without a workaround

PR #2768, opened 2022-04-12
GH-2759 - fix path to file api

sbishel:
#### Summary
Prior to permissions release, the `api/v1` was not included in the file upload path. This was fixed in the permission branch, but the octoclient was never updated. This PR updates the octoclient to include that in the path.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2759


harshilsharma63 (approved):
LGTM 👍

PR #2771, opened 2022-04-13
Fix #2770: Handle Windows path in import and tests

### LINKED ISSUE #2770: Bug: Windows App doesn't have templates on fresh install
## Steps to reproduce the behavior (on Windows)

1. Delete the `My Documents\Focalboard` folder if it exists
2. Install the latest build of the Windows App
3. Open it
4. No built-in templates exist

## Expected behavior

Standard built-in templates exist

## Edition and Platform

 - Edition: Personal Desktop
 - Version: 0.16 (main)
 - Browser and OS: Windows

A failure happens in the import templates method.

chenilim:
#### Summary
Use the OS-dependent function to manage paths, to avoid a failure in importing templates.

Also fixed Windows issues in server tests, and added CI server tests for Windows and SQLite.

#### Ticket Link
#2770


wiggin77:
@chenilim the windows test brought the CI time from 14 mins to 22 mins.  Do we need this every PR or can we do it only before release builds?

server/app/import.go
```diff
@@ -58,7 +58,7 @@ func (a *App) ImportArchive(r io.Reader, opt model.ImportArchiveOptions) error {
 			return err
 		}
 
-		dir, filename := path.Split(hdr.Name)
+		dir, filename := filepath.Split(hdr.Name)
```
chenilim, line 61: @wiggin77, while this fixes the issue for me, I think there may be more to it. I checked and this code shipped in v0.15, where it is working correctly. What changed is we now generate the archive at build-time, and I *think* that causes the reader to  use OS-dependent path separators, but I'm not 100% sure.

chenilim:
@wiggin77, on the CI times - there seems to be variation in the CI runtimes. On the initial CI run off my fork, it took 14 minutes. I've re-run it again to see. Agree we can disable it if it takes much longer, or (better yet) run a subset of tests on Windows. I was planning on adding single-user tests for Windows and Mac, and those could run a subset of the current unit tests.

server/app/import.go
wiggin77: Switching to `filepath` is what fixed it.  `path` only handles slash, but not backslash.  `filepath` handles both.  The hardcoded paths used slash, but the generated archive used whatever the filesystem decided.  In the case of generating the archive on windows it ended up with backslashes.  `filepath` will handle either case so import won't care which OS generated the archive.

PR #2739, opened 2022-04-06
Simplifying the categories migration

jespino:
Fixes #2717

PR #2774, opened 2022-04-13
Avoided using UUID() as function param for MySQL migration

### LINKED ISSUE #2773: Bug: Category mmigrations produce duplicate primary keys in MySQL 5.7

1. Switch to a Mattermost server using MySQL 5.7, ideally the MySQL version provided by default in Mattermost docker-compose.
1. Install a previous version of Focalboard, say, 0.15.
2. Create some boards in channels.
3. Upgrade to the latest main.
4. Notice the error produced by categories migration about duplicate values in the primary key. 

## Expected behavior

Categories data migration should complete successfully.

harshilsharma63:
Fixes #2773

@sbishel @wiggin77 This is the quickest and cleanest fix IMO. The data migration anyways updates the IDs to our standard UUID format, so it doesn't really matter what we use in the SQL migration as long as its a varchar.

Tested on both MySQL v5.7 and v8.

### DIFF
--- a/server/services/store/sqlstore/migrations/000018_populate_categories.up.sql
+++ b/server/services/store/sqlstore/migrations/000018_populate_categories.up.sql
@@ -26,7 +26,7 @@ CREATE TABLE {{.prefix}}categories (
             REPLACE(uuid_in(md5(random()::text || clock_timestamp()::text)::cstring)::varchar, '-', ''),
         {{ end }}
         {{ if .mysql }}
-            REPLACE(UUID(), '-', ''),
+            UUID(),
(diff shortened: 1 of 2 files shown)


wiggin77:
Ok. Doesn't this need to be fixed in the migration PR?

wiggin77 (approved):

harshilsharma63:
> Ok. Doesn't this need to be fixed in the migration PR?

This change was already in master, done as part of #2739, so not needed in migration cleanup PR.

sbishel (approved):

Issue #2744, opened 2022-04-07
Bug: Including a`#` in a card title stops the board export at that character


coltoneshaw:
## Steps to reproduce the behavior

1. Go to a board
2. Add a `#` to any card title
3. Attempt to run the export
4. View the `csv` and you'll see the last thing exported is the card containing the `#` and it stops right where the `#` character lives. 

## Expected behavior

Export the whole file properly.

## Example (optional)
![Screen Shot 2022-04-07 at 10 00 18 AM](https://user-images.githubusercontent.com/46071821/162216964-cc51159e-7618-4d33-bbbb-fe229c83db22.png)

[Onboarding-without-hash.csv](https://github.com/mattermost/focalboard/files/8443494/Onboarding-without-hash.csv)
[Onboarding-withhash.csv](https://github.com/mattermost/focalboard/files/8443497/Onboarding-withhash.csv)

### With a `#` in a card title

```csv
Name
"@mention teammates"
"Add new properties"
"Create a new card 
```

### Without a `#` in a card title. 

```csv
Name
"@mention teammates"
"Add new properties"
"Create a new card"
"Create a new view"
"Create your own board"
"Drag cards"
"Filter and sort cards"
"Manage tasks with cards"
"Share a board"
"Share cards on Channels"
```

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.15.0
 - Browser and OS: Mattermost Desktop v5.0.4

## Additional context (optional)

* Sev 3: Affects minor, non-critical functionality

PR #2778, opened 2022-04-13
GH-2744 Fix export with hash sign (#) in block title breaks export to CSV

wiggin77:
#### Summary
This PR fixes a bug where `#` in a block title breaks export to CSV.  The issue is that CSV export is done client side, and the technique used is to stuff the CSV contents into a data URL.  The data URL cannot contain `#` because that is a special character that ends the path (or in this case data) portion of the URL.  Everything after that gets truncated.

**BONUS:**  added an export archive unit test.

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/2744


sbishel (approved):
LGTM

PR #2784, opened 2022-04-13
CI updates: Speed up register / login, minimize Windows tests, and add Mac tests

chenilim:
#### Summary
* Dramatically speed up CI runs by reducing password hash complexity during unit tests.
* Remove unnecessary npm calls and server builds for server unit test actions
* Only run integration unit tests for Windows to speed up run times
* Add Mac to CI runs

Net effect is the total CI run should now take less than 10-12 minutes in GitHub, and local server unit test runs are under 3 minutes.


server/services/auth/password.go
```diff
@@ -23,9 +23,11 @@ const (
 	InvalidSymbolPassword    = "symbol"
 )
 
+var PasswordHashStrength = 10
```
chenilim, line 26: @wiggin77 and @jespino - added this variable so that unit tests can reduce it (making register and login calls go from around 0.8 to 1 second to near instantaneous). We should be careful that this variable isn't accidentally set during production, and open to other ways to refactor this.

server/services/scheduler/scheduler_test.go
```diff
@@ -35,8 +35,8 @@ func TestCreateTask(t *testing.T) {
 
 func TestCreateRecurringTask(t *testing.T) {
 	taskName := "Test Recurring Task"
-	taskTime := time.Millisecond * 200
-	taskWait := time.Millisecond * 100
+	taskTime := time.Millisecond * 500
+	taskWait := time.Millisecond * 200
```
chenilim, line 39: This test was intermittently failing on macOS. Increasing the duration seems to make it stable, but we can continue to monitor to see if it needs to be refactored more.

server/services/auth/password.go
wiggin77: There is the `FB_UNIT_TESTING` environment variable that is used in other places to change config for unit tests.

chenilim: I did consider that, but wasn't sure if that was better or worse than setting the variable directly. There's a (small) risk that environment variables (unintentional or malicious) could affect a production environment. Do you have a strong preference one way or another?

wiggin77 (approved):

server/services/auth/password.go
wiggin77: 0/5

Issue #2575, opened 2022-03-21
Bug: focus on title input field when adding card


sambody:
I would like the focus (cursor) automatically be in the card title input field. So that I can start adding the title right away, without having to click on it.

## How important this is to me and why

Importance: High

Having the focus automatically on the title would save a bit of time (one less click), especially when adding multiple cards in a row.

## Additional context/similar features

A good implementation of this is Trello: when adding a new card, the focus is directly on the input field. What's more is that when adding the title and hitting enter, the card is saved, a new card is presented, again with the focus in the input field. This makes adding multiple cards very fast, all with the keyboard. As if writing a list of items in a text document. I'm not sure that speed is possible in Focalboard, since, unlike trello, adding a new card pops up a lightbox window. But having the focus on the title would be one step forward to faster card creation.

Thanks.

kamre:
Hi @sambody, this was working before exactly like you have described. So some regression was introduced.

@chenilim FYI

chenilim:
Thanks @sambody and @kamre! I checked, and this is indeed a regression... from 0.11. So, it's been broken for a while, but agree that the behavior should be to auto-select the title field when opening a card with no title (such as a new card). cc @wuwinson, @asaadmahmood.

chenilim:
Also, as a note: Would be worth adding a (Cypress) unit test for this when fixed.

kamre:
Here is a possible fix:
```diff
Index: webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
===================================================================
diff --git a/webapp/src/components/markdownEditorInput/markdownEditorInput.tsx b/webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
--- a/webapp/src/components/markdownEditorInput/markdownEditorInput.tsx	(revision aa9ebc237fc162c1c98a9ff02cff4389bc539c83)
+++ b/webapp/src/components/markdownEditorInput/markdownEditorInput.tsx	(date 1648580066731)
@@ -125,10 +125,13 @@
     }, [editorState, onBlur])
 
     const onEditorStateChange = useCallback((newEditorState: EditorState) => {
+        const text = editorState.getCurrentContent().getPlainText()
         const newText = newEditorState.getCurrentContent().getPlainText()
-        onChange && onChange(newText)
-        setEditorState(newEditorState)
-    }, [onChange])
+        if (text !== newText) {
+            onChange && onChange(newText)
+            setEditorState(newEditorState)
+        }
[...]

PR #2795, opened 2022-04-14
Fix focusing title

kyeongsoosoo:
#### Summary

Sorry for opening a PR for #2575 again. 
After initial PR(#2772) was merged, I found that mention plugin is not working.
I found a wokraround for #2575, so I fixed.

#### Ticket Link
 #2575 
Should I open an issue for this PR??

Fixes https://github.com/mattermost/focalboard/issues/2816


sbishel (approved):
Thanks @kyeongsoosoo - we found this in our bug bash, so a ticket was already created.

Issue #2798, opened 2022-04-14
Bug: Newly created boards from templates should be private by default (not public)


chenilim:
## Steps to reproduce the behavior

1. Create a new board from a template, e.g. Personal Goals
2. Open the Share dialog
3. The board is open to everyone on the team

## Expected behavior

Newly created boards should be private by default.

## Screenshots (optional)

<img width="620" alt="Screen Shot 2022-04-14 at 8 15 43 AM" src="https://user-images.githubusercontent.com/46905241/163421042-678e2811-ce0d-49df-bd36-549a5781d70f.png">

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16 (bugbash)
 - Browser and OS: Chrome on Mac

## Additional context (optional)

Note that creating a new empty board sets it as private by default, but not for templates. This is likely because the permissions are duplicated from the template.

PR #2821, opened 2022-04-15
GH-2798 - make board private when created from template

sbishel:
#### Summary
Currently all boards created from public templates are public (because the templates are public). This PR updates the duplication logic to set boards created from a template to private boards.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/2798

### DIFF
--- a/server/services/store/sqlstore/boards_and_blocks.go
+++ b/server/services/store/sqlstore/boards_and_blocks.go
@@ -141,6 +141,11 @@ func (s *SQLStore) duplicateBoard(db sq.BaseRunner, boardID string, userID strin
 	if err != nil {
 		return nil, nil, err
 	}
+	// if original board is template
+	// make new board private
+	if board.IsTemplate {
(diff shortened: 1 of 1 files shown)


wiggin77 (approved):
LGTM  :+1:

server/services/store/sqlstore/boards_and_blocks.go
```diff
@@ -141,6 +141,11 @@ func (s *SQLStore) duplicateBoard(db sq.BaseRunner, boardID string, userID strin
 	if err != nil {
 		return nil, nil, err
 	}
+	// if original board is template
+	// make new board private
+	if board.IsTemplate {
+		board.Type = "P"
```
chenilim, line 147: Just a note that we should set all duplicated boards to private, based on the followup from the bugbash this week. I filed a separate bug to track that: https://github.com/mattermost/focalboard/issues/2822.

Issue #2808, opened 2022-04-14
Bug: Unexpected error in log every time an editor opens the Share dialog


wiggin77:
## Steps to reproduce the behavior

1. Be an editor on a board (not board creator/admin)
2. Click the `Share` button
3. Dialog displays fine, but an error is always generated in log:

```
{"timestamp":"2022-04-14 17:44:33.763 Z","level":"error","msg":"API ERROR","caller":"app/plugin_api.go:940","plugin_id":"focalboard","code":"403","error":"\"access denied to sharing the board\"","_msg":"","api":"/api/v2/boards/bq495ankeu7gr5jjagaa1h73rme/sharing"}
```

## Expected behavior

No error in log

## Edition and Platform

 - Edition: plugin
 - Version: v0.16 bug bash
 - Browser and OS: Linux FF 93

## Additional context (optional)

Add any other context about the problem here, and any notes about the severity:
* Sev 1: Affects critical functionality without a workaround
* Sev 2: Affects major functionality with a difficult or non-obvious workaround
* Sev 3: Affects minor, non-critical functionality

sbishel:
Possibly related to https://github.com/mattermost/focalboard/pull/2719 (v0.15)

PR #2827, opened 2022-04-15
GH-2808: fix some logging issues with share board.

sbishel:
#### Summary
When displaying shared board menu, check if user has permissions before calling API. Otherwise, API logs error and return error unnecessarily. 

If "sharing" is not found. That is an expected case and no reason to log error.

#### Ticket Link
https://github.com/mattermost/focalboard/issues/2808


jespino (approved):
Tests are failing, but the code looks good to me. Probably you need to update a tests around the new OK return instead of 404 return.

Pinjasaur (approved):
LGTM

Issue #2815, opened 2022-04-14
Bug: Follow button not working when opening card from @mention notification


wiggin77:
## Steps to reproduce the behavior

1. User1 mention User2 (who is not a member of the board)
2. User2 clicks card title in notification to open card
3. User2 cannot click greyed Follow button (should be automatically subscribed)
4. User2 clicks to another board, then back to User1's board
5. Follow button works again

## Expected behavior

Follow button accurately shows state and it clickable

## Screenshots (optional)

If applicable, add screenshots or a screen recording to elaborate on the problem.

## Edition and Platform

 - Edition: plugin
 - Version: v0.16 bug bash
 - Browser and OS: Linux FF 93

## Additional context (optional)

Add any other context about the problem here, and any notes about the severity:
* Sev 1: Affects critical functionality without a workaround
* Sev 2: Affects major functionality with a difficult or non-obvious workaround
* Sev 3: Affects minor, non-critical functionality

Issue #2783, opened 2022-04-13
Bug: Card links don't open the card dialog when opened in the same tab or in a new window


lindy65:
## Steps to reproduce the behavior

1. Post a link to a card in channels
2. Click on the link (not the card preview)
3. Observed: User is taken to the board but the card dialog doesn't open

## Expected behavior

User is taken to the boards with the linked card dialog open

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.16.0
 - Browser and OS: Win10/Chrome and macOS/Safari

This was fixed previously in https://github.com/mattermost/focalboard/issues/1702

PR #2833, opened 2022-04-18
GH-2815 fix for following button on card

sbishel:
#### Summary
Fixes setting the following button on a card. Currently if going directly to a card via a URL the following information is not loaded.

In addition in order to duplicate the issue, I had to provide a fix where if going directly to the card via a URL, you were automatically redirected to the team.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/2815
  Fixes https://github.com/mattermost/focalboard/issues/2783


wiggin77 (approved):
LGTM :+1:

Issue #2750, opened 2022-04-08
[Tech Debt]: Review log level at server side.


jespino:
We are logging a lot of stuff in the server side at Error, Info, Warn and Debug level.

We should have a clear vision of what is each level for and avoid any message that is not directly useful for the administrator (useful means he can do something or can to send that data to US to do something).

Any log that is not useful for the administrator (based on the previous definition) should be in the `Debug` level.

jespino:
@wiggin77 I assigned it to you because I'm out the whole week next week, feel free to reassign it or give it back to me if you want, but I'll do it, in the best scenario, after the next week.

wiggin77:
Also `BroadcastingXXXChange` are output at `INFO` level and should be `DEBUG.`
```
{"timestamp":"2022-04-14 17:33:47.130 Z","level":"info","msg":"BroadcastingBoardChange","caller":"app/plugin_api.go:937","plugin_id":"focalboard","teamID":"giyjmbra97nnfrxz71hoaoboco","boardID":"b35szx6f5o7n1iyn1mr5oz8xheo"}
{"timestamp":"2022-04-14 17:33:47.281 Z","level":"info","msg":"BroadcastingMemberChange","caller":"app/plugin_api.go:937","plugin_id":"focalboard","teamID":"giyjmbra97nnfrxz71hoaoboco","boardID":"b35szx6f5o7n1iyn1mr5oz8xheo","userID":"zc3jz8a393ghiret31ehrdfe8a"}
```

PR #2834, opened 2022-04-18
Standardize err not found

wiggin77:
#### Summary
This PR standardizes on `model.IsErrNotFound` for checking if an error was one of the `not found` variety.  Previously there was a mix of `store.IsErrNotFound` and `errors.Is(err, sql.ErrNoRows)` and another pattern for plugin API calls.

This simplifies error checking and reduces mistakes.  For example cases where the arguments were reversed for `errors.Is(err, sql.ErrNoRows)` have been fixed.

#### Ticket Link
Follow-up to https://github.com/mattermost/focalboard/issues/2750


Pinjasaur (approved):
LGTM, nice refactor

server/services/store/storetests/boards.go
```diff
@@ -106,7 +105,7 @@ func testGetBoard(t *testing.T, store store.Store) {
 
 	t.Run("nonexisting board", func(t *testing.T) {
 		rBoard, err := store.GetBoard("nonexistent-id")
-		require.ErrorIs(t, err, sql.ErrNoRows)
+		require.True(t, model.IsErrNotFound(err))
```
jespino, line 108: nit: A better error message would be great if not you get something like:

```
=== RUN   TestSQLStore/BoardStore/GetMemberForBoard/should_return_a_no_rows_error_for_nonexisting_membership
    boards.go:571: 
        	Error Trace:	boards.go:571
        	Error:      	Should be false
        	Test:       	TestSQLStore/BoardStore/GetMemberForBoard/should_return_a_no_rows_error_for_nonexisting_membership
```

jespino (approved):
LGTM, I added a comment that can be applied in multiple places, but other than that, it looks great.

server/services/store/storetests/boards.go
wiggin77: Added descriptions to the asserts.  https://github.com/mattermost/focalboard/pull/2834/commits/4fb77221e7f68c213be6de5865bb318dd20f2618

Issue #2506, opened 2022-03-09
Bug: Calendar week/month toggles are applying to all calendars views in the same board


wuwinson:
## Steps to reproduce the behavior

1. Go to Content Calendar template or any board with multiple calendar views
2. Go to the first calendar view
3. Select the Week toggle
4. Go to second calendar view
5. See error - second calendar should be on the Month view by default
6. Select the Month toggle on second calendar
7. Go back to first calendar view
8. See error - first calendar view is back on the Month view instead of the Week view
9. Error also occurs when switching to different months.

## Expected behavior

Other calendar views should not change based on changes to another calendar in the same board

## Screenshots

https://user-images.githubusercontent.com/93531870/157498742-53a2e62f-14e6-4fc8-b549-b454361216a3.mov

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.15 RC1
 - Browser and OS: Chrome 96 on macOS

sbishel:
I investigated and when switching between calendar views the calendar object is reused, therefore those settings are reused. If you go to a non-calendar view, when switching back to calendar these are set back to default.

Discussed with @chenilim and @wuwinson. Long-term we want to be able change views and only store to the view when the use chooses to "save". There is quite a bit to figure out and design. Therefore, short-term we have 3 choices. 
1. Do nothing, user switching between calendar views settings are shared. 
2. Have it works similar to "Search", reload the view so that when switching between views it is always the default.
3. Store as a property on the view. This is the most expensive and also means when I switch the view it changes for everyone. Similar to "Display by", "Properties", and "Filters".

Note: The selected month has the same issue.
Assigning to @wuwinson to reach a decision.

wuwinson:
Let's have calendar view work similar to option 2 "Search".

PR #2835, opened 2022-04-18
GH-2506 Calendar View

sbishel:
#### Summary
Adds key to FullCalendar component so it is reloaded rather then reused. If switching from one calendar view from another calendar view.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2506


harshilsharma63 (approved):
LGTM 👍

@sbishel the original bug is resolved, but I notice that the switch from month to week view (and vice-versa) is not saved anywhere. If I switch to week view and refresh, it lands upon month view always. Is this something to figure out later or a bug?

sbishel:
> LGTM 👍
> 
> @sbishel the original bug is resolved, but I notice that the switch from month to week view (and vice-versa) is not saved anywhere. If I switch to week view and refresh, it lands upon month view always. Is this something to figure out later or a bug?

That is the decision we made for now. Essentially, until we save with something like "user preferences", it is one way or the other. That should all be explained in the bug.

PR #2839, opened 2022-04-18
Local permissions test

chenilim:
#### Summary
Added local (Personal Server) version of permissions_test.
* Added missing permission check for duplicating board to another team
* Refactored permissions_test to handle variable user IDs (for local server tests)

Diff `permissinos_tests.go` with `local_permissions_tests.go` to see differences in expected responses.


server/integrationtests/local_permissions_test.go
```diff
@@ -0,0 +1,2047 @@
...
+}
+
+func TestLocalPermissionsRegister(t *testing.T) {
+	th := SetupTestHelperLocalMode(t)
+	defer th.TearDown()
+	clients := setupLocalClients(th)
+	testData := setupData(t, th)
+
+	ttCases := []TestCase{
+		{"/register", methodPost, "", userAnon, http.StatusInternalServerError, 0},
```
jespino, line 1592: Why register method is generating an `InternalServerError`?

jespino (changes requested):
Everything looks good to me, except the InternalServerError return on register.

server/integrationtests/local_permissions_test.go
chenilim: Thanks. I've updated the test case to make a successful register call.

jespino (approved):
LGTM

PR #2840, opened 2022-04-19
Updated update category-board mapping method to not rely on DB returning affected rows

### LINKED ISSUE #2799: Bug: Move board to category does not work

1. Create a new category in the sidebar (... next to Boards, "Create new category")
2. Select  ... next to an open board and select "Move to..." and select the new category
3. The board does not move

## Expected behavior

The board is moved to the selected category

## Screenshots (optional)

Screen Shot 2022-04-14 at 8 24 11 AM

## Edition and Platform

 - Version: v0.16 (bugbash)

harshilsharma63:
Fixes #2799

I'm not sure why the duplicate user-category-board rows got introduced in the database but I suspect it has to do with replying to the database returning the correct number of rows affected in an update operation. This is known to be inconsistent across databases and drivers.

In this PR I've removed the dependency on affected rows being rec=turned by DB in the update operation.

### DIFF
--- a/server/services/store/sqlstore/category_boards.go
+++ b/server/services/store/sqlstore/category_boards.go
@@ -2,17 +2,12 @@ package sqlstore
 
 import (
 	"database/sql"
-	"errors"
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/mattermost/focalboard/server/model"
 	"github.com/mattermost/focalboard/server/utils"
 
 	"github.com/mattermost/mattermost-server/v6/shared/mlog"
-)
-
-var (
-	errDuplicateCategoryEntries = errors.New("duplicate entries found for user-board-category mapping")
 )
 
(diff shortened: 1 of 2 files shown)


sbishel (changes requested):
Since this is now multiple queries (delete and add). It should be wrapped in a transaction. Also worries me a bit we don't know how the extra records got added, but if we are deleting first, that's probably ok.

harshilsharma63:
> Since this is now multiple queries (delete and add). It should be wrapped in a transaction. Also worries me a bit we don't know how the extra records got added, but if we are deleting first, that's probably ok.

Done. Wrapped the queries into a transaction.

Previously I was inserting the raw when the update query returns 0 as the affected number of rows, so it mush be something around that. But yeah, it works fine now. Previously if I moved a board between two categories about 8-10 times, I got the 500 error. With this change, I've done it more than 50 times (via JS, of course, 😛) and it worked fine.

sbishel (approved):

PR #2841, opened 2022-04-19
Run lint in CI for plugin;  fix plugin linter errors

wiggin77:
#### Summary
I recent PR slipped through with linter errors in plugin.  This PR ensures plugin gets linter run in CI.  Also fixes the linter errors.

There already was a `lint-plugin.yml` script that was supposed to lint the plugin, however it didn't work because the wrong working directory was specified.  It failed silently.  

Note:  purposely holding back one linter fix to ensure CI fails (as it should).

#### Ticket Link
NONE


wiggin77:
This PR went nowhere good.  Closing.

PR #2848, opened 2022-04-20
Include plugin unit tests in CI via makefile; fix unit tests

wiggin77:
#### Summary
This PR enables unit testing for plugin when running locally via `make server-test` or CI.  Also fixes broken unit tests.

#### Ticket Link
NONE

### DIFF
--- a/mattermost-plugin/server/plugin_test.go
+++ b/mattermost-plugin/server/plugin_test.go
@@ -6,13 +6,19 @@ import (
 	"net/http/httptest"
 	"testing"
 
-	"github.com/mattermost/mattermost-server/v6/model"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/mattermost/mattermost-server/v6/model"
 )
 
 func TestServeHTTP(t *testing.T) {
+	th, tearDown := SetupTestHelper(t)
+	defer tearDown()
+
 	assert := assert.New(t)
-	plugin := Plugin{}
+	plugin := Plugin{
+		server: th.Server,
+	}
 	w := httptest.NewRecorder()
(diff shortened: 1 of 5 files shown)


mattermost-plugin/server/plugin_test.go
```diff
@@ -25,11 +31,18 @@ func TestServeHTTP(t *testing.T) {
 	assert.Nil(err)
 	bodyString := string(bodyBytes)
 
-	assert.Equal("Hello, world!", bodyString)
+	// TODO: fix this; where is `Hello World` text generated?
+	// assert.Equal("Hello, world!", bodyString)
```
wiggin77, line 35: @jespino I left this as a TODO as I'm not sure of the intention.  This code comes from a very old PR https://github.com/mattermost/focalboard/pull/335.  I'm not seeing where the `Hello World` text is generated now.

jespino: It was originally the code that was contained in the `/` of the plugin http server, I guess this is not needed.

See it here: https://github.com/mattermost/mattermost-plugin-starter-template/blob/master/server/plugin.go#L25

Probably this tests just come with the starter template, see it here:

https://github.com/mattermost/mattermost-plugin-starter-template/blob/master/server/plugin_test.go#L27

We can use any other endpoint to verify that this is working and returning the expected that. or just keep the test as is without the TODO.

wiggin77: @jespino I added a simple `/hello` endpoint to act as a web service up check.  The unit test now uses that.

sbishel (approved):

Issue #2520, opened 2022-03-10
Bug: Ctrl+X and Ctrl+Z don't work in comments


chenilim:
## Steps to reproduce the behavior

1. Open a card
2. Enter a comment
3. Select some text
4. Press Ctrl+X (or Cmd+X on Mac)
5. Nothing happens

## Expected behavior

The text should be cut.

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.15 RC1
 - Browser and OS: Chrome on Mac

## Additional context

Similarly, Cmd+Z / Cmd+Shift+Z to undo / redo typed text in comments doesn't work.

chenilim:
Note: The issue is probably that the card keyboard handler is intercepting undo / redo keys even when the comment is being edited.

kyeongsoosoo:
@chenilim  Can I take this issue? 
I took a look and it seems that it's a bug from `markdownEditorInput`.
Maybe it's related to #2814, too.

sbishel:
Thanks @kyeongsoosoo - all yours.

PR #2852, opened 2022-04-20
GH-2520 make cut/undo/redo work in comments 

kyeongsoosoo:
#### Summary

1. I removed `useEffect` hook in `markdownEditorInput` which blocked the key command. 
I checked that mention plugin works and tests pass after deleting the hook. If I have to test something else, please let me know.
2. When we press redo/undo key command, handling logic is executed. 
3. `markdownEditorInput` is rendered only when `isEditing` state is true, so `editorState` can be re-initialized when `isEditing` state is changed.
4. added a cypress test.

#### Ticket Link

#2520


webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
```diff
@@ -7,7 +7,7 @@ import createMentionPlugin from '@draft-js-plugins/mention'
 import '@draft-js-plugins/mention/lib/plugin.css'
 import {ContentState, DraftHandleValue, EditorState, getDefaultKeyBinding} from 'draft-js'
 import React, {
-    ReactElement, useCallback, useEffect,
+    ReactElement, useCallback,
```
sbishel, line 10:
```suggestion
    ReactElement, useCallback, useEffect,
```

sbishel (approved):

Issue #2850, opened 2022-04-20
Bug: bad format on JSONB fields for postgres with binary_parameters=true


mgdelacroix:
## Steps to reproduce the behavior

1. Run a mattermost server with Postgres and `binary_parameters=true`
2. Install the focalboard plugin
3. The store throws an error when trying to save registries in tables with JSONB fields

## Additional context (optional)

We need to check for `binary_parameters=true` on the database DSN and append the 0x01 byte if present.

* Related PR on server: https://github.com/mattermost/mattermost-server/pull/19898

PR #2854, opened 2022-04-20
Handle jsonb fields on postgres with binary_parameters set

mgdelacroix:
#### Summary
This PR sets a flag on the store marking if it has the `binary_parameters` set and creates a utility function that should be used to marshal the contents of the `jsonb` fields and will append the `0x01` byte to them if required as per the spec.

As a follow up, I will abstract these methods in the server code and import them directly to avoid code repetition.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/2850

### DIFF
--- a/server/services/store/sqlstore/sqlstore.go
+++ b/server/services/store/sqlstore/sqlstore.go
@@ -2,6 +2,7 @@ package sqlstore
 
 import (
 	"database/sql"
+	"net/url"
 
 	"github.com/mattermost/mattermost-server/v6/plugin"
 
@@ -24,6 +25,7 @@ type SQLStore struct {
 	logger           *mlog.Logger
 	NewMutexFn       MutexFactory
 	pluginAPI        *plugin.API
+	isBinaryParam    bool
 }
 
 // MutexFactory is used by the store in plugin mode to generate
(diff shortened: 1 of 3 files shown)


server/services/store/sqlstore/sqlstore.go
```diff
@@ -49,7 +51,14 @@ func New(params Params) (*SQLStore, error) {
...
 
-	err := store.Migrate()
+	var err error
+	store.isBinaryParam, err = store.computeBinaryParam()
+	if err != nil {
+		params.Logger.Error(`Cannot compute binary parameter`, mlog.Err(err))
+		return nil, err
+	}
+
+	err = store.Migrate()
```
wiggin77, line 61: What happens if the binary flag is added to an existing database?

wiggin77 (approved):
LGTM :+1:   One non-blocking question.

server/services/store/sqlstore/sqlstore.go
mgdelacroix: AFAIK, the `binary_parameters` flag controls the way the driver communicates with the DB, but does not affect the final contents of the table, so it should work correctly.

Tested locally to confirm, but pinging @agnivade in case I'm missing anything.

agnivade: Yep, what Miguel said. `binary_parameters` just determines the client driver protocol. How postgres actuall stores the data is always the same.

agnivade (approved):

jespino (approved):
LGTM

Issue #2813, opened 2022-04-14
Bug: `<Editable>` has an `undefined` class if `className` isn't defined


Pinjasaur:
## Steps to reproduce the behavior

1. Go to a table view
2. Inspect one of the `<Editable>` components e.g. `<input class="Editable">` in the DOM
3. Note the addition of a literal `undefined` to the class e.g. `<input class="Editable undefined">`

## Expected behavior

Don't have a literal `undefined` end up in the class.

## Screenshots (optional)

![image](https://user-images.githubusercontent.com/6335792/163449246-b1045b9f-8013-4f5d-93cc-a3d4c4d88e08.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16.0 (bugbash)
 - Browser and OS: Firefox 100 on MacOS 12.3.1 (M1)

## Additional context (optional)

Sev 3: low importance.

Likely can be resolved with the addition of a check for the `className` prop in the `editable.tsx` component, similar to existing `readonly` prop. https://github.com/mattermost/focalboard/blob/923437cc573ca515e17c67b56b09436ae9014e0c/webapp/src/widgets/editable.tsx#L90

Rajat-Dabade:
Fixed
![Screenshot 2022-04-20 at 7 32 45 PM](https://user-images.githubusercontent.com/26198062/164251651-49456a61-bb58-4a17-add7-d9f326c25716.png)

PR #2855, opened 2022-04-20
Fix for undefined className prop issue in Editable component

Rajat-Dabade:
#### Summary

This PR is set the className prop with the default value as an empty string for all the Editable component references where this prop was not passed.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/2813


webapp/src/components/viewHeader/emptyCardButton.scss
```diff
@@ -0,0 +1,22 @@
+@-moz-document url-prefix('') {
```
Pinjasaur, line 1: It looks like this file change is duplicated from #2849, was that the intention?

webapp/src/components/content/checkboxElement.tsx
```diff
@@ -86,6 +86,7 @@ const CheckboxElement = (props: Props) => {
                 }}
                 readonly={readonly}
                 spellCheck={true}
+                className=''
```
Pinjasaur, line 89: alternatively, could check for an `undefined` `className` prop in the `Editable` component itself to avoid needing to pass in am empty string literal (similar to how `readonly` is handled https://github.com/mattermost/focalboard/blob/923437cc573ca515e17c67b56b09436ae9014e0c/webapp/src/widgets/editable.tsx#L90)

the primary benefit being we could wholly omit the `className` prop if it's not needed for a component

Pinjasaur (changes requested):
nice work! have 2 questions I typed as individual comments

webapp/src/components/viewHeader/emptyCardButton.scss
Rajat-Dabade: My bad, I have to checkout this branch from the branch I was working on rather than the main, I will make the changes and push it. Thank you.

webapp/src/components/content/checkboxElement.tsx
Rajat-Dabade: Yes, this can be the alternate and best solution. Thank you @Pinjasaur, I will make the changes and push it as you suggested.

jespino (approved):
LGTM

jespino:
Thanks @Rajat-Dabade! :tada:

PR #2882, opened 2022-04-21
Adding the permissions for commenter + viewer roles

jespino:
Adding the new permissions for add comments in the board. And showing the commenter role there.


chenilim:
@jespino, is the intention to check this in after the v0.16 branch? If so, the changes look good.

chenilim:
Thanks @jespino! I realized we may also need to check this for Delete Block, because a user can currently delete their own comment. In fact, we might want to consider refactoring that for comments, or even having a separate set of APIs for comments. That might also make it easier to refactor later, e.g. when we move to using channel threads. I'm out next week, so looping in @wiggin77 as an extra pair of eyes.

wiggin77:
> or even having a separate set of APIs for comments.

I'll review.  In the meantime, +1 for dedicated API for comments.  Will make it much easier to change how comments are stored later.

server/api/api.go
```diff
@@ -3021,9 +3035,16 @@ func (a *API) handleDuplicateBlock(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !a.permissions.HasPermissionToBoard(userID, boardID, model.PermissionManageBoardCards) {
-		a.errorResponse(w, r.URL.Path, http.StatusForbidden, "", PermissionError{"access denied to modify board members"})
-		return
+	if block.Type == model.TypeComment {
+		if !a.permissions.HasPermissionToBoard(userID, boardID, model.PermissionCommentBoardCards) {
+			a.errorResponse(w, r.URL.Path, http.StatusForbidden, "", PermissionError{"access denied to comment board cards"})
```
wiggin77, line 3040:
```suggestion
			a.errorResponse(w, r.URL.Path, http.StatusForbidden, "", PermissionError{"access denied to comment on board cards"})
```

server/api/api.go
```diff
@@ -3021,9 +3035,16 @@ func (a *API) handleDuplicateBlock(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !a.permissions.HasPermissionToBoard(userID, boardID, model.PermissionManageBoardCards) {
-		a.errorResponse(w, r.URL.Path, http.StatusForbidden, "", PermissionError{"access denied to modify board members"})
-		return
+	if block.Type == model.TypeComment {
```
wiggin77, line 3038: How would a user duplicate a comment unless they are duplicating a board or card?  Does this permission really apply to that case?

wiggin77:
> we may also need to check this for Delete Block, because a user can currently delete their own comment.

It would appear any user with board access can delete anyone else's comments via the DeleteBlock API.

server/api/api.go
jespino: From the API perspective, I think it doesn't matter, if you can duplicate a block, we should check that permission.

Pinjasaur:
> > we may also need to check this for Delete Block, because a user can currently delete their own comment.
> 
> It would appear any user with board access can delete anyone else's comments via the DeleteBlock API.

@wuwinson I chatted about this with @wiggin77 and we're good with creating a follow-up card to add logic to DeleteBlock since a user with board access can delete any comment via the API _directly_.

Pinjasaur:
@wuwinson I chatted about this with @jespino

> Yes, that makes a lot of sense, actually, the comment should be handled in a specific way there, in my opinion, you shouldn't be able to delete comments in general, doesn't matter if you are an editor
> 
> I would expect that maybe, and admin, can delete comments
> 
> but not other editor, for example

additionally, I was reviewing the [spec](https://docs.google.com/document/d/1hQAjl-4FFaEVCVYtU2BtyGxA97hcnNxCBDSnpnDT_cw/edit#heading=h.y3zb05cfwcw2) more and it makes mention of being able to _edit_ comments which I didn't realize initially (editing doesn't exist right now)

wuwinson:
> @wuwinson I chatted about this with @jespino
> 
> > Yes, that makes a lot of sense, actually, the comment should be handled in a specific way there, in my opinion, you shouldn't be able to delete comments in general, doesn't matter if you are an editor
> > I would expect that maybe, and admin, can delete comments
> > but not other editor, for example
> 
> additionally, I was reviewing the [spec](https://docs.google.com/document/d/1hQAjl-4FFaEVCVYtU2BtyGxA97hcnNxCBDSnpnDT_cw/edit#heading=h.y3zb05cfwcw2) more and it makes mention of being able to _edit_ comments which I didn't realize initially (editing doesn't exist right now)

@Pinjasaur Good catch! I updated the spec to remove the portion about editing comments since it's not supported yet. However, Commenters should be able to delete their own comments, but not others. This behavior would be consistent with Channels as well.

wiggin77 (approved):
LGTM :+1:

Pinjasaur (approved):
@wiggin77 thanks for the approval, could you give another quick once-over since I added a new permission `DeleteOthersComments` to implement the logic for only Admins being able to delete _all_ comments. if it all looks good, go ahead & merge :)

@ogi-m FYI that this will be merged soon—we can test it during the bugbash later this week (or if you're able to QA it prior, that's excellent too)

Pinjasaur:
Momentary hold to remove the `Commenter` role from a custom template.

PR #2884, opened 2022-04-21
Adding the board default role for public boards

jespino:


server/integrationtests/local_permissions_test.go
```diff
@@ -235,6 +235,90 @@ func TestLocalPermissionsPatchBoard(t *testing.T) {
 	runTestCases(t, ttCases, testData, clients)
 }
 
+func TestLocalPermissionsPatchBoardType(t *testing.T) {
```
mgdelacroix, line 238: The more I see the simplicity of this testing mechanism, the more I like it :)

server/integrationtests/board_test.go
```diff
@@ -1945,7 +1945,82 @@ func TestJoinBoard(t *testing.T) {
...
+		board, resp := th.Client.CreateBoard(newBoard)
+		th.CheckOK(resp)
+		require.NoError(t, resp.Error)
+		require.NotNil(t, board)
+		require.NotNil(t, board.ID)
+		require.Equal(t, title, board.Title)
+		require.Equal(t, model.BoardTypeOpen, board.Type)
+		require.Equal(t, teamID, board.TeamID)
+		require.Equal(t, me.ID, board.CreatedBy)
+		require.Equal(t, me.ID, board.ModifiedBy)
```
mgdelacroix, line 1964: Could we add a check on the default role, even if it's to assert that it's empty?

server/integrationtests/permissions_test.go
```diff
@@ -448,6 +448,90 @@ func TestPermissionsPatchBoard(t *testing.T) {
...
+}
+
+func TestPermissionsPatchBoardDefaultRole(t *testing.T) {
+	th := SetupTestHelperPluginMode(t)
+	defer th.TearDown()
+	clients := setupClients(th)
+	testData := setupData(t, th)
+
+	ttCases := []TestCase{
+		{"/boards/{PRIVATE_BOARD_ID}", methodPatch, "{\"defaultRole\": \"vieiwer\"}", userAnon, http.StatusUnauthorized, 0},
```
mgdelacroix, line 500: Same `vieiwer` typo

server/integrationtests/local_permissions_test.go
```diff
@@ -235,6 +235,90 @@ func TestLocalPermissionsPatchBoard(t *testing.T) {
...
+}
+
+func TestLocalPermissionsPatchBoardDefaultRole(t *testing.T) {
+	th := SetupTestHelperLocalMode(t)
+	defer th.TearDown()
+	clients := setupLocalClients(th)
+	testData := setupData(t, th)
+
+	ttCases := []TestCase{
+		{"/boards/{PRIVATE_BOARD_ID}", methodPatch, "{\"defaultRole\": \"vieiwer\"}", userAnon, http.StatusUnauthorized, 0},
```
mgdelacroix, line 287: typo: `s/vieiwer/viewer`, on this and all following lines

if the patch was being accepted, we should update the validation to prevent it

sbishel (commented):
I would suggest creating an enum for the role strings and use them instead of hard-coded strings. Could prevent bugs caused by typos.

mgdelacroix (approved):
LGTM! :+1:

jespino:
@mgdelacroix @sbishel I changed the approach to "minimum role". PTAL

sbishel (approved):

PR #2900, opened 2022-04-21
fix: personal server invite link omitted base URL

### LINKED ISSUE #2898: Bug: Personal Server > User Invite link doesn't include site url

1. Log in to Personal Server
2. Click on Focalboard in the top left
3. Click Invite Users
4. Observe the url - it's missing the site url and only shows `/register?t=[token]`

## Expected behavior

Full invite link should be shown

## Screenshots (optional)

image

## Edition and Platform

 - Edition: Personal Server 
 - Version: v0.16.0

Pinjasaur:
#### Summary

The Personal Server 'invite users' URL wasn't showing the base URL, just the `/register?t=xxx`. It looks like there was a change during the permissions branch that updated the behavior of `Utils.buildURL` to return _only_ the `path` argument if it's _not_ a plugin instance, even if the `absolute` arg is present. I followed the existing conventions to use the builtin `URL` type + constructor to create a properly formatted URL string.

image

#### Ticket Link

closes #2898

### DIFF
--- a/webapp/src/components/sidebar/registrationLink.tsx
+++ b/webapp/src/components/sidebar/registrationLink.tsx
@@ -44,7 +44,7 @@ const RegistrationLink = (props: Props) => {
         }
     }
 
-    const registrationUrl = new URL(Utils.buildURL('/register?t=' + signupToken), window.location.toString()).toString()
+    const registrationUrl = `${Utils.getBaseURL(true).replace(/\/$/, '')}/register?t=${signupToken}`
 
     return (
         <Modal
--- a/webapp/src/components/sidebar/registrationLink.tsx
+++ b/webapp/src/components/sidebar/registrationLink.tsx
@@ -44,7 +44,7 @@ const RegistrationLink = (props: Props) => {
         }
     }
 
-    const registrationUrl = Utils.buildURL('/register?t=' + signupToken, true)
+    const registrationUrl = new URL(Utils.buildURL('/register?t=' + signupToken), window.location.toString()).toString()
 
     return (
         <Modal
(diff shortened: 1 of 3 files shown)


webapp/src/components/sidebar/registrationLink.tsx
```diff
@@ -44,7 +44,7 @@ const RegistrationLink = (props: Props) => {
         }
     }
 
-    const registrationUrl = Utils.buildURL('/register?t=' + signupToken, true)
```
jespino, line 47: This looks really wrong, we should fix this behavior in the buildURL function or create another function to build the absolute url. Clearly, having an "absolute" parameter that doesn't work in certain context is totally missleading.

Pinjasaur: Agreed, I created #2901 as a code debt ticket to track that specifically

jespino: but I don't think it should be a different ticket, I think we should fix it properly right away, instead of fixing it wrongly here, and fix the original function, and probably miss to fix it here.

sbishel:
@Pinjasaur I also assigned you https://github.com/mattermost/focalboard/issues/2901.

Pinjasaur:
Discussing this with @sbishel and realized that there's more nuance to the problem related to subpaths i.e. if the `config.json` `serverRoot` key has some path e.g. `http://localhost:8000/focalboard` the current solution won't work.

Pinjasaur:
Changed to use the `Utils.getBaseURL` function which will respect a `baseURL` (if present) when building the invite URL.

sbishel (changes requested):
The code itself looks fine. But let's go ahead and add a unit test for this file.

sbishel (approved):
It doesn't really hurt anything, but you don't need to create so much state for this test. I went ahead and remove the unnecessary lines, because I want to merge. But wanted you to be aware.

Issue #2876, opened 2022-04-21
Bug: board members are not being created when updating to 0.16 on personal desktop


mgdelacroix:
## Steps to reproduce the behavior

1. Create some data on a 0.15 personal desktop boards install
2. Update to the current main
3. See how all the boards have been migrated, but no memberships were created for them, so they don't appear on the interface

## Expected behavior

Memberships for all the boards should have been created during the migration.

## Edition and Platform

 - Edition: Personal Desktop
 - Version: v0.16.0
 - Browser and OS: Firefox 99.0.1 on Linux

PR #2906, opened 2022-04-22
Initialise the store with a single user flag and pass it to the migrations

mgdelacroix:
#### Summary
This PR initialises the store with a `singleUser` parameter that is in the end passed to the migrations. It also modifies the board members initial migration to operate differently if running in single user mode.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/2876

### DIFF
--- a/linux/main.go
+++ b/linux/main.go
@@ -61,7 +61,8 @@ func runServer(port int) (*server.Server, error) {
 		AuthMode:                "native",
 	}
 
-	db, err := server.NewStore(config, logger)
+	singleUser := len(sessionToken) > 0
+	db, err := server.NewStore(config, singleUser, logger)
 	if err != nil {
 		fmt.Println("ERROR INITIALIZING THE SERVER STORE", err)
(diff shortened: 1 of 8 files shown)


linux/main.go
```diff
@@ -61,7 +61,8 @@ func runServer(port int) (*server.Server, error) {
 		AuthMode:                "native",
 	}
 
-	db, err := server.NewStore(config, logger)
+	singleUser := len(sessionToken) > 0
```
jespino, line 64: This makes no sense, it is going to be always true in this context.

jespino: maybe it makes sense to keep consistency with everything else

jespino (approved):
LGTM

linux/main.go
mgdelacroix: That was my reasoning, but I'm happy hardcoding `true` in this context as well :)

harshilsharma63 (approved):

Issue #2894, opened 2022-04-21
Bug: Shared boards URL do not contain read token


sbishel:
## Steps to reproduce the behavior

1. Go to Share
2. Select  Publish
3. Turn on Publish
4. Copy URL

## Expected behavior

URL should work in incognito, but doesn't because the read token is empty.

## Screenshots (optional)

https://bugbash-boards2.test.mattermost.cloud/plugins/focalboard/team/65tdp6qhxbdiicjr6y7yg47xdo/shared/b4om8q3he4tfkjjo963sbur1y6h/vtbwxirqm6jntpp3tngxsbus9zo?r=

## Edition and Platform

 - Edition:  Mattermost Boards (plugin)
 - Version: v0.16.0
 - Browser and OS:  Chrome on Mac

PR #2909, opened 2022-04-22
GH-2894- Fix missing readonly token

sbishel:
#### Summary
An update was made to reduce logging, so when a Sharing object isn't found, an empty object is returned rather than an error. However, the code that creates the token is dependent on the Sharing object being `undefined`, not an empty object.

This PR updates the return to return nothing, instead of an empty object, so the Sharing object is `undefined`.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2894


jespino (approved):
LGTM

Issue #2895, opened 2022-04-21
Bug: User does not auto-follow cards created from template 


wuwinson:
## Steps to reproduce the behavior

1. Go to template picker
2. Select a template
3. Open a card
4. See error - User is not following any cards created from the template. However, creating new cards on the board does properly auto-follow the cards.

## Expected behavior

User who loaded the template should auto follow all cards on the template. Confirmed that `Follow` button is working, and issue is not a matter of the button not reflecting the proper state.

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/164540154-e8bc90c0-2752-43d8-a5a8-cf7d0b3638d3.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16 Bugbash-2
 - Browser and OS: Chrome on Mac

wiggin77:
I wanted to check this out sooner than later to ensure this lack of auto-follow was not part of a bigger issue.  Turns out no board duplication was auto-subscribing the author because the notifications engine was not called for `DuplicateBoard`.

PR #2910, opened 2022-04-22
GH-2895  Call notification engine when duplicating boards

wiggin77:
#### Summary
This PR fixes an issue where authors are not auto-subscribed to cards when duplicating boards (from template or regular board).

This PR also ensures that boards app APIs use the same notifications thread pool as blocks.

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/2895

### DIFF
--- a/server/app/app.go
+++ b/server/app/app.go
@@ -19,7 +19,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


sbishel (approved):
LGTM

Issue #2896, opened 2022-04-21
Bug: Read Only server displays without read token


sbishel:
## Steps to reproduce the behavior

Follow up on #2894 
1. Go to Shared Link in Incognito mode without the read only token
https://bugbash-boards2.test.mattermost.cloud/plugins/focalboard/team/65tdp6qhxbdiicjr6y7yg47xdo/shared/b4om8q3he4tfkjjo963sbur1y6h/vtbwxirqm6jntpp3tngxsbus9zo?r=
2. The Add Board is displayed...partially

## Expected behavior

With no read token or invalid read token, unauthenticated user should be taken to MM login screen. Double check to see if it should redirect to error page. cc @wuwinson 

## Screenshots (optional)
![Screen Shot 2022-04-21 at 1 39 30 PM](https://user-images.githubusercontent.com/12704875/164540594-ca05b3d2-45a5-45a4-9e4a-e8d9335878a9.png)

## Edition and Platform

 - Edition:  Mattermost Boards (plugin)
 - Version:  v0.16.0]
 - Browser and OS: Chrome on Mac

## Additional context (optional)

Add any other context about the problem here, and any notes about the severity:
* Sev 1: Affects critical functionality without a workaround
* Sev 2: Affects major functionality with a difficult or non-obvious workaround
* Sev 3: Affects minor, non-critical functionality

PR #2911, opened 2022-04-22
GH-2896 - Error on invalid token

sbishel:
#### Summary
If a read-only token is invalid, an unauthorized user is left on the "Create Board" screen. This PR creates and handles an error when a read-only board is not returned. (ie. an invalid read-only token). 

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2896

#### ScreenShot
Screen Shot 2022-04-22 at 10 31 22 AM

### DIFF
--- a/webapp/src/errors.ts
+++ b/webapp/src/errors.ts
@@ -84,7 +84,9 @@ function errorDefFromId(id: ErrorId | null): ErrorDef {
         errDef.title = intl.formatMessage({id: 'error.invalid-read-only-board', defaultMessage: 'You don\’t have access to this board. Log in to access Boards.'})
         errDef.button1Enabled = true
         errDef.button1Text = intl.formatMessage({id: 'error.go-login', defaultMessage: 'Login'})
-        errDef.button1Redirect = window.location.origin + '/login'
+        errDef.button1Redirect = (): string => {
+            return window.location.origin
+        }
         errDef.button1Fill = true
         break
     }
--- a/webapp/src/errors.ts
+++ b/webapp/src/errors.ts
@@ -81,7 +81,7 @@ function errorDefFromId(id: ErrorId | null): ErrorDef {
         break
     }
     case ErrorId.InvalidReadOnlyBoard: {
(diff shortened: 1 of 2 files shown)


webapp/src/errors.ts
```diff
@@ -79,6 +80,14 @@ function errorDefFromId(id: ErrorId | null): ErrorDef {
         errDef.button1Fill = true
         break
     }
+    case ErrorId.InvalidReadOnlyBoard: {
+        errDef.title = intl.formatMessage({id: 'error.invalid-read-only-board', defaultMessage: 'You don\’t have access to this board. Log in to access Boards.'})
+        errDef.button1Enabled = true
+        errDef.button1Text = intl.formatMessage({id: 'error.go-login', defaultMessage: 'Login'})
+        errDef.button1Redirect = window.location.origin + '/login'
```
wiggin77, line 87: After login should they go to the board/card they were trying to access (like `ErrorId.NotLoggedIn`) ?

wiggin77: ... also is the `window.location.origin` needed, versus returning `/login` ?

sbishel: > After login should they go to the board/card they were trying to access (like `ErrorId.NotLoggedIn`) ?

Not sure, but my thought is no. The board they are going to is a read-only board (and may not exist or may be a bad (missing) token). We don't get back details about the error. Just that board wasn't retrieved. 

But, I can add it and try to send them back. I'll have to see how it works with an invalid board. I

sbishel: > ... also is the `window.location.origin` needed, versus returning `/login` ?

Yes, this was intention. `/login` will take them to `/plugins/focalboard/login`. So this was intentional. It is the same as what `ErrorId.TeamUndefined` does above. 

However, I need to test against Personal Server. (Forgot that one.)

wiggin77 (approved):
LGTM :+1:

webapp/src/errors.ts
sbishel: Works on Personal Server as well.

PR #2914, opened 2022-04-22
Add Go code coverage report to CI

chenilim:
#### Summary
Add go code coverage output for `make server-test-sqlite`, which can be seen in the CI job logs. Current coverage is 52%.

After running `make server-test-sqlite` locally, run `cd server; go tool cover -html=test-sqlite-profile.coverage` to view the detailed code coverage report.


Makefile
```diff
@@ -118,7 +118,8 @@ server-test: server-test-sqlite server-test-mysql server-test-postgres ## Run se
 server-test-sqlite: export FB_UNIT_TESTING=1
 
 server-test-sqlite: templates-archive ## Run server tests using sqlite
-	cd server; go test -tags '$(BUILD_TAGS)' -race -v -count=1 -timeout=30m ./...
+	cd server; go test -tags '$(BUILD_TAGS)' -race -v -coverpkg=./... -coverprofile=test-sqlite-profile.coverage -count=1 -timeout=30m ./...
```
wiggin77, line 121: Shouldn't coverage run against one of the plugin databases?

chenilim: Good question. I figured the coverage wouldn't be that different, but there's also no harm in adding coverage for all the tests, so I've updated that. Turns out the coverage for mysql and postgres are 54%, so those are the ones we should track.

wiggin77 (approved):
LGTM :+1:

Issue #2890, opened 2022-04-21
Bug: Welcome board has no images


chenilim:
## Steps to reproduce the behavior

1. Register a new account and login
2. Select "Take the tour" on the initial welcome page
3. The Welcome board opens
4. Open the first card
5. There is not image attached to the card (nor on any of the other cards)

## Expected behavior

The welcome board's cards should have images

## Screenshots (optional)

<img width="823" alt="Screen Shot 2022-04-21 at 12 17 48 PM" src="https://user-images.githubusercontent.com/46905241/164536997-f11bf6dc-4e4c-44b8-8e85-9867e93f183e.png">

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16 (bugbash2)
 - Browser and OS: Safari and Chrome on Mac

## Additional context (optional)

If you create a board from the welcome board template, there are no images either.

wuwinson:
Try to repro on Personal Server first to confirm if its a connectivity issue.

wiggin77:
This may be more than a S3 connectivity issue... while testing a PR I noticed these in my local logs after creating a board from template:
```
error [2022-04-22 11:56:33.209 -04:00] API ERROR                                     caller="app/plugin_api.go:940" plugin_id=focalboard code=500 error=""unable to open file wnnbqiiw13d7xnh3omhoqrdw3r/bn9oykhxmdb8b7eeip4cm1t7dto/77pe9r4ckbin438ph3f18bpatua.png: open data/wnnbqiiw13d7xnh3omhoqrdw3r/bn9oykhxmdb8b7eeip4cm1t7dto/77pe9r4ckbin438ph3f18bpatua.png: no such file or directory"" msg= api=/api/v2/files/teams/wnnbqiiw13d7xnh3omhoqrdw3r/bn9oykhxmdb8b7eeip4cm1t7dto/77pe9r4ckbin438ph3f18bpatua.png
debug [2022-04-22 11:56:33.307 -04:00] attachSession                                 caller="app/plugin_api.go:934" plugin_id=focalboard single_user=false
[...]

PR #2915, opened 2022-04-22
GH-2890  Copy files when duplicating boards

wiggin77:
#### Summary
This PR fixes the issue where duplicating boards does not copy files (images).  The "copy images" code was left out when the way boards are duplicated was refactored (permissions branch).

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/2890


server/app/boards.go
```diff
@@ -145,6 +147,38 @@ func (a *App) DuplicateBoard(boardID, userID, toTeam string, asTemplate bool) (*
 	if err != nil {
 		return nil, nil, err
 	}
+
+	// copy any file attachments from the duplicated blocks.
+	if err = a.CopyCardFiles(boardID, bab.Blocks); err != nil {
+		return nil, nil, fmt.Errorf("could not copy files while duplicating board %s: %w", boardID, err)
+	}
```
mgdelacroix, line 154: In case of an error copying the files, should we delete the incomplete board and its blocks? Maybe something like:

```suggestion
	if err = a.CopyCardFiles(boardID, bab.Blocks); err != nil {
		if err := a.store.DeleteBoardsAndBlocks(bab, userID); err != nil {
			a.logger.Error("Cannot delete board after duplication error when copying block's files", mlog.String("boardID", bab.Boards[0].ID), mlog.Err(err)) 
		}
		return nil, nil, fmt.Errorf("could not copy files while duplicating board %s: %w", boardID, err)
	}
```

server/app/boards.go
```diff
@@ -145,6 +147,38 @@ func (a *App) DuplicateBoard(boardID, userID, toTeam string, asTemplate bool) (*
...
+	a.logger.Debug("Duplicate boards patching file IDs", mlog.Int("count", len(blockIDs)))
+
+	if len(blockIDs) != 0 {
+		patches := &model.BlockPatchBatch{
+			BlockIDs:     blockIDs,
+			BlockPatches: blockPatches,
+		}
+		if err = a.store.PatchBlocks(patches, userID); err != nil {
+			return nil, nil, fmt.Errorf("could not patch file IDs while duplicating board %s: %w", boardID, err)
+		}
```
mgdelacroix, line 179: Same here:

```suggestion
		if err = a.store.PatchBlocks(patches, userID); err != nil {
			if err := a.store.DeleteBoardsAndBlocks(bab, userID); err != nil {
				a.logger.Error("Cannot delete board after duplication error when updating block's file info", mlog.String("boardID", bab.Boards[0].ID), mlog.Err(err)) 
			}
			return nil, nil, fmt.Errorf("could not patch file IDs while duplicating board %s: %w", boardID, err)
		}
```

mgdelacroix (approved):
LGTM! :+1: A couple of questions left

mgdelacroix (approved):

sbishel (approved):

Issue #2874, opened 2022-04-21
Bug: Last update by and created by fields have different representation in the board cards


jespino:
## Steps to reproduce the behavior

In the board cards and gallery cards the representation of the created by and last modified by has different representaiton. One shows the user with the avatar, the other don't. It should be consistent

## Expected behavior

Consistent user representation in the cards

## Screenshots (optional)

![image](https://user-images.githubusercontent.com/290303/164416835-7d02507d-ccab-4295-bc17-a7ba9d828b8f.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.17.0
 - Browser and OS: Firefox/Linux

jespino:
@asaadmahmood can you confirm what do you want for the correct behavior? My guess is the user representation with the avatar.

asaadmahmood:
You are right @jespino We should have the avatar.

PR #2937, opened 2022-04-27
Last updated by aligement fixed

Rajat-Dabade:
#### Summary
The PR fixes the alignment of the last updated field in the card view.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/2874

### DIFF
--- a/webapp/src/components/properties/lastModifiedBy/lastModifiedBy.tsx
+++ b/webapp/src/components/properties/lastModifiedBy/lastModifiedBy.tsx
@@ -20,7 +20,6 @@ type Props = {
 }
 
 const LastModifiedBy = (props: Props): JSX.Element => {
-    const boardUsersById = useAppSelector<{[key:string]: IUser}>(getBoardUsers)
     const lastContent = useAppSelector(getLastCardContent(props.card.id || '')) as Block
     const lastComment = useAppSelector(getLastCardComment(props.card.id)) as Block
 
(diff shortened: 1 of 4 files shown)


jespino (changes requested):
Why the snapshot is not longer showing the username?

sbishel (changes requested):

jespino (approved):
Now looks good to me

webapp/src/components/properties/lastModifiedBy/lastModifiedBy.tsx
```diff
@@ -32,8 +30,12 @@ const LastModifiedBy = (props: Props): JSX.Element => {
     }
 
     return (
-        <div className='LastModifiedBy octo-propertyvalue readonly'>
-            {(boardUsersById && boardUsersById[latestBlock.modifiedBy]?.username) || latestBlock.modifiedBy}
+        <div>
```
jespino, line 33: Is there any reason to have a container `div` here? Looks like not needed unless you want to add something like `LastModifiedBy` class. Actually, you are importing the `lastModifiedBy.scss` but you are not using it, because the class is no longer available.

jespino (changes requested):
I would provide the className `LastModifiedBy` to be caonsistent with our components class names approach, and ensure that the CSS in the `lastModifiedBy.scss` is needed and correct, if not, we can just remove the file not import it.

Rajat-Dabade:
@jespino, Thank you for your review, just tested it, and it looks like we don't need `lastModifiedBy.scss`. Have removed it.

jespino (commented):
LGTM

jespino (approved):

sbishel (approved):

Issue #2863, opened 2022-04-21
Bug: Pressing enter should create a category


asaadmahmood:
## Steps to reproduce the behavior

1. Try to create a new category
2. Category modal opens up
3. Type a name, and press enter
4. Does not create a category, you have to manually press the create button

## Expected behavior
Pressing enter after you've typed a name should create a category.

## Screenshots (optional)
<img width="656" alt="Screenshot 2022-04-21 at 1 23 17 PM" src="https://user-images.githubusercontent.com/11034289/164412751-511f83a6-cb50-45c0-9bf3-185e45993970.png">

## Edition and Platform
 - Edition: Mattermost Boards (plugin)
 - Version: 0.16
 - Browser and OS: Chrome on Mac

PR #2940, opened 2022-04-27
Feature added: Pressing enter should create a category

Rajat-Dabade:
#### Summary

This PR adds a feature of creating categories by hitting the enter button inside Category model

#### Ticket Link
Fixes: https://github.com/mattermost/focalboard/issues/2863

### DIFF
--- a/webapp/src/components/createCategory/createCategory.tsx
+++ b/webapp/src/components/createCategory/createCategory.tsx
@@ -1,9 +1,9 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, { useState } from 'react'
+import React, {useState, KeyboardEvent} from 'react'
 
-import { useIntl } from 'react-intl'
+import {useIntl} from 'react-intl'
 
 import Dialog from '../dialog'
 import Button from '../../widgets/buttons/button'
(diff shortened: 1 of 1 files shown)


webapp/src/components/createCategory/createCategory.tsx
```diff
@@ -32,35 +32,38 @@ const CreateCategory = (props: Props): JSX.Element => {
...
-                    <Button
-                        size={'medium'}
-                        filled={Boolean(name)}
-                        onClick={() => props.onCreate(name)}
-                        disabled={!name}
-                    >
-                        {props.initialValue ? updateText : createText}
-                    </Button>
+            <form>
+                <div className='CreateCategory'>
```
harshilsharma63, line 36: We follow the convention of having the component's name as a classname on the outermost HTML element being returned. So to continue following it, you can either set the classname on the `form` or wrap everything inside the `<div className='CreateCategory'>` in a form. Up to you.

harshilsharma63:
A few issues-

1. I'm on Firefox and pressing enter reloads the page for me. This reload shouldn't happen. The category should be created the dialog box should close instead-

https://user-images.githubusercontent.com/18575143/165685637-a79ab094-6f1d-4e5d-9f44-d34f3f4084d6.mov

2. The "Create" button's styling when the input field is empty looks odd. It's probably disabled or something. It happens on `main` as well, so not caused by this PR. @asaadmahmood any thoughts on this or does it look fine to you?
Screenshot 2022-04-28 at 11 03 55 AM

3. The vertical position of the "cancel" and "create" buttons seem to have been changed unintentionally.

before

after

webapp/src/components/createCategory/createCategory.tsx
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useState} from 'react'
+import React, { useState } from 'react'
```
harshilsharma63, line 4: Were these changes caused by the linter? If not, we should revert them.

harshilsharma63 (changes requested):

webapp/src/components/createCategory/createCategory.tsx
Rajat-Dabade: Thanks for the review @harshilsharma63, sure I will fix it.

webapp/src/components/createCategory/createCategory.tsx
Rajat-Dabade: This was because of my mistake, I have used vscode built-in linter, I will revert this back. Thank you.

Rajat-Dabade:
Thank you @harshilsharma63 for the review. Yes, I can see that the page is getting refreshed every time the enter button is hit. This was because of the built-in HTML form element. I have made the changes according. Now I can see the page is not refreshing and the alignment of the button for cancel and create are also in the proper position. Please review it from your end.

webapp/src/components/createCategory/createCategory.tsx
```diff
@@ -56,6 +63,7 @@ const CreateCategory = (props: Props): JSX.Element => {
                         filled={Boolean(name)}
                         onClick={() => props.onCreate(name)}
                         disabled={!name}
+                        submit={true}
```
harshilsharma63, line 66: We don't need this as we're no longer using forms.

harshilsharma63 (approved):
LGTM 👍 Only one minor change and we're good to merge!

Issue #2941, opened 2022-04-27
Bug: DMs and GMs don't have a team_id after the database migrations


mgdelacroix:
## Steps to reproduce the behavior

1. Install the plugin version 0.15 on a server
2. Create boards for DMs and GMs
3. Update plugin to version 0.16
4. Boards for DMs and GMs would not appear on the interface.

## Expected behavior

Boards should be correctly migrated.

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16.0

PR #2928, opened 2022-04-26
Add migration to ensure all boards are either open or private

mgdelacroix:
#### Summary
This PR adds a migration to take boards that come from either group channels or direct message channels and make them private boards.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/2923

PR #2942, opened 2022-04-27
Move teamless data migration with the rest of data migrations and update teamless boards query

mgdelacroix:
#### Summary
The teamless boards data migration was relying on boards of type `D` and `G`, that were removed as part of https://github.com/mattermost/focalboard/pull/2928. This PR reverts said patch, fixes the query that fetches teamless boards and cleans a bit the data migration code, moving the migration itself to the `data_migrations.go` file and ensuring migration and schema version keys are consistently named and placed.

The actual fix for the teamless data migration in MySQL is https://github.com/mattermost/focalboard/pull/2942/commits/5c5a2bb8988cb7b781ec39754352d87482682db5

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/2941


jespino:
Can you separate the fix and the movement in two different commits, this is going to be really hard to review if not.

mgdelacroix:
@jespino done!

sbishel (approved):
LGTM

PR #2971, opened 2022-04-28
Showing only the right boards in the board switcher

### LINKED ISSUE #2956: Bug: Board search are returning public boards from teams that I'm not a member of

1. Go to the board switcher
2. Introduce a generic character like `a`
3. If the board exists in multiple teams (and you are not part of one of them) the results keep blank.

## Expected behavior

I only get results for my team, and I see the results in the screen

## Screenshots (optional)

imagen

## Edition and Platform

 - Version: 0.16.0
 - Browser and OS: Firefox on Linux

* Sev 1: Over expose data

### LINKED ISSUE #3013: Bug: /teams API returns teams which user left

1. Join two teams (A and B).
2. Noe leave team B.
3. Open Boards and keep the browser network inspector open and monitor the API `focalboard/api/v2/teams`.
4. Notice that in the response of this API, you'll see team B being returned as well.

## Expected behavior

API should not return teams that user left.

## Edition and Platform

 - Edition:  Mattermost Boards (plugin)
 - Version: 0.16

jespino:
Fixes #2956 
Fixes #3013


mgdelacroix:
:thinking: I might be missing something. Why not ensuring all the boards are valid via the query? If we don't allow to find boards from teams that the user is not a member of, we could do something like this in the mattermost auth layer:

```go
query := s.getQueryBuilder(db).
	Select(boardFields("b.")...).
	Distinct().
	From(s.tablePrefix + "boards as b").
	LeftJoin(s.tablePrefix + "board_members AS bm ON b.id=bm.board_id").
	LeftJoin(s.tablePrefix + "TeamMembers AS tm ON b.team_id=tm.team_id").
	Where(sq.Eq{"tm.user_id": userID}).
	Where(sq.Eq{"b.is_template": false}).
	Where(sq.Eq{"b.delete_at": 0}).
	Where(sq.Or{
		sq.Eq{"b.type": model.BoardTypeOpen},
		sq.And{
			sq.Eq{"b.type": model.BoardTypePrivate},
			sq.Eq{"bm.user_id": userID},
		},
	})
```

jespino:
Was the quick solution to not introduce extra complexity in the cherry-pick but that would need to copy over the method into the mattermostauthlayer

harshilsharma63:
The changes work fine but I prefer @mgdelacroix 's approach of fixing it in the query itself. Currently, it is only the API layer that's fixed, and app and store layers are still buggies. Anywhere where we call the app or store methods, we always get incorrect results.

I'm fine with the current implementation for saving us time but we should open a card for fixing this after 0.16 is released.

harshilsharma63 (changes requested):
Code changes look fine for now except the CI tests are not passing.

harshilsharma63 (approved):
Tested and works fine!

Issue #2946, opened 2022-04-28
Bug: Community clone migration failed.


sbishel:
## Steps to reproduce the behavior

This occurred when upgrading our community clone. 

- go to plugins.
- upload v0.16 plugin

Plugin cannot restart due to migration error.
We will have to get the database reset in order to duplicate. SRE is resetting.

## Expected behavior

MIgration and plugin restart are successfule

## Additional context (optional)
Error -
```
"error":"\"driver: mysql, message: failed when applying migration, command: apply_migration, originalError: invalid connection, query:
```

```
[...]

PR #2974, opened 2022-04-28
Remove timeout from MySQL migrations

mgdelacroix:
#### Summary
This PR reuses the server mechanism to remove the MySQL timeout from the connection string to bypass timeouts when running migrations.

A couple of follow up PRs will come later making the server method public for it to be used in Focalboard instead of duplicating the implementation.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/2946

### DIFF
--- a/server/services/store/sqlstore/migrate.go
+++ b/server/services/store/sqlstore/migrate.go
@@ -57,6 +57,16 @@ func appendMultipleStatementsFlag(connectionString string) (string, error) {
 	return config.FormatDSN(), nil
 }
 
+// resetReadTimeout removes the timeout contraint from the MySQL dsn.
+func resetReadTimeout(dataSource string) (string, error) {
+	config, err := mysqldriver.ParseDSN(dataSource)
+	if err != nil {
+		return "", err
+	}
+	config.ReadTimeout = 0
+	return config.FormatDSN(), nil
+}
+
 // migrations in MySQL need to run with the multiStatements flag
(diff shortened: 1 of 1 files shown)


sbishel (approved):
Woohoo...Couple questions. This only affects mySQL? Do we need to worry about timeouts on large Postgres DBs? And it make sense, but just want to make sure...only ReadTimeOut is an issue. Thanks Miguel.

mgdelacroix:
Only ReadTimeouts should be an issue as WriteTimeouts controls the time when the server is writing to the client. In any case, I tested the fix with both timeouts below migration time and it worked fine. As per postgres, I've tested migrations above the timeout set in the connection string and it seems to respect `morph`s timeout, so we're good only making changes to MySQL

mgdelacroix:
@sbishel ^

mgdelacroix:
After digging a bit deeper, the `connect_timeout` of Postgres only measures the time until the connection is established, but not the time it takes to run the queries. It has `statement_timeout`, which I've tested and can break migrations, but it's not set in our default parameters and it's default value is 0, which disables the timeout. Is as well [recommended not to be set globally in the official docs](https://www.postgresql.org/docs/current/runtime-config-client.html), so an admin would need to explicitly add it to the server connection string for it to cause problems.

I'd propose to follow what server is doing and assume `statement_timeout` is not set, and we can go back to it later if needed.

Issue #2970, opened 2022-04-28
Bug: Typing a second character into the search box results in an error on a particular board


chenilim:
## Steps to reproduce the behavior

1. Join the [Compass Design System channel](https://community-daily.mattermost.com/core/channels/compass-design-system)
2. Open [The Compass Design System board](https://community-daily.mattermost.com/boards/workspace/di1j7wzg67b4bfnihbz9if4nnw/bz5upto7c6ty8bn4j57zjf4s4ch/vso8gngdx4bbs5d657sd78ii1iw) (or import and open the attached board archive below), and open any view
3. Select the Search field, and type 2 characters (e.g. "aa")
4. Redirected to error page immediately

## Expected behavior

No error.

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.15.0
 - Browser and OS: Chrome on Mac

## Additional context (optional)

This also repros on an import of the board, attached here:
[Compass Design System Board archive-2022-4-28.boardarchive.zip](https://github.com/mattermost/focalboard/files/8583986/Compass.Design.System.Board.archive-2022-4-28.boardarchive.zip)

PR #2986, opened 2022-04-28
GH-2970 - Check for propertyValue not null

sbishel:
#### Summary
Some older cards may have Select property set to null. Newer cards have the property removed rather than its value set to null. Add a check for the propertyValue being null.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2970


chenilim (approved):
Makes sense. LGTM!

Issue #2995, opened 2022-04-28
Bug: Follow button doesn't always work


chenilim:
## Steps to reproduce the behavior

1. Create a new board
2. Create a new card, and name it
3. Click on the Follow button
4. Nothing happens (sometimes)

## Expected behavior

Should unfollow card

## Screenshots

https://user-images.githubusercontent.com/46905241/165834882-77fe4a4e-f6c2-443b-aff1-fbb9e6c3e55d.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16 (bugbash2)
 - Browser and OS: Chrome on Mac

## Additional context (optional)

The Follow button works sometimes, but appears to get into a broken state on repeated attempts.

The Network tab shows an error calling `DELETE https://bugbash-boards2.test.mattermost.cloud/boards/team/65tdp6qhxbdiicjr6y7yg47xdo/bu1cpgggjetg1fbzupz9pj94m4c/vgabgs17j8pndtpfgt9oamh8orc/cjjb6ykqekffy3gq8cuwzpkohja`

PR #2997, opened 2022-04-28
GH-2995- Fix for following

sbishel:
#### Summary
When following/unfollowing card, the subscriber id is compared to me?id. However, we weren't rebuilding that function after me.id gets set.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2995


jespino (approved):
LGTM

Issue #2996, opened 2022-04-28
Bug: Text change for Not all team members are listed on the permissions modal search 


wuwinson:
## Steps to reproduce the behavior

1. Go to board
2. Open the permissions modal
3. Click on the search bar
4. Scroll down to see team members to add to the board
5. See error - only select team members appear on the dropdown

## Expected behavior
Search bar placeholder text should display as "Search for people" - see comments below for context.

## Screenshots (optional)
**List of team members**
![Screen Shot 2022-04-28 at 3 48 48 PM](https://user-images.githubusercontent.com/93531870/165834289-31f97edd-e62a-47e0-aec2-f4970e9580d2.png)

**Members who appear on dropdown**

https://user-images.githubusercontent.com/93531870/165835123-2c22d6ba-e047-4297-ab98-9548f4ce9d03.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16 (bugbash 2)
 - Browser and OS: Chrome on Mac

sbishel:
From closed PR for reference - 

@sbishel I think the limit of 10 users was intentional and expected. Depending on the team, it can be a huge list. I think this is actually a UX issue in that it's not communicated to the user that you need to search for the items you're looking for. Happened to me too.

@wuwinson why is it required to show all users in that scrolling dropdown? If we make it more explicit that you need to search for things, would that be fine as well? cc @asaadmahmood 

@wuwinson also, the member's list view in channels you demoed in the video in the issue shows only 100 users max. It feels it's showing all users because there are <100 users on the test server or local server. Try opening that thing for Town Square on community.

The API call used for that list in channels has a hardcoded 100 users limit param being sent, for the same reason we added the 10 user limit in Boards-

```
https://community-daily.mattermost.com/api/v4/users?in_channel=so48j8ynojdtmbr8kdzzpzuztw&page=so48j8ynojdtmbr8kdzzpzuztw&per_page=100&sort=
```

We can certainly increase our limits to match what Channel has but no limits is not a good idea IMO.

I'm 3/5 on keeping the hardcoded 10 users limit.

sbishel:
From closed PR for reference - 
Will discuss with @asaadmahmood @chenilim in our PM/Design meeting this Thursday (May 5th) to figure out some short term and long term solutions to address this. 

- Need to make it clearer to users that they can search for users outside the list of 10 recents displayed. 
- Consider whether we or not we need to increase the limit above 10.
[...]

PR #2998, opened 2022-04-28
GH-2996 - add all members to list, except already members

sbishel:
#### Summary
Fixes issue where only 10 members were being returned for the filtered select list. Also removes those members that are already members of the board.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2996

### DIFF
--- a/webapp/src/components/shareBoard/shareBoard.tsx
+++ b/webapp/src/components/shareBoard/shareBoard.tsx
@@ -300,7 +300,7 @@ export default function ShareBoardDialog(props: Props): JSX.Element {
                     <div className='share-input'>
                         <SearchIcon/>
                         <Select
-                            key={membersCount}
+                            key={JSON.stringify(membersCount)}
                             styles={styles}
                             value={selectedUser}
                             className={'userSearchInput'}
--- a/webapp/src/components/shareBoard/shareBoard.tsx
+++ b/webapp/src/components/shareBoard/shareBoard.tsx
@@ -6,10 +6,7 @@ import React, {useState, useCallback, useEffect} from 'react'
 import {useIntl, FormattedMessage} from 'react-intl'
 import {generatePath, useRouteMatch} from 'react-router'
 import Select from 'react-select/async'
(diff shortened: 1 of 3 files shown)


webapp/src/components/shareBoard/shareBoard.tsx
```diff
@@ -294,11 +300,12 @@ export default function ShareBoardDialog(props: Props): JSX.Element {
                     <div className='share-input'>
                         <SearchIcon/>
                         <Select
+                            key={JSON.stringify(membersCount)}
```
sbishel, line 303: This forces the Async Select list to reload. I tried wrapping `searchAndFilterTeamUsers` with useCallback() but could not get it to work. I'm open to other suggestions. This is the solution suggested [here](https://github.com/JedWatson/react-select/issues/1581).

Without this, when a user is added. If you click on the drop down list again the user is still listed. Once you perform a search, the list is updated.

jespino: Why not use the `filterOptions` property of the `Select` component? looks like the right approach here.

sbishel: Way better...Thanks.

harshilsharma63:
@sbishel I think the limit of 10 users was intentional and expected. Depending on the team, it can be a huge list. I think this is actually a UX issue in that it's not communicated to the user that you need to search for the items you're looking for. Happened to me too.

@wuwinson why is it required to show all users in that scrolling dropdown? If we make it more explicit that you need to search for things, would that be fine as well? cc @asaadmahmood 

I'm 3/5 on keeping the hardcoded 10 users limit.

wuwinson:
Will discuss with @asaadmahmood @chenilim in our PM/Design meeting this Thursday (May 5th) to figure out some short term and long term solutions to address this. Need to make it clearer to users that they can search for users outside the list of 10 recents displayed. Also consider whether we or not we need to increase the limit above 10.

sbishel:
Closing this PR. I will move comments over the issue.

PR #3004, opened 2022-05-02
Fixed incorrect navigation bug

### LINKED ISSUE #2945: Bug: Deleting a shared board puts other users in an unrecoverable state

1. Create a board (as user1)
2. Share it with user2 (set user2 as an Editor)
3. Log in as user2 (e.g. in a separate browser)
4. Create a category (e.g. "Favorites")
5. Move the board to the category
6. As user1, delete the board
7. User2 sees an error (board not found), and clicking on "Back to team" continues to show the error

## Expected behavior

No error, and certainly not an unrecoverable error

## Edition and Platform

 - Version: v0.16 (bugbash2)
 - Browser and OS: Chrome and Safari on Mac

The problem is that the board is stored in user2's category data, and the loading sequence tries to join the deleted board, which results in a 404 error, and redirects to the error page without the ability to recover.

harshilsharma63:
Fixes #2945

1. The issue wasn't just with other users but with your own use as well once you start switching between teams.
2. The issue occurs when you delete the last viewed board of a team, or the first board in the first category (in order of appearance in the sidebar).

### DIFF
--- a/mattermost-plugin/server/manifest.go
+++ b/mattermost-plugin/server/manifest.go
@@ -21,7 +21,7 @@ const manifestStr = `
   "release_notes_url": "https://github.com/mattermost/focalboard/releases",
   "icon_path": "assets/starter-template-icon.svg",
   "version": "0.17.0",
-  "min_server_version": "6.7.0",
+  "min_server_version": "6.0.0",
   "server": {
     "executables": {
       "darwin-amd64": "server/dist/plugin-darwin-amd64",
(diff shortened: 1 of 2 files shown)


sbishel (approved):

harshilsharma63:
I'll cherrypick this after this discussion - https://community-daily.mattermost.com/private-core/pl/md9i1xxnmj8nfg5axraq98gxgc

PR #3005, opened 2022-05-02
[MM-43781] Feature: boards insights

plant99:
#### Summary

This PR introduces 2 endpoints to get boards insights.

`/users/<user-id>/boards/insights?duration<time>`
`/teams/<team-id>/boards/insights?duration=<time>`

`time` can be of the format

- 1 day
- 3 minute
- 2 year

#### Ticket Link

JIRA: https://mattermost.atlassian.net/browse/MM-43781


server/services/store/sqlstore/sqlstore.go
```diff
@@ -119,6 +120,41 @@ func (s *SQLStore) escapeField(fieldName string) string { //nolint:unparam
 	return fieldName
 }
 
+func (s *SQLStore) durationSelector(interval string) string { //nolint:unparam
```
plant99, line 123: Will add tests for these three functions.

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,129 @@
...
+	/**
+	Some squirrel issues to note here are
+	1. https://github.com/Masterminds/squirrel/issues/285 - since we're using 1+ sub queries. When placeholders are counted for second query, the placeholder names are repeated.
+		This is the reason to not use conditional operators in Where clauses which would eventually parametrize the variables.
+	*/
+
+	boardsHistoryQuery := s.getQueryBuilder(db).Select("boards.id", "boards.title", "count(boards_history.id) as count", "boards_history.modified_by", "boards.created_by").
+		From(s.tablePrefix + "boards_history as boards_history").
+		Join(s.tablePrefix + "boards as boards on boards_history.id = boards.id").
+		Where(fmt.Sprintf("boards_history.insert_at > %s and boards.team_id = '%s' and boards_history.modified_by != 'system' and boards.delete_at = 0",
```
srkgupta, line 24: Hi @plant99 
There are no SQL sanitizations in place while using `fmt.Sprintf` statements. As a result these are vulnerable to SQL Injections. Instead please use the Query Generator offered by Squirrel client. Please check other code in sqlstore for examples. 
Thanks.

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,129 @@
...
+			s.durationSelector(duration), teamID)).
+		GroupBy("boards_history.id, boards.id, boards_history.modified_by")
+	blocksHistoryQuery := s.getQueryBuilder(db).Select("boards.id", "boards.title", "count(blocks_history.id) as count", "blocks_history.modified_by", "boards.created_by").
+		From(s.tablePrefix + "blocks_history as blocks_history").
+		Join(s.tablePrefix + "boards as boards on blocks_history.board_id = boards.id").
+		Where(fmt.Sprintf("blocks_history.insert_at > %s and boards.team_id = '%s' and blocks_history.modified_by != 'system' and boards.delete_at = 0",
```
srkgupta, line 30: Hi @plant99 
There are no SQL sanitizations in place while using `fmt.Sprintf` statements. As a result these are vulnerable to SQL Injections. Instead please use the Query Generator offered by Squirrel client. Please check other code in sqlstore for examples. 
Thanks.

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,129 @@
...
+
+	boardsHistoryQuery := s.getQueryBuilder(db).Select("boards.id", "boards.title", "count(boards_history.id) as count", "boards_history.modified_by", "boards.created_by").
+		From(s.tablePrefix + "boards_history as boards_history").
+		Join(s.tablePrefix + "boards as boards on boards_history.id = boards.id").
+		Where(fmt.Sprintf("boards_history.insert_at > %s and boards_history.modified_by != 'system' and boards.delete_at = 0", s.durationSelector(duration))).
+		GroupBy("boards_history.id, boards.id, boards_history.modified_by")
+	blocksHistoryQuery := s.getQueryBuilder(db).Select("boards.id", "boards.title", "count(blocks_history.id) as count", "blocks_history.modified_by", "boards.created_by").
+		From(s.tablePrefix + "blocks_history as blocks_history").
+		Join(s.tablePrefix + "boards as boards on blocks_history.board_id = boards.id").
+		Where(fmt.Sprintf("blocks_history.insert_at > %s and blocks_history.modified_by != 'system' and boards.delete_at = 0", s.durationSelector(duration))).
```
srkgupta, line 76: Same here

server/services/store/sqlstore/board_insights.go
plant99: @srkgupta Thanks for the feedback! Due to [this](https://github.com/mattermost/focalboard/pull/3005/commits/7c2d3134c734f04d6421126a20ceb74d24c5e64e#diff-456b30574cd5964ecde484e5d156d4c8878aa6274bb257588fabc073a9f72538R17) we can't use parameters in this query. So `where` can't be constructed with `sq.Conditional`.

Also I can't find any similar use-cases in focalboard repository. 

1. I'll check mattermost-server or [this](https://github.com/mattermost/mattermost-server/pull/19953/files) PR for how to sanitise fields in a manually constructed query.
2. While I find a workaround, please suggest me fixes

harshilsharma63:
@plant99 can you share any requirement doc for this? That'll help in the review.

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,134 @@
...
+	Some squirrel issues to note here are
+	1. https://github.com/Masterminds/squirrel/issues/285 - since we're using 1+ sub queries. When placeholders are counted for second query, the placeholder names are repeated.
+		This is the reason to not use conditional operators in Where clauses which would eventually parametrize the variables.
+	*/
+
+	boardsHistoryQuery := s.getQueryBuilder(db).Select("boards.id", "boards.title", "boards.icon",
+		"count(boards_history.id) as count", "boards_history.modified_by", "boards.created_by").
+		From(s.tablePrefix + "boards_history as boards_history").
+		Join(s.tablePrefix + "boards as boards on boards_history.id = boards.id").
+		Where(fmt.Sprintf("boards_history.insert_at > %s and boards.team_id = '%s' and boards_history.modified_by != 'system' and boards.delete_at = 0",
```
harshilsharma63, line 25: Did you meant `boards.delete_at != 0",`? (`!=` vs `=`)

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,134 @@
...
+			s.durationSelector(duration), teamID)).
+		GroupBy("boards_history.id, boards.id, boards_history.modified_by")
+	blocksHistoryQuery := s.getQueryBuilder(db).Select("boards.id", "boards.title", "boards.icon",
+		"count(blocks_history.id) as count", "blocks_history.modified_by", "boards.created_by").
+		From(s.tablePrefix + "blocks_history as blocks_history").
```
harshilsharma63, line 30: This would be too much data to fetch into memory at once. It won't impact new users but the more you'll use Boards, the more this query will suffer. Can we perform this operation in batch or think of any other way to make it more performant?

Where exactly and how frequently will this API be called BTW?

harshilsharma63 (changes requested):

harshilsharma63:
Added 2 comments for now as I haven't gone through the entire PR yet. Will review this again.

plant99:
@harshilsharma63 https://github.com/mkraft/suite-users/blob/main/tech-specs/activity-and-insights-widgets/top-boards.md this is a brief design doc.

I'm still working on two things.
 - https://github.com/mattermost/focalboard/pull/3005/files#diff-b9c5dee0af94791cf10c365cac0a7b054a4cc0d3e10b329fe9a4942d42f590c0 there are more cases to be added here, and they'd be made stricter with more `assert` statements.
 - https://github.com/mattermost/focalboard/pull/3005/files#diff-4c7fcc4f7fcc084199c81be767b488432860acc05e6d5d29bdb1dadf66064b52R127 tests for the new utility functions in `sqlstore_test.go`

server/services/store/sqlstore/board_insights.go
plant99: My assumptions were wrong. I'd wait for @mkraft 's comments again. 

Deleting a board isn't a soft-delete.

server/services/store/sqlstore/board_insights.go
mkraft: @harshilsharma63 Yeah, I'm sure we can think of lots of ways to make it more performant; the sky's the limit. I've been testing this against one of our loadtest instances which has a very small dataset. Before this feature can be enabled for customers we'll need to add loadtest data and tweak the performance to an acceptable level.

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func (s *SQLStore) getUserBoardsInsights(db sq.BaseRunner, userID string, duration string) ([]*model.BoardInsight, error) {
+	/**
+	Some squirrel issues to note here are
+	1. https://github.com/Masterminds/squirrel/issues/285 - since we're using 1+ sub queries. When placeholders are counted for second query, the placeholder names are repeated.
+	2. No handlers at the moment for nested conditions with 'in' operator in squirrel - for the final where clause to shortlist user's boards.
+	*/
+	insightsQueryStr := fmt.Sprintf(`select *
+	from (select
```
jespino, line 111: Isn't there any other way to do it right with squirrel, this queries looks like a lot of complexity, actually, I can see it more aceptable if the subqueries are generated independently and then you build the final query using that subquery already calculated SQL. This wall of formatted text is too much for me.

server/api/api.go
```diff
@@ -2039,6 +2040,129 @@ func (a *API) handleUploadFile(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	jsonBytesResponse(w, http.StatusOK, data)
+
+	auditRec.AddMeta("boardInsightCount", len(boardsInsights))
+	auditRec.Success()
+}
+
+func (a *API) handleUserBoardsInsights(w http.ResponseWriter, r *http.Request) {
+	// swagger:operation GET /teams/{teamID}/boards/insights getTeamUsers
```
jespino, line 2109: this comment is wrong

server/api/api.go
```diff
@@ -2039,6 +2040,129 @@ func (a *API) handleUploadFile(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	jsonBytesResponse(w, http.StatusOK, data)
+
+	auditRec.AddMeta("boardInsightCount", len(boardsInsights))
+	auditRec.Success()
+}
+
+func (a *API) handleUserBoardsInsights(w http.ResponseWriter, r *http.Request) {
```
jespino, line 2108: This looks leaky, there is no permission check to verify if you can see other users insights. I would expect at least to check the permissions for that (I don't know if a system_admin permission, or you can check your own insights).

In any case, I expect here some permission checks.

server/api/api.go
```diff
@@ -2039,6 +2040,129 @@ func (a *API) handleUploadFile(w http.ResponseWriter, r *http.Request) {
...
+	//     schema:
+	//       "$ref": "#/definitions/ErrorResponse"
+
+	vars := mux.Vars(r)
+	teamID := vars["teamID"]
+	userID := getUserID(r)
+	query := r.URL.Query()
+	duration := query.Get("duration")
+
+	if !a.permissions.HasPermissionToTeam(userID, teamID, model.PermissionViewTeam) {
```
jespino, line 2082: I just want to confirm. If I can see the team, I can see the team insights?

mkraft: Guests won't have access to insights for now so @plant99 you can add another OR condition here also return this error response if the user is a guest.

wiggin77 (changes requested):
Let me know if/when this is ready for review.

mkraft (approved):

mkraft (commented):
Just the question about verifying the start time is correct compared to similar Mattermost APIs.

server/model/board_insights.go
```diff
@@ -0,0 +1,40 @@
...
+	"encoding/json"
+	"io"
+)
+
+// BoardInsight gives insight into activities in a Board
+// swagger:model
+type BoardInsight struct {
+	// ID of the board
+	// required: true
+	BoardID string `json:"boardID"`
```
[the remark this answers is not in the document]
plant99: I agree, also let's keep it this way, since last version of focalboard insights had camel case. So it'd mean less changes for frontend.

server/api/api.go
```diff
@@ -2105,6 +2110,201 @@ func (a *API) handleUploadFile(w http.ResponseWriter, r *http.Request) {
...
+	if aErr != nil {
+		a.errorResponse(w, r.URL.Path, http.StatusBadRequest, "Error getting time zone of user", aErr)
+		return
+	}
+	userLocation, _ := time.LoadLocation(userTimezone)
+	if userLocation == nil {
+		userLocation = time.Now().UTC().Location()
+	}
+	// get unix time for duration
+	startTime := mmModel.StartOfDayForTimeRange(timeRange, userLocation)
```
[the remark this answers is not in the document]
plant99: Verfied this.

plant99:
Thanks for the review @mkraft , the queries are written with squirrel fully now. Addressed other concerns as well.

server/api/api.go
plant99: Added guests and license checks.

server/services/store/sqlstore/board_insights.go
plant99: Fixed this.

plant99:
> Let me know if/when this is ready for review.

@wiggin77 it is now ready for review.

server/api/api.go
```diff
@@ -2119,6 +2124,201 @@ func (a *API) handleUploadFile(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	jsonBytesResponse(w, http.StatusOK, data)
+
+	auditRec.AddMeta("teamBoardsInsightCount", len(boardsInsights.Items))
+	auditRec.Success()
+}
+
+func (a *API) handleUserBoardsInsights(w http.ResponseWriter, r *http.Request) {
+	// swagger:operation GET /teams/{teamID}/boards/insights getUserBoardsInsights
```
wiggin77, line 2229: This doc path is a different path than what is registered above.  It should be `/users/me/boards/insights`

server/app/insights.go
```diff
@@ -0,0 +1,72 @@
...
+	boardIDs, err := getUserBoards(userID, teamID, a)
+	if err != nil {
+		return nil, err
+	}
+	return a.store.GetUserBoardsInsights(teamID, userID, opts.StartUnixMilli, opts.Page*opts.PerPage, opts.PerPage, boardIDs)
+}
+
+func licenseAndGuestCheck(a *App, userID string) (bool, error) {
+	licenseError := errors.New("invalid license/authorization to use insights API")
+	guestError := errors.New("Guests aren't authorized to use insights API")
```
wiggin77, line 37:
```suggestion
	guestError := errors.New("guests aren't authorized to use insights API")
```

server/model/board_insights.go
```diff
@@ -0,0 +1,62 @@
...
+	return boardInsights
+}
+
+// GetTopBoardInsightsListWithPagination adds a rank to each item in the given list of BoardInsight and checks if there is
+// another page that can be fetched based on the given limit and offset. The given list of BoardInsight is assumed to be
+// sorted by ActivityCount(score). Returns a BoardInsightsList.
+func GetTopBoardInsightsListWithPagination(boards []*BoardInsight, limit int) *BoardInsightsList {
+	// Add pagination support
+	var hasNext bool
+	if (limit != 0) && (len(boards) == limit+1) {
```
wiggin77, line 56:
```suggestion
	if limit != 0 && len(boards) == limit+1 {
```

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,147 @@
...
+	boardsActivity := boardsHistoryQuery.SuffixExpr(blocksHistoryQuery)
+
+	insightsQuery := s.getQueryBuilder(db).Select(
+		fmt.Sprintf("id, title, icon, sum(count) as activity_count, %s as active_users, created_by", s.concatenationSelector("distinct modified_by", ",")),
+	).
+		FromSelect(boardsActivity, "boards_and_blocks_history").
+		GroupBy("id, title, icon, created_by").
+		OrderBy("activity_count desc").
+		Offset(uint64(offset)).
+		Limit(uint64(limit))
```
wiggin77, line 49: Pagination here looks great but doesn't help the unbounded query of the history table above.  If I were to log in with one of my load test accounts to check results this query could return millions of rows.  @mkraft are you certain we'll address this before enabling in a large instance?

wiggin77 (changes requested):
Looks great @plant99 .  Just a few minor fixes requested and a concern about the unbounded query of blocks_history table.

server/services/store/sqlstore/board_insights.go
plant99: Thanks for pointing this out @wiggin77 , genuine concern indeed.

All I could think of is, [this line](https://github.com/mattermost/focalboard/pull/3005/files/194e1c28e9a4d9428f9677de8c7a16574ec7cd69#diff-456b30574cd5964ecde484e5d156d4c8878aa6274bb257588fabc073a9f72538R33) limits the `blocks_history` entries to be in the last `28 days` at maximum. Do you think this information helps reduce the number of rows?

plant99:
@BenCookie95 could you please update the plugin frontend if necessary? similar to [this commit](https://github.com/mattermost/focalboard/pull/3086/commits/722a488975f9a2377c8039152b14b39fc7bf3920)

Let's merge that commit into this PR.

harshilsharma63:
@plant99 sorry for the delay. I'll be reviewing this today.

harshilsharma63 (approved):
Looks good other then few minor things. We definitally need to try out performance on community clone.

webapp/src/insights/index.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export type TopBoard = {
+    boardID: string;
+    icon: string;
+    title: string;
+    activityCount: number;
+    activeUsers: string;
```
mkraft, line 9: Does it make sense for `activeUsers` to be a `string`?

server/services/store/sqlstore/board_insights.go
wiggin77: That 28 day bound certainly helps.  My load test scenario is a pathological case, however it could also happen if someone imports a large archive.  That said we can revisit this knowing the 28 day bound is there.

wiggin77 (approved):
Great work @plant99  :+1:

webapp/src/insights/index.ts
plant99: it's the `string_agg(user_id, ',')` value frontend uses to populate participants. It should be fixed, but in my opinion we can take some time after 7.2 release to work on this.  [created a ticket](https://mattermost.atlassian.net/browse/MM-46251)

plant99:
I understand that we had to rush this because of 7.2 deadline, but if there are any other improvements you'd like to see in the merged code, please feel free to create tickets and assign it to me. Thanks for all your reviews.

Issue #2978, opened 2022-04-28
Bug: Duplicated board has same name as original


chenilim:
## Steps to reproduce the behavior

1. Open a board, and duplicate it
2. The duplicated board has the same name as the original board

## Expected behavior

The duplicated name should be named "<Original name>  copy" (as it did in v0.15)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16 (bugbash)
 - Browser and OS: Chrome on Mac

## Additional context (optional)

The problem is it's difficult to tell which is the original and which is the copy, and it can be easy to subsequently modify the wrong one.

chenilim:
@wuwinson, I don't remember if we had discussed this previously, if this is captured in a spec or issue already, or if this was an intentional change.

chenilim:
Verified this is fixed the latest bugbash today.

Note that if a board with the name "<Original name> copy" already exists, there could be two of them if you dup the one with "<original name>" again, but this is expected.

PR #3011, opened 2022-05-02
GH- 2987 - Add "copy" to board title when duplicated

sbishel:
#### Summary

When duplicating a board, a "copy" suffix should be added. Followed the same logic as is v0.15.  So when a template is created from a board the title is changed to "New board template".

As we do not have server localization yet. These are unlocalized.  However, they were in the release-0.15 code as well.
https://github.com/mattermost/focalboard/blob/release-0.15/webapp/src/mutator.ts#L793

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2978

### DIFF
--- a/server/services/store/sqlstore/boards_and_blocks.go
+++ b/server/services/store/sqlstore/boards_and_blocks.go
@@ -141,6 +141,16 @@ func (s *SQLStore) duplicateBoard(db sq.BaseRunner, boardID string, userID strin
 	if err != nil {
 		return nil, nil, err
 	}
+
+	// todo: server localization
+	if asTemplate == board.IsTemplate {
+		// board -> board or template -> template
+		board.Title += " copy"
+	} else if asTemplate {
+		// template from board
+		board.Title = "New board template"
+	}
+
 	// make new board private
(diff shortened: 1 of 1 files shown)


harshilsharma63 (approved):
LGTM 👍

harshilsharma63:
@sbishel for an untitled board, this sets the new board name as `copy`. I feel it should be `(Untitled Board) copy` as that's the name user sees everywhere in the UI (1/5 on this). Thoughts? cc @wuwinson

wuwinson:
> @sbishel for an untitled board, this sets the new board name as `copy`. I feel it should be `(Untitled Board) copy` as that's the name user sees everywhere in the UI (1/5 on this). Thoughts? cc @wuwinson

@harshilsharma63 Agree that would be a better UX, but would not hold up this PR over that unless it's super quick to do. Otherwise we can address that at a later time.

sbishel:
@harshilsharma63, @wuwinson - I'm merging without the above change. If we want this change we can add it later, but people copying an untitled board would not be a very common scenario.

Issue #2979, opened 2022-04-28
Bug: Duplicate board fails if file (image) attachments are not found


chenilim:
## Steps to reproduce the behavior

1. Unzip and import the attached board below (it has cards with file attachments that are missing)
[Design Options archive-2022-4-28.boardarchive.zip](https://github.com/mattermost/focalboard/files/8584865/Design.Options.archive-2022-4-28.boardarchive.zip)
3. Try to duplicate it
4. Duplicate fails

## Expected behavior

Duplicate succeeds, even if some files are missing.

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16
 - Browser and OS: Chrome on Mac

## Additional context (optional)

This also repros on Personal Server. The server error log is "CopyCardFiles failed to copy file".
Alternatively, import should handle and clean up any missing files in the archive.

sbishel:
Keep Error in log. Allow duplication to succeed. 

Show error to user.(v.,0.17)

PR #3012, opened 2022-05-03
allow duplicating cards/baords if filecopy fails

sbishel:
#### Summary
Don't throw errors when copy files fails. Log a warning and allow duplication to complete.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/2979


jespino (approved):
LGTM

PR #3015, opened 2022-05-03
Don't return teams user is not a member of

### LINKED ISSUE #3013: Bug: /teams API returns teams which user left

harshilsharma63:
Fixes #3013


harshilsharma63:
Closing as it's already fixed in #2971

PR #3018, opened 2022-05-03
update min version in Plugin and build

sbishel:
#### Summary
Update plugin.json to have updated `min_server-version`. Then ran build to ensure manifest.go got updated.


harshilsharma63 (approved):
LGTM 👍

Issue #434, opened 2021-05-17
Server Unit Tests - ./web/webserver.go


sbishel:
Write Unit Tests for /web/webserver.go

prakharporwal:
@sbishel I thought I won't be able to do this. So unassigned. But I think I should give one more try. But need a little guidance on how to go approaching the test.  I am new to unit tests.

#### Their can be multiple ways :
- For NewServer func :
1. On NewServer should I check if the function returns a webserver object.

is this enough!
- For Start( ):
1. Should I actually try pinging the server on Start( ) .
 
Need guidance for this one.

PR #3037, opened 2022-05-06
Added Tests for NewServer() , snapshot matching

prakharporwal:
#### Summary
Added unit test for NewServer functions in web/webserver.go
- [x] snapshot matching given properties should match the properties returned by the function.

#### Ticket Link
#434   
Fixes https://github.com/mattermost/focalboard/issues/434


server/web/webserver_test.go
```diff
@@ -0,0 +1,89 @@
...
+			baseURL := ""
+			url, err := url.Parse(test.serverRoot)
+			if err != nil {
+				test.logger.Error("Invalid ServerRoot setting", mlog.Err(err))
+			}
+			baseURL = url.Path
+
+			ws := NewServer(test.rootPath, test.serverRoot, test.port, test.ssl, test.localOnly, test.logger)
+
+			require.NotNilf(t, ws, "The webserver object is nil!")
```
jespino, line 72: nit: the `NotNilf` can be `NotNil` because you are not using formatting there. The same for `Equalf` later. Anyway, this is just a tiny thing that doesn't matter too much.

server/web/webserver_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+			require.Equalf(t, baseURL, ws.baseURL, "baseURL does not match")
+			require.Equalf(t, test.rootPath, ws.rootPath, "rootPath does not match")
+			require.Equalf(t, test.port, ws.port, "rootPath does not match")
+			require.Equalf(t, test.ssl, ws.ssl, "logger pointer does not match")
+			require.Equalf(t, test.logger, ws.logger, "logger pointer does not match")
+
+			if test.localOnly == true {
+				expectedServerAddr := "localhost:" + strconv.Itoa(test.port)
```
jespino, line 81: Can we move this `expectedServerAddr` to the test cases itself, looks more clear to me to understand the test cases.

server/web/webserver_test.go
```diff
@@ -0,0 +1,89 @@
...
+			port:       9999, // fake port number
+			localOnly:  true,
+			logger:     &mlog.Logger{},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			baseURL := ""
+			url, err := url.Parse(test.serverRoot)
```
jespino, line 64: Looks like this part here is extracted from the `NewServer` code... so we are not testing the underneath behavior, we are testing that the behavior of that chunks of code are equivalent. I would suggest to add the `ExpectedBaseURL` to the test cases.

Also, here we are only testing the happy path. Actually if you look for a non-happy path, for example, passing something that is not a valid URL, you would find a bug there (If I'm right, it is going to panic in that situation).

My suggestion would be add that test cases, and fix the problem how you consider is the right approach (that can be changing the NewServer to return the server and the error, change the NewServer behavior to return nil on error (probably that is unexpected), or maybe change the NewServer to set the baseURL to a defult value (for example `/`) in that case, probably also unexpected.

jespino (changes requested):
It is looking good, but I think we can add some improvements there. Thanks @prakharporwal!

server/web/webserver_test.go
prakharporwal: @jespino 
What would be a non-happy path be like .. 
/@#$  - in this case it returns   / 
///path  -  runs correct
/...path  - runs correct

sbishel (approved):
Thanks @prakharporwal

server/web/webserver_test.go
jespino: Sorry for the late reply, but I think you nailed it.

jespino (approved):
Now looks good to me. Thanks @prakharporwal! :tada:

jespino:
@ogi-m this is only affecting unit tests so I going to skip QA for this PR.

PR #3043, opened 2022-05-06
Refactor notify init

wiggin77:
#### Summary
This PR refactors the notification service initialization to reduce dependencies needed and to eliminate requirements for websocket notifications.  This simplifies the service and removes duplicated code. 

#### Ticket Link
https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/150d1464-6d59-4c39-810f-ee201439dbaf/ch83fx6874f8dfcs8qf1s9r5zhc

### DIFF
--- a/mattermost-plugin/server/notifications.go
+++ b/mattermost-plugin/server/notifications.go
@@ -2,18 +2,20 @@ package main
 
 import (
 	"fmt"
+	"time"
 
+	"github.com/mattermost/focalboard/server/app"
+	"github.com/mattermost/focalboard/server/model"
 	"github.com/mattermost/focalboard/server/services/config"
 	"github.com/mattermost/focalboard/server/services/notify/notifymentions"
 	"github.com/mattermost/focalboard/server/services/notify/notifysubscriptions"
 	"github.com/mattermost/focalboard/server/services/notify/plugindelivery"
 	"github.com/mattermost/focalboard/server/services/permissions"
 	"github.com/mattermost/focalboard/server/services/store"
-	"github.com/mattermost/focalboard/server/ws"
 
 	pluginapi "github.com/mattermost/mattermost-plugin-api"
 
-	"github.com/mattermost/mattermost-server/v6/model"
+	mm_model "github.com/mattermost/mattermost-server/v6/model"
 
 	"github.com/mattermost/mattermost-server/v6/shared/mlog"
 )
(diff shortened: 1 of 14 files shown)


mattermost-plugin/server/notifications.go
```diff
@@ -90,38 +89,97 @@ func createDelivery(client *pluginapi.Client, serverRoot string) (*plugindeliver
...
+func (da *pluginAPIAdapter) CreateMember(teamID string, userID string) (*mm_model.TeamMember, error) {
 	return da.client.Team.CreateMember(teamID, userID)
 }
+
+// appAPI provides app and store APIs for notification services. Where appropriate calls are made to the
+// app layer to leverage the additional websocket notification logic present there, and other times the
+// store APIs are called directly.
+type appAPI struct {
+	store store.Store
+	app   *app.App
```
jespino, line 135: Can we change this to not directly depend on app package? I think we can just add an `AppIface` that only received what is needed for `appAPI`.

mattermost-plugin/server/notifications.go
```diff
@@ -90,38 +89,97 @@ func createDelivery(client *pluginapi.Client, serverRoot string) (*plugindeliver
...
 
-func (da *pluginAPIAdapter) CreateMember(teamID string, userID string) (*model.TeamMember, error) {
+func (da *pluginAPIAdapter) CreateMember(teamID string, userID string) (*mm_model.TeamMember, error) {
 	return da.client.Team.CreateMember(teamID, userID)
 }
+
+// appAPI provides app and store APIs for notification services. Where appropriate calls are made to the
+// app layer to leverage the additional websocket notification logic present there, and other times the
+// store APIs are called directly.
+type appAPI struct {
```
jespino, line 133: What is the reasoning combining this things together? looks like their are used independently in the code.

jespino (changes requested):
Some doubts around putting store and app together in the same struct and also about adding an unneeded package dependency.

mattermost-plugin/server/notifications.go
wiggin77: The service should not be concerned about knowing whether to call into store or app.  Calling into app is a detail brought on by putting so much logic there that ought to be shared.  An alternative is to add all the needed store APIs into app, which perpetuates the god-like design of app.  If we get around to refactoring app then this struct can be modified to do the right thing without impacting any services.

mattermost-plugin/server/notifications.go
wiggin77: I created an appIface - app package is no longer imported for this file.

wiggin77:
Ticket created to address the `app` layer refactor:  https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vd89yph84ajbkufkbgts9m3sb5a/cjswhj4gnzfnsietdryk68u6wjy

jespino (approved):
LGTM

PR #3052, opened 2022-05-11
Added fileinfo tracking

harshilsharma63:
#### Summary
1. Added a mechanism to populate fileinfo table on uploading a new file.
2. Added check for archived file via fileinfo table (if exists)
3. Displayed archived file UI if the file is archived.

#### Ticket Link
1. https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vch9s93uimjrfik7fisjczzpfkc/cy5ixizibo7yutfccxrq4urzb7r
2. https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vch9s93uimjrfik7fisjczzpfkc/crkymxr43fpf48x5udywketptfa

Mattermost server PR - https://github.com/mattermost/mattermost-server/pull/20164


server/app/files.go
```diff
@@ -1,34 +1,91 @@
...
+	parts := strings.Split(filename, ".")
+	fileInfoID := parts[0][1:]
+	fileInfo, err := a.store.GetFileInfo(fileInfoID)
+	if err != nil {
+		return nil, err
+	}
+
+	if fileInfo == nil {
+		return nil, nil
+	}
```
sbishel, line 86: Nit - Unnecessary check.

webapp/src/components/content/imageElement.tsx
```diff
@@ -8,26 +8,38 @@ import octoClient from '../../octoClient'
...
             loadImage()
         }
-    })
+    }, [])
+
+    if (fileInfo.archived) {
+        return (
+            <ArchivedFile fileInfo={fileInfo}/>
+        )
+    }
```
sbishel, line 42: Should we add a test for this?

webapp/src/widgets/icons/brokenFile.tsx
```diff
@@ -0,0 +1,15 @@
...
+import CompassIcon from './compassIcon'
+
+export default function BrokenFile(): JSX.Element {
+    return (
+        <CompassIcon
+            icon='file-image-broken-outline'
+            className='BrokenFile'
+        />
+    )
+}
```
sbishel, line 15: Is this still necessary? I thought we could just use <CompassIcon> directly.

server/app/files.go
```diff
@@ -1,34 +1,91 @@
...
+		Extension:       fileExtension,
+		Size:            fileSize,
+		MimeType:        " ",
+		Width:           0,
+		Height:          0,
+		HasPreviewImage: false,
+		MiniPreview:     nil,
+		Content:         "",
+		RemoteId:        nil,
+	}
```
sbishel, line 60: Saving a string with a space (" "), seems suspicious to me. I assume an empty string ("") won't work. I would almost prefer "empty" or "unused" to a space string.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -500,3 +503,81 @@ func (s *MattermostAuthLayer) CreatePrivateWorkspace(userID string) (string, err
...
+		}
+
+		s.logger.Error("error fetching fileinfo", mlog.String("id", id), mlog.Err(appErr))
+		return nil, appErr
+	}
+
+	return fileInfo, nil
+}
+
+func (s *MattermostAuthLayer) SaveFileInfo(fileInfo *mmmodel.FileInfo) error {
```
sbishel, line 524: Why is `GetFileInfo` a pluginAPI, but `SaveFileInfo` we do directly. I would rather see both as plugin api calls.

server/app/files.go
```diff
@@ -1,34 +1,91 @@
 package app
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"path/filepath"
 	"strings"
 
+	"github.com/mattermost/mattermost-server/v6/model"
```
sbishel, line 10: May want to alias this (even though not strictly necessary). May be easier to understand which model we're using.

sbishel (changes requested):

server/services/store/mattermostauthlayer/mattermostauthlayer.go
ashishbhate: I'm working on the usages API - https://docs.google.com/document/d/19yqSXo-8dahwITBYrkrP9Qy1S28J0o9u8ZBrmCugJv8/edit#

I need to track the storage usage for cloud freemium. 

Doing a `select sum(size) from FileInfo` on the server every time the API is called could end up being a problem for a large `FileInfo` table. 

I was going to store this value in the systems table and update it on every file upload, but if plugins directly write to the FileInfo table, then the files uploaded by the plugins wont be counted in the storage usage. 
If we have a `SaveFileInfo` plugin API endpoint, then I can correctly track storage usage.

For now I can create a cached version that does a `select sum(size) from FileInfo` every hour. I don't think we need real-time file usage tracking just yet.

server/app/files.go
harshilsharma63: Done.

webapp/src/components/content/imageElement.tsx
harshilsharma63: Done.

webapp/src/widgets/icons/brokenFile.tsx
harshilsharma63: Sorry, didn't get it. You mean directly use `<CompassIcon/>` instead of creating this file?

server/app/files.go
harshilsharma63: Yeah, an empty string causes Mattermost server to crash when retrieving fileinfo from DM. The columns are nullable in DB but not nullable in go structs, hence the dummy string.

I've replaced this with `"empty"`.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
harshilsharma63: @ashishbhate we've discussed this in DM. LMK if there is anything else I can do.

server/app/files.go
harshilsharma63: Makes sense, done!

webapp/src/widgets/icons/brokenFile.tsx
sbishel: Yeah. Asking to make sure I'm doing it correctly.

sbishel:
@harshilsharma63 - Still have some conflicts...

webapp/src/widgets/icons/brokenFile.tsx
harshilsharma63: Here's the card for reference - https://community-daily.mattermost.com/boards/workspace/b1ni7nprjpfwpb566k1gk8z9nw/886d2280-e88e-4065-8c85-1b62578f192e/f30dcb19-a0e8-45d2-8768-a9489f6e7fe8/ch47b5ot8bfrq5jpktjhccu3rqw

sbishel (approved):

PR #3055, opened 2022-05-11
Update version to 7.0.0

sbishel:
#### Summary
Update version to 7.0.0


harshilsharma63 (approved):
LGTM 👍

harshilsharma63:
@sbishel for reference, do we need to make all these changes manually or is there a command for it?

sbishel:
> @sbishel for reference, do we need to make all these changes manually or is there a command for it?

Unfortunately, its manual right now.

PR #3058, opened 2022-05-12
Add an endpoint and client method to fetch the boards cloud limits

mgdelacroix:
#### Summary
This PR updates the server dependency to be able to use the new cloud limit plugin APIs and adds an endpoint and a client method to generate and fetch the Boards version of Cloud Limits.

A follow up PR will populate the two fields of the new struct (cardLimitTimestamp and usedCards).


sbishel (approved):
Just wondering if non plugin should be error or not otherwise. LGTM

server/app/cloud.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/mattermost/focalboard/server/model"
+)
+
+var ErrNilPluginAPI = errors.New("server not running in plugin mode")
+
+func (a *App) GetBoardsCloudLimits() (*model.BoardsCloudLimits, error) {
+	if a.pluginAPI == nil {
+		return nil, ErrNilPluginAPI
+	}
```
sbishel, line 17: It depends on the UI implementation, but should this just return an unlimited `BoardCloudLimits` object, so that the UI code works the same.

mgdelacroix: 0/5 on this, I've updated the response to return a `Not Implemented` to follow what Channels is doing for unlicensed and non-cloud instances, but I'm happy to return an empty object. `Limit = 0` should effectively be the same as no limit, so we can make no distinction and do it like that

harshilsharma63:
I'm cherrypicking this to `main`.

PR #3070, opened 2022-05-13
Adding search bar instead of search button for board

asaadmahmood:
#### Summary
Adding search bar instead of search button for board

#### Screenshot
Screenshot 2022-05-14 at 12 47 06 AM


Pinjasaur (approved):
LGTM other than the failing test(s)—looks like the DOM queries need to be updated because there's no longer a `<button>`.

Would it make sense for the placeholder text to just be `Search` vs `Search text`? I suggest it because it matches the dashboard search component, otherwise I'm 1/5 on it.

image

PR #3074, opened 2022-05-16
Notification box for card limit

jespino:


Pinjasaur:
@jespino in my PR I used the `limited` field from the `Block` type: https://github.com/mattermost/focalboard/blob/10dcc071fe231d9747498736eaf1144e167a271a/webapp/src/blocks/block.ts#L44-L63

double checking if we need both `isLimited` and `limited` or if we can get away with only one

mgdelacroix (commented):
LGTM! :+1:

jespino:
Merging with the cypress test failing, I'll fix it after merge.

Issue #2992, opened 2022-04-28
Bug: Table - dragging rows to under another group doesn't always work


chenilim:
## Steps to reproduce the behavior

1. Open a table, group by a property
2. Drag a row from one group to **under** another group
3. It repositions under the same group

## Expected behavior

The row should be moved under the other group

## Screenshots (optional)

https://user-images.githubusercontent.com/46905241/165830699-40030431-8ec2-4c31-8d95-304ab2dec69e.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16 (bugbash2)
 - Browser and OS: Chrome on Mac

## Additional context (optional)

Dragging to another group appears to always work if you drop over the new group's title. But dropping over a row under the new group doesn't always work. This repros in v0.15 as well.

PR #3076, opened 2022-05-16
GH-2992 - Drag drop card to groups doesn't work

sbishel:
#### Summary
Dragging a card from one group to anther and then dragging back to the original group did now work because OnDropToCard did not get updated when cards collection changed. Therefore is was using the old collection and thus had its group to original group. 

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/2992


harshilsharma63 (approved):

harshilsharma63:
@sbishel do we also need to cherrypick this for 7.0?

sbishel:
> @sbishel do we also need to cherrypick this for 7.0?

No, this was only an issue in v7.2

Issue #3053, opened 2022-05-11
Feature Idea: Remove "List: Epics" view from global "Roadmap" template


wuwinson:
## Summary

In v7.0, Cloud freemium will only support up to 5 views per board, but the Roadmap template currently has 6 views. We need to remove 1 view ("List: Epics") from the template so we avoid any issues with freemium.

## How important this is to me and why

Importance: Medium

Use cases:
1. Users on freemium using the Roadmap template

## Additional context/similar features
![Screen Shot 2022-05-11 at 11 52 10 AM](https://user-images.githubusercontent.com/93531870/167893564-f943593d-cf33-4a83-b2bd-c29890e7e93f.png)

wuwinson:
cc @sbishel @chenilim

PR #3080, opened 2022-05-16
update template, remove view

sbishel:
#### Summary
Removes one view from the Roadmap Template.

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/3053


sbishel:
@wiggin77 - I ran `make template-archive`, but it is excluded via https://github.com/mattermost/focalboard/blob/main/.gitignore#L48

However, that file is currently checked in, so I'm a little confused if I need to update and include it or not. 
https://github.com/mattermost/focalboard/blob/main/server/app/templates.boardarchive

wiggin77:
> @wiggin77 - I ran `make template-archive`, but it is excluded via https://github.com/mattermost/focalboard/blob/main/.gitignore#L48
> 
> However, that file is currently checked in, so I'm a little confused if I need to update and include it or not. https://github.com/mattermost/focalboard/blob/main/server/app/templates.boardarchive

Current main now builds `template.boardarchive` dynamically and the resulting file is no longer checked in.  The 7.0 branch still uses the old method where the file is generated manually and checked in.  I merged an updated .gitignore that had latest entries from main since moving back and forth between branches was causing issues with file/dirs being left that would get inadvertently added to git.  

You can `git add -f` to override the gitignore for that file.

wiggin77 (commented):
LGTM :+1:

wiggin77 (approved):

PR #3086, opened 2022-05-17
Feature: board insights - against 7.0 release

plant99:
#### Summary

This PR introduces 2 endpoints to get boards insights.

`/users/me/<team-id>/boards/insights?duration<time>`
`/teams/<team-id>/boards/insights?duration=<time>`

`time` can be of the format

- 1 day
- 3 minute
- 2 year

#### Ticket Link

JIRA: https://mattermost.atlassian.net/browse/MM-43781


plant99:
Not writing integration tests for this endpoint until 7.2 release because the current test setup doesn't allow testing of focalboard-as-plugin. In `main` branch, [there are the necessary tools](https://github.com/mattermost/focalboard/blob/9431a36886a08b9845e1dbe1659883a22412c518/server/integrationtests/clienttestlib.go#L180).

Until then, hopefully store and app tests should suffice.

server/api/api.go
```diff
@@ -101,6 +101,10 @@ func (a *API) RegisterRoutes(r *mux.Router) {
 
 	apiv1.HandleFunc("/workspaces", a.sessionRequired(a.handleGetUserWorkspaces)).Methods("GET")
 
+	// Insights APIs
+	apiv1.HandleFunc("/teams/{teamID}/boards/insights", a.sessionRequired(a.handleTeamBoardsInsights)).Methods("GET")
+	apiv1.HandleFunc("/users/me/{teamID}/boards/insights", a.sessionRequired(a.handleUserBoardsInsights)).Methods("GET")
```
BenCookie95, line 106: Should team be an optional parameter because there might be a point in future where these insights aren't team scoped?

wiggin77: @mkraft, @chenilim, can you answer this?

server/services/store/sqlstore/sqlstore.go
```diff
@@ -99,6 +104,58 @@ func (s *SQLStore) getLicense(db sq.BaseRunner) *mmModel.License {
 	return nil
 }
 
+func (s *SQLStore) durationSelector(interval string) int64 {
+	intervalMagnitudeString := strings.Fields(interval)[0]
+	intervalMagnitude, err := strconv.Atoi(intervalMagnitudeString)
+	if err != nil {
+		// handle error: buggy, change function to introduce err
```
wiggin77, line 111: Is this a TODO?

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,169 @@
...
+		count(blocks_history.id) as count, blocks_history.modified_by, blocks.created_by`).
+		From(s.tablePrefix + "blocks_history as blocks_history").
+		Join(s.tablePrefix + "blocks as blocks on blocks_history.root_id = blocks.id").
+		Where(sq.Gt{"blocks_history.update_at": s.durationSelector(duration)}).
+		Where(sq.Eq{"blocks_history.workspace_id": channelIDs}).
+		Where(sq.NotEq{"blocks_history.modified_by": "system"}).
+		Where(sq.Eq{"blocks.delete_at": 0}).
+		GroupBy("blocks.title, blocks.created_by, blocks.id, blocks_history.modified_by")
+
+	userInsightsQuery := qb.Select("*").FromSelect(qb.Select(fmt.Sprintf("id, title, sum(count) as activity_count, %s as active_users, created_by",
```
wiggin77, line 76: Why the `fmt.Sprintf` here?  Select supports an arbitrary number of string args.

server/app/insights.go
```diff
@@ -0,0 +1,63 @@
...
+		channelIDs[index] = channel.ID
+	}
+	return a.store.GetUserBoardsInsights(userID, duration, channelIDs)
+}
+
+func licenseAndGuestCheck(a *App, userID string) (bool, error) {
+	licenseError := errors.New("invalid license/authorization to use insights API")
+	lic := a.store.GetLicense()
+	if lic == nil {
+		fmt.Println("license is nil")
```
wiggin77, line 51: If the intention is to log this then the `app.logger` can be used for that, with some context added.  Otherwise this should be removed.

server/model/board_insights.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// BoardInsight gives insight into activities in a Board
+// swagger:model
+type BoardInsight struct {
+	// ID of the board
+	// required: true
+	BoardID string `json:"boardID"`
+
+	// icon of the board
```
wiggin77, line 15:
```suggestion
	// Icon of the board
```

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,169 @@
...
+		count(blocks_history.id) as count, blocks_history.modified_by, blocks.created_by`).
+		From(s.tablePrefix + "blocks_history as blocks_history").
+		Join(s.tablePrefix + "blocks as blocks on blocks_history.root_id = blocks.id").
+		Where(sq.Gt{"blocks_history.update_at": s.durationSelector(duration)}).
+		Where(sq.Eq{"blocks_history.workspace_id": channelIDs}).
+		Where(sq.NotEq{"blocks_history.modified_by": "system"}).
+		Where(sq.Eq{"blocks.delete_at": 0}).
+		GroupBy("blocks.title, blocks.created_by, blocks.id, blocks_history.modified_by")
+
+	insightsQuery := qb.Select(fmt.Sprintf("id, title, sum(count) as activity_count, %s as active_users, created_by",
```
wiggin77, line 31: I don't think Sprintf is needed here.

server/services/store/sqlstore/board_insights.go
```diff
@@ -0,0 +1,169 @@
...
+func populateIcons(s *SQLStore, db sq.BaseRunner, boardsInsights []*model.BoardInsight) ([]*model.BoardInsight, error) {
+	qb := s.getQueryBuilder(db)
+	for _, boardInsight := range boardsInsights {
+		// querying raw instead of calling store.GetBoardsFromSameID needs container, and this function has no context on channel ID
+		// performance wise, this is better since 1) it's querying for only fields 2) it's querying for only one row.
+		boardID := boardInsight.BoardID
+		iconQuery := qb.Select("COALESCE(fields, '{}')").From(s.tablePrefix + "blocks").Where(sq.Eq{"id": boardID})
+		iconQueryString, args, err := iconQuery.ToSql()
+		if err != nil {
+			s.logger.Error(`Query parsing error while getting icons`, mlog.Err(err))
```
wiggin77, line 147: Should return here since the query is invalid.

wiggin77 (changes requested):
Looks good overall @plant99.  Great work.  A few changes requested.

plant99:
hey @BenCookie95 now the api matches that of mm-server insights endpoints. here's a sample curl.
```bash
 curl -L -X GET 'http://localhost:8065/plugins/focalboard/api/v1/users/me/boards/insights?time_range=28_day&page=0&per_page=10&team_id=y6az7uth57nwbcqbekz8uo9weo'\ 
    -H 'X-Requested-With: XMLHttpRequest'\
    -H 'Authorization: Bearer f1sxryt7f3n5xcfkao6mzwataw'
```

plant99:
@BenCookie95 @mkraft if you move a block from one column to another, it's counted as 2 changes. If you edit the block, it's counted as one change. Ben and I were thinking the activity count was too high, should we modify this counting of `history-append` as activity count?

server/api/api.go
mkraft: Going from required to optional isn't a breaking change to an API so I'm fine with us addressing that if it comes up in the future.

BenCookie95:
> @BenCookie95 @mkraft if you move a block from one column to another, it's counted as 2 changes. If you edit the block, it's counted as one change. Ben and I were thinking the activity count was too high, should we modify this counting of `history-append` as activity count?

I think the only change we need to make is moving a card from one column to another should only count as one activity. Not sure what code change that involves

plant99:
> > @BenCookie95 @mkraft if you move a block from one column to another, it's counted as 2 changes. If you edit the block, it's counted as one change. Ben and I were thinking the activity count was too high, should we modify this counting of `history-append` as activity count?
> 
> I think the only change we need to make is moving a card from one column to another should only count as one activity. Not sure what code change that involves

@BenCookie95  This will be tricky i.e to modify the query for conditional counting. Otherwise, we can just write another aggregate query. But Martin was of the opinion that the `ActivityCount` would be same for everybody. Even if they count higher, we can either normalise that value by a global average, or let it be higher in magnitude. I agree with this, and we can always come back and fix it if people complain about this. What do you think?

BenCookie95:
> > > @BenCookie95 @mkraft if you move a block from one column to another, it's counted as 2 changes. If you edit the block, it's counted as one change. Ben and I were thinking the activity count was too high, should we modify this counting of `history-append` as activity count?
> > 
> > 
> > I think the only change we need to make is moving a card from one column to another should only count as one activity. Not sure what code change that involves
> 
> @BenCookie95 This will be tricky i.e to modify the query for conditional counting. Otherwise, we can just write another aggregate query. But Martin was of the opinion that the 'higher' `ActivityCount` would be same for everybody. Even if they count higher, we can either normalise that value by a global average, or let it be higher in magnitude. I agree with this, and we can always come back and fix it if people complain about this. What do you think?

Yeah I'm ok with that approach!

server/api/api.go
```diff
@@ -125,7 +130,186 @@ func (a *API) RegisterRoutes(r *mux.Router) {
...
+	if !a.app.CheckUserIDInTeam(userID, teamID) {
+		a.errorResponse(w, r.URL.Path, http.StatusForbidden, "Access denied to team", PermissionError{"access denied to team"})
+		return
+	}
+
+	auditRec := a.makeAuditRecord(r, "getTeamBoardsInsights", audit.Fail)
+	defer a.audit.LogRecord(audit.LevelRead, auditRec)
+	page, err := strconv.Atoi(query.Get("page"))
+	if err != nil {
+		a.errorResponse(w, r.URL.Path, http.StatusInternalServerError, "error converting page parameter to integer", err)
```
wiggin77, line 199: A invalid parameter should respond with a 400 "Bad Request".

server/api/api.go
```diff
@@ -125,7 +130,186 @@ func (a *API) RegisterRoutes(r *mux.Router) {
...
+		return
+	}
+	perPage, err := strconv.Atoi(query.Get("per_page"))
+	if err != nil {
+		a.errorResponse(w, r.URL.Path, http.StatusInternalServerError, "error converting per_page parameter to integer", err)
```
wiggin77, line 204: A invalid parameter should respond with a 400 "Bad Request".

server/api/api.go
```diff
@@ -125,7 +130,186 @@ func (a *API) RegisterRoutes(r *mux.Router) {
...
+	session := ctx.Value(sessionContextKey).(*model.Session)
+	userID := session.UserID
+	query := r.URL.Query()
+	teamID := query.Get("team_id")
+	timeRange := query.Get("time_range")
+
+	// get unix time for duration
+	startTime, aErr := mmModel.GetStartUnixMilliForTimeRange(timeRange)
+	if aErr != nil {
+		a.errorResponse(w, r.URL.Path, http.StatusInternalServerError, "Error parsing time_range="+timeRange, aErr)
```
wiggin77, line 279: A invalid parameter should respond with a 400 "Bad Request".

server/api/api.go
```diff
@@ -125,7 +130,186 @@ func (a *API) RegisterRoutes(r *mux.Router) {
...
+	auditRec := a.makeAuditRecord(r, "getUserBoardsInsights", audit.Fail)
+	defer a.audit.LogRecord(audit.LevelRead, auditRec)
+	page, err := strconv.Atoi(query.Get("page"))
+	if err != nil {
+		a.errorResponse(w, r.URL.Path, http.StatusInternalServerError, "error converting page parameter to integer", err)
+		return
+	}
+	perPage, err := strconv.Atoi(query.Get("per_page"))
+	if err != nil {
+		a.errorResponse(w, r.URL.Path, http.StatusInternalServerError, "error converting per_page parameter to integer", err)
```
wiggin77, line 291: A invalid parameter should respond with a 400 "Bad Request".

plant99:
Thanks @wiggin77 for catching them, I've updated the status codes.

BenCookie95:
@plant99 One thing I noticed today is I can’t see the board of a public channel on team insights if I’m not part of the channel. Is that an easy fix?

plant99:
> @plant99 One thing I noticed today is I can’t see the board of a public channel on team insights if I’m not part of the channel. Is that an easy fix?

I realize now, thanks for checking this @BenCookie95 . I'll push a fix within today. Shouldn't be too much work.

plant99:
@BenCookie95  this is fixed.

> @plant99 One thing I noticed today is I can’t see the board of a public channel on team insights if I’m not part of the channel. Is that an easy fix?

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -428,6 +430,92 @@ func (s *MattermostAuthLayer) GetUserWorkspaces(userID string) ([]model.UserWork
...
+	case mysqlDBType:
+		nonTemplateFilter = nonTemplateFilterMySQL
+	case postgresDBType:
+		nonTemplateFilter = nonTemplateFilterPostgres
+	default:
+		return nil, fmt.Errorf("GetUserWorkspaces - %w", errUnsupportedDatabaseError)
+	}
+
+	query = s.getQueryBuilder().
+		Select("Channels.ID", "Channels.DisplayName", "COUNT(focalboard_blocks.id), Channels.Type, Channels.Name").
```
wiggin77, line 448:
```suggestion
		Select("Channels.ID", "Channels.DisplayName", "COUNT(focalboard_blocks.id)", "Channels.Type", "Channels.Name").
```

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -428,6 +430,92 @@ func (s *MattermostAuthLayer) GetUserWorkspaces(userID string) ([]model.UserWork
...
+	case mysqlDBType:
+		nonTemplateFilter = nonTemplateFilterMySQL
+	case postgresDBType:
+		nonTemplateFilter = nonTemplateFilterPostgres
+	default:
+		return nil, fmt.Errorf("GetUserWorkspaces - %w", errUnsupportedDatabaseError)
+	}
+
+	query = s.getQueryBuilder().
+		Select("pc.ID", "pc.DisplayName", "COUNT(focalboard_blocks.id), 'O' as Type, pc.Name").
```
wiggin77, line 498:
```suggestion
		Select("pc.ID", "pc.DisplayName", "COUNT(focalboard_blocks.id)", "'O' as Type", "pc.Name").
```

wiggin77 (approved):
LGTM :+1:

furqanmlk:
Hi @plant99,  Please add me when Dev review has done.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -428,6 +430,92 @@ func (s *MattermostAuthLayer) GetUserWorkspaces(userID string) ([]model.UserWork
...
+
+	rows, err := query.Query()
+	if err != nil {
+		s.logger.Error("ERROR GetUserWorkspaces", mlog.Err(err))
+		return nil, err
+	}
+
+	defer s.CloseRows(rows)
+
+	accessibleWorkspaces, err := getPublicWorkspacesInTeam(teamID, s)
```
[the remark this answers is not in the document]
plant99: We can make it better that way, surely. However, here are some of the catches that made me decide against it.

1) In this version of code, there aren't tests for `plugin-version` of focalboard. So the blocks defined here should be as clean as possible.
2) Since the norm is to use Squirrel in focalboard, the only way we can `UNION` is to do something like [this](https://github.com/Masterminds/squirrel/issues/20#issuecomment-508680438), which I've added in my most recent commit.

This way seemed cleaner to me than use of `suffixExpr` because of the parameter counting issues with squirrel.

Please let me know what you think. @mkraft

plant99:
@mkraft did the changes you pointed out, @wiggin77 have changed go mods, sums so it'd be nice if you both could please review the recent commits.

mkraft (approved):

wiggin77 (approved):

furqanmlk (approved):
Working as expected

PR #3087, opened 2022-05-17
Add cloud limits

mgdelacroix:
#### Summary
This PR adds the cloud limits for cards

#### Ticket Link
[Card link](https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vch9s93uimjrfik7fisjczzpfkc/cu8fzugji6idgjpad3i6yrzwazh)


mattermost-plugin/server/plugin.go
```diff
@@ -153,6 +152,17 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 
+	limits, err := p.API.GetCloudLimits()
+	if err != nil {
+		fmt.Println("ERROR FETCHING CLOUD LIMITS WHEN STARTING THE PLUGIN", err)
```
jespino, line 157: Debug string here, probably, shouldn't be it a logger call?

mgdelacroix: I'm following the pattern of other prints on the method, probably not to rely on the logger before activating the plugin

mattermost-plugin/server/plugin.go
```diff
@@ -493,3 +503,7 @@ func isBoardsLink(link string) bool {
 	workspaceID, boardID, viewID, cardID := returnBoardsParams(pathSplit)
 	return workspaceID != "" && boardID != "" && viewID != "" && cardID != ""
 }
+
+func (p *Plugin) OnCloudLimitsUpdated(limits *mmModel.ProductLimits) {
+	p.server.App().SetCloudLimits(limits)
```
jespino, line 508: Should we log an error, at least?

server/app/cloud.go
```diff
@@ -4,28 +4,133 @@
...
+	}
+
+	// ToDo:
+	// 1-get limited cards only on a map
+	// 2-iterate through all the blocks, limiting those that either
+	// are limited cards or are linked to them
+
+	limitedBlocks := make([]model.Block, len(blocks))
+	for i, block := range blocks {
+		if block.Type != model.TypeBoard &&
```
jespino, line 126: You are limiting the content/comments blocks too? I think is enough to limit the Cards.

server/model/block.go
```diff
@@ -62,6 +62,10 @@ type Block struct {
 	// The workspace id that the block belongs to
 	// required: true
 	WorkspaceID string `json:"workspaceId"`
+
+	// Indicates if the card is limited
+	// required: false
+	Limited bool `json:"limited,omitempty"`
```
jespino, line 68: I think it was `IsLimited`, it is from the client side, I don't mind to change that in my side anyway, but we should be consistent.

server/app/cloud.go
mgdelacroix: Agree. The approach will be updated on a follow up PR that will as well whitelist template blocks from being limited

jespino (approved):
Nothing mayor stopping us from merging it. I would like a second reviewer for this PR because it is big enough to need it.

server/model/block.go
mgdelacroix: Let's leave `limited` then for consistency with other branches as well

sbishel (approved):
Some ToDos, and merge conflicts, otherwise LGTM

Issue #2988, opened 2022-04-28
Review log entry


sbishel:
## Summary

Review the following log entry. You can cause it by updating the Channels Configuration. 

```
mm-ddux-776547fc89-bvvhz mattermost {"timestamp":"2022-04-28 13:47:38.327 Z","level":"warn","msg":"cannot determine action from cluster message data","plugin_id":"focalboard","id":"websocket_message","payload":"map[enablePublicSharedBoards:false featureFlags:map[:true AddChannelButton: AppsEnabled:true AutoTour: BoardsUnfurl:false CallsMobile:false CloudDelinquentEmailJobsEnabled:false CollapsedThreads:true DownloadAppsCTA: EnableRemoteClusterService:false GlobalHeader:false NewAccountNoisy:false PermalinkPreviews:true PluginApps:off PluginFocalboard:off PluginPlaybooks: PrewrittenMessages: TestBoolFeature:false TestFeature:off] telemetry:true telemetryid:rcaxxez5wfbgmq885iq5oy4yoa]"}
```

[Conversation](https://community.mattermost.com/private-core/pl/6oybtxb31fgi5kb4h5553orrde) for more details.

PR #3090, opened 2022-05-18
GH-2988  Don't redundantly send config chg messages to cluster nodes

wiggin77:
#### Summary
A unexpected warning is appearing in community logs. After investigation it was determined that the config change message was being rebroadcast to cluster nodes as a Focal Board plugin message, which could not be processed.

This PR fixes the issue by only sending the config change message to clients via the PublishWebSocketEvent API which handles node propagation. 

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/2988


sbishel (approved):
LGTM

PR #3094, opened 2022-05-19
Feature to show hidden card count

Rajat-Dabade:
This PR will add a feature to show the hidden card count with the label `Cards Hidden`


webapp/src/store/cards.ts
```diff
@@ -322,3 +322,8 @@ export const getCurrentCard = createSelector(
     (state: RootState) => state.cards.cards,
     (current, cards) => cards[current],
 )
+
+export const getHiddenCard = createSelector(
```
sbishel, line 326: This should be named `GetHiddenCards`

webapp/i18n/en.json
```diff
@@ -257,6 +257,8 @@
   "error.relogin": "Log in again",
   "login.log-in-button": "Log in",
   "login.log-in-title": "Log in",
+  "limitedCard.id": "hidden-card-group-id",
```
sbishel, line 260: Are you sure this needs localized?

webapp/src/components/gallery/gallery.test.tsx
```diff
@@ -129,4 +134,49 @@ describe('src/components/gallery/Gallery', () => {
...
+                    cards={[card1, card3]}
+                    activeView={activeView}
+                    readonly={false}
+                    addCard={jest.fn()}
+                    selectedCardIds={[card1.id]}
+                    onCardClicked={jest.fn()}
+                />
+            </ReduxProvider>,
+        ))
+        expect(getByTitle('limited-card-count').innerHTML).toBe('<span>2</span>')
```
sbishel, line 180: Add a snapshot check here as well.

webapp/src/components/gallery/gallery.tsx
```diff
@@ -90,6 +99,10 @@ const Gallery = (props: Props): JSX.Element => {
                     />
                 </div>
             }
+            {limited.length > 0 && <div className='limitedCard'>
+                <div className='limitedCard-title'>{intl.formatMessage({id: 'limitedCard.title', defaultMessage: 'Cards Hidden'})}</div>
+                <Button title='limited-card-count'>{limited.length}</Button>
+            </div>}
```
sbishel, line 105: Move to a component, then just call component where desired. Encapsulates the limit card message feature

webapp/src/components/kanban/__snapshots__/kanban.test.tsx.snap
```diff
@@ -279,6 +279,7 @@ exports[`src/component/kanban/kanban return kanban and change title on KanbanCol
             </span>
           </div>
           <button
+            title="limited-card-count"
```
sbishel, line 282: Based on the code changes, nothing should change since your checking 
`if (limitedCard.length > 0) {`
and there are no changes to test data.
In addition, this item is `<span>0</span` which shouldn't happen.

webapp/src/components/gallery/gallery.tsx
```diff
@@ -25,6 +31,8 @@ type Props = {
 
 const Gallery = (props: Props): JSX.Element => {
     const {activeView, board, cards} = props
+    const limited = useAppSelector(getHiddenCard)
```
sbishel, line 34: nit - `limited` sounds like a `boolean` variable. Why not name this `hiddenCards`?

webapp/src/components/kanban/kanban.tsx
```diff
@@ -46,6 +50,20 @@ const vStrength = createVerticalStrength(Utils.isMobile() ? 60 : 250)
 
 const Kanban = (props: Props) => {
     const {board, activeView, cards, groupByProperty, visibleGroups, hiddenGroups} = props
+    const limitedCard = useAppSelector(getHiddenCard)
+    const intl = useIntl()
+    const hiddenCardGroupId = 'hidden-card-group-id'
+
+    if (limitedCard.length > 0) {
+        props.hiddenGroups.push({
```
sbishel, line 58: I wouldn't confuse "LimitedCards" with "HiddenGroups". I would keep those things separate. If you create a "LimitedCards" component, you can simply add that compoent after the "hiddenGroups". see below, Line 331. None of this code would be necessary.

webapp/src/components/kanban/kanban.test.tsx
```diff
@@ -80,6 +80,7 @@ describe('src/component/kanban/kanban', () => {
         comments: {
             comments: {},
         },
+        boards: {},
```
sbishel, line 83: Add a test for displaying the component.

webapp/src/components/kanban/kanbanHiddenColumnItem.test.tsx
```diff
@@ -99,4 +100,22 @@ describe('src/components/kanban/kanbanHiddenColumnItem', () => {
...
+                group={{
+                    option,
+                    cards: [card, card2],
+                }}
+                readonly={false}
+                onDrop={jest.fn()}
+                intl={intl}
+            />,
+        ))
+        expect(getByTitle('limited-card-count')).toHaveTextContent('2')
```
sbishel, line 119: Nice, add snapshot check as well.

webapp/src/components/kanban/kanbanHiddenColumnItem.tsx
```diff
@@ -55,16 +55,16 @@ export default function KanbanHiddenColumnItem(props: Props): JSX.Element {
                 >
                     {group.option.value}
                 </Label>
-                <Menu>
+                {props.group.option.id !== intl.formatMessage({id: 'limitedCard.id'}) && <Menu>
```
sbishel, line 58: This code shouldn't be needed if you separate LimitedCards and HiddenGroups

webapp/src/components/table/table.test.tsx
```diff
@@ -153,6 +154,56 @@ describe('components/table/Table', () => {
...
+                    readonly={true}
+                    cardIdToFocusOnRender=''
+                    showCard={callback}
+                    addCard={addCard}
+                    onCardClicked={jest.fn()}
+                />
+            </ReduxProvider>,
+        )
+        const {getByTitle} = render(component)
+        expect(getByTitle('limited-card-count')).toHaveTextContent('2')
```
sbishel, line 205: Check snapshot as well.

webapp/src/components/kanban/kanbanHiddenColumnItem.tsx
```diff
@@ -55,16 +55,16 @@ export default function KanbanHiddenColumnItem(props: Props): JSX.Element {
...
                         id='show'
                         icon={<ShowIcon/>}
                         name={intl.formatMessage({id: 'BoardComponent.show', defaultMessage: 'Show'})}
                         onClick={() => mutator.unhideViewColumn(activeView, group.option.id)}
                     />
-                </Menu>
+                </Menu>}
             </MenuWrapper>
-            <Button>{`${group.cards.length}`}</Button>
+            <Button title='limited-card-count'>{`${group.cards.length}`}</Button>
```
sbishel, line 67: This is why your snapshot is changing...

sbishel (changes requested):
@Rajat-Dabade You should create a `<HiddenCardCount>` component. Then in the various views, you can simply call that component. This will encapsulate that functionality. Also, we should check the cloud license first, so we don't do anything else if there are no limits. An on-premise server should not ever need to call `getHiddenCards`. 

Let me know if any of my comments need further explanation. Thanks

webapp/src/components/kanban/__snapshots__/kanban.test.tsx.snap
sbishel: Realize now, the 0 is because its a "Hidden Column" not "limited-card-count"

sbishel:
@Rajat-Dabade - Could you also add screen shots to the PR. Thanks.

webapp/src/blocks/card.ts
```diff
@@ -12,6 +12,7 @@ type CardFields = {
 
 type Card = Block & {
     fields: CardFields
+    limited: boolean
```
Pinjasaur, line 15: question: could this just be added to the Block type? similar to in my PR: https://github.com/mattermost/focalboard/blob/a863e43fb0876d21d7996d92f5c3452a99c25437/webapp/src/blocks/block.ts#L44

webapp/src/blocks/card.ts
```diff
@@ -36,6 +37,7 @@ function createCard(block?: Block): Card {
             contentOrder,
             isTemplate: block?.fields.isTemplate || false,
         },
+        limited: false,
```
Pinjasaur, line 40: I don't think we want to default it to false, right? It should just be passed through? https://github.com/mattermost/focalboard/blob/a863e43fb0876d21d7996d92f5c3452a99c25437/webapp/src/blocks/block.ts#L63

similar to my comment above, could this be moved to `createBlock` instead of being in `createCard`?

webapp/src/test/testBlockFactory.ts
```diff
@@ -112,14 +112,14 @@ class TestBlockFactory {
...
-    static createCard(board?: Board): Card {
+    static createCard(board?: Board, limited = false): Card {
         const card = createCard()
         card.parentId = board ? board.id : 'parent'
         card.rootId = board ? board.rootId : 'root'
         card.title = 'title'
         card.fields.icon = 'i'
         card.fields.properties.property1 = 'value1'
-
+        card.limited = limited
```
Pinjasaur, line 122: Same question as my other comments, can it just be passed through?

webapp/src/blocks/card.ts
Rajat-Dabade: Yes, nice catch. I will update it.

webapp/src/blocks/card.ts
Rajat-Dabade: Yes. Fixed

webapp/src/test/testBlockFactory.ts
Rajat-Dabade: Done

Rajat-Dabade:
Here are the screenshot, how what hidden cards look like:

For Status view:

Screenshot 2022-05-23 at 5 47 55 PM
Screenshot 2022-05-23 at 5 48 08 PM

For List view: 
Screenshot 2022-05-23 at 5 48 38 PM

For Gallery view:
Screenshot 2022-05-23 at 5 49 14 PM

webapp/src/store/cards.ts
```diff
@@ -322,3 +322,8 @@ export const getCurrentCard = createSelector(
     (state: RootState) => state.cards.cards,
     (current, cards) => cards[current],
 )
+
+export const GetHiddenCard = createSelector(
```
mgdelacroix, line 326: nit: to keep the naming and casing consistent, I'd suggest something in plural and a bit more explicit, derived from `getCurrentBoardCards` as it's actually getting the inforrmation from that selector:

```suggestion
export const getCurrentBoardHiddenCards = createSelector(
```

mac/Focalboard/whatsnew.txt
```diff
@@ -1,8 +1,12 @@
-Welcome to Focalboard v0.11.0!
```
mgdelacroix, line 1: Don't think this file should be updated as part of this PR

webapp/src/components/gallery/gallery.tsx
```diff
@@ -25,6 +30,7 @@ type Props = {
 
 const Gallery = (props: Props): JSX.Element => {
     const {activeView, board, cards} = props
+    const hiddenCards = useAppSelector(GetHiddenCard)
```
mgdelacroix, line 33: Reviewing the information flow for the `gallery`, `kanban` and `table` components, the `board`, `cards` and other pieces of the store are being fetched in the `CenterContent` component, so I think the `hiddenCards` should be fetched there as well, and passed through

mgdelacroix (commented):
Looking really good!

webapp/src/components/kanban/kanbanHiddenColumnItem.tsx
```diff
@@ -55,16 +58,17 @@ export default function KanbanHiddenColumnItem(props: Props): JSX.Element {
                 >
                     {group.option.value}
                 </Label>
-                <Menu>
+                {props.group.option.id !== hiddenCardGroupId && <Menu>
```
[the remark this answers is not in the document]
Rajat-Dabade: Fixed

webapp/src/store/cards.ts
Rajat-Dabade: Sure, will change it.

webapp/src/components/gallery/gallery.tsx
Rajat-Dabade: True, I will make the changes. Thanks

mac/Focalboard/whatsnew.txt
```diff
@@ -2,7 +2,7 @@ Welcome to Focalboard v0.11.0!
 
 Thank you contributors! There are a number of improvements in this release, and one major new feature:
 
-* Calendar view. Thanks @sbishel!
+* Calendar view. Thanks @sbishel!*
```
mgdelacroix, line 5: nit: minor change left behind

mgdelacroix (approved):
Thanks for applying the changes @Rajat-Dabade! LGTM! :rocket:

mac/Focalboard/whatsnew.txt
Rajat-Dabade: Fixed. Thanks

webapp/src/components/workspace.tsx
```diff
@@ -36,6 +36,7 @@ function CenterContent(props: Props) {
     const groupByProperty = useAppSelector(getCurrentViewGroupBy)
     const dateDisplayProperty = useAppSelector(getCurrentViewDisplayBy)
     const clientConfig = useAppSelector(getClientConfig)
+    const hiddenCards = useAppSelector(GetCurrentBoardHiddenCards)
```
sbishel, line 39: We should wrap this with the normal Cloud checks. `utils.IsCloudLicense() and GetCloudLimits`

This will slow down on-premise and paid accounts. by looping through cards unnecessarily.

webapp/src/components/hiddenCardCount/hiddenCardCount.tsx
```diff
@@ -0,0 +1,26 @@
...
+import React from 'react'
+import {useIntl} from 'react-intl'
+
+import Button from '../../widgets/buttons/button'
+
+import {Card} from '../../blocks/card'
+import './hiddenCardCount.scss'
+
+type Props = {
+    hiddenCards: Card[]
```
sbishel, line 13: Didn't notice before, but you are only using the `.length` of array. Probably better to just get and pass an `Integer` rather than a `Card[]`.

sbishel (changes requested):
I'm sorry for not noticing this before, but since you moved the retrieval of hidden cards into `workspace.tsx`, I noticed that you don't need the HiddenCards (Cards[]) at all. You only use the Len of that array. So, I would suggest changing the `HIddenCards` property from a Card[] to `HIddenCardsCount` as an Int. Defaulting the property to 0, will allow you to not need to change the test code as well.

Let me know if I missed where the actually cards are needed. Also, we should be aware of performance implications for non-limited installations and wrap those with checks.

webapp/src/components/hiddenCardCount/hiddenCardCount.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+const HiddenCardCount = (props: Props): JSX.Element => {
+    const intl = useIntl()
+    return (
+        <div className='HiddenCardCount'>
+            <div className='hidden-card-title'>{intl.formatMessage({id: 'limitedCard.title', defaultMessage: 'Cards Hidden'})}</div>
```
Pinjasaur, line 20: looks like you'll want to run `npm run i18n-extract` to update the `en.json` file

I'd recommend updating your branch with `release-7.0` before you do that as my PR that was merged yesterday updated the `npm run i18n-extract` command

Pinjasaur (approved):
looks good @Rajat-Dabade! small note re: updating the i18n file(s) and it looks like there's a couple merge conflicts, but otherwise LGTM after addressing that

webapp/src/components/hiddenCardCount/hiddenCardCount.tsx
Rajat-Dabade: Okay, got it.

Rajat-Dabade:
@sbishel, Just had a word with @jespino, he will be taking care of wrapping the necessary code with the `isCloudLicense` whenever it is needed and optimising the selector code. Thank you.

sbishel (approved):

PR #3107, opened 2022-05-24
Improve apply cloud limits to leave templates unlimited

mgdelacroix:
#### Summary
This PR adds testing to the apply cloud limit logic and improves it to limit only non-template cards


jespino (approved):
LGTM

PR #3127, opened 2022-05-25
Forbid updating limited cards

mgdelacroix:
#### Summary
This PR checks if cloud limits should apply before accepting patch requests, and rejects them if they refer to a card that is limited


jespino (approved):
LGTM

jespino (approved):
LGTM

Issue #3116, opened 2022-05-24
Bug: Duplicate view doesn't trigger "Limit reached" notification


wuwinson:
## Steps to reproduce the behavior

1. Go to a board with 5 views
2. On the view drop down, select "Duplicate view"
3. See error - nothing happens

## Expected behavior

"Views per board limit reached" modal should appear

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/170141362-3a492fb6-9207-4f0e-a955-91ff0bddf9e3.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.0-RC1
 - Browser and OS: Chrome on Mac

PR #3129, opened 2022-05-25
GH 3116 Check cloud limits when duplicating view

sbishel:
#### Summary
When duplicating a view, the cloud limits should be checked. Just as when adding a view.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/3116


Pinjasaur (approved):
LGTM—I believe failing Cypress tests are unrelated to this PR

PR #3126, opened 2022-05-25
`main` Cherrypick: Hidden card view via direct URL & unfurl (#3071)

Pinjasaur:
#### Summary

This is an initial cherrypick of #3071 to `main` with merge conflicts resolved. However, other PRs got merged before mine so changes e.g. `limited` on a `Block` type and additions to the `WSClient` prevent this from fully working until they're cherrypicked into `main` as well. CI is expected to fail at this point.

Once `main` has the changes, I'm expecting to need to `npm run i18n-extract` along with verifying tests and everything still working as expected.

#### Ticket Link

#3071

PR #3071, opened 2022-05-13
Hidden card view via direct URL & unfurl

Pinjasaur:
#### Summary

If viewing a card that is `limited` directly via a URL or unfurl it will render less data.

#### Ticket Link

- https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vch9s93uimjrfik7fisjczzpfkc/cekkdizfr5b8fjmfstwgb7knoyy
- https://community-daily.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vch9s93uimjrfik7fisjczzpfkc/ck1kek4ffx3du38uissq7oqnb9c

#### Screenshots

image

image

To validate this locally, I'd set `limited: true` in `block.ts#createBlock`:

```diff
- limited: Boolean(block?.limited),
+ limited: true,
```

PR #3156, opened 2022-05-31
ci: test & lint `mattermost-plugin`

Pinjasaur:
#### Summary

In #3071 tests were added to `BoardsUnfurl` along with an addition to the `make ci` step to simulate a CI run locally. At the time, I didn't realize that `make ci` wasn't explicitly ran in CI so this PR updates `ci.yml` to lint + test `mattermost-plugin/webapp` in addition to the normal `webapp`.

I initially created this PR in https://github.com/mattermost/focalboard/tree/plugin-ci which was branched off `release-7.0`. You can view a successful run here: https://github.com/mattermost/focalboard/runs/6616768660?check_suite_focus=true

If it's desired, I can create another PR to target `release-7.0` as well. This is blocked by #3126 as it includes the aforementioned changes + tests that are actually _ran_.

#### Ticket Link

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,10 +40,14 @@ jobs:
       uses: actions/checkout@v3
 
     - name: npm ci
-      run: cd webapp; npm ci
+      run: |
+        cd webapp && npm ci && cd -
+        cd mattermost-plugin/webapp && npm ci
 
     - name: ESLint
-      run: cd webapp; npm run check
+      run: |
+        cd webapp && npm run check && cd -
+        cd mattermost-plugin/webapp && npm run lint
 
     - name: Set up Go
       uses: actions/setup-go@v3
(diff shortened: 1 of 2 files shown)


wiggin77:
Should we switch to using the make file for running tests in CI?  Otherwise we will continue to change the makefile and forget to update the ci.yml

Pinjasaur:
(I typed this all up earlier and managed to lose the entire textarea before commenting so we're doing this twice)

@sbishel I noticed that `npm run lint` was failing for the plugin `webapp` after what I _believe_ is the dependency changes (specifically the mattermost eslint plugin) in #3317. The root `webapp` wasn't failing so I tracked changes back to @jespino's #2642 and applied the eslint config changes, which mainly swapped the mattermost eslint plugin. I'm not positive the state of the mattermost eslint plugin, but this fixes it for now.

@wiggin77 I modified the Makefile & ci.yml, most notably with a `make webapp-ci` recipe which is used locally in `make ci` along with in the `ci.yml` to handle linting & testing (jest + cypress) which should address your comment above.

wiggin77 (commented):
LGTM :+1:

jespino (approved):
LGTM

PR #3164, opened 2022-06-02
Forward-porting fileinfo limits

harshilsharma63:
Ported file info limits code to main.

### DIFF
--- a/webapp/src/octoClient.ts
+++ b/webapp/src/octoClient.ts
@@ -563,16 +563,16 @@ class OctoClient {
             path += `?read_token=${readToken}`
         }
         const response = await fetch(this.getBaseURL() + path, {headers: this.headers()})
-        let fileURL: FileInfo = {}
+        let fileInfo: FileInfo = {}
 
         if (response.status === 200) {
             const blob = await response.blob()
-            fileURL.url = URL.createObjectURL(blob)
+            fileInfo.url = URL.createObjectURL(blob)
         } else if (response.status === 400) {
-            fileURL = await this.getJson(response, {}) as FileInfo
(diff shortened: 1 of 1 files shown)


server/services/store/sqlstore/migrations/000017_add_teams_and_boards.up.sql
```diff
@@ -13,6 +13,7 @@ ALTER TABLE {{.prefix}}blocks_history ADD COLUMN board_id VARCHAR(36);
 {{- /* cleanup incorrect data format in column calculations */ -}}
 {{- /* then move from 'board' type to 'view' type*/ -}}
 {{if .mysql}}
+SELECT 'I am mysql';
```
jespino, line 16: debug strings here

webapp/src/octoClient.ts
```diff
@@ -556,18 +556,23 @@ class OctoClient {
...
+        if (response.status === 200) {
+            const blob = await response.blob()
+            fileURL.url = URL.createObjectURL(blob)
+        } else if (response.status === 400) {
+            fileURL = await this.getJson(response, {}) as FileInfo
         }
-        const blob = await response.blob()
-        return URL.createObjectURL(blob)
+
+        return fileURL
```
jespino, line 575: nit: a bit misleading name of variable.

jespino (approved):
LGTM

server/services/store/sqlstore/migrations/000017_add_teams_and_boards.up.sql
harshilsharma63: Done.

PR #3168, opened 2022-06-02
Do not check rowCount when updating the card limit

mgdelacroix:
#### Summary
This PR modifies the `updateCardLimitTimestamp` store method not to rely on `rowCount`, as `MySQL` seems not to send back `rowCount` if several updates happen in quick succession


harshilsharma63 (approved):
LGTM 👍

PR #3172, opened 2022-06-02
Git attributes for linguist

kamre:
#### Summary
Git attributes for ignoring `website` and `swagger docs` in linguist added.

Currently this repo page shows this distribution of the languages:
image
which is a bit misleading because of auto generated documentation.

With the proposed changes it looks like this:
image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+website/** linguist-documentation
+server/swagger/** linguist-generated


wiggin77 (approved):
LGTM :+1:   I agree @kamre this change makes the list more accurate.

kamre:
Hi @jespino, @chenilim what do you think about this minor improvement?

chenilim:
Super! Thanks so much @kamre! This makes sense. The "HTML" bar was throwing me off too. @wiggin77 not sure if you were waiting for more reviews, but you could have merged this after you approved it. I'll merge it now.

PR #2819, opened 2022-04-15
Issue 2410: checkboxes wrap long text

igordsm:
#### Summary

Change checkbox widget to `EditableArea`, which uses `<textarea>` (from `Editable` which uses `<input>`). `<input>` does not support text wrapping but `<textarea>` does.

image

#### Ticket Link
Fix #2410

PR #3179, opened 2022-06-03
Prevent wrapping of text in checkbox

harshilsharma63:
Clone of #2819


sbishel:
@harshilsharma63 - What is the status of this PR?

PR #3189, opened 2022-06-06
Updated go version

harshilsharma63:


linux/go.mod
```diff
@@ -1,12 +1,119 @@
...
+go 1.18
 
 replace github.com/mattermost/focalboard/server => ../server
 
+replace github.com/mattermost/mattermost-server/v6 => github.com/mattermost/mattermost-server/v6 v6.0.0-20220601144141-e54977931ccd
+
 require (
 	github.com/google/uuid v1.3.0
-	github.com/mattermost/focalboard/server v0.0.0-20220325164658-33557093b00d
-	github.com/mattermost/mattermost-server/v6 v6.5.0
```
jespino, line 10: Is there any reason for this downgrade?

harshilsharma63: go mod tidy does it for some unknown reason. Its a know issue in server team and no one knows why it happens. Just living with it :P. There's a replace on top of the file to fix this and use the right version.

jespino (approved):
LGTM

PR #3196, opened 2022-06-07
Migration fixes

wiggin77:
#### Summary
Fixes issues with migrations from `release-7.0` to main

- start migrations at slot 18 because `release-7.0` goes up to 17
- ensure all create table statements check for `not exists`
- simplify migration `root_id --> board_id` 
    - I ensured all block records in community are populated (no nulls) for root_id.
    - tested against 500K block instance; migration took 3.5 minutes for MySQL

#### Ticket Link
NONE


server/services/store/sqlstore/migrations/000018_add_teams_and_boards.up.sql
```diff
@@ -274,137 +274,15 @@ CREATE TABLE {{.prefix}}boards_history (
...
-         {{.prefix}}blocks_history as GP,
-         {{.prefix}}blocks_history as GGP
-   WHERE B.parent_id=P.id
-     AND P.parent_id=GP.id
-     AND GP.parent_id=GGP.id
-     AND GGP.type = 'board';
-{{end}}
-
+UPDATE {{.prefix}}blocks as B SET B.board_id=B.root_id;
+UPDATE {{.prefix}}blocks_history as B SET B.board_id=B.root_id;
```
harshilsharma63, line 278: This deletion look a bit dangerous. Do we not need all the old stuff because `B.root_id` points to the board already?

wiggin77: All the deleted lines were to populate the board_id based on either the parent_id of a card, or the parent's parent for content blocks, down to parent's parent's parent.  However, cards and all children, plus views, have a root_id that points to the board.   There are no root_id is null anywhere in community which means we've been diligent in populating it.

harshilsharma63: That's awesome! Always love when we can remove code 👍

harshilsharma63 (approved):

mgdelacroix (approved):
Awesome! It's going to be much simpler & faster now :rocket:

PR #3212, opened 2022-06-11
[GH-3211] Fix switching of views

### LINKED ISSUE #3211: Bug: filtered out rows are visible on the first render of the table

1. Use `Project Tasks` template to create new board
2. Switch to `Task Overview` table view
3. Add **new filter** (for example **include only** tasks which have `Status` value `In Progress`)
4. Switch between views in the sidebar (from `Task Overview` and back)
5. When switching back to the table view with filter there is a flicker because initial rendering shows all the rows and only later some rows are removed

## Expected behavior
No flickering and first rendering of the table view shows only rows according to the filter

## Edition and Platform

 - Edition: Personal Desktop
 - Version: `main` branch

Can use CPU throttling in Crome Dev Tools to notice the flicker

kamre:
#### Summary
Fix flickering when switching views: use redux state for rendering instead of the router match params.

#### Ticket Link
Fixes #3211


harshilsharma63 (approved):
Works like magic!

Issue #3221, opened 2022-06-14
Bug: Gives dependency error when doing linux developer setup following the developer setup guide


imasdekar:
### Below is the error received 

```shell npm run pack

> focalboard@7.2.0 pack
> cross-env NODE_ENV=production webpack --config webpack.prod.js

sh: line 1: cross-env: command not found
make: *** [Makefile:158: webapp] Error 127
```

## Steps to reproduce the behavior

Following [dev-setup guide](https://developers.mattermost.com/contribute/focalboard/personal-server-setup-guide/), 
I did the following steps.
1. git clone the repo git@github.com:mattermost/focalboard.git
2. cd into the cloned repo folder
3. run the command `make linux-app`

## Expected behavior

It should build the standalone linux desktop application without giving the npm dependencies error(mentioned above)

## Screenshots (optional)

If applicable, add screenshots or a screen recording to elaborate on the problem.

## Edition and Platform

 - Edition: Personal Desktop 
 - Version: latest commit from the main branch
 - Browser and OS: ArcoLinux

wuwinson:
cc @chenilim @justinegeffen

PR #3222, opened 2022-06-14
fix: fixes the dev setup dependencies issues when doing make linux-app

imasdekar:
#### Summary

- This PR adds a run for prebuild step that installs all the npm dependencies when doing make linux-app.

Signed-off by: imasdekar <imasdekar@disroot.org>

#### Ticket Link
Related Issue: #3221

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,12 +28,12 @@ endif
 all: webapp server ## Build server and webapp.
 
 prebuild: ## Run prebuild actions (install dependencies etc.).
-	cd webapp; npm install
+	npm install --prefix webapp
 
 ci: server-test
-	cd webapp; npm run check
-	cd webapp; npm run test
-	cd webapp; npm run cypress:ci
+	npm run check --prefix webapp
+	npm run test --prefix webapp
+	npm run cypress:ci --prefix webapp
 
 templates-archive: ## Build templates archive file
 	cd server/assets/build-template-archive; go run -tags '$(BUILD_TAGS)' main.go --dir="../templates-boardarchive" --out="../templates.boardarchive"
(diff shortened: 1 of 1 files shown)


Pinjasaur (commented):
Hey @imasdekar, thanks for the PR!

I'm not familiar with using `--prefix` with `npm`. Do you have have any docs you could reference? I wasn't able to find much, it seems undocumented: https://github.com/npm/cli/issues/1368 I like that it cleans up the commands but would be concerned about making the commands themselves less clear.

And adding `prebuild` as a prereq to `linux-app` in the Makefile is a good addition. @chenilim do you know if it would be appropriate to add it as a prereq to for the Mac & Windows apps too?

imasdekar:
@Pinjasaur the --prefix option is documented [here](https://docs.npmjs.com/cli/v7/using-npm/config#prefix)

Also, I fell that ideally the prebuild command should be added as a dependency to both windows and mac commands as well but I did not want to make a change which I was unable to test fully so I did not include it in my PR.

wiggin77:
@imasdekar thanks for your PRs.  I suspect this PR stems from issues in our developer docs which we need to address.  I apologize for any frustration that caused.  Regarding this PR:

1. the `prebuild` step was intentionally kept separate to speed up the build, run, repeat cycle.  We do need to make it more clear in the docs.

2. The `--prefix` option for npm is undocumented for the use case here.  It is documented for `install` only.  I'm wary of relying on it versus sticking with something that works for all versions of npm past and future.

Let me know if you strongly disagree with any of the above.
Thanks.

imasdekar:
@wiggin77 I do agree with both your points and we can close this MR.
We can update the docs with relevant info.
If no one is up for it at the moment then I can send an MR for the doc update.

imasdekar:
@wiggin77 I have created an MR #3245 where I have updated the readme with the instructions for running the prebuild step.
I will shortly send the same update to the mattermost-developer-documentation as well.

PR #3223, opened 2022-06-14
Fixed a bug where rmeoved board members showed up in comment list

### LINKED ISSUE #3002: Bug: Removed board member appears in user autocomplete list

1. Open a private board as user1
2. Open a card, and type `@` in the comment, see just user1 (good)
3. Open the board share dialog and add user2
4. Open a card, and type `@` in the comment, see user1 and user2 (good)
5. Open the board share dialog and **remove** user2
6. Open a card, and type `@` in the comment, see user1 and user2 (not good)

## Expected behavior

The user autocomplete list on a private board must always match the board members in the share dialog.

## Screenshots (optional)

https://user-images.githubusercontent.com/46905241/166007456-d2c251db-e8c3-4b92-86a8-50f9804ae9ab.mov

## Edition and Platform

 - Version: v0.16 (bugbash2)

The autocomplete list should be up to date each time it is opened.

harshilsharma63:
Fixes #3002


webapp/src/mutator.ts
```diff
@@ -384,9 +385,17 @@ class Mutator {
         await undoManager.perform(
             async () => {
                 await octoClient.deleteBoardMember(member)
+                const user = await octoClient.getUser(member.userId)
+                if (user) {
+                    store.dispatch(removeBoardUsers([user]))
```
jespino, line 390: Probably implementing this as `removeBoardUserById` or something like that, it would be less "homogeneous", but it would bet one server request less.

webapp/src/mutator.ts
```diff
@@ -384,9 +385,17 @@ class Mutator {
...
+                }
             },
             async () => {
                 await octoClient.createBoardMember(member)
+                const user = await octoClient.getUser(member.userId)
```
jespino, line 395: Probably here we can save of doing this call receiving the user directly in the `deleteBoardMember` function, actually that remove the need of the two requests, and this is only done in one place in the code where you have the user in the context, so it is not hard to do.

webapp/src/store/boards.ts
```diff
@@ -81,7 +81,9 @@ export const updateMembersEnsuringBoardsAndUsers = createAsyncThunk(
             }
             const user = await client.getUser(m.userId)
             if (user) {
-                thunkAPI.dispatch(addBoardUsers([user]))
+                const deleted = !m.schemeAdmin && !m.schemeEditor && !m.schemeViewer && !m.schemeCommenter
+                const processor = deleted ? removeBoardUsers : addBoardUsers
```
jespino, line 85: Personal opinion: This looks like DRY, but for me it is just less readable. For me is way cleaner to have something like:
```ts
if (delete) {
    thunkAPI.dispatch(removeBoardUsers([user]))
} else {
    thunkAPI.dispatch(adBoardUsers([user]))
}
```

jespino (changes requested):
Some things that I think can be improve but in general looks good to me.

webapp/src/mutator.ts
harshilsharma63: Makes sense. Done.

webapp/src/store/boards.ts
harshilsharma63: Done.

webapp/src/store/users.ts
```diff
@@ -59,6 +59,11 @@ const usersSlice = createSlice({
                 state.boardUsers[user.id] = user
             })
         },
+        removeBoardUsers: (state, action: PayloadAction<IUser[]>) => {
```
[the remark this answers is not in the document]
harshilsharma63: Done. Kept it as array because its just a nit.

webapp/src/mutator.ts
harshilsharma63: Done.

Pinjasaur (approved):
LGTM, going to defer merging immediately—not sure if you had any further feedback pending an approval @jespino?

jespino:
No more feedback from my side, everything looks good now.

jespino:
No more feedback from my side, everything looks good now.

jespino (approved):
LGTM

Issue #3217, opened 2022-06-13
Bug: "Share" button shown on Personal Desktop


ogi-m:
## Steps to reproduce the behavior

1. Run Boards Personal Desktop
2. Note the "Share" button in the top right

## Expected behavior

"Share" button shouldn't show on Personal Desktop

## Screenshots (optional)

![173419040-719e253c-b778-4192-9bec-6c862adfb8cf](https://user-images.githubusercontent.com/15314183/173421539-b2397481-fb71-4cbb-adf0-e55b0087af39.png)

## Edition and Platform

 - Edition: Personal Desktop
 - Version: v7.0
 - Browser and OS: Windows

Issue #3216, opened 2022-06-13
Bug: "Get your own free cloud server" banner won't dismiss on Personal Desktop


ogi-m:
## Steps to reproduce the behavior

1. Run Boards Personal Desktop
2. Note the "Get your own free cloud server"
3. Click 'x' to dismiss - nothing happens

## Expected behavior

The banner should be removed

## Screenshots (optional)

![image](https://user-images.githubusercontent.com/15314183/173419040-719e253c-b778-4192-9bec-6c862adfb8cf.png)

## Edition and Platform

 - Edition: Personal Desktop
 - Version: v7.0
 - Browser and OS: Windows

ogi-m:
This issue is specific to Personal Desktop 
cc @sbishel @wuwinson

PR #3227, opened 2022-06-14
Personal server fixes for cloud message

sbishel:
#### Summary
Fixes two issues - Personal server was not storing setting for hiding the cloud message, so the message never went away and the "Share" button was also being displayed.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/3217
  Fixes https://github.com/mattermost/focalboard/issues/3216


harshilsharma63 (approved):
LGTM 👍

PR #3232, opened 2022-06-15
Making postgres aware of schema name

jespino:


mgdelacroix (changes requested):
LGTM besides a var naming change required

mgdelacroix (approved):

PR #3238, opened 2022-06-16
Added constant template IDs for tracking template usage

harshilsharma63:
Fixes https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/0/cafgi9y86ifftzc3rkospofr14c/


wiggin77:
@harshilsharma63 this is already a field for this called `defaultTemplateId`

harshilsharma63:
@wiggin77 `defaultTemplateId` is used for storing the default card template of a view. Its different from what we're doing here. Please correct me if I'm missing something.

wiggin77:
> @wiggin77 `defaultTemplateId` is used for storing the default card template of a view. Its different from what we're doing here. Please correct me if I'm missing something.

Ok, I may be mis-remembering the name used, but I definitely saw some work to ensure all boards had a constant id coming from the template used to create the board.  This was done so telemetry could track template usage.  

The main issue with this PR, aside from whether or not it is duplicating work, is that when the templates are modified and re-exported they will lose the ids that were manually added here.  The templates are coming from a board on community where @wuwinson edits them.

wiggin77:
@wuwinson are you ok with the template ids added directly to the exported default templates?  If you modify those templates via Boards then export again, those ids will be gone.

sbishel:
> but I definitely saw some work to ensure all boards had a constant id coming from the template used to create the board. This was done so telemetry could track template usage.

@wiggin77 Any idea of how this was supposed to be handled. My (quick) look through the code, I don't see where we handle anything like that. 

> are you ok with the template ids added directly to the exported default templates? If you modify those templates via Boards then export again, those ids will be gone.

If we did it this way, it should happen automatically, or some other way than manually adding these settings to the exported JSON.

Chen had the idea of hashing the title, but also wants to some day translate the templates, so not sure that is a great solution either.

wiggin77:
> Any idea of how this was supposed to be handled. My (quick) look through the code, I don't see where we handle anything like that.

@sbishel  I can't find it either. I do recall code that copied the template's id into a field in a new board created by the template, and that field was specifically for telemetry.  I don't recall if that was one of my PR's or one from @jespino that I reviewed.  Perhaps he remembers.  It may not have made it past permissions branch.

The intention was that via telemetry we could count all the boards created with any default template.  Since the default templates are exported from a workspace that @wuwinson uses, the ids would stay the same through edits.  Also there would be no work required when default templates are added or deleted.

I cannot find the code now that sets the template id field when creating boards from a template.  It either was replaced/removed or didn't get merged from permissions branch.

sbishel:
@wiggin77 @harshilsharma63 @jespino - I found the original PR that went into the permissions branch. It looks like it part of this didn't make it into `Main`. However, I'm not sure it is complete. It appears to rely on the "TemplateID" or "BuiltInTemplateID" field to be included in the JSON as well. The templates themselves don't appear to have been updated. 

https://github.com/mattermost/focalboard/pull/2060/files

webapp/src/components/boardTemplateSelector/boardTemplateSelector.tsx
```diff
@@ -98,7 +98,8 @@ const BoardTemplateSelector = (props: Props) => {
     }
 
     const handleUseTemplate = async () => {
-        await mutator.addBoardFromTemplate(currentTeam?.id || Constants.globalTeamId, intl, showBoard, () => showBoard(currentBoardId), activeTemplate.id, currentTeam?.id)
+        const templateID: string = activeTemplate.properties.templateID as string
+        await mutator.addBoardFromTemplate(currentTeam?.id || Constants.globalTeamId, intl, showBoard, () => showBoard(currentBoardId), templateID, currentTeam?.id)
```
jespino, line 102: I think this is not what I expect. I understand that from one side we have the `template.id` that refers the id in the instance, and we should have a custom `trackingTemplateId` or something like that, that is only set for the built-in templates. And then, use that whenever we are tracking that information. So, the trackingID is only used for tracking purposes, not for using the template in any way.

harshilsharma63: Done.

wiggin77: Isn't this going to pollute telemetry with the template ID of every board created from templates, not just the default (built-in) templates?  I don't think the requirement is to track user generated templates.

harshilsharma63: @wiggin77 I see your point.

@wuwinson do we wanna track user-generated templates? I'm thinking that we should so we know how much users use the board template feature.

chenilim:
@harshilsharma63, it's not a requirement to track user generated template IDs. If we do, it should be non-identifiable. cc @wuwinson.

wiggin77 (changes requested):
Let me know when the changes are done.  Thanks.

harshilsharma63:
@wiggin77 you can review it now.

wiggin77 (approved):
LGTM :+1:

PR #3240, opened 2022-06-16
Fix failure to upgrade from 7.0 to 7.2

### LINKED ISSUE #3241: Bug: Failed to upgrade from 7.0 to 7.2

1. Install Boards plugin 7.0
2. Upgrade to the latest main and notice migration error

## Expected behavior

Upgrade from 7.0 to 7.2 should work smoothly.

## Edition and Platform

 - Version: 7.2

The error occurred because the SQL migration for adding `focalboard_file_info` table was numbered `17` in 7.0 but it was 23 in `7.2`, causing the migration to be attempted again and fail.

harshilsharma63:
Fixes #3241


mgdelacroix (commented):
I think we should be able to safely remove the `placeholder` migration and not modify the constants, as that file was created just to be in place of the number 17

mgdelacroix (approved):
LGTM! :+1:

PR #3242, opened 2022-06-16
Board switcher keyboard shortcut navigation

### LINKED ISSUE #2755: Permissions Branch: Board switcher keyboard nav not working

1. Go to board switcher
2. Try using the keyboard to navigate through the switcher
3. See error

## Expected behavior
See [spec](https://docs.google.com/document/d/1Kcc5KVGLwBzry_mVGglY7kDpm1Zr1TQ3m54eUpJeEmc/edit#bookmark=id.ogjvzt85mfao) for details
1. `Cmd+K` should open the search dialog (This is working as expected, but 2-4 below are not working)
2. `Up/Down` keys should move up and down through the search results
3. `Enter` key should select the board and dismiss the modal
4. `Esc` key should dismiss the modal and return user to the view below the modal

## Screenshots (optional)
Screen Shot 2022-04-08 at 3 04 45 PM

## Edition and Platform

Pinjasaur:
#### Summary

Implements esc/up/down/enter keyboard shortcuts for better navigating the Boards switcher—similar to Channels.

#### Ticket Link

closes #2755 

https://user-images.githubusercontent.com/6335792/174140943-b6774788-25b1-49cd-9d05-bb9dcbd940eb.mov


webapp/src/components/boardsSwitcher/boardsSwitcher.tsx
```diff
@@ -40,12 +40,21 @@ const BoardsSwitcher = (props: Props): JSX.Element => {
...
+    const handleEscKeyPress = (e: KeyboardEvent) => {
+        if (Utils.isKeyPressed(e, Constants.keyCodes.ESC)) {
+            e.preventDefault()
+            setShowSwitcher(false)
+        }
+    }
+
     useEffect(() => {
         document.addEventListener('keydown', handleQuickSwitchKeyPress)
+        document.addEventListener('keydown', handleEscKeyPress)
```
jespino, line 52: Is there any reason to not use the react-hotkeys-hook library? I think is more straight forward.

webapp/src/components/boardsSwitcherDialog/boardSwitcherDialog.tsx
```diff
@@ -81,6 +85,7 @@ const BoardSwitcherDialog = (props: Props): JSX.Element => {
             title={title}
             subTitle={subTitle}
             searchHandler={searchHandler}
+            refs={refs}
```
jespino, line 88: All this refs passings looks a bit convoluted to me. Couldn't it be something like an state at this level with the selected id? and an onSelect callback or something like that?

I always find passing refs down the components really hard to track.

webapp/src/components/searchDialog/searchDialog.tsx
```diff
@@ -36,6 +41,41 @@ const SearchDialog = (props: Props): JSX.Element => {
...
+            e.preventDefault()
+            props.refs.current[selected].current.click()
+        }
+    }
+
+    useEffect(() => {
+        if (selected >= 0)
+            props.refs.current[selected].current.parentElement.focus()
+
+        document.addEventListener('keydown', handleUpDownKeyPress)
```
jespino, line 69: Again, probably I would prefer to use the hook `useHotkeys` here.

webapp/src/components/boardsSwitcher/boardsSwitcher.tsx
Pinjasaur: IIRC I tried that and similar to the Cmd/Ctrl + K above, that library has difficulty with shortcuts that are native to the browser/OS.

webapp/src/components/boardsSwitcherDialog/boardSwitcherDialog.tsx
Pinjasaur: I agree, it wasn't as straightforward as I was expecting it to be. My reasoning for using `ref`s was it let me use the native `.focus()` and `.click()` methods on the underlying native elements.

Since the parent component (this) defines the search handler which returns the JSX components and passes that definition into the child component which calls it and does the rendering, it seemed to make sense to define the `ref`s in the parent so the child could use the native methods I mentioned above. `onSelect` did come to mind, but it's not available for all elements, only `<input>` and `<textarea>`.

webapp/src/components/searchDialog/searchDialog.tsx
```diff
@@ -36,6 +41,41 @@ const SearchDialog = (props: Props): JSX.Element => {
 
     const emptyResult = results.length === 0 && !isSearching && searchQuery
 
+    const handleUpDownKeyPress = (e: KeyboardEvent) => {
+        if (Utils.isKeyPressed(e, Constants.keyCodes.DOWN)) {
+            e.preventDefault()
+            if (results.length > 0)
+                setSelected(Utils.clamp(selected + 1, 0, results.length - 1))
```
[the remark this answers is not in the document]
Pinjasaur:
```suggestion
                setSelected(((selected + 1) <= (results.length - 1)) ? selected + 1 : selected)
```

e.g. a ternary operator or something else?

Honestly, this might have been tech debt from when I was implementing this PR. I thought initially I would need the up/down functionality to wrap so I was using a modulo operator (`%`) but then realized that's now how Channels switcher worked so I changed to the `clamp` approach and didn't think twice. Happy to remove code though. 🙂

webapp/src/components/searchDialog/searchDialog.tsx
Pinjasaur: Yeah, good call out. Like what I mentioned above, after fighting with native browser/OS shortcuts I did the event listeners natively and didn't consider the hotkey library. I'll see about getting it refactored.

webapp/src/components/boardsSwitcher/boardsSwitcher.tsx
Pinjasaur: Yep, verified again locally and it doesn't work with hotkeys because of being focused on an input field.

webapp/src/components/searchDialog/searchDialog.tsx
Pinjasaur: Tried locally here too and same issue as above, hotkeys doesn't work in this case because focused on an `<input>`.

Pinjasaur:
@jespino friendly ping to re-review when you have time 🙂

webapp/src/components/searchDialog/searchDialog.tsx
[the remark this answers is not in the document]
jespino: Yes, that would work for me, also I would use a bit simpler expression like `(selected+1) < results.length`.

webapp/src/components/boardsSwitcherDialog/boardSwitcherDialog.tsx
jespino: Ok, what I meant is. move the `selected` state to the `BoardSwitcherDialog` component, update the `selected` state using a callback that is passed to the SearchDialog component, and then, you can use a `useEffect` call to update the focus on every `selected` change. The syntethic click, is not going to be needed, because you are already there, and you have access to the `selectBoard` function. You are going to still need the refs, but you are not passing them around that is what makes it more confusing.

webapp/src/components/searchDialog/searchDialog.tsx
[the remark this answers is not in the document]
Pinjasaur: Done. 👍🏻

sbishel:
@jespino - Would you review this again?

jespino (approved):
It still looks to me a bit complicated code, but at least no passing a list of refs makes it more contained. LGTM

harshilsharma63 (approved):
LGTM 👍

asaadmahmood (approved):

Issue #2933, opened 2022-04-26
Enhancement: Add User Full Name, Nickname


sbishel:
## Summary

We need to retrieve and add the user's online status and full name in order to match Channels. Currently, we do not retrieve this information from channels.

## How important this is to me and why

Importance: Medium - match channels UI

## Current Implementation
![Screen Shot 2022-04-26 at 4 46 24 PM](https://user-images.githubusercontent.com/12704875/165405072-827a0425-491c-40b3-b35e-77ccd54dd426.png)

Any examples of good implementations of this capability.
Per spec

https://www.figma.com/file/bu8YbzRGfO1AEw9pmQyUSR/Focalboard-Permissions?node-id=34%3A9382

![Screen Shot 2022-04-26 at 4 44 46 PM](https://user-images.githubusercontent.com/12704875/165404964-ebd2bed8-8b80-4e7a-b37d-da840b4211f8.png)

sbishel:
There is also a configuration setting `TeamSettings.TeammateNameDisplay` that we should honor.

sbishel:
Created a separate issue for status portion as it is more difficult and would be easier to implement after multiproduct architecture. https://github.com/mattermost/focalboard/issues/3255

PR #3244, opened 2022-06-16
User updates

sbishel:
#### Summary
Returns LastName, FirstName, and Nickname for channels users.
Returns config setting for `TeamSettings.TeammateNameDisplay`
Uses setting and displays users in Permissions and User property as requested.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/2933


sbishel:
Screen Shot 2022-06-20 at 12 44 51 PM
Screen Shot 2022-06-20 at 12 41 42 PM
Screen Shot 2022-06-20 at 12 41 33 PM

harshilsharma63 (approved):

harshilsharma63:
@sbishel it works fine but I was confused with it because I thought it would read the user's Teammate Name Display" setting. If it's possible, it would be nicer to use the user's preference of how they wanna view a teammate's name rather than the team-level setting. Channels seem to be doing the same - preferring user-level settings over team-level.

This can be done later too.

sbishel:
Thanks @harshilsharma63 - I didn't realize there was a personal setting. I agree we should be using this on. I will update.

harshilsharma63:
@sbishel I'm not 100% sure whats the exact logic channels use to determine which setting to use finally, so worth checking channel's implementation.

harshilsharma63:
Also another extension for this would be the @ mention autocomplete list in comment adn description.

sbishel:
Added to @mentions as well.
Screen Shot 2022-07-05 at 4 59 31 PM

harshilsharma63 (approved):
LGTM 👍

Issue #3050, opened 2022-05-10
Bug: Only able to delete property if specific criteria is met


j6b72:
I am not able to delete a property on a block unless I fill its value when editing a block directly and then delete it

## Steps to reproduce the behavior

1. Create a new standalone focalboard instance with the docker image ( `podman run docker.io/mattermost/focalboard:latest` )
2. Register a new user and create an empty board
3. Create a new property of any type either inside a new block template or on a new block
4. Try to delete that property in any other way than 1) create a new block, 2) fill in a value and then 3) attempt to delete it

## Expected behavior

either
1) The property also gets deleted when attempting to do that
- without having a value set or
- from the template editor

or

2) The correct error is shown instead of a success message

## Screenshots (optional)

https://user-images.githubusercontent.com/65040021/167655581-49c66b06-7932-41a5-b153-d1d044aacd37.mp4

This request is shown as failed in the network tab of the dev tools everytime such a non-working deletion is attempted:

<img src="https://user-images.githubusercontent.com/65040021/167656315-57ed8fd5-e205-4791-ae52-fe27cb68da92.png" width=400 />
<img src="https://user-images.githubusercontent.com/65040021/167656327-fa48e8a1-027c-4ff3-a73d-d56e865c2e9a.png" width=400 />
<img src="https://user-images.githubusercontent.com/65040021/167656347-882f16ed-965b-4ab9-9988-5d512c6f7788.png" width=400 />

## Edition and Platform

 - Edition: Personal Server
 - Version: docker image tags: latest, also reproducible on 0.16.0
 - Browser and OS: Latest brave on linux / [Version 1.38.111 Chromium: 101.0.4951.54]

## Additional context (optional)
nginx in front of the focalboard server with
1) the options provided here: https://www.focalboard.com/download/personal-edition/ubuntu/
2) certificates for https 
3)  `add_header X-Robots-Tag "noindex, nofollow, nosnippet, noarchive" always;`
4) `server_tokens off;`

PR #3249, opened 2022-06-17
GH-3050 - Fix deleting empty property

sbishel:
#### Summary
When trying to delete a new property with no values, an error is thrown because there are no blocks to patch only a board. Remove the check for blocks length > 0

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/3050


wiggin77 (approved):
LGTM :+1:

PR #3252, opened 2022-06-20
WIP: Ported view limits to main

harshilsharma63:


server/api/api.go
```diff
@@ -34,6 +34,8 @@ const (
 	ErrorNoTeamMessage = "No team"
 )
 
+var errEmpty = errors.New("")
```
jespino, line 37: This looks really wrong, I would like to now why something is failing, In this case I would say something like "Not supported API in standalone mode" or something like that.

mattermost-plugin/webapp/src/components/cloudUpgradeNudge/cloudUpgradeNudge.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <div className='PostTypeCloudUpgradeNudge'>
+            <span>{props.post.message}</span>
+            <div
+                style={ctaContainerStyle}
+            >
+                <button
+                    onClick={ctaHandler}
+                    style={ctaBtnStyle}
+                >
+                    {'View upgrade options'}
```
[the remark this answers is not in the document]
harshilsharma63: Custom post types are not easy to translate. There's a separate ticket tracking this task as its wasn't quick and easy to be done as part of 7.2.

server/api/api.go
harshilsharma63: Done.

mattermost-plugin/server/plugin.go
```diff
@@ -109,7 +109,7 @@ func (p *Plugin) OnActivate() error {
 		return fmt.Errorf("error initializing the DB: %w", err)
 	}
 	if cfg.AuthMode == server.MattermostAuthMod {
-		layeredStore, err2 := mattermostauthlayer.New(cfg.DBType, sqlDB, db, logger, p.API)
+		layeredStore, err2 := mattermostauthlayer.New(cfg.DBType, sqlDB, db, logger, p.API, client)
```
jespino, line 112: We don't need to pass it here, we can create a new client inside using the p.API, the client it is just a wrapper that provide some helpers. so passing it around all the time is not specially useful.

My suggestion here would be, or we pass the APIClient directly instead of p.API, and then use it, or we create it inside.

Probably I prefer the former, but I would do the later for this PR, and then convert it in to the other one in another PR.

webapp/src/components/viewLImitDialog/viewLimitDialog.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+    return (
+        <Dialog
+            className='ViewLimitDialog'
+            onClose={props.onClose}
+        >
+            <div className='ViewLimitDialog_body'>
+                <img
+                    src={Utils.buildURL(upgradeImage, true)}
+                    alt={'upgrade'}
```
jespino, line 100: not translated text

webapp/src/components/viewLImitDialog/viewLimitDialog.tsx
```diff
@@ -0,0 +1,128 @@
...
+                    {subtext}
+                </p>
+            </div>
+            <div className='ViewLimitDialog_footer'>
+                <Button
+                    size={'medium'}
+                    className='cancel'
+                    onClick={props.onClose}
+                >
+                    {intl.formatMessage({id: 'ConfirmationDialog.cancel-action', defaultMessage: 'Cancel'})}
```
jespino, line 115: Why intl.formatMessage instead of `FormattedMessage`?

jespino (changes requested):
Some changes suggested, but nothing mayor.

mattermost-plugin/server/plugin.go
jespino: We talked it trhoguh mattermost and the fact that we also need the `p.Driver` inside, makes no sense to this change until we make a refactor to use always the APIClient all the time, and that should be in another PR.

harshilsharma63: 👍

webapp/src/components/viewLImitDialog/viewLimitDialog.tsx
harshilsharma63: Done.

jespino (approved):
now LGTM. Thanks!

mattermost-plugin/server/plugin.go
wiggin77: @jespino @harshilsharma63 , FYI we are no longer using client (plugin api wrapper) anywhere in FocalBoard.  For the multi-product architecture I've created a ServicesAPI that has all the methods needed to interact with mmserver.  There are adapters that allow the plugin API and the new product APIs to implement the ServicesAPI.

Issue #3059, opened 2022-05-12
Bug: Name of Property is not saved the first time you add it


lindy65:
Reproduced in Rainforest 

## Steps to reproduce the behavior

1. Create a new card in Table View of a board
2. Click on "+ Add a property" then select "Text"
3. Replace "Text" in the field with "Estimate" then hover over "Type: Text" and select "Number"
4. Observed: The field where you replaced "Text" with "Estimate" still shows "Text"

NOTE: If I go back and edit the "Text" field and replace "Text" with "Estimate", the word "Estimate" is saved

## Expected behavior

The field where you replaced the value of "Text" with "Estimate", shows "Estimate" on first attempt

## Screenshots (optional)

![image](https://user-images.githubusercontent.com/13020833/168137194-13576255-b0ef-41b4-b021-14e2f2b15991.png)

![image](https://user-images.githubusercontent.com/13020833/168137413-4ecd5324-a638-4c2f-9e65-c41e7be3a87f.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: [e.g. v0.17.0]
 - Browser and OS: Chrome on Windows

PR #3256, opened 2022-06-20
Store current input while user is changing.

sbishel:
#### Summary
Receives callback to store current input as user types. Uses this value when saving a type change, in case the property name change hasn't occurred.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/3059


jespino (changes requested):
I'm having doubts about it, is there any reason to not do it in more normal way. for example, we can change the `onValueChanged` property to `onConfirmValue` or something like that, and create a new `onValueChanged` that is triggered with the `setValue` in the input's `onChange`, that way this is more a self contained component and you don't need to do weird ref stuff.

sbishel:
> I'm having doubts about it, is there any reason to not do it in more normal way. for example, we can change the `onValueChanged` property to `onConfirmValue` or something like that, and create a new `onValueChanged` that is triggered with the `setValue` in the input's `onChange`, that way this is more a self contained component and you don't need to do weird ref stuff.

Updated based on your suggestion, much cleaner....

jespino (approved):
LGTM

Issue #2645, opened 2022-03-25
Reenable Cyclomatic Complexity check


sbishel:
## Summary
Determine if and how we can reimplement the cyclomatic complexity check.

The cyclomatic complexity check was removed here. 
https://github.com/mattermost/focalboard/pull/2557

it was complaining about cyclomatic complexity exceeding the default threshold of 30 for the data migration function (which currently handles about 4-5 kinds of data migrations). IN the future, I see only more data migrations being added there, so increasing the linter's threshold value won't work in the long term.

I can optimize the code to reduce the conditions and returns paths but it won't work for long when we add the next data migration.

Hence, I removed the linter altogether. I'm open to any alternate solution.

sbishel:
This was disabled in the permissions branch. `release-7.0` its still enabled.

PR #3257, opened 2022-06-20
add cyclo complexity, update migrate() to use defer

sbishel:
#### Summary

Reimplement Cyclomatic complexity check 
Update migrate() to use defer in order to reduce CC.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/2645

### DIFF
--- a/mattermost-plugin/.golangci.yml
+++ b/mattermost-plugin/.golangci.yml
@@ -67,6 +67,7 @@ linters:
     - unconvert
     - unused
     - whitespace
+    - gocyclo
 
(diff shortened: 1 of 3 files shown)


jespino (approved):
LGTM

jespino:
@sbishel Actually, seeing the code I think it can be related to the initial startup problem in the cloud, becuase the log was removed before the "ApplyAll", so maybe that was causing other nodes to start applying the migrations before the other node has all the migrations applied.

Issue #3259, opened 2022-06-21
Bug: Creating/deleting/opening a board in the linux desktop app opens a blank window with console error


imasdekar:
## Steps to reproduce the behavior

1. Follow the [dev-setup guide](https://developers.mattermost.com/contribute/focalboard/personal-server-setup-guide/#building-and-running-standalone-desktop-apps) for Linux,
2. run `focalboard-app`
3. Click on use this template button or on one of the already existing boards or delete an existing board from the left side bar. User is navigated to a blank window  with the following error in the background

``` shell

http://localhost:43029/static/main.js?1e4d077b3e1976e51dfd:321:82523: CONSOLE ERROR TypeError: null is not an object (evaluating 'n.featureFlags.disableTour')
http://localhost:43029/static/main.js?1e4d077b3e1976e51dfd:28:41701: CONSOLE JS ERROR Unhandled Promise Rejection: TypeError: null is not an object (evaluating 'n.featureFlags.disableTour')
 
```

## Expected behavior

Depending on the user action i.e. create/delete/open the user should be shown the respective successful or predefined/expected erroneous next step.

## Screenshots (optional)
![Screenshot_20220621_122422](https://user-images.githubusercontent.com/29521808/174737626-56fd2587-a711-482c-a6bd-3f449538495b.png)

## Edition and Platform

 - Edition: Personal Desktop
 - Version: latest commit from the main branch
 - Browser and OS: ArcoLinux

PR #3260, opened 2022-06-21
fix: handles clientConfig causing board error

imasdekar:
#### Summary

Adds a check to see if the clientConfig object is null or empty.
If null or empty clientConfig then sets the disableTour to false.

#### Ticket Link
Fixes: #3259

Signed-off by: imasdekar <imasdekar@disroot.org>


sbishel (approved):
Thanks @imasdekar, LGTM

PR #3315, opened 2022-07-06
Refactor websockets state and lifecycle

mgdelacroix:
#### Summary
This PR moves the state of the authentication and subscriptions to the websockets client, allowing for multiple components to communicate with it and request subscriptions independently. With this change, the lifecycle of the websockets client is now managed on a component, and a hook is provided for easy access to it from individual components.

These changes allow as well for components exported through the plugin API to use the Focalboard websockets client.


jespino (commented):
I like what I see here, the interface is much cleaner

server/services/store/sqlstore/board.go
```diff
@@ -684,6 +684,52 @@ func (s *SQLStore) searchBoardsForUser(db sq.BaseRunner, term, userID string) ([
 	return s.boardsFromRows(rows)
 }
 
+// searchBoardsForUserInTeam returns all boards that match with the
+// term that are either private and which the user is a member of, or
+// they're open, regardless of the user membership.
+// Search is case-insensitive.
+func (s *SQLStore) searchBoardsForUserInTeam(db sq.BaseRunner, teamID, term, userID string) ([]*model.Board, error) {
```
harshilsharma63, line 691: This is almost completely a copy of `SearchBoardsForUser` in MM auth layer. Can one function re-use the other one as the only difference is including team id check?

harshilsharma63: This is for both store and app layers.

harshilsharma63 (changes requested):

harshilsharma63 (approved):

harshilsharma63:
The comments can be addressed in a separate PR, independent from this.

PR #3317, opened 2022-07-06
update webapp dependencies

sbishel:
Update webapp dependencies


Pinjasaur (approved):
LGTM. Is the `experiments/webext` used directly anywhere? I did a quick search and it didn't appear to be, just curious

PR #3321, opened 2022-07-06
Refactor "..." menu into component & add to Calendar view

### LINKED ISSUE #808: Update Menu to common component
The menus in CardDialog.tsx, GalleryCard.tsx, and KanbanCard.tsx are all very similar and in two cases identical. 

We should create a common component for the menu and reuse on each card. All cards should support Delete, Duplicate, Copy Link. Additional menu items such as for CardDialog should be able to be added.

### LINKED ISSUE #2324: UX: Add ... menu to cards in Calendar view
## Summary

Cards shown in the Calendar view should have a ... menu on hover similar to the Kanban and Gallery view, with the same options (delete, duplicate, copy link).

Importance: Medium

Pinjasaur:
#### Summary

Refactors the menus from CardDialog.tsx, GalleryCard.tsx, and KanbanCard.tsx into a single component + adds it to Calendar view.

#### Ticket Link

closes #2324 
closes #808 

image


jespino (changes requested):
The only thing that I see is needed is to rename the `optionsMenu` to something that is more meaningful. Options menu can be any menu in the system that has options, so... basically any menu. I would use something like `cardActionsMenu` or something like that. Apart from that, I really like the refactor here. :)

Pinjasaur:
> The only thing that I see is needed is to rename the `optionsMenu` to something that is more meaningful. Options menu can be any menu in the system that has options, so... basically any menu. I would use something like `cardActionsMenu` or something like that. Apart from that, I really like the refactor here. :)

@jespino done 🚀

harshilsharma63 (approved):
Looks awesome! LGTM 👍

jespino (approved):
Great, now LGTM. Thanks

PR #3322, opened 2022-07-06
Update telemetryClient.ts

wuwinson:

### DIFF
--- a/webapp/src/telemetry/telemetryClient.ts
+++ b/webapp/src/telemetry/telemetryClient.ts
@@ -10,7 +10,7 @@ export const TelemetryActions = {
     ClickChannelHeader: 'clickChannelHeader',
     ClickChannelIntro: 'channelIntro_boardLink',
     ViewBoard: 'viewBoard',
-    CreateBoard: 'createBoard',
+    CreateEmptyBoard: 'createBoard',
     DuplicateBoard: 'duplicateBoard',
     DeleteBoard: 'deleteBoard',
     DeleteBoardTemplate: 'deleteBoardTemplate',


jespino:
It is not that simple, if you explain me the intention here, I can update the whole code to accept properly this change.

wuwinson:
@jespino I was trying to update to telemetry ID so it's a bit clearer on what is actually being tracked through that ID. I assumed only the section between apostrophe was referenced in the code and not the part I updated. Not worth the trouble if this requires updating the whole code.

jespino:
@wuwinson actually it is all the way around, there is nothing wrong in updating any side, but I going to explain. The left side, is the code part, everywhere you use that in the code, you should be using the left side. The right side is the event name that is send for telemetry, if we change the right side, the code doesn't gets affected, but the telemetry is going to start counting it as a different event, and is going to "break" the stats (break here means, start counting from zero the new event type, and leave untouch the old event type).

Renaming this in the code can be really useful from our side, and I'm ok doing it. And I guess we want to preserve the event name to not mess the stats, right?

wuwinson:
Yep, that makes sense, thanks for clarifying @jespino! We do indeed want to preserve the event name to not mess up the stats. If it's not too much trouble renaming it in the code, then it would be great to have it updated so it's clearer on what the ID is tracking. Thanks!

chenilim:
Hi @wuwinson, late to this, but in general we should avoid changing existing event IDs, because tracking them across the change will be very difficult. In this case, as long as we know that "CreateBoard" means "an empty board was created", that's ok. I advise not making this change.

In the future, let's say we need to add a "CreateNonEmptyBoard" (hypothetically), we can add that.

wuwinson:
Sounds good @chenilim, closing this out.

PR #3261, opened 2022-06-21
Cherry pick hidden card btn actions

Rajat-Dabade:
#### Summary

Cherry pick to `main` for this PR: https://github.com/mattermost/focalboard/pull/3180

PR #3157, opened 2022-05-31
fix: delete `.action-cell` CSS bg color

Pinjasaur:
#### Summary

Deletes the `.action-cell` CSS `background-color` to allow for a proper cascade—previously a UI regression.

#### Ticket Link

Closes #3081.

PR #3323, opened 2022-07-06
fix: delete `.action-cell` CSS bg color

Pinjasaur:
#### Summary

Re-doing #3157 which regressed in #3261. Deleting this CSS allows for a proper cascade which otherwise causes unwanted styling depending on what theme you're viewing the UI in.

#### Ticket Link

n/a


harshilsharma63 (approved):
LGTM 👍

PR #3326, opened 2022-07-07
WIP: Memomised views

harshilsharma63:


wiggin77:
@harshilsharma63 is this something we want to merge even though it does not fix the Cypress test?

PR #3332, opened 2022-07-07
Refactor `z-index` values to a SCSS map

### LINKED ISSUE #2746: Code debt: refactor `z-index`s into SCSS map
## Summary

Refactor CSS `z-index` properties into a single SCSS `map`.

When reviewing #2727 I suggested it as a future refactor / code debt item. The benefit is it consolidates all the `z-index` occurrences into a single place to help manage them as ultimately they're effectively context-less arbitrary integers otherwise.

Importance: Low

1. Create a new file (or re-use an existing centrally `@import`-ed one) e.g. `_z-index.scss` which contains a map e.g.

	  ```scss
	  // Keep in descending order to aid in visual grepping
	  //
	  // Usage:
	  // map-get($z-index, <key>)
	  
	  $z-index: (
	    // key: value
	  );
	  ```

2. Refactor `z-index` occurrences to the new `map-get($z-index, key)` pattern

Links:
- https://jonsuh.com/blog/organizing-z-index-with-sass/
- https://css-tricks.com/handling-z-index/

Pinjasaur:
#### Summary

Refactors `z-index` values to a SCSS map that can be referenced directly, through a function, or via a mixin e.g.

```scss
.selector {
	z-index: map-get($z-index, key);
	// or
	z-index: z-index(key);
	// or
	@include z-index(key);
}
```

Note that this PR doesn't de-duplicate the actual values.

#### Ticket Link

closes #2746


jespino (approved):
I really like it, would be also great to have a css linter check to avoid usage of z-index directly.

asaadmahmood (approved):
Nice, I think this looks good!

Pinjasaur:
> I really like it, would be also great to have a css linter check to avoid usage of z-index directly.

Poked around with `stylelint` a bit and came up with a basic solution—the side effect being the `@include z-index(key)` mixin must be used to avoid `z-index` property directly. Lmk if you had something different in mind or if that works for you. @jespino

harshilsharma63 (approved):
Love it! LGTM 👍

Issue #3036, opened 2022-05-05
Bug: Custom template's default card template doesn't work


chenilim:
## Steps to reproduce the behavior

1. Create a custom board template
2. Add a card, and convert it to a card template
3. Make that card template the default one
4. Create a board from the board template
5. Click "New" to add a card
6. Nothing happens

## Expected behavior

A card with the default template is created

## Screenshots (optional)

https://user-images.githubusercontent.com/46905241/167014986-e2cb77f3-638d-4a10-b761-096191f11607.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v0.16 (bugbash2 server)
 - Browser and OS: Chrome on Mac

## Additional context (optional)

There is a 404 error in the network tab:
<img width="1281" alt="Screen Shot 2022-05-05 at 12 55 34 PM" src="https://user-images.githubusercontent.com/46905241/167015272-a9853ee3-2fbe-4a0c-9054-45102e2f2cef.png">

Guess is that the default card template ID isn't correctly updated on duplicate.

Rajat-Dabade:
@chenilim, I was trying to reproduce this issue but was not able to, can you please check if the issue still exists. If yes, please brief the step to reproduce this issue. Thank you.

https://user-images.githubusercontent.com/26198062/174249092-2e4b2978-a159-4330-a1cd-32168c344d5f.mov

chenilim:
@Rajat-Dabade, did you try to repro this on a main (7.2 target) build? The original repro was on 0.16, which was a branch off main.

thomaslc66:
I will open a new Issue, but this look almost the same: 
- If you create a template or save one as a template
- Then you set it as default
1) If you delete something in a task based on that template. The template is also modified
2) If you click on the + New at the bottom of a column, the template will not be selected but it will show something else.

PR #3335, opened 2022-07-08
Custom template's default card template works

Rajat-Dabade:
#### Summary
This PR will make the Custom template's default card template to be used when the new button is clicked for the first time

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/3036


Rajat-Dabade:
@mgdelacroix I have set all the defaultTemplateID to an empty string so that it will point to the empty card whenever someone uses these predefined templates as suggested by @wuwinson.

sbishel (approved):
Go ahead and assign yourself a ticket to create unit tests for `blockid.go`. It can be done in the next week. 

Also, I think you need to run 
`make templates-archive` 
when you update the templates. And update version [here](https://github.com/mattermost/focalboard/blob/main/server/assets/templates-boardarchive/version.json)

@wiggin77 can confirm. 

Otherwise LGTM.

mgdelacroix (approved):
Code LGTM! :+1: I was just about to suggest creating an unit test before I saw @sbishel's comment

PR #3344, opened 2022-07-09
[GH-3318] Fix flaky cypress test

### LINKED ISSUE #3318: Bug: Cypress test failing for main.
`webapp/cypress/integration/cardURLProperty.ts` is failing consistently in CI.  PR https://github.com/mattermost/focalboard/pull/3316 has temporarily disabled the problematic part of the test.  The test should be fixed and re-enabled.

kamre:
#### Summary
Fix Cypress test `cardURLProperty`:
- during the creation of a new board view and switching to it there are many API requests executed in effects for `BoardPage` component
- some requests trigger additional rerendering of the `BoardPage` component
- Cypress starts to work with the `BoardPage` immediately when on its first render
- additional rerenders after requests can lead to remount of some components related to menu
- so in order to workaround the issue waiting for the last request (which is `getUser`) was introduced in the Cypress test
- after this last request `BoardPage` component doesn't have additional rerenders and Cypress should not have issues with detached elements 

#### Ticket Link
Fixes #3318

### DIFF
--- a/webapp/cypress/integration/cardURLProperty.ts
+++ b/webapp/cypress/integration/cardURLProperty.ts
@@ -72,7 +72,6 @@ describe('Card URL Property', () => {
         cy.findByRole('button', {name: 'Edit'}).should('exist')
         cy.findByRole('button', {name: 'Copy'}).should('not.exist')
 
-        /*        
         // Add gallery view
         addView('Gallery')
         showURLProperty()
@@ -82,7 +81,6 @@ describe('Card URL Property', () => {
         cy.findByRole('link', {name: changedURL}).realHover()
         cy.findByRole('button', {name: 'Edit'}).should('not.exist')
         cy.findByRole('button', {name: 'Copy'}).should('exist')
-*/
 
         // Add calendar view
         addView('Calendar')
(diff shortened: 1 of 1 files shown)


kamre:
Cypress test [still fails](https://github.com/mattermost/focalboard/runs/7265891811?check_suite_focus=true) but at least the error is now the original one [described here](https://github.com/mattermost/focalboard/issues/3318#issuecomment-1177174629).

So the problem with opening submenu is not reproduced anymore, and the changes here fixed only the submenu problem. Looks like webapp itself should be fixed _avoiding redundant rerenderings_ with remounting react components.

wiggin77:
> Cypress test [still fails](https://github.com/mattermost/focalboard/runs/7265891811?check_suite_focus=true) but at least the error is now the original one [described here](https://github.com/mattermost/focalboard/issues/3318#issuecomment-1177174629).
> 
> So the problem with opening submenu is not reproduced anymore, and the changes here fixed only the submenu problem. Looks like webapp itself should be fixed _avoiding redundant rerenderings_ with remounting react components.

Thanks for looking this @kamre.  

cc @harshilsharma63

kamre:
@wiggin77 @harshilsharma63 I have decided to make the test more robust to rerenders via introducing wait for the last API request, so that `BoardPage` component doesn't rerender unexpectedly while Cypress is performing user actions (look at the updated summary for pull request for details).

P.S. do you know why workflows are blocked and awaiting for approval?

wiggin77:
> P.S. do you know why workflows are blocked and awaiting for approval?

I'll look into this.

wiggin77 (approved):
Great work @kamre.  I'll raise a ticket to look at why we are doing redundant renders.

PR #3347, opened 2022-07-11
Cypress wait

harshilsharma63:


harshilsharma63:
This is not needed. See this thread for reference - https://community.mattermost.com/core/pl/8djkbk9o97yozr5tod3oedfq1y

PR #3356, opened 2022-07-13
Refactor of prorperties

jespino:
This is a pretty big PR. I'm going to try to explain the logic here.

I'm moving all the properties to a specific place that makes them more or less "agnostic" from the rest of the app. The property define itself, and the rest of the application use that definition to show the behavior of the property. For example, you can define that a property can be filter by text, or by a set of options. Or you can change the component that is rendered for that specific property.

There are missing parts here. The two more important parts that we have to implement later, are calculations and grouping. That two parts are now hardcoded, but we should abstract them and find the way to define what we expect for calculations and grouping just defining data or functions in the properties.


harshilsharma63 (approved):
LGTM 👍

PR #3364, opened 2022-07-14
Prevented empty name in create/rename categories

### LINKED ISSUE #3363: Bug:

1. Create a category with name containing only spaces. Same for renaming an existing category.
2. Notice the API call returns a 500 error

## Expected behavior

The webapp shouldn't even make the API call and not allow using empty names

harshilsharma63:
Fixes #3363


Pinjasaur (approved):
LGTM 👍🏻

PR #3384, opened 2022-07-19
Add `disable_notify` flag for InsertBlocks API

wiggin77:
#### Summary
This PR adds an optional `disable_notify` flag for InsertBlocks rest API.  Defaults to false.  When true, subscription notifications are disabled (users won't be notified of board/card changes).  Websocket notifications are unaffected.

This allows bulk add tools, load testers, etc to insert blocks without a notification storm.

#### Ticket Link
NONE

### DIFF
--- a/mattermost-plugin/product/boards_product.go
+++ b/mattermost-plugin/product/boards_product.go
@@ -70,7 +70,7 @@ type boardsProduct struct {
 }
 
 //nolint:gocyclo
-func newBoardsProduct(mmServer *app.Server, services map[app.ServiceKey]interface{}) (app.Product, error) {
+func newBoardsProduct(_ *app.Server, services map[app.ServiceKey]interface{}) (app.Product, error) {
 	boards := &boardsProduct{}
 
 	for key, service := range services {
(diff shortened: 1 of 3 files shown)


wiggin77:
> I'd more expect the _existence_ of the query parameter to equate to a truthy value

It is common for libraries making the call to always include the parameter and enable/disable via the value.  It's more convenient when generating the URL versus "if" statements to conditionally include it.  Someone will expect `disable_notify=false` to do the right thing, but if existence means true then it won't.  It would require: missing is false, `disable_notify` is true, `disable_notify=false` is false, and then `disable_notify=true` would be true (it would be expected that if you can set false, you must be able to set true).  Instead of all that special handling, `disable_notify=true` is true, and anything else is false is the easiest to reason about.  I guess a utility method to allow for true, 1, yes, etc would be ok, but would not be standard boolean according to Swagger docs.

Pinjasaur (approved):
> I guess a utility method to allow for true, 1, yes, etc would be ok, but would not be standard boolean according to Swagger docs.

Ahhh, didn't consider Swagger—thanks!

PR #3388, opened 2022-07-20
Added logic to add board member to my memberships is it is for me

### LINKED ISSUE #3294: Bug: New membership not reflected live
When I add a user I expect that to be reflected in the interface on the fly, but I need to switch board to get the new board.

1. Create one board
2. With other user, open Boards
3. Add this other user to the board
4. The board is not shown in the added user interface.

## Expected behavior

The sidebar of the new added user should show the board on the fly.

## Edition and Platform

 - Version: 7.2
 - Browser and OS: Firefox/Linux

harshilsharma63:
Fixes #3294


webapp/src/store/boards.ts
```diff
@@ -72,6 +72,7 @@ export const updateMembersEnsuringBoardsAndUsers = createAsyncThunk(
             /* eslint-enable no-await-in-loop */
 
             thunkAPI.dispatch(updateBoards(boardsToUpdate))
+            thunkAPI.dispatch(updateMembers(members))
```
Pinjasaur, line 75: what's the reason for this addition?

harshilsharma63: To update board-memberships in redux store. We want it to stay up to date with any WS event.

harshilsharma63: Anything specific you're looking for?

webapp/src/pages/boardPage/boardPage.tsx
```diff
@@ -71,6 +72,13 @@ const BoardPage = (props: Props): JSX.Element => {
     const viewId = match.params.viewId
     const me = useAppSelector<IUser|null>(getMe)
 
+    // used for websocket event callback functions
+    // to not get stuck with initial value of `me`, which is null.
+    const meRef = useRef<IUser | null>(null)
```
jespino, line 77: I think `useWebsockets` support extra dependencies to load it, I think if you add the `[me]`, list after the handler, is going to refresh the context every time is needed, and because `me` shouldn't be changing often, it shouldn't be a problem. That way you don't need the reference here.

webapp/src/store/boards.ts
jespino: I'm not sure if this is needed.

Pinjasaur: I was wondering because I was having a hard time understanding where it fit into the PR and the addition of `addMyBoardMemberships`—was it causing a bug or edge case in state before?

harshilsharma63: That's right, no need to dispatch `updateMembers`. Removed now.

I had to create `addMyBoardMemberships` because the existing store function that updates memberships for the current user (`updateMembersHandler`) assumes the user-board membership to already exist in the DB, with won't if the user is just being added to the board. The assumption of the existence of the same entry in store is used to check if the BoardMember payload deals with the current user.

Here's a snipped of the `updateMembersHandler()` function I'm talking about -

```
for (const member of action.payload) {
        if (state.myBoardMemberships[member.boardId] && state.myBoardMemberships[member.boardId].userId === member.userId) {
            state.myBoardMemberships[member.boardId] = member
        }
    }
```
`state.myBoardMemberships[member.boardId]` won't exist when a user is newly added to a board, and nothing works after that. 

Hence, the need for a new store function that exclusively deals with `state.myBoardMemberships` doesn't need to rely on any existing data.

Please let me know if this is still a bit confusing :)

webapp/src/pages/boardPage/boardPage.tsx
harshilsharma63: `me?.id` was already being used in the code below with `useWebsocket`. The issue still occurred as we are creating the function `incrementalBoardMemberUpdate` in there (slightly below this line of code) and we're refering to `me` in that function. That function again gets passed to some other function as well. So while `useWebsocket` body will receive the new value of me, it would be too late as the old value was already locked in the `incrementalBoardMemberUpdate` function as passed to `wsClient.addOnChange`.

jespino: I don't think that's the case, the `useWebsocket`, works like an effect, and it have its setup, and teardown, and because the function creation and adding the handler is part of the setup process and the remove of the handler is part of the tear down, whenever the me changes, it should remove the old handler and use a new fresh one with the right context.

harshilsharma63: You're right. It does indeed without ref as well. I see that when `me` changes, we unregister and re-register the WS handlers, so that's taking care of this. I've removed the use of ref now

jespino (approved):
LGTM

webapp/src/store/boards.ts
Pinjasaur: > state.myBoardMemberships[member.boardId] won't exist when a user is newly added to a board, and nothing works after that.

Makes sense, thanks for breaking it down! 👍🏻

Pinjasaur (approved):
Thanks @harshilsharma63!

PR #3390, opened 2022-07-20
Converted synthetic membership to natuaral for board's channel members

### LINKED ISSUE #3369: Bug: Cannot add explicit board members when a channel is linked to a board

1. Go to a board thats linked to a channel
2. Open Share modal
3. Add a team member to the board
4. See error - user is not added
5. Remove channel as a member of the board
6. Add the same team member from Step 3 - member can now be added

## Expected behavior

Users can be added to a board regardless if a channel is linked to the board or not.

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/179058094-dc834cfa-e106-480b-9c51-c4ad94e5bf48.mov

## Edition and Platform

 - Version: v7.2 Bug Bash

harshilsharma63:
Fixes #3369

### DIFF
--- a/server/app/boards_test.go
+++ b/server/app/boards_test.go
@@ -1,0 +1,107 @@
+package app
+
+import (
+	"testing"
+
+	"github.com/mattermost/focalboard/server/model"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddMemberToBoard(t *testing.T) {
+	th, tearDown := SetupTestHelper(t)
+	defer tearDown()
+
+	t.Run("base case", func(t *testing.T) {
+		const boardID = "board_id_1"
+		const userID = "user_id_1"
+
+		boardMember := &model.BoardMember{
+			BoardID:      boardID,
+			UserID:       userID,
+			SchemeEditor: true,
+		}
+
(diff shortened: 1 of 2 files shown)


mgdelacroix (changes requested):
Code is looking good! could you please add an integration test for the use case?

harshilsharma63:
@mgdelacroix I've added unit test for app later to cover this case.

server/app/boards_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+		th.Store.EXPECT().GetBoard(boardID).Return(&model.Board{
+			TeamID: "team_id_1",
+		}, nil)
+
+		th.Store.EXPECT().GetMemberForBoard(boardID, userID).Return(&model.BoardMember{
+			UserID:    userID,
+			BoardID:   boardID,
+			Synthetic: true,
+		}, nil)
```
mgdelacroix, line 89: nit: you can return directly the board member you've created:

```suggestion
		th.Store.EXPECT().GetMemberForBoard(boardID, userID).Return(boardMember, nil)
```

mgdelacroix (approved):
LGTM! :+1: Minor, nonblocking nit

harshilsharma63:
> LGTM! 👍 Minor, nonblocking nit

Actually, the two are not the same. One is synthetic and the other isn't.

PR #3395, opened 2022-07-21
Duplicate board retain category

### LINKED ISSUE #2897: Feature Idea: duplicating a board maintains the category
## Summary

When duplicating a board that is in a non-default category (`Boards` being the default) it will be duplicated into the default (`Boards`) category.

Importance: Low

Use cases:
1. As a user, I would expect 'duplicate' to equate to a familiar 'copy + paste' workflow e.g. retain the category.
2. As a user, if I have many boards / categories and I duplicate a board I could visually lose it.
3.

Any examples of good implementations of this capability.

harshilsharma63:
Fixes #2897


sbishel (changes requested):
Although it ends up correct, there is an obvious flash when the category is updated after being copied. See attached video.

https://user-images.githubusercontent.com/12704875/181312887-378a2b08-3111-40fd-a547-48d763c193e7.mov

harshilsharma63:
@sbishel That's happening because the WS events for new board creation and moving the board to the category is sent to the client very close to each other, and sometimes it ends up processing one before the other. Need to see how to handle this.

sbishel:
> @sbishel That's happening because the WS events for new board creation and moving the board to the category is sent to the client very close to each other, and sometimes it ends up processing one before the other. Need to see how to handle this.

Seems we should try to do this in a single update or possibly not send the websocket message until both updates are complete.

webapp/src/components/sidebar/sidebarBoardItem.tsx
```diff
@@ -113,6 +113,26 @@ const SidebarBoardItem = (props: Props) => {
...
+        // then jumping to the custom category.
+        // The jump would happen because when server clones a board from a custom category,
+        // two WS events are sent - first to indicate the new board belongs to the specific category,
+        // second, to indicate the new board is created. Depending on the order of execution of the two
+        // WS event handlers, if the handler for second events executes first, it will show the new board
+        // in default category in LHS, then when the handler for first events gets executed, it moves the board
+        // to the correct category.
+        // By not waiting for the board-category WS event and setting the right category for the board,
+        // we avoid the jumping behavior.
+        if (props.categoryBoards.id !== '') {
```
Pinjasaur, line 129: Thanks for the explanation! 🙂

Pinjasaur (approved):
LGTM @harshilsharma63. I'm going to defer merging in case @sbishel had any further feedback.

sbishel (approved):

harshilsharma63:
> LGTM @harshilsharma63. I'm going to defer merging in case @sbishel had any further feedback.

Merging as Scott approved.

PR #3415, opened 2022-07-25
Improve the board creation from channels

### LINKED ISSUE #3367: Feature Idea: Creating a board from the channel intro button should automatically link the board to the channel
## Summary

I think it makes sense that the board is automatically linked to the channel if created from the "Create a board" button in the channel.
Without that, the user has to go through the extra steps of linking it.

image

Importance: Medium

jespino:
Fixes #3367


sbishel (approved):

sbishel:
I just noticed this is going directly to `release-7.2`, do we need this in `main` as well?

jespino:
@sbishel yes, but main was broken for me because the multiproduct architecture and some dependency problems. It needs to be cherry picked to main

PR #3419, opened 2022-07-25
Remove opengraph dependency

agnivade:


agnivade:
Ah okay, so only server update won't work. We need to update plugin first. Which means plugin-api needs to be updated first. Will send a PR later.

jespino (approved):
Lgtm

PR #3420, opened 2022-07-25
Replace query with exec on database update queries

mgdelacroix:
#### Summary
This PR uses optimizes the category migration to use `.Exec` on update queries instead of `.Query`, as we're not expecting any data in return


sbishel (approved):

harshilsharma63 (approved):
LGTM 👍

PR #3429, opened 2022-07-27
Not requesting for channels in personal server mode

jespino:


Pinjasaur (approved):
Nice fix @jespino, checked it locally with Personal Server real quick and it LGTM! 👍🏻

PR #3437, opened 2022-07-27
Move category creation and UUID update to a more performant data migration

mgdelacroix:
#### Summary
This PR replaces the approach of creating the categories and category boards on a migration and updating their IDs on a data migration with one that directly generates that data in a data migration. This new approach inserts the data in batches of 1000 values at a time, which combined with the speed of the inserts makes the whole migration way faster.

### DIFF
--- a/server/services/store/sqlstore/data_migrations.go
+++ b/server/services/store/sqlstore/data_migrations.go
@@ -16,7 +16,7 @@ import (
 const (
 	// we group the inserts on batches of 1000 because PostgreSQL
 	// supports a limit of around 64K values (not rows) on an insert
-	// query, so we want to stay safely below
+	// query, so we want to stay safely below.
 	CategoryInsertBatch          = 1000
 
 	TemplatesToTeamsMigrationKey = "TemplatesToTeamsMigrationComplete"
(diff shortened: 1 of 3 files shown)


jespino (approved):
LGTM

server/services/store/sqlstore/data_migrations.go
```diff
@@ -14,6 +14,11 @@ import (
 )
 
 const (
+	// we group the inserts on batches of 1000 because PostgreSQL
+	// supports a limit of around 64K values (not rows) on an insert
+	// query, so we want to stay safely below.
+	CategoryInsertBatch = 1000
```
sbishel, line 20: Is there any reason an admin may want to tweak this number?

server/services/store/sqlstore/data_migrations.go
```diff
@@ -122,6 +127,10 @@ func (s *SQLStore) runUniqueIDsMigration() error {
 	return nil
 }
 
+// runCategoryUUIDIDMigration takes care of deriving the categories
+// from the boards and its memberships. The name references UUID
+// because of the preexisting purpose of this migration, and has been
+// preserved for compatibility with already migrated instances.
```
sbishel, line 133: Nice. I was going to ask if this should be changed...

sbishel (approved):
Should we add any logging, so we/admins know what point in the sequence we are at? Or is it performant enough that its unnecessary.

LGTM

server/services/store/sqlstore/data_migrations.go
mgdelacroix: I don't think there is, being the migration a one time operation the admin is not likely to tweak the value. We tested with 5000 entities per batch and it took a couple more seconds, so sadly more is not always faster

Issue #3275, opened 2022-06-23
Bug: Default task template 


thomaslc66:
## Steps to reproduce the behavior

1. Go to a board that you've previously created
2. Create a new task and add some properties
![image](https://user-images.githubusercontent.com/9827392/175296835-c5baad14-3de8-4c07-aa0b-f64d8f0213a6.png)

3. Save the modification
4. Set this template as "Default"
![image](https://user-images.githubusercontent.com/9827392/175297070-dff9ccff-c502-43dc-8b7d-47b8723c75ff.png)

## Actual behavior

1) if you click on the + New button at the bottom of a column, the default template is not selected
![image](https://user-images.githubusercontent.com/9827392/175297194-e9772dd9-f10e-48a8-864d-6d8cf00420f3.png)

2) if you click on the New button (top right) 
![image](https://user-images.githubusercontent.com/9827392/175297288-6668f99b-a1c0-4150-a444-6cccb7222d77.png)
If you fill the template with your info but for a specific task you don't want (let's say) a "due date" in it and delete this property (in your new task not the template), when you go back to the template edition, the "Due date" property is also delete from the template.

## Expected behavior

1) If I set default template, I should be able to have the default template when I click on all the "New" button and not only the blue one but also the ones under each column

2) Modification on a specific task, should not be reflected in the template.

## Edition and Platform

 - Edition: Personal Server
 - Version: [6.5.1]
 - DB Schema Version: [6.5.1]
 - Database: postgres
 - Browser and OS: Chrome
 
 
 
 Is it something that is fixed in Latest Release: 7.0.0? Didn't see anything in the changelog.
 
 Thank you for the update

sbishel:
For Item 1. The New button at the bottom of a Kanban column should use the template and then apply the proper property for the column.
This will apply to all views. "New" should use template and apply field (as necessary based on view).

Item 2 - Is by design. All cards for a board share columns and deleting a property deletes on all cards. (There is warning message to this effect.)

thomaslc66:
@sbishel Thank you for your answer.

For Item 1. As stated in our tests "New" does not always take the default template.

For Item 2. Makes no sense to have a default template then if all cards share the same properties and that every card can edit the template. Then what happen if I have 2 different template with different properties ? Will the default one delete the properties no existing inside the other one. And also will the modifications in one of my template edit my default template?
That's really weird and I can see no real use case where this is useful. 

[...]

PR #3444, opened 2022-07-28
New button using default template in kanban view.

Rajat-Dabade:
#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/3275


wiggin77:
@Rajat-Dabade there is a compile error.

webapp/src/components/kanban/kanban.test.tsx
```diff
@@ -88,6 +89,12 @@ describe('src/component/kanban/kanban', () => {
                 board_id_1: {userId: 'user_id_1', schemeAdmin: true},
             },
         },
+        views: {
+            views: {
+                boardView: activeView,
+            },
+            current: 'boardView',
+        },
```
sbishel, line 97: Add tests that will verify `addCardFromTemplate` gets called as appropriate. Similar to 
`test('return kanban and click on New', () => {`

webapp/src/components/kanban/kanbanCard.tsx
```diff
@@ -115,6 +115,7 @@ const KanbanCard = (props: Props) => {
                                 false,
                                 'duplicate card',
                                 false,
+                                undefined,
```
sbishel, line 118: With my suggested changes you may need to pass an empty object `{}` here.

webapp/src/octoClient.ts
```diff
@@ -665,15 +665,20 @@ class OctoClient {
         return this.getJson<BoardsAndBlocks>(response, {} as BoardsAndBlocks)
     }
 
-    async duplicateBlock(boardID: string, blockID: string, asTemplate: boolean): Promise<Block[] | undefined> {
+    async duplicateBlock(boardID: string, blockID: string, asTemplate: boolean, propertiesThatMeetFilters?: Record<string, string>): Promise<Block[] | undefined> {
```
sbishel, line 668: These changes shouldn't be necessary.

sbishel (changes requested):
Your webapp changes look good, however, I think you can/should handle this on the webapp side and revert all your server changes.

Eventually, we do want to handle these things server-side, this will help performance, but we want to move everything when we do that.

So to stay consistent with how everything else works, you can make a single line change in `mutator.ts` and remove all the server side changes. 

Let me know if you want to meet to discuss.

webapp/src/mutator.ts
```diff
@@ -975,12 +975,13 @@ class Mutator {
         fromTemplate = false,
         description = 'duplicate card',
         asTemplate = false,
+        propertiesThatMeetFilters?: Record<string, string> | undefined,
```
sbishel, line 978: Let's rename this to something like `propertyOverrides`. That name has context in `kanban.tsx`, but not here.

webapp/src/mutator.ts
```diff
@@ -975,12 +975,13 @@ class Mutator {
...
         afterRedo?: (newCardId: string) => Promise<void>,
         beforeUndo?: () => Promise<void>,
     ): Promise<[Block[], string]> {
         return undoManager.perform(
             async () => {
-                const blocks = await octoClient.duplicateBlock(boardId, cardId, asTemplate)
+                const blocks = await octoClient.duplicateBlock(boardId, cardId, asTemplate, propertiesThatMeetFilters)
```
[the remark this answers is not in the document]
Rajat-Dabade: Got it.

webapp/src/components/kanban/kanban.test.tsx
Rajat-Dabade: Sure

webapp/src/octoClient.ts
Rajat-Dabade: Yes, got it.

webapp/src/mutator.ts
Rajat-Dabade: Done.

webapp/src/components/kanban/kanbanCard.tsx
Rajat-Dabade: Done

wiggin77 (approved):
LGTM :+1:

PR #3454, opened 2022-07-28
Add the creator of the board as a administrator of the board if it still belongs to the channel

jespino:
The boards creators weren't added as administrators of the boards during the
migration (only the admin administrators), but this PR adds also the creator as
a new admin.

This is done only if the user is still a member of the channel, keeping this
way the original behavior of boards (In 7.1 if you leave the channel, doesn't
matter if you were the creator, you are no longer member of the board).


wiggin77 (approved):
LGTM :+1:    Note, this won't help anyone that has already run migration 18.

jespino:
Yes, but migration 18 shouldn't be executed yet by anybody (in theory)

mgdelacroix (approved):
LGTM! :+1:

Issue #1007, opened 2021-08-16
Bug: popup menu disappears when mouse is not over


kamre:
**Summary:**
Some components contain a button that appears only on hover and allows to show popup menu. Such popup menus disappear when the mouse is moved away and reappear when mouse is over the component.

**Expected behavior:**
Popup menu should be visible until user clicks somewhere.

**Screenshots:**
Kanban card:
![PopupMenuKanbanCard](https://user-images.githubusercontent.com/5587620/129602031-800438b4-8e3c-4dce-ac7e-1ddf2233fee2.gif)

Gallery card:
![PopupMenuGalleryCard](https://user-images.githubusercontent.com/5587620/129602085-93d8cc02-83b7-481c-8ad9-9afaba28fe90.gif)

Card comment:
![PopupMenuComment](https://user-images.githubusercontent.com/5587620/129602144-e5128b3d-2b0b-4caf-8f8e-09f6e36098b6.gif)

Card description:
![PopupMenuCardDescription](https://user-images.githubusercontent.com/5587620/129602173-926ff843-efa3-4ac3-9ad9-a3ee23d46914.gif)

**Platform:**
 - Browser and version: Chrome 92
 - OS and version: Windows 10

chenilim:
Thanks @kamre, good catch! Not sure why this happens for some menus as you noted, but not others (e.g. board menu, board column menu). In any case, the menus should remain visible until the user selects an option, clicks away, or presses Esc. cc @asaadmahmood, @jespino, @sbishel.

Putting this up for grabs.

Pinjasaur:
> Wouldn't a better solution be just to add a class to the card who's menu is open, and then just increase the z-index on it?

> I think we can then also use that functionality to make sure that the menu remains even if you hover outside of that card, and only remove that class on a blur event or on a click outside.

Taken from https://github.com/mattermost/focalboard/pull/3486#issuecomment-1201035635.

kamre:
Hi @wuwinson, I am working on this bug and will soon create a pull request with the fix. So you can assign it to me.

wuwinson:
Awesome, thanks @kamre!

PR #3486, opened 2022-07-29
fix: Calendar card "..." action menu overlapping behind adjacent cards

### LINKED ISSUE #3425: Bug: Calendar View - "More Options" modal displays behind tasks
Found by `@krmh04` QA community contributor bug bashing on `https://boards-bugbash.test.mattermost.cloud/` server

1. Switch to calendar view
2. Add tasks on a date
3. Click on the more options (three dots) button
4. Observed: "More options" modal displays behind tasks

## Expected behavior

Not sure - perhaps display the "More options" modal a little to the right of the tasks?

## Screenshots (optional)

image

## Edition and Platform

 - Version: v7.2.0
 - Browser and OS: Windows 10 / Chrome

Pinjasaur:
#### Summary

This PR `unset`s some `z-index` values that Full Calendar sets to prevent the "..." action menu from being overlapped by adjacent events in Calendar view.

I went through a couple iterations to get to this solution, but it ended up being the last "hacky" (subjectively) even though it is overriding CSS from a vendored dependency. For some context, other "solutions" I tried:

1. Reverse the event render order in Full Calendar and then use flexbox to reverse it. This didn't work with multi-day events along with the "show more" calendar widget.
2. Use a SCSS `@for` loop to set descending `:nth-child` `z-index` values e.g. https://stackoverflow.com/a/36609688

Before:

image

After:

image

#### Ticket Link

closes #3425

### DIFF
--- a/webapp/src/components/calendar/fullCalendar.tsx
+++ b/webapp/src/components/calendar/fullCalendar.tsx
@@ -253,11 +253,13 @@ const CalendarFullView = (props: Props): JSX.Element|null => {
                 events={myEventsList}
                 editable={isEditable()}
                 eventResizableFromStart={isEditable()}
+                // this is the reverse of the default ordering (https://fullcalendar.io/docs/eventOrder)
+                // which gets reversed in CSS so the stacking contexts play nicely w/ "..." menus
+                eventOrder='-start,duration,-allDay,-title'
                 headerToolbar={toolbar}
                 buttonText={buttonText}
                 eventContent={renderEventContent}
                 eventChange={eventChange}
-
                 selectable={isSelectable}
                 selectMirror={true}
                 select={onNewEvent}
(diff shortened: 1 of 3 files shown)


sbishel (approved):

asaadmahmood:
@Pinjasaur Wouldn't a better solution be just to add a class to the card who's menu is open, and then just increase the z-index on it?

I think we can then also use that functionality to make sure that the menu remains even if you hover outside of that card, and only remove that class on a blur event or on a click outside.

harshilsharma63 (approved):

Pinjasaur:
> @Pinjasaur Wouldn't a better solution be just to add a class to the card who's menu is open, and then just increase the z-index on it?
> 
> I think we can then also use that functionality to make sure that the menu remains even if you hover outside of that card, and only remove that class on a blur event or on a click outside.

@asaadmahmood We don't have a lot of control over how the Calendar view is rendered since it's mostly done by the Full Calendar library component. I went the `unset` route for the existing `z-index` values because—from how I understand it—when they were set a CSS stacking context was being created so the adjacent elements would naturally stack over the preceding siblings.

I think your idea is worth exploring, but perhaps as part of #1007 since it might also solve that same issue that bug is tracking. Let me know your thoughts—we're looking to get this for 7.2 or otherwise revert the functionality of the "..." menu for the release since it's broken as-is. https://github.com/mattermost/focalboard/issues/3425#issuecomment-1203281336

asaadmahmood (approved):
LGTM for now.

PR #3490, opened 2022-07-30
Update query to use alias

sbishel:
#### Summary
UserList query was failing because the query was not using the alias in the JOIN clause.


sbishel:
@Rajat-Dabade - Take a look, but I'm going to go ahead and merge to fix issue.

PR #3516, opened 2022-08-02
Modify go.work file to include multi-product architecture repositories by default

pfltdv:
#### Summary
We do not include multi product repositories by default that break the developer builds. PR contains changes to include those repositories always and exclude them by using environment variables.

fixes https://github.com/mattermost/focalboard/pull/3516


wiggin77:
@pfltdv have a look at https://github.com/mattermost/focalboard/pull/3517

pfltdv:
> @pfltdv have a look at #3517

Thanks alot @wiggin77. So for docker builds we need to exclude everything. But if developer has changes for server and focalbard at the same time on his local work environment. I assume we need to include mattermost-server in docker build.

wiggin77:
> But if developer has changes for server and focalbard at the same time on his local work environment. I assume we need to include mattermost-server in docker build.

Ideally yes, but I don't know how we would include the `mattermost-server` repo in the docker build since it is outside the docker-compose context.  Docker doesn't allow that.  We would have to re-org the projects.

For now I decided that docker builds will fetch whatever version of mattermost-server we are pinned to.

pfltdv:
Lets find a way to include in docker builds, cause excluding that is increase complexity for everything and it will not support multi-product architecture.

@agnivade What are your toughts about excluding mattermost-server on docker builds.

agnivade:
We need to include it. Without it, it will introduce a discrepancy on how the dev environment works and how docker builds are created.

> but I don't know how we would include the mattermost-server repo in the docker build since it is outside the docker-compose context.

How do we do it right now for enterprise? It should be the same.

I should say that the final build should only be created from the mattermost-server repo as that's the base repo which includes everything. Focalboard repo should become like the enterprise repo which is just a library and has no standalone meaning. I realize that focalboard has to support standalone builds as well which makes this a whole lot more complicated. I don't know if the tradeoffs of supporting standalone builds justify the complexity of the build pipeline. IMO, I'd prefer to discontinue standalone builds to keep things simple. But that's not my call to make :)

Tldr is:
1. builds from server should include server+enterprise+focalboard.
2. builds from focalboard ideally should not be possible (same as enterprise), but since we need to support standalone builds, we need to include server+enterprise+focalboard as well.

This 2-way building might become hard to maintain in the future.

pfltdv:
Thanks alot i will adopt Docker compose and dockerfile to support that architecture. 

But if our goal is to build everything from everywhere. Maybe we should consider some build tool alternatives that make everyone's life easier since we will not go with monorepo solution. (ie. Bazel can track multiple workspaces and sync them all together which also ensure hermetic builds. https://blog.bazel.build/2018/07/09/bazel-sync-and-resolved-file.html)

agnivade:
> But if our goal is to build everything from everywhere. 

That is not the goal. Focalboard is the only exception because you can have focalboard as a standalone product as well. Other than that, server is the only repo to build, which pulls in other repos and makes the final build.

pfltdv:
So if focalboard can be standalone, can we revert go.work and multi-product architecture changes in focalboard repository. So it can be independent of mattermost-server and uses go.mod to resolve dependencies.

And change mattermost-server & enterprise builds to include boards at pipelines and makefiles ?

agnivade:
Right now, temporarily for the time being, focalboard needs to work in the plugin mode until the webapp multi-product arch is complete. So we need the go.work file in the plugin directory (only temporarily).

But whether we need go.work in the linux directory or not, I'll defer to @wiggin77 on that. If we do that, developers just need to be aware that the server version in go.mod needs to be constantly updated for every bug fix/release.

pfltdv:
Thanks alot Agniva, for releases we need to have go.mod updated to have hermetic builds and we can not use go.work. Our released artifacts should be generated from tags and all dependencies need to be versioned. 

For other pipelines and docker builds better to use go.work in this case to make developers live's easier.

agnivade:
> Thanks alot Agniva, for releases we need to have go.mod updated to have hermetic builds and we can not use go.work. 

We are already using go.work for main server releases. This is just an extension of how we used to build enterprise. It's the same way, we are building focalboard now.

I think perhaps there's some confusion somewhere. Please ping me and I can explain further.

pfltdv:
We fixed makefiles to support below:

1. It will always include enterprise and mattermost-server repositories in local builds, so go.work will work as requested. Developers need to make sure that they cloned and switched to correct branches in mattermost-server and enterprise. 
2. CI pipelines configured to clone server repo whenever needed so PR's depending on Server branches will work. (it is same for release branches)
3. At docker compose builds we excluded server repo since build context grows to 2GB, since we include every repo.

pfltdv:
@wiggin77 @agnivade PR is ready for review and it should fix docker builds as well.

```
mmm:/srv/github/focalboard/docker (fix/focalboard-pipelines)> docker-compose build
Building app
Sending build context to Docker daemon  86.01MB
Step 1/21 : FROM node:16.3.0@sha256:ca6daf1543242acb0ca59ff425509eab7defb9452f6ae07c156893db06c7a9a4 as nodebuild
 ---> 7105279fa2ab
Step 2/21 : WORKDIR /webapp
 ---> Using cache
 ---> b825aad7d032
Step 3/21 : ADD webapp/ /webapp
 ---> Using cache
 ---> bbdf5a1fada5
 ....
 ....
 ...
 qlite3-binding.c:128049:10: warning: function may return address of local variable [-Wreturn-local-addr]
128049 |   return pNew;
       |          ^~~~
sqlite3-binding.c:128009:10: note: declared here
128009 |   Select standin;
       |          ^~~~~~~
Removing intermediate container a8909953da99
 ---> 26dbd260d5a3
Step 9/21 : RUN mkdir /data
 ---> Running in f08037a7b2af
Removing intermediate container f08037a7b2af
 ---> 57ea290161fe
Step 10/21 : FROM gcr.io/distroless/base-debian10@sha256:d2ce069a83a6407e98c7e0844f4172565f439dab683157bf93b6de20c5b46155
 ---> 3f147b3ac70c
Step 11/21 : WORKDIR /opt/focalboard
 ---> Using cache
```

agnivade (approved):
Thanks Mustafa!

wiggin77 (approved):

wiggin77:
@pfltdv I merged the dependency upgrades to this PR, and also added a fix for the standalone docker build.

pfltdv:
> @pfltdv I merged the dependency upgrades to this PR, and also added a fix for the standalone docker build.

Thanks alot @wiggin77 for fixes and updating the branch. I will test locally to understand the changes then merge after.

PR #3526, opened 2022-08-03
Hide board feature

### LINKED ISSUE #3447: Bug: Cannot remove/hide boards from LHS

1. With User A, add User B to a board. Notice the board shows up in LHS for User B.
2. With User B, try to remove that board from the LHS. YOu can neither leave the board from anywhere nor remove the board from the sidebar.

## Expected behavior

User should either be allowed to leave a board or at least hide it from the LHS. Channels' behavior is to leave a channel to remove it from LHS and for DMs, one can "close conversation" which keeps the messages but removes the DM from LHS.

Lack of this feature would be an issue after v7.2 is deployed on the community and everyone ends up with dozens of boards on LHS>

## Edition and Platform

 - Version: 7.2

harshilsharma63:
Card link - https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vuswsftt1rb873datosmmbtwkpa/cmebjs18pyifuuy3rojeyw9o1ry

Fixes #3447

Notes

* Storing hidden board IDs in `user.props` and in webapp not rendering them in LHS if board is in that list.
* On loading user on load, even though the API sends an array of hidden board IDs, I store it in redux store as a map of board id to a boolean value, to make querying faster.
* @sbishel initially we discussed storing the hidden property in the focalboard_category_boards table but if we delete a category, we should be deleting the corresponding entries from that table as well, so we'd lose the hidden board data. Currently, we don't delete from focalboard_category_boards on deleting a category, but we should, and I've create a code debt card for the same at https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vuswsftt1rb873datosmmbtwkpa/cgphy38ph3ifkpfbhhus4mr5chy


sbishel (commented):
I really don't like storing these in user properties. I think this would probably be better saved to a separate table. Channels uses the `preferences` table for this.

harshilsharma63:
@sbishel @jespino and I discussed this on Tuesday and after 7.2 we'll migrate from using user.props to preferences table. Why we didn'y use preferencesd table? We'll have to create the migration to create corresponding focalboard table, add store functions for accessing that table, new APIS etc. So I don't want to use preferences table for 7.2.

sbishel:
I figured that was the case. As I wrote the comment, I thought it would be too big to implement in that manner. Let's make sure we create a card to track.  Thanks.

sbishel (approved):

harshilsharma63:
@sbishel added the card for 7.3 here - https://community.mattermost.com/boards/workspace/qgsck6cts3fwpqwyjiupjm5cde/b4ctmm917q3rzdfi451qtuxjzcw/vuswsftt1rb873datosmmbtwkpa/cmpcwh7mpi7gcf8qx3pmbdkkoph

harshilsharma63:
I've cherrypicked this to 7.2 as well.

PR #3552, opened 2022-08-04
Fix panic caused by returning nil,nil in GetMemberForBoard

wiggin77:
#### Summary
mattermostauthlayer.GetMemberForBoard was returning nil,nil when an error occurred as it was only checking the IsErrNotfound case.  This caused a panic further up the stack (HasPermissionToBoard) in that it expected a valid member returned if err==nil.

This should be added to 7.2 as well. 

#### Ticket Link
NONE

### DIFF
--- a/server/services/store/mattermostauthlayer/mattermostauthlayer.go
+++ b/server/services/store/mattermostauthlayer/mattermostauthlayer.go
@@ -712,6 +712,8 @@ func (s *MattermostAuthLayer) GetMemberForBoard(boardID, userID string) (*model.
 				Synthetic:       true,
 			}, nil
 		}
+	} else {
+		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


stylianosrigas (approved):

wiggin77:
Thanks for fixing the linter issue @sbishel.  Strange that it made it past CI linter before. The `else` that was added should not have changed anything relative to the variable naming.

sbishel:
Yeah. I guess because the `else` caused the err variable to be used in a different block, variable renaming isn't allowed. I removed the change and linter did indeed pass without the change.

sbishel (approved):

wiggin77:
@sbishel I'll manually cherry pick this and resolve the conflict.

PR #3556, opened 2022-08-04
Show the active board even when the category is hidden

jespino:
#### Summary
Show the active board even when the category is hidden.


asaadmahmood (approved):
LGTM! Thanks @jespino !

sbishel (approved):

PR #3579, opened 2022-08-05
Persist category state

### LINKED ISSUE #3448: Bug: LHS category collapsed/expanded state resets on switching teams

1. Be part of more than one team.
2. Open team 1, create a category in Boards LHS and collapse it.
3. Switch to team 2, then back to team 1.
4. Notice the collapsed category is now expanded.

## Expected behavior

The collapsed/expanded state of a category should not reset on switching teams. Ideally, it should even persist refreshes and be part of user's preferences.

## Edition and Platform

 - Version: 7.2

### LINKED ISSUE #3465: Bug: Categories should be collapsed by default

1. Go to the [Community Clone server](https://community-dr.staging.cloud.mattermost.com/)
2. On Channels, go to the `R&D Meeting` channel
3. Click on the Boards Apps Bar button to load the new RHS
4. Click on the `ice-breakers` board
5. See error - Board loads, but the board is not visible on the LHS until you scroll down because so many categories are open after the migration

## Expected behavior

A better UX might be to collapse all categories by default after migration and then possible do #3449 in a future release.

## Edition and Platform

 - Version: v7.2 Bug Bash

harshilsharma63:
Fixes #3465 and fixes #3448

Summary-

1. Added a column `collapsed boolean` to `categories` table to store if the user category is collapsed or not.
2. The SQL migration for MySQL and PostgreSQL is idempotent but that of SQLite is not as I couldn't find a SQL only way to conditionally add a column to a table. Need to dig more into this but created the PR anyways as it doesn't block the plugin release.
3. Added new and missing tests for category and board category store methods.
4. Since the default "Boards" category is not stored in the database, for now, its state is not persisted. If needed, this can be added later.


server/services/store/sqlstore/migrations/000023_persist_category_collapsed_state.up.sql
```diff
@@ -0,0 +1,29 @@
+{{if .mysql}}
+    SET @dbname = DATABASE();
```
jespino, line 2: This is to allow "CREATE IF NOT EXISTS" in the migration to make this "idenpotent", but... if we put (as we decided) on ALTER/CREATE alone in each migration, you should have that knowledge about idenpotency already. If the transaction succeeded you have the migration applied, so the column is there and the migration is not going to be executed again. If the translaction failed its get not applied at tall. I don't know how much benefit we can obtain here in exchange of readability.

I'm not against this, but looks awful to me.

jespino (approved):
LGTM. I think the code is tested enough and simple enough to get in to the release. But here I also would love to know @sbishel's opinion

server/services/store/sqlstore/migrations/000023_persist_category_collapsed_state.up.sql
harshilsharma63: You're right. I completely forgot about this - having only, and one DDL per migration. I'll update the migration.

sbishel (approved):

Issue #2965, opened 2022-04-28
Feature Idea: Dont close board property selector menu on selecting a property


harshilsharma63:
## Summary

Problem 1-
Currently, when you wanna select multiple properties to be displayed in board view, you need to open the "Properties" menu again and again for each property as it closes as soon as you select a property.

Proposed solution-
Don't close the properties menu on selecting a property, and close if the user clicks on the "Properties" menu again or clicks anywhere outside.

It would also be nice if we showed which properties in that menu are already displayed. Such as by showing a checkmark with those already displayed.

## How important this is to me and why

Importance: Medium

chenilim:
Related to https://github.com/mattermost/focalboard/issues/2989

chenilim:
Amy recently ran into this issue when working on the FY23 Goals board.

Also filed https://github.com/mattermost/focalboard/issues/3361 to track a similar issue for the filter menu.

kamre:
Hi @chenilim, I have started looking into this issue as well as #3361 

My idea is to add a new flag to props of Menu Switch Item that will avoid closing the menu when Switch Item is toggled. And use this flag for Switch Items in menus for board properties and filter values.

Issue #3361, opened 2022-07-14
Feature Idea: Filter selection menu should not close on select


chenilim:
## Summary

If you want to set a filter to multiple values, you need to re-open the filter value menu each time. E.g. see video:

https://user-images.githubusercontent.com/46905241/178857962-4cd36d80-b07a-47aa-90ef-9bd8fe46b418.mov

Instead, the menu should not close until dismissed (e.g. clicked outside). This may apply elsewhere as well.

## How important this is to me and why

Importance: Medium

Use cases:
1. Filtering for multiple values

sbishel:
closing as fixed

PR #3587, opened 2022-08-05
[GH-2965] Don't close property selector menu when item is clicked

kamre:
#### Summary
Property selector menu is not closed when some property is selected or deselected:
- applied to menus in `ViewHeaderPropertiesMenu` and `FilterValue` components
- new property `suppressItemClicked` added to `SwitchOption` component to avoid dispatching `menuItemClicked` event
- unit tests and jest snapshots updated
- cypress tests updated
- new unit tests for checking that menu is not closed added

Here is how it looks for properties menu:

and for filtering value:

#### Ticket Link
Fixes: #2965 and #3361


kamre:
Hi @chenilim @harshilsharma63 , please review the changes.

harshilsharma63:
@kamre any way to make the changes not update so many snapshots? If they absolutely must, can you please describe why so?

kamre:
@harshilsharma63 almost all snapshot differences come from this change in the `MenuWrapper` component:
image
 As you can see on the right side I needed additional wrapper for the button that activates the menu. This was required because otherwise any clicks from the menu with items will lead to `toggle` call and close the menu.

But this was before I have added `stopPropagation` call to every clickable menu item. So I will check if this wrapper is still needed.

kamre:
@harshilsharma63 The changes to `MenuWrapper` structure are not needed. So I have reverted them and now the diff is really small. Please review.

harshilsharma63 (approved):

harshilsharma63:
LGTM 👍

PR #3590, opened 2022-08-08
[MM-43781] Cherry pick and patches

plant99:
Cherry-pick of https://github.com/mattermost/focalboard/pull/3005 to release-7.2.


mattermost-plugin/webapp/src/index.tsx
```diff
@@ -319,6 +319,21 @@ export default class Plugin {
                 ),
                 false
             )
+
+            // Insights handler
+            if (this.registry?.registerInsightsHandler) {
+                this.registry?.registerInsightsHandler(async (timeRange: string, page: number, perPage: number, teamId: string, insightType: string) => {
+                    if (insightType === 'MY') {
```
jespino, line 326: nit: Probably a constant storing the types somewhere can be more explicit an clear.

server/api/api.go
```diff
@@ -2166,6 +2170,216 @@ func (a *API) handleUploadFile(w http.ResponseWriter, r *http.Request) {
...
+	//     schema:
+	//       "$ref": "#/definitions/ErrorResponse"
+
+	vars := mux.Vars(r)
+	teamID := vars["teamID"]
+	userID := getUserID(r)
+	query := r.URL.Query()
+	timeRange := query.Get("time_range")
+
+	if !a.permissions.HasPermissionToTeam(userID, teamID, model.PermissionViewTeam) {
```
jespino, line 2222: This only makes sense in plugin mode, so I would do a check here to verify that only takes place in plugin mode because if not is always going to fail when it tries to get the user timezone. The same happens in the other API call implemented. Both has to return a Not Implement HTTP status code in desktop or standalone server.

server/services/store/storetests/board_insights.go
```diff
@@ -0,0 +1,99 @@
...
+		UserID:      userID,
+		BoardID:     "board-id-2",
+		SchemeAdmin: true,
+	}
+
+	_, _ = store.SaveMember(bm)
+
+	boardsUser1, _ := store.GetBoardsForUserAndTeam(testUserID, testTeamID)
+	boardsUser2, _ := store.GetBoardsForUserAndTeam(testInsightsUserID1, testTeamID)
+	// t.Run("team insights", func(t *testing.T) {
```
jespino, line 72: Why is this commented out?

jespino (changes requested):
Some things to address, nothing to strongly stop the PR to be merged, but needs to be solved before the release "IMO".

mattermost-plugin/webapp/src/index.tsx
plant99: cc: @BenCookie95 could you please make this small change in [this](https://github.com/mattermost/focalboard/pull/3596) PR?

plant99:
> Some things to address, nothing to strongly stop the PR to be merged, but needs to be solved before the release "IMO".

Thanks for catching the issues @jespino , I hope it's fine if we merge this and then I can proceed with [this PR](https://github.com/mattermost/focalboard/pull/3596) and the corresponding cherry-pick.

mattermost-plugin/webapp/src/index.tsx
plant99: The branch is in upstream focalboard for convenience.

BenCookie95: Done!

server/api/api.go
plant99: https://github.com/mattermost/focalboard/pull/3596/commits/cf42959b53526f54ca58373951dee8488bd80d2a

server/services/store/storetests/board_insights.go
plant99: https://github.com/mattermost/focalboard/pull/3596/commits/64cb8fba794845a115dbcdd61aa2a42d2c4b7196

mattermost-plugin/webapp/src/index.tsx
plant99: https://github.com/mattermost/focalboard/pull/3596/commits/9e8cab2b83cc30df826b87d2e40a6aa2e7230934

jespino (approved):
LGTM

jespino:
> > Some things to address, nothing to strongly stop the PR to be merged, but needs to be solved before the release "IMO".
> 
> Thanks for catching the issues @jespino , I hope it's fine if we merge this and then I can proceed with [this PR](https://github.com/mattermost/focalboard/pull/3596) and the corresponding cherry-pick.

Yes, that works for me too :)

Issue #3591, opened 2022-08-08
Doc: How do I build this without having mattermost/enterprise


clmnin:
## Summary

I tried building the app using the following commands
```shell
make prebuild
make
```

But it fails when it can't find the `mattermost/enterprise` repo - which is not open source. And so I tried building a stand alone linux app using `make linux-app` but this fails with the same error.

## Link to documentation page

Link to the doc I followed - [link](https://developers.mattermost.com/contribute/focalboard/personal-server-setup-guide/#building-via-the-terminal)

noxer:
Hi @clmnin, thanks for reporting this. It's a bug in our Makefile.

For now you can mitigate it by running `make setup-go-work`, opening the `go.work` file in the root folder of the repository, and removing the line `use ../mattermost-server`.

clmnin:
Thank you

PR #3593, opened 2022-08-08
Enforce passing the environment variable to the gowork generator

noxer:
#### Summary
When generating the `go.work` file, the environment variable `EXCLUDE_ENTERPRISE` was not passed to the generator (at least on Linux) which lead to the inclusion of the enterprise repo for external contributors. This PR forces `make` to pass the env variable to the generator.

Thanks to @clmnin for reporting this.

#### Ticket Link
#3591

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ ci: webapp-ci server-test ## Simulate CI, locally.
 
 setup-go-work: export EXCLUDE_ENTERPRISE ?= true
 setup-go-work: ## Sets up a go.work file
-	go run ./build/gowork/main.go
+	EXCLUDE_ENTERPRISE=$(EXCLUDE_ENTERPRISE) go run ./build/gowork/main.go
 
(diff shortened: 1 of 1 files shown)


wiggin77:
That environment var was exported for the task 2 lines above the change.  Was that not working?  I tested for Linux.

noxer:
Huh, funny. I retried it and it just works without the change. Earlier I had the exact same behavior as in #3591... Even with setting `export EXCLUDE_ENTERPRISE=true` it wouldn't work.

wiggin77:
@noxer  If the var is already defined with a false (or even blank) string then it won't be exported with a true value.  If it's not working in some case then I would look for it being defined somewhere.

noxer:
That doesn't explain why it wouldn't work with `export EXCLUDE_ENTERPRISE=true`.

noxer:
I tried it a few more times and can't reproduce it. Now it correctly ignores enterprise, even when I do the exact same steps as before.

PR #3594, opened 2022-08-08
Restricted scope of board-category change WS event to correct user

### LINKED ISSUE #3559: Bug: Moving boards to another category causes the board to disappear for other users until they refresh

1. Login with 2 users at the same time or have another person check on their Boards during testing
2. User 1: Select a board that is visible to both users (i.e., "Focalboard Developer Meeting Agenda" under the `Focalboard` category in the Contributors team)
3. User 1: Move the board to another category
4. User 2: See error - the "Focalboard Developer Meeting Agenda" is now gone for User 2 until the user refreshes the browser or switches teams and comes back

## Expected behavior

Moving the board to a new category from User 1 should not impact the board for User 2.

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/182892056-ba55db99-1a74-4214-b47a-8b58b325a637.mov

## Edition and Platform

 - Version: v7.2 RC1

harshilsharma63:
Fixes #3559

The issue was that the WS event for notifying change of a board's category was being sent to all users in team. It should be sent only to the user who performed the action. In this PR I've reduced the WS event's scope only to the correct user for both direct WS event and events received via cluster.

### DIFF
--- a/server/ws/plugin_adapter.go
+++ b/server/ws/plugin_adapter.go
@@ -486,8 +486,8 @@ func (pa *PluginAdapter) BroadcastCategoryChange(category model.Category) {
 
 	go func() {
 		clusterMessage := &ClusterMessage{
-			Payload:     payload,
-			EnsureUsers: []string{category.UserID},
+			Payload: payload,
+			UserID:  category.UserID,
 		}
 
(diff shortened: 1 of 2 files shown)


jespino (approved):
LGTM

PR #3598, opened 2022-08-08
Making more robust websocket notifications on synthetic memberships

### LINKED ISSUE #3477: Bug: board gets deleted from frontend if my membership gets removed but I'm still a member through an associated channel

1. From the perspective of user A, associate a board with a channel and add user B to the board, which is a member of the channel
2. In an incognito window, login with user B and go to said board
3. As the user A, remove user B membership
4. User B will lose the board on both the LHS and main window if it's there, and will need to refresh to see it again

## Expected behavior

As user B is still a member of the board, the board should not get removed from the frontend

## Edition and Platform

 - Version: v7.2
 - Browser and OS: Firefox 102.0.1

## Additional context

When we get a membership deletion message, if it's for the `me` user, we should check if the user has a synthetic membership before removing its membership from the store, and just update it with the synthetic one

jespino:
#### Summary
Syntetic membeships weren't properly propagate through the websockets events,
this PR try to address that problem, sending information to the websockets when
something happen with the channel linking/unlinking and respecting the
non-syntetic meberships too.

#### Ticket Link
Fixes #3477


server/app/boards.go
```diff
@@ -250,20 +250,67 @@ func (a *App) CreateBoard(board *model.Board, userID string, addMember bool) (*m
 		if addMember {
 			a.wsAdapter.BroadcastMemberChange(newBoard.TeamID, newBoard.ID, member)
 		}
+		if newBoard.ChannelID != "" {
```
mgdelacroix, line 253: Won't this send again the `member` membership? I think we could fix this moving the `if addMember` condition to be the `else` of this line

server/app/boards.go
```diff
@@ -250,20 +250,67 @@ func (a *App) CreateBoard(board *model.Board, userID string, addMember bool) (*m
...
+		if patch.ChannelID != nil && *patch.ChannelID != "" {
+			members, err := a.GetMembersForBoard(updatedBoard.ID)
+			if err != nil {
+				a.logger.Error("Unable to get the board members", mlog.Err(err))
+			}
+			for _, member := range members {
+				a.wsAdapter.BroadcastMemberChange(updatedBoard.TeamID, member.BoardID, member)
+			}
+		} else if patch.ChannelID != nil && *patch.ChannelID == "" {
+			members, err := a.GetMembersForBoard(updatedBoard.ID)
```
mgdelacroix, line 293: I'm not sure I'm understanding this piece of logic. If the non-synthetic membership always takes precedence over the synthetic one, how is it that I'm going to get a synthetic membership on `oldMembers` and a non-synthetic one after removing the channel link? If the non-synthetic membership was there, I would have fetched it on the `oldMembers`, so there are no chances for the `syntheticOnly` condition to be `true`, right?

server/app/boards.go
jespino: Good point, the answer is yes, not a big deal, but can be improved, let me fix it.

server/app/boards.go
jespino: I see what you mean, really good catch, I'll fix that.

jespino: Done

server/app/boards.go
jespino: Done

mgdelacroix (approved):
Awesome, LGTM! :+1: Thanks for working on the changes!

PR #3607, opened 2022-08-08
fix bad translation causing parse errors

sbishel:
#### Summary
Fixes a bad line that is causing parsing errors in `en-AU.json`. 

Also ran `npm run i18n-extract`.


cwarnermm (approved):
Thanks, @sbishel!

Issue #3601, opened 2022-08-08
Bug: Delete option shows in read-only table-view rows


chenilim:
## Steps to reproduce the behavior

1. Open a board with a table view and share it publicly
2. Open the public link in another browser
3. Hover over a row in a table view
4. The drag grip appears with a delete option

## Expected behavior

No drag grip, nor delete option should appear when read-only

## Screenshots (optional)

https://user-images.githubusercontent.com/46905241/183457928-ad8dac64-f4ab-4483-878e-44022c92108a.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.2 bugbash (8/8/22)
 - Browser and OS: Chrome on Mac

## Additional context (optional)

Selecting delete from the delete dialog doesn't do anything, but it should not be available in the first place.

PR #3611, opened 2022-08-08
check readonly flag before displaying row menu

sbishel:
#### Summary
Don't display drag and delete menu in table view if readonly

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/3601


Rajat-Dabade (approved):
LGTM 👍

PR #3613, opened 2022-08-08
Update teamless boards migration to check for deleted entities

### LINKED ISSUE #3588: Bug: Lost board when new permissions rolled out

After permissions and the new LHS was added, I am no longer able to find the board I had in my DM with myself. (I believe the board name was ```katie.wiersgalla,katie.wiersglla```  - that is how I found it when searching for it from the prior Boards UI. The view I believe was named "Personal Tasks", but I am not completely certain. 

I have a bookmark to it on Chrome, and when I select the bookmark, I get a white page.  Here's the URL: https://community.mattermost.com/boards/workspace/furnzjswi3rf7cg65ix9iuhfrw/074aca72-00c3-4708-852c-a6289f096f54/e40b0d8f-a66a-434a-806e-572c94ed8c6f.  

I am unable to find the board listed under any other new category in the Boards LHS including the generic "Boards" category. I also seem to be missing a shared board in a DM I have with Corey Hulen for managing 1:1s. 

## Screenshots (optional)

image

## Edition and Platform

 - Edition: Community.mattermost.com / Mattermost Boards (I believe 7.2)

mgdelacroix:
#### Summary
After this changes, the `TeamLessBoardsMigration` will check the team and the team membership of the users to discard deleted entities when cosidering what team to assign a board to.

Function names updated for consistency with the rest of the migrations.

PR labelled as `Do Not Merge` as I want to run some more tests focused on deleted memberships tomorrow. Please review the approach as it's code completed.

#### Ticket Link
Fixes #3588

### DIFF
--- a/server/services/store/sqlstore/data_migrations.go
+++ b/server/services/store/sqlstore/data_migrations.go
@@ -347,7 +347,7 @@ func (s *SQLStore) createCategoryBoards(db sq.BaseRunner) error {
 // We no longer support boards existing in DMs and private
 // group messages. This function migrates all boards
 // belonging to a DM to the best possible team.
-func (s *SQLStore) migrateTeamLessBoards() error {
+func (s *SQLStore) runTeamLessBoardsMigration() error {
 	if !s.isPlugin {
 		return nil
 	}
@@ -376,7 +376,7 @@ func (s *SQLStore) migrateTeamLessBoards() error {
 
 	tx, err := s.db.BeginTx(context.Background(), nil)
 	if err != nil {
-		s.logger.Error("error starting transaction in migrateTeamLessBoards", mlog.Err(err))
+		s.logger.Error("error starting transaction in runTeamLessBoardsMigration", mlog.Err(err))
 		return err
 	}
 
@@ -417,7 +417,7 @@ func (s *SQLStore) migrateTeamLessBoards() error {
 	}
 
 	if err := tx.Commit(); err != nil {
-		s.logger.Error("failed to commit migrateTeamLessBoards transaction", mlog.Err(err))
+		s.logger.Error("failed to commit runTeamLessBoardsMigration transaction", mlog.Err(err))
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


server/services/store/sqlstore/data_migrations.go
```diff
@@ -514,10 +514,15 @@ func (s *SQLStore) getBestTeamForBoard(tx sq.BaseRunner, board *model.Board) (st
 
 func (s *SQLStore) getBoardUserTeams(tx sq.BaseRunner, board *model.Board) (map[string][]string, error) {
 	query := s.getQueryBuilder(tx).
-		Select("TeamMembers.UserId", "TeamMembers.TeamId").
-		From("ChannelMembers").
-		Join("TeamMembers ON ChannelMembers.UserId = TeamMembers.UserId").
-		Where(sq.Eq{"ChannelId": board.ChannelID})
+		Select("tm.UserId", "tm.TeamId").
```
wiggin77, line 517: Would it make sense to also modify `getBestTeamForBoard` line 490 so that instead of just returning the first team left after filtering, it selects the team based on highest number of overall users.  In other words, a very small team is more likely to be deleted versus a large one.  On that note, what happens if/when the team selected is deleted later?  Do we move the board to some arbitrary team like we do at migration time?  Should we ask the user to pick?

mgdelacroix: On the first question, I agree on a more populated board being a better guess than just the first, although at this point in time I'm a bit reluctant to make more changes than the strictly required. As per the second question, once the migration has finished, the board is just a private board part of the team that it belongs to, with no special considerations.

I think the best solution for both cases is to provide the user with the ability to move boards between teams, so independently of the decision that this migration will take on a specific board, the user will have the final say on where the board should be.

sbishel (approved):
I do like Doug's suggestion. But looking through the "other" teams we have, most are either empty or have test users. So, I do think this fix would be fine for community. 

That said, I'm not sure how others installations use teams and if they are commonly deleted. It would be good to have a defined process for determining which team it got moved to. But every change is a risk. 

Let's move forward with this fix, but @mgdelacroix if you could see how easy/risky it may be to check the team counts.

harshilsharma63 (approved):
LGTM 👍

PR #3624, opened 2022-08-09
Rest API reorganization

wiggin77:
#### Summary
This PR re-organizes the Rest API by breaking up the large api.go file and grouping APIs in separate files. 

This PR contains no functional changes.  It is simply moving code within a package.

#### Ticket Link
NONE


server/api/auth.go
```diff
@@ -19,123 +17,16 @@ import (
...
-	}
-	return isValidPassword(rd.NewPassword)
-}
-
-func isValidPassword(password string) error {
-	if len(password) < MinimumPasswordLength {
-		return ParamError{fmt.Sprintf("password must be at least %d characters", MinimumPasswordLength)}
-	}
-	return nil
+	r.HandleFunc("/users/{userID}/changepassword", a.sessionRequired(a.handleChangePassword)).Methods("POST")
```
jespino, line 29: This call is only allow if it is a plugin, so... probably we can move it inside the `isPlugin` block, clearly... it is not something that needs to be in this PR. actually I prefer a different one for that, because this is already huge change of restructuring.

jespino (approved):
LGTM. I can't read this whole thing, but as far as I understand this is just moving code from one place to another, and the main real change is in the register route functions that looks good to me. And also the integration tests are exercising most (if not all) of them.

server/api/auth.go
wiggin77: I have made note of this.  Thanks!

wiggin77:
> as far as I understand this is just moving code from one place to another, and the main real change is in the register route functions

Correct on both counts.

PR #3634, opened 2022-08-10
Move rest API for changepassword to personal server route

wiggin77:
#### Summary
The rest api for changing password is not needed for plugin mode.  This PR ensures it does not get registered when in plugin mode.

See https://github.com/mattermost/focalboard/pull/3624#discussion_r942309325

#### Ticket Link
NONE


jespino (approved):
LGTM

PR #3635, opened 2022-08-10
Adds data migration to clean badly assigned boards

mgdelacroix:
#### Summary
This data migration fetches all boards whose owner has a deleted membership on the board's team and fixes them by processing them again, this time with the fix applied to the `getBestTeamForBoard` function that skips deleted teams and team memberships

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/3588

### DIFF
--- a/server/services/store/sqlstore/data_migrations.go
+++ b/server/services/store/sqlstore/data_migrations.go
@@ -570,6 +570,14 @@ func (s *SQLStore) runDeletedMembershipBoardsMigration() error {
 		return err
 	}
 
+	if len(boards) == 0 {
+		s.logger.Debug("No boards with owner not anymore on their team found, marking runDeletedMembershipBoardsMigration as done")
+		if err := s.SetSystemSetting(DeletedMembershipBoardsMigrationKey, strconv.FormatBool(true)); err != nil {
+			return fmt.Errorf("cannot mark migration as completed: %w", err)
+		}
+		return nil
+	}
+
 	s.logger.Debug("Migrating boards with owner not anymore on their team", mlog.Int("count", len(boards)))
 
 	tx, err := s.db.BeginTx(context.Background(), nil)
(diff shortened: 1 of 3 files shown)


server/services/store/sqlstore/data_migrations.go
```diff
@@ -547,3 +549,91 @@ func (s *SQLStore) getBoardUserTeams(tx sq.BaseRunner, board *model.Board) (map[
...
+	if !s.isPlugin {
+		return nil
+	}
+
+	setting, err := s.GetSystemSetting(DeletedMembershipBoardsMigrationKey)
+	if err != nil {
+		return fmt.Errorf("cannot get deleted membership boards migration state: %w", err)
+	}
+
+	// If the migration is already completed, do not run it again.
```
wiggin77, line 563: Is this a migration that would be safe and useful to run every time?

server/services/store/sqlstore/data_migrations.go
```diff
@@ -547,3 +549,91 @@ func (s *SQLStore) getBoardUserTeams(tx sq.BaseRunner, board *model.Board) (map[
...
+	}
+
+	boards, err := s.getDeletedMembershipBoards(s.db)
+	if err != nil {
+		return err
+	}
+
+	s.logger.Debug("Migrating boards with owner not anymore on their team", mlog.Int("count", len(boards)))
+
+	tx, err := s.db.BeginTx(context.Background(), nil)
```
wiggin77, line 575: Perhaps don't create a transaction if len(boards) == 0?

server/services/store/sqlstore/data_migrations.go
mgdelacroix: That's a very good point, but I don't think we should run this migration every time, as with the permissions system in place, there are "legal" ways of getting to that point. What we should do probably is to spec out what happens when a user leaves a team / is completely deleted, and that user is the only admin of a board.

We could probably mimic the same thing that channels is doing for channel admins right now

server/services/store/sqlstore/data_migrations.go
wiggin77: I guess this doesn't matter if we're not going to run this migration every time.

wiggin77 (approved):
LGTM :+1:

server/services/store/sqlstore/data_migrations.go
mgdelacroix: Anyhow, fixed in https://github.com/mattermost/focalboard/pull/3635/commits/d23e6e41b09cf82775dbc646b66f908312875677 :)

Issue #3625, opened 2022-08-09
Bug: Channel role groups should not be visible on the template editor Share dialog


wuwinson:
## Steps to reproduce the behavior

1. Go to template editor
2. Open the Share dialog
3. Search for a channel
4. See error - channel role groups should not be available on the template editor
5. Assigning a channel to the template doesn't actually do anything as well

## Expected behavior

Channel role groups are only available on boards, but not on templates. Templates should only allow team access or explicit membership for editing.

## Screenshots (optional)

![Screen Shot 2022-08-09 at 4 17 30 PM](https://user-images.githubusercontent.com/93531870/183752678-27a45f9c-ed04-4884-aa71-e207020dca12.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.2 RC2
 - Browser and OS: Chrome on Mac

PR #3639, opened 2022-08-10
don't display channel roles if board is a template

sbishel:
#### Summary
Don't display channel roles in ShareBoard if board is a template.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/3625


harshilsharma63 (approved):
LGTM 👍

PR #3651, opened 2022-08-11
fix: don't re-.push() the URL if already on teamId

### LINKED ISSUE #3530: Bug: Switching boards takes you to the template picker.
See video attached below.
https://user-images.githubusercontent.com/11034289/182689961-1a7f1403-d36f-4ab5-9cf4-ecd1987ddd61.mp4

## Expected behavior
Boards should switch normally.

## Edition and Platform

 - Version: 7.2
 - Browser and OS: Chrome Mac

Pinjasaur:
#### Summary

This has been a painful bug.

The root cause is we have logic within `mattermost-plugin/webapp` to do `.subscribe()`-ing to both the Mattermost redux store & the Focalboard redux store and keep a few specific items in sync. For this bug, we we're focusing on the current team ID.

Previously, when syncing, if the current team ID changed we would re-`.push()` the URL to go to the team, even if we were already on a URL for the same team. This was causing a sort of race/circular condition where depending on how the stores sync'd up it might associate the `lastBoardId` with an incorrect `teamId` and it all breaks subtly by rendering the template selector.

This change _should_ fix the above case however we've seen some instances of navigating Boards -> Channels -> Boards not rendering the last viewed board + view for the team. Again, difficult to reproduce so we're getting this merged in to able to test in a bug bash environment.

#### Ticket Link

closes #3530

### DIFF
--- a/mattermost-plugin/webapp/src/index.tsx
+++ b/mattermost-plugin/webapp/src/index.tsx
@@ -241,7 +241,9 @@ export default class Plugin {
             const currentTeamID = mmStore.getState().entities.teams.currentTeamId
             if (currentTeamID && currentTeamID !== prevTeamID) {
                 if (prevTeamID && window.location.pathname.startsWith(windowAny.frontendBaseURL || '')) {
-                    browserHistory.push(`/team/${currentTeamID}`)
+                    // Don't re-push the URL if we're already on a URL for the current team
+                    if (!window.location.pathname.startsWith(`${(windowAny.frontendBaseURL || '')}/team/${currentTeamID}`))
+                        browserHistory.push(`/team/${currentTeamID}`)
                 }
                 prevTeamID = currentTeamID
                 store.dispatch(setTeam(currentTeamID))


Pinjasaur:
@lindy65 it's possible this might break some Rainforest tests—likely manifesting in either 1) seeing the template selector rendered when you weren't expecting it or 2) _not_ seeing the template selector render when you were expecting it previously.

lindy65:
Thanks @Pinjasaur - this is likely to break all the tests as the starting point is the template selector page

>likely manifesting in either 1) seeing the template selector rendered when you weren't expecting it or 2) not seeing the template selector render when you were expecting it previously.

Can you help me understand why the resultant behavior will be an "either or" situation?  It's really difficult to deal with these types of scenarios in Rainforest as the automation cannot "understand" either do this;  or do that

Pinjasaur:
> Can you help me understand why the resultant behavior will be an "either or" situation? It's really difficult to deal with these types of scenarios in Rainforest as the automation cannot "understand" either do this; or do that

Sure thing @lindy65! Apologies for not explaining better initially:

Previously, there was scenarios where if you changed to a board on a different team than you're currently on either via the boards switcher dialog or the linked boards<->channels dialog it would render the template selector. This PR aims to fix that exact bug.

However, when testing this PR locally we were sometimes seeing behavior where if we changed from Boards -> Channels -> Boards the template selector would be rendered instead of the board & view that was previously shown.

So it's less of an "either or" and more of you _might_ see behavior of the template selector being rendered when it previously was not (or vice versa). We have this slated to be released on our [bugbash server](https://boards-bugbash.test.mattermost.cloud/) and tested later today which should give us more confidence on behavior of this PR.

Lmk if anything above didn't make sense—happy to help! 🙂

lindy65:
Thanks so much for the great explanation @Pinjasaur 😄 

>However, when testing this PR locally we were sometimes seeing behavior where if we changed from Boards -> Channels -> Boards the template selector would be rendered instead of the board & view that was previously shown.

Am I correct in saying that this is perhaps a bug that needs to be looked at?  Or is it more dependent on team / permissions / etc when switching from Boards > Channels > Boards?

>and more of you might see behavior of the template selector being rendered when it previously was not (or vice versa).

Unfortunately, this scenario would still break tests as it would be difficult to know whether or not to include a step in the test to expect the template screen

I think what I'm trying to ask (in a not so eloquent way!) is whether this issue is expected to be fixed down the line to a scenario where we will definitely see the template screen or definitely not see it?

Of course, I'm sure you'll have more info / feedback once this PR's been tested on your bugbash server - happy to wait until then for a response :)

Pinjasaur:
> Am I correct in saying that this is perhaps a bug that needs to be looked at? Or is it more dependent on team / permissions / etc when switching from Boards > Channels > Boards?
> 
> Of course, I'm sure you'll have more info / feedback once this PR's been tested on your bugbash server - happy to wait until then for a response :)

The bugbash went well—we didn't see the Boards -> Channels -> Boards behavior so I believe we can chock it up to a fluke locally. 

> I think what I'm trying to ask (in a not so eloquent way!) is whether this issue is expected to be fixed down the line to a scenario where we will definitely see the template screen or definitely not see it?

Excellent question. I _think_ the answer is it should only be shown when you're explicitly wanting to select a template i.e. you're creating a new board or as the default case of you're 1) viewing a team for the first time or 2) for some reason we don't have access to the last board you viewed for a team. Anything obvious I missed there @wuwinson?

The reality is we have a number of bugs/regressions (like this PR) that were causing unwanted behavior. Another one that I'm looking to solve ASAP is #3654.

Sorry for the vague answer @lindy65. Appreciate your patience. 🙂

wuwinson:
@lindy65 @Pinjasaur I documented the expected behavior in [this spec](https://docs.google.com/document/d/1HOnoB87WZojqplpB6XGbtKsLuqjG8h-8CbJpmWsRfM4/edit#bookmark=id.92yhyce3rfrz) and just updated it now to reflect the app switcher (top left grid button) instead of the original channel header button:

- If accessing Boards for the first time, then display FTUE starting with Welcome screen
- If not a first time user, but user has not viewed any boards, then display **template picker**
- Otherwise, display last viewed board
- If last view board does not exist or user is no longer a member of the board, display the first board/view in his list. If user has no boards, then he can land to the **template picker**

lindy65:
Thanks @Pinjasaur and @wuwinson - this has cleared things up for me :) Appreciate all the info and spec!  Glad the bug-bash went well!

Looping @lindalumitchell in on the conversation for visibility into the issue and spec

PR #3670, opened 2022-08-12
Add foundation tests

mgdelacroix:
#### Summary
This PR adds the necessary instrumentation code to allow for SQL and data migrations tests, and creates a couple of initial tests (one an sql-only test and another one for a data migration) as an example.

The tests can run in "normal" mode, that is, only with the Boards migrations, or in Plugin mode, as the ones included in the PR, in which case they'll run first all the `mattermost-server` database migrations and then will use the Boards ones.


wiggin77 (approved):
Very nice!  :+1:

harshilsharma63 (approved):
LGTM 👍

PR #3672, opened 2022-08-12
Only allow viewer role in temaPermissionRow for templates

### LINKED ISSUE #3543: Bug: Custom templates shared with the team are misleading with the role assignment

1. Open the template picker
2. Create new template
3. Share the template with the team by setting the team role as "Editor"
4. Log-in as another team member
5. Open the template picker and hover over the template created in step 2
6. See error - User from step 4 is not able to edit the template even though they are "Editors" of the template

## Expected behavior

The team role should display something less misleading, so it's clear to the template admin that sharing with the team only gives the team "viewer" access and not edit access.

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/182745971-5b8fc448-13bd-4974-8047-bdb1fbb2a4cc.mov

## Edition and Platform

 - Version: v7.2

jespino:
#### Summary
Only allow viewer role in temaPermissionRow for templates. Also I show "Viewer"
instead of "Editor" if you are currently stablish as editor. We can change that
in the future and add a migration for that.

#### Ticket Link
Fixes #3543


sbishel (approved):

sbishel:
@jespino - Look's like some tests failures. Also when this gets cherry-picked, we will have to add back the `Viewer` role as it was removed in `release-7.2`
https://github.com/mattermost/focalboard/blob/release-7.2/webapp/src/components/shareBoard/teamPermissionsRow.tsx#L73-L88

Issue #3669, opened 2022-08-12
Bug: 3rd tour point (Manage boards) for v7.2 does not load


wuwinson:
## Steps to reproduce the behavior

1. Go to Boards for the first time on Community (RC3 as of Friday Aug 12th)
2. Click through 1st and 2nd tour point, then click "Next"
3. See error - the 3rd tour point does not appear. Even if I expand the first category or other categories, I still don't see the 3rd tour point. **Edit**: Looks like the 3rd tour point briefly appears after a browser refresh, but then disappears again. See attached recording.

## Expected behavior

3rd tour point "Manage boards" should pop-over on the Options Menu button from the first board under first category. Potentially related to https://github.com/mattermost/focalboard/pull/3579 since all migrated categories are now collapsed by default. Can we force the first board under the first category to be visible or selected?

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/184387003-3b96555b-c2c5-4cad-847c-084dcc1484ae.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.2 RC3
 - Browser and OS: Chrome on Mac

wuwinson:
cc @sbishel @Rajat-Dabade @harshilsharma63

PR #3681, opened 2022-08-15
fix: `Viewer` only on a template, not board

Pinjasaur:
#### Summary

Add a conditional to only render `Viewer` on a (custom) template sharing dialog, not on a board. Notice this is direct to 7.2 w/ no cherrypick. I discussed this with @wuwinson and we agreed that `main` is working as expected for 7.3, which will have more roles e.g. `Commenter`.

image

image

#### Ticket Link

closes #3678

PR #3684, opened 2022-08-15
Remove manageBoards Tour Point

sbishel:
#### Summary
Removes the ManageBoards Tour Point.

Fixing this issue was more than a quick fix. Based on the way we load the sidebar, it's not easy to tell which is the first visible board (cannot simply use index 0, since board may be hidden. Also these are loaded looping through "all boards", which means index(0) is the first board alphabetically.

I tried several quick fixes, but nothing worked correctly. Talked to @wuwinson and decided to pull the tour point vs spending more time on it.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/3669


Pinjasaur:
@sbishel confirming we're looking to remove it on `main` _and_ cherrypick to `release-7.2` too? wasn't sure if it could be removed in 7.2 and fixed for 7.3? 0/5

wuwinson:
We should try to fix this in `v7.3` if possible, but don't want to hold up the `v7.2` release over this.

sbishel:
> confirming we're looking to remove it on main and cherrypick to release-7.2 too? wasn't sure if it could be removed in 7.2 and fixed for 7.3? 0/5

Yes, if we want to reimplement for 7.3, we can "revert" this, but I don't want to leave dead code in there.

webapp/src/components/shareBoard/teamPermissionsRow.tsx
```diff
@@ -91,13 +91,15 @@ const TeamPermissionsRow = (): JSX.Element => {
...
                                     onClick={() => updateBoardType(board, BoardTypeOpen, 'commenter')}
                                 />}
-                            <Menu.Text
-                                id='Viewer'
-                                check={board.minimumRole === 'viewer'}
-                                icon={board.type === BoardTypeOpen && board.minimumRole === 'viewer' ? <CheckIcon/> : null}
-                                name={intl.formatMessage({id: 'BoardMember.schemeViwer', defaultMessage: 'Viewer'})}
-                                onClick={() => updateBoardType(board, BoardTypeOpen, 'viewer')}
-                            />
+                            {board.isTemplate &&
```
Pinjasaur, line 94: We don't need this conditional—it's why the snapshot tests are failing. It only needed to be fixed in `release-7.2` as I did earlier in #3681.

Pinjasaur (approved):
@sbishel one note, otherwise it LGTM once that's fixed & CI is happy 🎉

webapp/src/components/shareBoard/teamPermissionsRow.tsx
sbishel: Yep...bad merge. This change shouldn't be here.

Issue #3676, opened 2022-08-13
Bug: Clicking + New on an existing board does not create a card


beastarman74:
## Steps to reproduce the behavior

1. Go to a board that existed pre v7.3 in Focalboard
2. Click + New at the bottom of any status
3. No new card item is generated
4. The following error is recorded in the docker log

```today at 7:44:47 AMerror [2022-08-13 07:44:47.198 +01:00] API ERROR                                caller="api/api.go:179" code=404 error= msg= api=/api/v2/boards/bzr39487fc7rx9jazrxgti9pmbr/blocks/czw9es1e89fdpjr7cqptr1xq7qh/duplicate
today at 7:44:47 AMdebug [2022-08-13 07:44:47.201 +01:00] attachSession                            caller="api/auth.go:349" single_user=false
today at 7:44:47 AMdebug [2022-08-13 07:44:47.252 +01:00] PATCH Block                              caller="api/blocks.go:566" boardID=bzr39487fc7rx9jazrxgti9pmbr blockID=v43abcfqoxjf8zrhb9wgnih81hy
today at 7:44:47 AMdebug [2022-08-13 07:44:47.255 +01:00] Broadcast block change                   caller="ws/server.go:549" teamID=0 blockID=v43abcfqoxjf8zrhb9wgnih81hy remoteAddr="192.168.1.21:59642"
today at 7:44:47 AMdebug [2022-08-13 07:44:47.261 +01:00] Block change event                       caller="notifylogger/logger_backend.go:48" action=update board="Home Lab Tasks " card= block_id=v43abcfqoxjf8zrhb9wgnih81hy
today at 7:45:00 AMdebug [2022-08-13 07:45:00.168 +01:00] attachSession                            caller="api/auth.go:349" single_user=false
today at 7:45:00 AMerror [2022-08-13 07:45:00.186 +01:00] API ERROR                                caller="api/api.go:179" code=404 error= msg= api=/api/v2/boards/bzr39487fc7rx9jazrxgti9pmbr/blocks/czw9es1e89fdpjr7cqptr1xq7qh/duplicate
today at 7:45:00 AMdebug [2022-08-13 07:45:00.199 +01:00] attachSession                            caller="api/auth.go:349" single_user=false
[...]

PR #3696, opened 2022-08-17
Fixed the new button error in kanban

Rajat-Dabade:
#### Summary
The PR fixes the `+ New` button issue in the kanban view which was not allowing the create a card.

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/3676

### DIFF
--- a/webapp/src/components/kanban/kanban.tsx
+++ b/webapp/src/components/kanban/kanban.tsx
@@ -58,13 +58,15 @@ const Kanban = (props: Props) => {
     const currentView = useAppSelector(getCurrentView)
     const cardTemplates: Card[] = useAppSelector(getCurrentBoardTemplates)
     const {board, activeView, cards, groupByProperty, visibleGroups, hiddenGroups, hiddenCardsCount} = props
-    const [defaultTemplateID, setDefaultTemplateID] = useState<string>()
+    const [defaultTemplateID, setDefaultTemplateID] = useState('')
 
     useEffect(() => {
-        if(currentView && currentView.fields && currentView.fields.defaultTemplateId) {
-            if(cardTemplates.find(ct => ct.id === currentView.fields.defaultTemplateId)) {
-                setDefaultTemplateID(currentView.fields.defaultTemplateId)
-            }
+        if(currentView.fields && currentView.fields.defaultTemplateId) {
+            cardTemplates.forEach((cardTemplate) => {
+                if(cardTemplate.id === currentView.fields.defaultTemplateId) {
+                    setDefaultTemplateID(currentView.fields.defaultTemplateId)
+                }
+            })
         }
     }, [currentView, currentView.fields.defaultTemplateId])
 
@@ -308,7 +310,7 @@ const Kanban = (props: Props) => {
                             <BoardPermissionGate permissions={[Permission.ManageBoardCards]}>
                                 <Button
                                     onClick={() => {
-                                        if(defaultTemplateID) {
+                                        if(defaultTemplateID !== '') {
                                             props.addCardFromTemplate(defaultTemplateID, group.option.id)
                                         } else {
                                             props.addCard(group.option.id, true)
--- a/webapp/src/components/kanban/kanban.tsx
+++ b/webapp/src/components/kanban/kanban.tsx
@@ -61,7 +61,7 @@ const Kanban = (props: Props) => {
     const [defaultTemplateID, setDefaultTemplateID] = useState('')
 
     useEffect(() => {
-        if(currentView.fields.defaultTemplateId) {
+        if(currentView.fields && currentView.fields.defaultTemplateId) {
             cardTemplates.forEach((cardTemplate) => {
                 if(cardTemplate.id === currentView.fields.defaultTemplateId) {
                     setDefaultTemplateID(currentView.fields.defaultTemplateId)
--- a/webapp/src/components/kanban/kanban.tsx
+++ b/webapp/src/components/kanban/kanban.tsx
@@ -60,13 +60,15 @@ const Kanban = (props: Props) => {
     const {board, activeView, cards, groupByProperty, visibleGroups, hiddenGroups, hiddenCardsCount} = props
     const [defaultTemplateID, setDefaultTemplateID] = useState<string>()
 
-    useEffect(() => {
-        if(currentView && currentView.fields && currentView.fields.defaultTemplateId) {
-            if(cardTemplates.find(ct => ct.id === currentView.fields.defaultTemplateId)) {
-                setDefaultTemplateID(currentView.fields.defaultTemplateId)
+    if(currentView) {
+        useEffect(() => {
+            if(currentView.fields.defaultTemplateId) {
+                if(cardTemplates.find(ct => ct.id === currentView.fields.defaultTemplateId)) {
+                    setDefaultTemplateID(currentView.fields.defaultTemplateId)
+                }
             }
-        }
-    }, [currentView.fields.defaultTemplateId])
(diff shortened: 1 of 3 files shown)


webapp/src/components/kanban/kanban.tsx
```diff
@@ -55,7 +56,19 @@ const vStrength = createVerticalStrength(Utils.isMobile() ? 60 : 250)
...
     const {board, activeView, cards, groupByProperty, visibleGroups, hiddenGroups, hiddenCardsCount} = props
+    const [defaultTemplateID, setDefaultTemplateID] = useState('')
+
+    useEffect(() => {
+        if(currentView.fields && currentView.fields.defaultTemplateId) {
+            cardTemplates.forEach((cardTemplate) => {
+                if(cardTemplate.id === currentView.fields.defaultTemplateId) {
+                    setDefaultTemplateID(currentView.fields.defaultTemplateId)
+                }
+            })
```
sbishel, line 69: Suggest changing this to a `.find()`. Breaking out of a `.foreach` isn't possible, so you will continue to loop even after finding the expected value. (more important on a large set). `.find` will return the first element it finds and we only expect one. 

```suggestion
            if(cardTemplates.find(ct => ct.id === currentView.fields.defaultTemplateId)){
                setDefaultTemplateID(currentView.fields.defaultTemplateId)
            }
```

webapp/src/components/kanban/kanban.tsx
```diff
@@ -55,7 +56,19 @@ const vStrength = createVerticalStrength(Utils.isMobile() ? 60 : 250)
 
 const Kanban = (props: Props) => {
     const currentView = useAppSelector(getCurrentView)
+    const cardTemplates: Card[] = useAppSelector(getCurrentBoardTemplates)
     const {board, activeView, cards, groupByProperty, visibleGroups, hiddenGroups, hiddenCardsCount} = props
+    const [defaultTemplateID, setDefaultTemplateID] = useState('')
```
sbishel, line 61: If state isn't set, I would prefer it be empty, rather than empty string. Will require change to line 313.
```suggestion
    const [defaultTemplateID, setDefaultTemplateID] = useState<string>();
```

webapp/src/components/kanban/kanban.tsx
```diff
@@ -297,8 +310,8 @@ const Kanban = (props: Props) => {
                             <BoardPermissionGate permissions={[Permission.ManageBoardCards]}>
                                 <Button
                                     onClick={() => {
-                                        if(currentView.fields.defaultTemplateId) {
-                                            props.addCardFromTemplate(currentView.fields.defaultTemplateId, group.option.id)
+                                        if(defaultTemplateID !== '') {
```
sbishel, line 313:
```suggestion
                                        if(defaultTemplateID) {
```

sbishel (changes requested):
Just some minor improvements, looks good in general approach.

sbishel:
We will want to cherry pick this back to `release-7.2` as we want to include it in the next dot release. Please set Labels and Milestone appropriately.

webapp/src/components/kanban/kanban.tsx
Rajat-Dabade: Got it.

webapp/src/components/kanban/kanban.tsx
Rajat-Dabade: Okay

webapp/src/components/kanban/kanban.tsx
Rajat-Dabade: Done 👍

webapp/src/components/kanban/kanban.tsx
```diff
@@ -55,7 +56,17 @@ const vStrength = createVerticalStrength(Utils.isMobile() ? 60 : 250)
...
     const {board, activeView, cards, groupByProperty, visibleGroups, hiddenGroups, hiddenCardsCount} = props
+    const [defaultTemplateID, setDefaultTemplateID] = useState<string>()
+
+    useEffect(() => {
+        if(currentView && currentView.fields && currentView.fields.defaultTemplateId) {
+            if(cardTemplates.find(ct => ct.id === currentView.fields.defaultTemplateId)) {
+                setDefaultTemplateID(currentView.fields.defaultTemplateId)
+            }
+        }
+    }, [currentView])
```
sbishel, line 69: Since you are only using `currentView.fields.defaultTemplateId` within the use effect. That should be your dependency. If other fields within `currentView` or `currentView.fields` change, we don't need to rerun this effect, only if `currentView.fields.defaultTemplateId` changes. Sorry missed that earlier. 

```suggestion
    useEffect(() => {
        if(currentView.fields.defaultTemplateId) {
            if(cardTemplates.find(ct => ct.id === currentView.fields.defaultTemplateId)) {
                setDefaultTemplateID(currentView.fields.defaultTemplateId)
            }
        }
    }, [currentView.fields.defaultTemplateId])
```

sbishel (changes requested):
Sorry, one other fix I missed earlier.

webapp/src/components/kanban/kanban.tsx
Rajat-Dabade: Yes, the dependency must be changed, but the condition in the if statement should remain the same because the component `kanban` get mounted on the application when it runs for the first without any board, that time the value of `currentView` will be `undefined` which will result into the error `Cannot access the property fields of undefined`, that the reason why we need to check all the 3 values.

Rajat-Dabade: I have made some more changes because it was failing the git CI. Please have a look, thanks.

Rajat-Dabade: I guess we need to keep it the same way because whenever the application is loaded for the first time, the component `kanban` will be mounted and the value of the `currentView` will be `undefined` and it will throw an error `cannot find the property fields of undefined.`. And I guess the dependency will not have an impact because the `defaultTemplateView` will only change if the `currentView` changes.  I have tried your suggested changes but it seems that some test case of cypress was failing due to `cannot read the property fields of undefined`. Thoughts.

sbishel (commented):
@Rajat-Dabade  Sorry, I thought that should work even with a null object. However, it looks like you reverted the wrong commit. As it appears back to the original code. I will try to take a look today, but may not have time.

sbishel:
Pulled it down. It does work if you update only the dependency portion. However, the check still needs to have a check for each object in the chain. 

```tsx
    useEffect(() => {
        if(currentView && currentView.fields && currentView.fields.defaultTemplateId) {
            cardTemplates.forEach((cardTemplate) => {
                if(cardTemplate.id === currentView.fields.defaultTemplateId) {
                    setDefaultTemplateID(currentView.fields.defaultTemplateId)
                }
            })
        }
    }, [currentView.fields.defaultTemplateId])
```

Rajat-Dabade:
@sbishel, I have also tested by keeping the dependency as `currentView.fields.defaultTemplateId` but it seems to be failing in the cypress test in git CI. Locally it worked but in git CI it was failing.

Rajat-Dabade:
@sbishel, I have reverted the changes back.

sbishel:
Weird that it works locally and not via CI. I was testing locally. I started investigating why and noticed that we don't even need to be retrieving/setting `currentView`. The property `activeView` is passed in and based on code in `workspace.tsx`, it should always be set. So we can just replace where `currentView` is being used with `activeView`.

Rajat-Dabade:
@sbishel, it does work with `activeView`, Thanks for the suggestion.

sbishel (approved):

PR #3698, opened 2022-08-17
Remove duplicated store and app method to get blocks that belong to a board

mgdelacroix:
#### Summary
Removes duplicated logic to get blocks that belong to a board.


jespino (approved):
Nice catch.

PR #3700, opened 2022-08-17
Register API router in product mode

wiggin77:
#### Summary
This PR fixes an issue setting up Boards in product mode.  The call to register the API router was missing.

This change has no effect when running as plugin.

#### Ticket Link
NONE


mgdelacroix (approved):
LGTM! :+1:

Issue #3600, opened 2022-08-08
Bug: Non-board admin should not be able to unlink board from channel RHS


wuwinson:
## Steps to reproduce the behavior

1. Go to a channel with linked boards where you are not an admin of the board
2. Click on the options "..." menu for the board
3. See error - I am able to click on the "Unlink board" button even though I am not an admin of the board (although nothing happens from clicking on the button)

## Expected behavior

The "Unlink board" option should be disabled for non-board admins, and a subtitle should be displayed underneath the "Unlink board" option: "You are not an admin of the board"

Related to https://github.com/mattermost/focalboard/issues/3403

## Screenshots (optional)

![Screen Shot 2022-08-08 at 10 54 47 AM](https://user-images.githubusercontent.com/93531870/183449506-7f81e5b6-56cb-4031-9750-3e4b6ad71260.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.2 RC2
 - Browser and OS: Chrome on Mac

wuwinson:
cc @asaadmahmood On where to display the subtitle.

asaadmahmood:
@wuwinson Added it, we can ignore the other options for now.
![image](https://user-images.githubusercontent.com/11034289/183450849-c2631697-f2c4-46a4-90bd-ddaaf43084b6.png)

PR #3703, opened 2022-08-18
Added role so that the non-admin will not be able to unlink the board

Rajat-Dabade:
#### Summary
This PR will add functionality to not allow non-admin to unlink the board from rhs-channel

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/3600

### DIFF
--- a/mattermost-plugin/webapp/src/components/rhsChannelBoardItem.tsx
+++ b/mattermost-plugin/webapp/src/components/rhsChannelBoardItem.tsx
@@ -17,9 +17,8 @@ import {SuiteWindow} from '../../../../webapp/src/types/index'
 
 import {Permission} from '../../../../webapp/src/constants'
 
-import {useHasPermissions} from '../../../../webapp/src/hooks/permissions'
-
 import './rhsChannelBoardItem.scss'
+import BoardPermissionGate from '../../../../webapp/src/components/permissions/boardPermissionGate'
 
 const windowAny = (window as SuiteWindow)
 
@@ -35,8 +34,6 @@ const RHSChannelBoardItem = (props: Props) => {
     if (!team) {
         return null
     }
-
-    const allowed = useHasPermissions(team.id, board.id, [Permission.ManageBoardRoles])
 
     const handleBoardClicked = (boardID: string) => {
         window.open(`${windowAny.frontendBaseURL}/team/${team.id}/${boardID}`, '_blank', 'noopener')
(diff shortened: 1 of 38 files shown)


sbishel (changes requested):
Is there a reason you didn't use <BoardPermissionGate>? I think you simply need to wrap the <MenuWrapper> in a <BoardsPermissionGate> like was done here.
https://github.com/mattermost/focalboard/blob/main/webapp/src/components/shareBoard/userPermissionsRow.tsx#L57

Also we should avoid mixing Bug fixes with UI Enhancements.

Rajat-Dabade:
Thanks for the feedback @sbishel, it was my bad I was not aware of the `invert` prop. I have made the changes and pushed it. And regarding the UI changes, It was needed for this PR and also confirmed it with @asaadmahmood. Should we go forward with it?

asaadmahmood:
I think @sbishel is more referring to some other minor UI changes I did on the side.
Noted Scott. I also fixed some UI bugs as well in this, that may have taken more time to create an issue and a PR (like the linked boards in the channels RHS not being theme compatible, etc).

And some spacing/sizing issues.

When I work on a part of the code, I also see if the UI on that part can also be improved if its just a few lines of fix.

mattermost-plugin/webapp/src/components/rhsChannelBoardItem.tsx
```diff
@@ -55,18 +58,41 @@ const RHSChannelBoardItem = (props: Props) => {
...
+                                id='unlinkBoard'
+                                disabled={true}
+                                name={intl.formatMessage({id: 'rhs-boards.unlink-board1', defaultMessage: 'Unlink board Hello'})}
+                                icon={<DeleteIcon/>}
+                                onClick={() => {
+                                    onUnlinkBoard(board)
+                                }}
+                                subText={intl.formatMessage({id: 'rhs-board-non-admin-msg', defaultMessage:'You are not an admin of the board'})}
+                            />
+                        </BoardPermissionGate>
```
sbishel, line 95: This section should be deleted.

sbishel (changes requested):
@rajat @asaadmahmood - Please revert all of the UX fixes in this PR. This PR should be essentially a 3 code line/3 file change. Instead there are 38 files that have changed. I would like the UX changes pulled out separately into another PR. 

I'm ok with doing minor UI fixes in with a PR that is implementing a new feature. But when it is a bug fix, I would like to restrict those to fixing the issue. When a "minor" UI fix causes 30+ snapshot files to change, then it should be in its own PR. These changes in this PR cause it go from a couple files being changed to 38 files. On this PR, I feel 5/5 on this.

sbishel:
@Rajat-Dabade @asaadmahmood - Sorry, I didn't read what the entire fix entailed. I just expected the menu item to be hidden. Sorry for that confusion. I don't like the feature. We don't disable menus anywhere else that I know of or tell users they can't do something because they are not an admin. But ignore my comments for now. It looks good base on the instructions in the PR....sorry.

sbishel (approved):
Approving from a code perspective.

sbishel:
@Rajat-Dabade - Ready to merge

rajat:
Kindly, remove me from the github.
My github handle is rajat.
I am not part of this thread.


Thanks and Regards,
Rajat Kumar

PR #3708, opened 2022-08-19
Do not shortcircuit the schema table populate process

mgdelacroix:
#### Summary
The order in which the migration slice is sorted is dependant on the `ReadDir` implementation of the filesystem. It should come back sorted by name, but I'm adding this patch just to be sure we register all the required migrations instead of shortcircuiting early if the order is not the one we're expecting.


harshilsharma63 (approved):

mgdelacroix:
Closed in favor or https://github.com/mattermost/focalboard/pull/3709, that has the same change and refactors the table migration mechanism

PR #3709, opened 2022-08-19
Refactor schema table migration mechanism

mgdelacroix:
#### Summary
The old schema table migration code was initializing the migration engine before changing the migrations table and was mixing queries on two different database connections (the `s.db` connection and the migrations `db` connection), both things potentially causing race conditions on the applications of the migrations.

The changes on this commit take care of altering the migrations table before the morph migration engine is initialized, so we're sure that it's using the new table cleanly, and they use the same connection for all operations, better isolating the schema table migration process.

The methods `isSchemaMigrationNeeded`, `isSchemaMigrationNeededSQLite`, `getLegacySchemaVersion`, `createTempSchemaTable`, `useNewSchemaTable` and `deleteOldSchemaMigrationTable` have only been moved to the new `schema_table_migration.go` file, so they have no changes for the purpose of the review.

### DIFF
--- a/server/services/store/sqlstore/migrate.go
+++ b/server/services/store/sqlstore/migrate.go
@@ -219,11 +219,11 @@ func (s *SQLStore) doMigrate(engine *morph.Morph, driver drivers.Driver) error {
 	}
 
 	if mErr := s.RunTeamLessBoardsMigration(); mErr != nil {
-		return mErr
+		return fmt.Errorf("error running teamless boards migration: %w", mErr)
 	}
 
 	if mErr := s.RunDeletedMembershipBoardsMigration(); mErr != nil {
-		return mErr
+		return fmt.Errorf("error running deleted membership boards migration: %w", mErr)
 	}
 
 	if mErr := s.ensureMigrationsAppliedUpToVersion(engine, driver, categoriesUUIDIDMigrationRequiredVersion); mErr != nil {
--- a/server/services/store/sqlstore/migrate.go
+++ b/server/services/store/sqlstore/migrate.go
@@ -202,10 +202,12 @@ func (s *SQLStore) Migrate() error {
 		engine.Close()
 	}()
 
-	return s.doMigrate(engine, driver)
+	return s.runMigrationSequence(engine, driver)
 }
 
-func (s *SQLStore) doMigrate(engine *morph.Morph, driver drivers.Driver) error {
+// runMigrationSequence executes all the migrations in order, both
+// plain SQL and data migrations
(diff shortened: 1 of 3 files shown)


server/services/store/sqlstore/migrate.go
```diff
@@ -74,6 +70,18 @@ func (s *SQLStore) getMigrationConnection() (*sql.DB, error) {
 }
 
 func (s *SQLStore) Migrate() error {
+	if err := s.EnsureSchemaMigrationFormat(); err != nil {
+		return err
+	}
+	defer func() {
```
wiggin77, line 76: Just want to confirm the order of defers is correct.  The code has defers executed in the following order:
1. Unlock mutex 
2. Morph engine close
3. Database close
4. deleteOldSchemaMigrationTable

Shouldn't the mutex be released last?

mgdelacroix: I don't think we need to wait for database and morph engine closing to release the mutex (2/5), but we do definitely need to run the `deleteOldSchemaMigrationTable` before releasing. Thanks a lot for catching this, I'll put table deletion as part of the migration process. The sequence will be:

1. Run migrations && `deleteOldSchemaMigrationTable`
2. Unlock cluster mutex
3. Morph engine close
4. Migration specific database connection close

Would you make any other change to this order?

mgdelacroix: Fixed on [0c2342b](https://github.com/mattermost/focalboard/pull/3709/commits/0c2342b496d9cb7a58cdb762e8566b3ee3274186)

server/services/store/sqlstore/migrate.go
```diff
@@ -181,26 +202,12 @@ func (s *SQLStore) Migrate() error {
...
-	}
-
-	if mErr := s.migrateSchemaVersionTable(src.Migrations()); mErr != nil {
-		return mErr
-	}
+	return s.runMigrationSequence(engine, driver)
+}
 
+// runMigrationSequence executes all the migrations in order, both
+// plain SQL and data migrations
```
wiggin77, line 209:
```suggestion
// plain SQL and data migrations.
```

wiggin77 (approved):
LGTM :+1:

server/services/store/sqlstore/migrate.go
mgdelacroix: 😅 thanks for pushing the fix!

harshilsharma63 (approved):
Awesome! LGTM 👍

sbishel:
@mgdelacroix - Can this be merged?

PR #3711, opened 2022-08-19
Print log server information on plugin run

mgdelacroix:
#### Summary
This change prints the same information that the standalone instance does on plugin activation

### DIFF
--- a/mattermost-plugin/server/plugin.go
+++ b/mattermost-plugin/server/plugin.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 
 	"github.com/mattermost/focalboard/mattermost-plugin/server/boards"
+	"github.com/mattermost/focalboard/server/model"
 
 	pluginapi "github.com/mattermost/mattermost-plugin-api"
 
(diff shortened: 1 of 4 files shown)


sbishel (approved):

mattermost-plugin/server/plugin.go
```diff
@@ -52,6 +53,8 @@ func (p *Plugin) OnActivate() error {
 		return fmt.Errorf("cannot activate plugin: %w", err)
 	}
 
+	model.LogServerInfo(logger)
```
wiggin77, line 56: @mgdelacroix can you add this to product mode as well? (https://github.com/mattermost/focalboard/blob/feb49eaf197afed8c082735837ebf95e04210621/mattermost-plugin/product/boards_product.go#L200)

mgdelacroix: Done on [b087f27](https://github.com/mattermost/focalboard/pull/3711/commits/b087f27ddf8277471c8405f296b5a42d8c732969), please let me know if it's the right place to add it

mgdelacroix: Oups, just saw that the comment had a line included, quickly fixing

mgdelacroix: Done on 4d5d16a, now log happens right after the `New` call

PR #3725, opened 2022-08-23
Adds a check that errors on store start if MariaDB is detected

mgdelacroix:
#### Summary
These changes causes the app to error on start if the database is detected to be MariaDB

### DIFF
--- a/server/services/store/sqlstore/sqlstore.go
+++ b/server/services/store/sqlstore/sqlstore.go
@@ -2,8 +2,10 @@ package sqlstore
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
 	"net/url"
+	"strings"
 
 	sq "github.com/Masterminds/squirrel"
 
@@ -14,6 +16,8 @@ import (
 	mmModel "github.com/mattermost/mattermost-server/v6/model"
 	"github.com/mattermost/mattermost-server/v6/shared/mlog"
 )
+
+var InvalidMariaDBErr = errors.New("MariaDB database is not supported")
 
 // SQLStore is a SQL database.
 type SQLStore struct {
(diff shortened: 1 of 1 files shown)


wiggin77:
@mgdelacroix this feels like we're telling customers MariaDB is officially dropped as an unofficially supported database.  In other words, we're saying if you're on MariaDB you can't use Mattermost going forward.  Is that really what we want?

sbishel:
We already say that here - https://docs.mattermost.com/install/software-hardware-requirements.html#database-software

> MariaDB v10+ no longer functions as a MySQL drop-in replacement, and it’s not supported for Mattermost due to the requirement of MySQL 5.7.12. Prior versions of MariaDB were not officially supported but may have functioned in older Mattermost releases. If you are running MariaDB now, migrating from MariaDB to the MySQL equivalent is recommended.

mgdelacroix:
We could tone down the error to say something less "radical" like `this version of Boards doesn't support MariaDB` if preferred, and even add a link to the code debt item, so users will know that we plan to make an effort to remove the non-compatible code.

The effect will still be that Boards is not going to work as of now, but they'll have a pointer to follow besides the error message.

wiggin77:
> and even add a link to the code debt item,

@mgdelacroix I guess adding the link sbishel posted (https://docs.mattermost.com/install/software-hardware-requirements.html#database-software) would give the user some info.

wiggin77 (approved):

mgdelacroix:
Error message updated on [59d6739](https://github.com/mattermost/focalboard/pull/3725/commits/59d6739a269b76c3cac0b3d635e67e064ead960d)

sbishel (approved):

Issue #3694, opened 2022-08-16
Bug: message banners stack


Pinjasaur:
## Steps to reproduce the behavior

1. Start up a fresh version of e.g. personal server
2. Notice version & cloud banner stack

## Expected behavior

Not 100% sure, perhaps different colors or not stacking? Just felt odd to have them as shown in the screenshot.

## Screenshots (optional)

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/6335792/184994807-ccd2d931-4764-4626-92b7-b202d70637a8.png">

## Edition and Platform

 - Edition: Personal Desktop / Personal Server / Mattermost Boards (plugin)
 - Version: main
 - Browser and OS: FF104 Mac

## Additional context (optional)

The cloud message is per-user (once dismissed, never comes back), and the version message is per-version (comes back next version) so this worst-case screenshot would be a brand new user.

Pinjasaur:
This might be more of an enhancement, not a bug.

cc @asaadmahmood for UI/UX and @sbishel since I _believe_ you implemented both banners.

asaadmahmood:
@wuwinson I don't think both banners should appear, the What's new banner should overlap the get your own free cloud, and when that is dismissed, we can show the get your own cloud server one (but even that not immediately, maybe on the next refresh or something?)

wuwinson:
Agreed, the "What's New" banner should take precedence over the "Free cloud server" banner. Not sure how frequent a user refreshes the desktop app, but what if we delay the the cloud server banner until 24 hrs after the "What's New" banner is dismissed?

wuwinson:
To clarify, this shouldn't be appearing on Personal Desktop, but the "What's New" banner should appear on Personal Server.

PR #3731, opened 2022-08-23
stack banners if multiple displayed

sbishel:
#### Summary
Stack second banner message, on top of one if one exists.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/3694

https://user-images.githubusercontent.com/12704875/186273159-33d4da17-5cb2-490a-9f47-0efc6693aa0f.mov

### DIFF
--- a/webapp/src/pages/boardPage/boardPage.scss
+++ b/webapp/src/pages/boardPage/boardPage.scss
@@ -5,7 +5,7 @@
     .VersionMessage:not(:first-child) {
         position: absolute;
         top: 0;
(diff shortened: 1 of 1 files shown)


Pinjasaur (approved):
LGTM @sbishel 🎉 works as expected when testing locally w/ personal server

my only thoughts is if there is an additional banner added in the future that `class` would need to be added to the selector. maybe having a more generic e.g. `BannerMessage` class or similar. won't block over it, just curious if you have any thoughts

PR #3741, opened 2022-08-25
Added property to select multiperson

Rajat-Dabade:
#### Summary
This PR will add the property to select multiple persons.


sbishel (approved):
When testing, the person list did not change when making the board public. It still only displayed the explicit team members. (I thought an "open" board should display all members of that team.) However, this was true for the original "person" property as well, so it may be a separate bug. I would just investigate and make sure before merging. If so, create another ticket for that issue.

jespino (approved):
LGTM

Issue #3293, opened 2022-06-30
Bug: Users are not removed from person property type after getting removed from the board


mgdelacroix:
## Steps to reproduce the behavior

1. Create a board
2. Add a user to the board
3. On a person property, check that the added user appears as an option 
4. Remove the user membership from the board
5. Check that on the person property, the user still appears as an option
6. Refresh and check again. The user will not appear this time

## Expected behavior

The options on the selector should be consistent with the current state of the board.

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.2
 - Browser and OS: Firefox 101.0.1 on Linux

wuwinson:
cc @asaadmahmood Thoughts on expected behaviour?

PR #3743, opened 2022-08-25
GH-3293 - Fix issue with deleted members not being updated via websocket

sbishel:
#### Summary
Separated logic to handle deleted members before handling adding members. Previously deleted items would return prematurely because they already existed in `boardUsers`.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/3293


mgdelacroix (approved):
LGTM! :+1:

PR #3748, opened 2022-08-25
MM-46274 Web App Product POC

hmhealey:
This is a proof of concept of the work required to ship Focalboard as a "product" using Webpack Module Federation. 

What works:
1. Boards is ran as part of the web app without being a plugin
2. In development, the web app checks for Focalboard's dev server on startup to know whether or not Focalboard should be loaded on startup. The web app runs both ways without any errors!
3. The Focalboard plugin is loaded and appears to run despite going through this new system.
4. As far as I can tell, it's loading shared libraries through module federation properly, so initial load should be slightly sped up

Short term TODOs (needed for 7.4 release):
1. ~~Images and other public files don't load. Their path appears correct (eg. `http://localhost:8065/plugins/boards/public/app-bar-icon.png`), but the MM server doesn't know where to serve the files from. (https://mattermost.atlassian.net/browse/MM-46646)~~ This has been fixed in this PR
2. Focalboard needs to actually be tested. I've just done some smoke testing of making sure the menu item for Boards appears, the Boards RHS appears, and that there's no errors in the console related to it on page load.
3. I haven't done the part to actually ship a product with a release of MM, but this covers the development case well enough. (https://mattermost.atlassian.net/browse/MM-46275)
4. There are some warnings from module federation to sort out around package versions. I'm not certain where they're coming from, but they seem harmless at the moment. (https://mattermost.atlassian.net/browse/MM-46647)

Longer term TODOs (after 7.4 release):
1. Having a new interface for a product rather than just instantiating the Focalboard plugin in a new way
2. Finding a way to standardize some of the plugin/product setup logic since parts of the Webpack config are going to be shared with the web app and other products.

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/20888
https://github.com/mattermost/mattermost-webapp/pull/11001

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46274


mattermost-plugin/webapp/webpack.config.js
```diff
@@ -120,30 +122,80 @@ module.exports = {
             },
         ],
     },
-    externals: {
```
hmhealey, line 123: I need to re-add this below when we're not building this as a product

Makefile
```diff
@@ -186,6 +186,9 @@ watch-plugin: modd-precheck ## Run and upload the plugin to a development server
 live-watch-plugin: modd-precheck ## Run and update locally the plugin in the development server
 	cd mattermost-plugin; make live-watch
 
+watch-product: ## Run the product as something the Mattermost web app will watch for
```
hmhealey, line 189: I tried to follow the naming scheme of the existing Makefile targets. Let me know if this works or if there's something better to use

mattermost-plugin/webapp/package.json
```diff
@@ -78,18 +79,19 @@
...
     "core-js": "3.21.1",
     "glob-parent": "6.0.2",
     "marked": ">=4.0.12",
     "mattermost-redux": "5.33.1",
-    "react": "^17.0.2",
-    "react-dom": "17.0.2",
-    "react-intl": "^5.24.7",
-    "react-redux": "^7.2.8",
-    "react-router-dom": "5.2.0",
+    "react": "^16.13.0",
```
hmhealey, line 90: As mentioned in a commit message, these needed to be downgraded temporarily to match the web app. Once Nevy's React 17 PR goes in, we can revert these. They should all be matched with fuzzy versions though so that we don't need to match the React versions exactly

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -14,9 +14,8 @@ import {selectTeam} from 'mattermost-redux/actions/teams'
 import {SuiteWindow} from '../../../webapp/src/types/index'
 import {UserSettings} from '../../../webapp/src/userSettings'
 
-
 const windowAny = (window as SuiteWindow)
-windowAny.baseURL = '/plugins/focalboard'
+windowAny.baseURL = process.env.TARGET_IS_PRODUCT ? '/plugins/boards' : '/plugins/focalboard'
```
hmhealey, line 18: I added a new environment variable for whenever we need the plugin to work differently than the product. For now, it's just used here since Boards is still mostly a plugin

mattermost-plugin/webapp/src/index.tsx
```diff
@@ -408,11 +407,3 @@ export default class Plugin {
...
     }
 }
-
-declare global {
-    interface Window {
-        registerPlugin(id: string, plugin: Plugin): void
-    }
-}
-
-window.registerPlugin(manifest.id, new Plugin())
```
hmhealey, line 418: This isn't needed as a product because the web app doesn't use that part of the plugin architecture to load Boards as a product

hmhealey:
@sbishel @jespino This should be ready for review now. There's some outstanding issues as noted in the description, but those should only occur while Focalboard is running as a product, so it shouldn't affect anyone else even if this is merged.

I think I've figured out a fix for https://mattermost.atlassian.net/browse/MM-46646 as well which I'll be adding to this branch

wiggin77:
@hmhealey did you see the CI error here:
```
[webpack-cli] Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
 - configuration has an unknown property 'outputs'.
```

wiggin77 (approved):
LGTM :+1:

hmhealey:
Right. I had a fix for this and forgot to push it while I was looking at the rest of these changes. There's also another fix for the plugin version included as well

hmhealey:
I've added a fix for https://mattermost.atlassian.net/browse/MM-46646 as well now. Some of the handling of the public paths for images seems a bit wonky to me since I wouldn't expect Focalboard itself to have to modify the URL for image files, but I guess that's a side effect of the many ways it can be built.

I've done some brief testing on both the product and plugin versions of Boards as well, and at the very least, there are no obvious errors or broken images

hmhealey:
The last commit should address https://mattermost.atlassian.net/browse/MM-46647 as well. It makes Boards as a product always match the web app for the version of shared packages

Previously, the higher `version` value set for React would cause the whole web app to use the newer version of React which caused those errors. Setting it to `false` means that Webpack just defers to other containers for which version to use.

Similarly, I also changed the `requiredVersion` to `false` as well. If I didn't do that, Boards and the rest of the web app would end up with different versions of `React` which caused a bunch of things to break.

mattermost-plugin/webapp/src/index.tsx
jespino: But what happens when we compile it in plugin mode?

jespino: nevermind, I've seen the other file.

jespino (approved):
LGTM

sbishel (approved):

Issue #3704, opened 2022-08-18
Bug: Card Template Not Carrying All Checkboxes Into New Card on Boards


jtb-jtb:
## Steps to reproduce the behavior

1. Create Card Template
2. Add multiple checkboxes at bottom of card
3. Exit out of card template editor
4. Create new card with template
5. See only first checkbox is present on new card
## Expected behavior

The new card should have all checkboxes from the template included when a new card is created

## Screenshots (optional)

![image](https://user-images.githubusercontent.com/111531635/185468720-b73854b4-49b1-4db0-866f-e5345ee7ebe9.png)
![image](https://user-images.githubusercontent.com/111531635/185468834-ccd742a2-bae5-4c03-96ff-f44481e6ca81.png)

## Edition and Platform

 - Edition: Personal Desktop / Cloud Server / Mattermost Boards (plugin)
 - Version: v7.2
 - Browser and OS: Chrome on Windows]

## Additional context (optional)

* Sev 2: Affects major functionality with a difficult or non-obvious workaround

wuwinson:
@jtb-jtb We can't repro this on [Community](https://community.mattermost.com/boards/) with the latest v7.2 release. Can you try to upgrade and see if this is still happening for you? Thanks!

jtb-jtb:
We're on cloud and I verified the plugin is v7.2. I tried it on Community as well, and it gave me the same results :( . I created a card template, added a checkbox at the bottom (not on top with the properties but below the comments), then added a couple more. Exited the template manager, created a card from that template, and only the first checkbox is there. If I open the template the checkboxes are there, but it doesn't create them when creating a new card.

PR #3757, opened 2022-08-30
GH-3704 - Fix for not copying all content items.

sbishel:
#### Summary
Fix issue returning from function early without full processing changes.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/3704

### DIFF
--- a/webapp/src/store/contents.ts
+++ b/webapp/src/store/contents.ts
@@ -30,7 +30,6 @@ const contentsSlice = createSlice({
                     for (let i = 0; i < state.contentsByCard[content.parentId].length; i++) {
                         if (state.contentsByCard[content.parentId][i].id === content.id) {
                             state.contentsByCard[content.parentId][i] = content
-                            return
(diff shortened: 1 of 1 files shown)


webapp/src/store/contents.ts
```diff
@@ -30,7 +30,6 @@ const contentsSlice = createSlice({
                     for (let i = 0; i < state.contentsByCard[content.parentId].length; i++) {
                         if (state.contentsByCard[content.parentId][i].id === content.id) {
                             state.contentsByCard[content.parentId][i] = content
-                            return
```
jespino, line 33: Probably a slightly more efficient approach is to replace it with a break. And also, I think we should avoid to push to the parent contentsByCard if it is already updated here.

jespino (changes requested):
Changed slightly the expected behavior (I think)

jespino (approved):
LGTM

Issue #3746, opened 2022-08-25
Bug: `@mention` not filtering in private boards


Pinjasaur:
## Steps to reproduce the behavior

1. Go to a private board and start typing e.g. comment to mention someone
2. Notice it doesn't filter as you type

## Expected behavior

Users to filter as typing.

## Screenshots (optional)

![image](https://user-images.githubusercontent.com/6335792/186743889-c8f06d55-0966-46b0-929a-dc51c480880b.png)

## Edition and Platform

 - Edition: Personal Desktop / Personal Server / Mattermost Boards (plugin)
 - Version: [e.g. v0.15.0]
 - Browser and OS: [e.g. Chrome on Mac, Edge on Windows]

## Additional context (optional)

Might be related to `synthetic` members through associated channels.

Pinjasaur:
@sbishel lmk your findings, #2604 has some related work I need to do for Guest users, so there may be overlap

wuwinson:
Might potentially be resolved with the [@mention autocomplete](https://docs.google.com/document/d/1IiIgtXI_lYtTSpm6XfqPvSe_jD0E0rajqerWSRLC6OQ/edit?usp=sharing) feature @jespino worked on. Can verify after it's been merged.

wuwinson:
Waiting on #3724 before testing.

wuwinson:
Confirmed this is fixed on v7.3 Bug Bash server. Users are filtered as you type on both private and public boards, and we also indicate which users are not explicit board members. Closing out.

PR #3761, opened 2022-08-30
Fix @-mentions in boards & for guests

### LINKED ISSUE #2604: Bug: Guest User can @mention users from other teams

1. Create a User in TeamA.
2. Create a guest user and add to TeamB
3. Login as Guest User
4. Go to boards
5. Select a card. Attempt to @mention User in TeamA

## Expected behavior
Guest users should not be able to @mention users in other teams.

Pinjasaur:
#### Summary

Fixing bugs related to `@mention`s and Guest accounts:
- filter the in-memory board users list for private boards (where the Redux state is used, no API calls)
- fix in an invalid DB query for searching users as a guest
- handle `includePublicBoards` when getting boards for a user & team
- don't use the API search for users when `@mention`ing as a guest

#### Ticket Link

closes #2604 

cc @wuwinson @jespino this should handle #3746 as-is too (intentionally not using a closing keyword there 🙂)

### DIFF
--- a/server/services/store/mattermostauthlayer/mattermostauthlayer.go
+++ b/server/services/store/mattermostauthlayer/mattermostauthlayer.go
@@ -380,8 +380,8 @@ func (s *MattermostAuthLayer) SearchUsersByTeam(teamID string, searchQuery strin
 			boardsIDs = append(boardsIDs, board.ID)
 		}
 		query = query.
-			Join(s.tablePrefix + "board_members as bm ON bm.UserID = u.ID").
-			Where(sq.Eq{"bm.BoardId": boardsIDs})
+			Join(s.tablePrefix + "board_members as bm ON bm.user_id = u.ID").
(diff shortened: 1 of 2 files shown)


server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -860,13 +860,21 @@ func (s *MattermostAuthLayer) GetBoardsForUserAndTeam(userID, teamID string, inc
...
+	if includePublicBoards {
+		boards, err := s.SearchBoardsForUserInTeam(teamID, "", userID)
+		if err != nil {
+			return nil, err
+		}
+		for _, b := range boards {
+			boardIDs = append(boardIDs, b.ID)
+		}
+	}
+
```
Pinjasaur, line 877: @jespino any problems you see with this approach?

sbishel (approved):

PR #3762, opened 2022-08-30
fix: filter out bots for sharing, @mention-ing, and Person property

### LINKED ISSUE #3656: Bug: Bots show up in Share Board Search

1. Go to Share on a board you are an admin
2. Type 'B'
3. Notice the user "Boards", which is a bot is available

## Expected behavior

Bots should not be displayed as users for board access.

## Screenshots (optional)
Screen Shot 2022-08-11 at 12 17 30 PM

## Edition and Platform

 - Version: v7.2.0-rc3
 - Browser and OS: Chrome on Mac,

Pinjasaur:
#### Summary

Filters out bots when sharing a board, `@mention`-ing a user, and the Person property by adding an `?exclude_bots=true` query param to the `/teams/{teamID}/users` API endpoint.

#### Ticket Link

fixes #3656


sbishel (changes requested):
One issue I see with this approach is we limit the returned items to 10. If one is a bot, then we will only display 9 users, which is not a huge deal. However, if you were to search on 'bot' and there were a lot of "bot" accounts. Then we may display only display say 2 and a legitimate name containing `bot` would not be returned. Therefore, we probably need to move this server side.

Pinjasaur:
> One issue I see with this approach is we limit the returned items to 10. If one is a bot, then we will only display 9 users, which is not a huge deal. However, if you were to search on 'bot' and there were a lot of "bot" accounts. Then we may display only display say 2 and a legitimate name containing `bot` would not be returned. Therefore, we probably need to move this server side.

Ahh, good point @sbishel. Is there a preference towards how it's implemented API-wise e.g. `?no_bots=true` query param or something else? cc @wiggin77 for API opinions 🙂

wiggin77:
> `?no_bots=true` query param or something else? cc @wiggin77 for API opinions

Query string is fine for now, but I suspect we'll move these into a QueryUserOptions struct that will be passed in the body.  This better matches how the Channels equivalent APIs work, and allows us to easily add more search criteria.

Currently the getTeamsUsers on the backend does not support a filter to exclude bot.  That will need to be added, but the underlying store API `getUsersByCondition` can have any condition added. `Users` has an `isBot` flag to filter on.

IMO the searchUsersByTeam and getUsersByTeam should be passed a QueryUsersOptions struct that contains TeamID, SearchQuery, ExcludeBots, and we can add to it.  This can be done later.   Thanks @Pinjasaur

Makefile
```diff
@@ -100,7 +100,7 @@ server-linux-package-docker:
 	rm -rf package
 
 generate: ## Install and run code generators.
-	cd server; go get github.com/golang/mock/mockgen
+	cd server; go install github.com/golang/mock/mockgen@v1.6.0
```
Pinjasaur, line 103: the install instructions changed for recent golang versions https://github.com/golang/mock#go-116

server/services/store/sqlstore/user.go
```diff
@@ -218,7 +218,7 @@ func (s *SQLStore) getUsersByTeam(db sq.BaseRunner, _ string, _ string) ([]*mode
 	return s.getUsersByCondition(db, nil, 0)
 }
 
-func (s *SQLStore) searchUsersByTeam(db sq.BaseRunner, _ string, searchQuery string, _ string) ([]*model.User, error) {
+func (s *SQLStore) searchUsersByTeam(db sq.BaseRunner, _ string, searchQuery string, _ string, _ bool) ([]*model.User, error) {
```
Pinjasaur, line 221: there are no bot users in non-plugin mode so the only real business logic change is in mattermostauthlayer.go

server/api/teams.go
```diff
@@ -217,6 +222,7 @@ func (a *API) handleGetTeamUsers(w http.ResponseWriter, r *http.Request) {
 	userID := getUserID(r)
 	query := r.URL.Query()
 	searchQuery := query.Get("search")
+	excludeBots := r.URL.Query().Get("exclude_bots") == True
```
wiggin77, line 225: Looking good.  It's unfortunate that the previous behaviour was to include bots, since I think most usage of this API will exclude bots and it would be nice to have that as the default (i.e. no param provided).

Pinjasaur: @wiggin77 could we make that change in the v3 you're working on?

wiggin77 (approved):
LGTM :+1:

sbishel (approved):

Issue #3695, opened 2022-08-16
Bug: unable to edit/delete a template in personal server/desktop


Pinjasaur:
## Steps to reproduce the behavior

1. Go to the template selector in personal server or desktop
2. Create a new custom template
3. Go back to the template selector and notice you cannot edit/delete your custom template

## Expected behavior

Be able to edit/delete a custom template.

## Screenshots (optional)

Personal Server:

<img width="1223" alt="image" src="https://user-images.githubusercontent.com/6335792/184999250-bbefc330-a691-43de-95f6-5aa278b67f36.png">

Personal Desktop:

<img width="1182" alt="image" src="https://user-images.githubusercontent.com/6335792/184999925-3c8935c9-9f6c-4a8d-a3be-3639f4dfeead.png">

## Edition and Platform

 - Edition: Personal Desktop / Personal Server
 - Version: main
 - Browser and OS: FF104 Mac

## Additional context (optional)

Looking at the code, we're doing a conditional to see if the template belongs to the `globalTeamId` of `0` which we use for the builtin templates, but it's also the team used on personal server & desktop.

https://github.com/mattermost/focalboard/blob/393b961a6b6bce0d8e60f74f7beae14b407dfdbb/webapp/src/components/boardTemplateSelector/boardTemplateSelectorItem.tsx#L46-L47

Pinjasaur:
@wuwinson can you confirm that this in a bug? if it's not and I'm simply mistaken, please close. thanks!

Pinjasaur:
> @Pinjasaur Definitely a bug and very similar to what we saw on plugin a few releases ago on #3061

Aha, that's why it felt so familiar...

wuwinson:
cc @sbishel

PR #3766, opened 2022-08-31
check createdBy instead of team id to determine global templates

sbishel:
#### Summary
Personal desktop/server does not support teams. Therefore the only teamId in these products is "GlobalTeamID" = 0. When we determine if a template can be edited, we check the teamId field which is fine for plugin because only the global templates will use GlobalTeamID. However, because the only team on Personal Desktop/Server is 0. We need another way to check for GlobalTemplates, moved to using the `CreatedBy` field and added a constant for 'system' which is listed as the creator of global templates.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/3695


Pinjasaur (approved):
LGTM, didn't realize we didn't have that constant used on the frontend 'til this point 👍🏻 

I'm going to defer merging immediately because I'm not able to personally confirm functionality locally as I'm getting migration errors. Potentially related to #3774 cc @harshilsharma63 I'll post my logs in that issue

Pinjasaur:
still broken for me locally after #3774 merge https://community.mattermost.com/boards/team/qghzt68dq7bopgqamcnziq69ao/2b3ba61b-2434-4892-8fe0-f22fe5af5e68/a43d6c7d-bf99-4ee2-8697-32071ab11207/csxa4s4yjmbdwpecbyj4omp4hgr

sbishel:
That issue appears unrelated to this PR

PR #3774, opened 2022-09-02
Extra safe MySQL migration

harshilsharma63:
There was an error on bug bash server `Invalid JSON path expression` when running props to preferences migration. A lot of data was successfully migrated, however.

The exact data item that caused the error is unknown and since then the cloud instance is unresponsive.

I've made some changes in migration to make it extra safe for MySQL.


mgdelacroix (approved):

mgdelacroix (approved):
LGTM! :thumbsup:

sbishel (approved):

Issue #3785, opened 2022-09-06
Feature Idea: Update confirmation message to account for Guest Accounts when linking boards to a channel


wuwinson:
## Summary

When linking a board to a channel, Guest accounts from the channel are not automatically granted access to the board. The board admin should be informed of this so they know who will be getting access to their board.

The copy on the 4 confirmation messages below need to be updated to:

>When you link a channel to a board, all members of the channel (existing and new) will be able to edit it. This excludes members who are guests

See [thread](https://community.mattermost.com/private-core/pl/dpw8uj3dbpdsijdobueacp7kyc) for reference.

## How important this is to me and why

Importance: Medium

Use cases:
#### Boards share dialog
1. Linking a channel to the board
![Screen Shot 2022-09-06 at 5 03 44 PM](https://user-images.githubusercontent.com/93531870/188738282-df41b216-f5d0-4f2d-8c8a-9b12c100f1e5.png)

2. Linking another channel to the same board
![Screen Shot 2022-09-06 at 5 03 25 PM](https://user-images.githubusercontent.com/93531870/188738673-a65c15f9-e3c1-4d6d-826c-af75d4cd5407.png)

#### Channel RHS Link Boards dialog
1. Linking a board to the channel
![Screen Shot 2022-09-06 at 5 04 31 PM](https://user-images.githubusercontent.com/93531870/188738938-a73d8413-f5d7-4554-ba3e-3da9be20ddcf.png)

2. Linking a board that is already linked to another channel
![Screen Shot 2022-09-06 at 5 04 50 PM](https://user-images.githubusercontent.com/93531870/188738961-443e57ca-ddf2-4563-9660-928165897743.png)

wuwinson:
cc @justinegeffen @asaadmahmood @Pinjasaur for input.

Pinjasaur:
If we're able to keep it identical for all 4 modals that'll simplify the i18n/translation aspect of it.

Perhaps:

> all members of the channel (existing and new) _**except guests**_

0/5 otherwise

wuwinson:
@justinegeffen Any updates on the copy for this? Thanks!

wuwinson:
Thanks @justinegeffen! @sbishel @Pinjasaur The text has been finalized.

Pinjasaur:
I'll plan on getting it in for #3786.

PR #3786, opened 2022-09-06
Remove synthetic/implicit memberships for guests

### LINKED ISSUE #3771: Bug: Guests can access boards linked to a channel

1. Invite a guest to a channel
2. Link boards to the same channel
3. Login as the guest
4. Go to the channel and click on the boards linked on the channel
5. See error - guest is able to access those boards even though they're not an explicit member

## Expected behavior

Guests should not have access any board unless they were added as an explicit member. See [this thread](https://community.mattermost.com/private-core/pl/dpw8uj3dbpdsijdobueacp7kyc) for reference.

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/188013397-d4bc393e-b5ca-40e7-9d2c-bae016aa9b68.mov

## Edition and Platform

 - Version: v7.3

Pinjasaur:
#### Summary

See #3771 for details, but the concept of having a synthetic membership for a _guest_ is removed. They will need to be explicitly added to a board via the Share modal.

See #3785 for tracking of the modal dialog copy to reflect this change. cc @wuwinson 

#### Ticket Link

closes #3771


server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -739,6 +739,14 @@ func (s *MattermostAuthLayer) implicitBoardMembershipsFromRows(rows *sql.Rows) (
 }
 
 func (s *MattermostAuthLayer) GetMemberForBoard(boardID, userID string) (*model.BoardMember, error) {
+	// No synthetic memberships for guests
+	user, err := s.GetUserByID(userID)
+	if err != nil {
+		return nil, err
+	}
+	if user.IsGuest {
+		return nil, nil
```
wiggin77, line 748: I would like to see us move away from methods that return `nil,nil`.  It is very common for developers to expect if error=nil then the other value is valid.  

Some alternatives would be to return an error if IsGuest, or return three values (*model.BoardMember, bool, error) where the bool signifies if member was found.

Pinjasaur: Thanks @wiggin77, I was suspecting this might be an anti-pattern. I think I'd personally prefer keeping with the tuple return unless there's some side effect I'm not considering. Any preference on a specific error type or just do an `errors.New`?

jespino: This is not correct, a Guest can be member of a board, in this case you are saying that doesn't matter what board are you querying that if the member is a guest, it has no membership.

It should check for the guest only in the case of the explicit membership is not found.

jespino: About what to return, I would expect a "NotFound" error here.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -868,7 +886,12 @@ func (s *MattermostAuthLayer) GetMembersForBoard(boardID string) ([]*model.Board
 		existingMembers[m.UserID] = true
 	}
 	for _, m := range implicitMembers {
-		if !existingMembers[m.UserID] {
+		// No synthetic memberships for guests
+		user, err := s.GetUserByID(m.UserID)
```
jespino, line 890: This is going to be really awful in performance, this is generating one DB roadtrip for each implicit memberships here.

jespino: My suggestion here is to use the original query to extract the data, another extra join to get the User information, now know if the user there is a guest or not.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
Pinjasaur: Thanks @jespino, good catch.

Pinjasaur:
Would like someone's input on the addition to the query to make sure that looks good. Additionally the `/boards/search` API will need to be updated to only include boards that guests are explicit members of. See #3771.

wiggin77 (approved):
LGTM :+1:

Pinjasaur:
> @Pinjasaur - Can this be merged?

@sbishel nearly, just need to update the `/boards/search` endpoint as mentioned above. That was my intention today along with a PR for #3785 to update the modal copy based on these changes.

jespino (approved):
LGTM

Pinjasaur:
@sbishel should be good to go, if you could give the changes to `/boards/search` endpoint a once-over: [`fe0121b` (#3786)](https://github.com/mattermost/focalboard/pull/3786/commits/fe0121b33ea032f9564743e0bc85eb4f68bbd653)

sbishel (approved):

PR #3795, opened 2022-09-07
test: add snapshot tests to CommentsList

Pinjasaur:
#### Summary

I remember briefly looking at this for #2882 but I recall seeing the behavior @Rajat-Dabade was seeing e.g. non-deterministic `data-offset-key`. Doing some searching, looks like the solution (that I see elsewhere in the codebase) is to mock draftjs: https://github.com/facebook/draft-js/issues/702#issuecomment-283169324

Anyways, I'm not sure what changed, but it looks like snapshots now work fine for `<CommentsList>`.

#### Ticket Link

n/a


Pinjasaur:
Hmm. Maybe I'm miss-remembering and it was the relative date/times that I ended up mocking to be deterministic for the tests.

harshilsharma63 (approved):
Neat!

PR #3822, opened 2022-09-09
Update version banner URL to redirect permalink

Pinjasaur:
#### Summary

@justinegeffen reached out to get the URL updated in the "what's new" version banner to point to a mattermost.com permalink redirect. Because of version/release reasons, I'm setting the cherrypick to 7.2 so it rolls out with the upcoming 7.2.1 (any concerns @sbishel?) and we'll need to manually run a `/cherry-pick release-7.3` so it makes it to 7.3 as well.

I'm not sure if this is how all mattermost.com/pl/ URLs work, but I'm concerned that it's a HTTP 301 (permanent) redirect when IMHO I think it should be a HTTP 302 (temporary) redirect. I'm 5/5 on this because web browser clients will typically make the 301 request _once_ and then any subsequent requests will just be redirected at the client level i.e. cached so if the redirect location changes e.g. new docs URL then the expected redirect won't happen. cc @emdecr maybe you know?

#### Ticket Link

n/a, thread conversation: https://community.mattermost.com/core/pl/nqer1j7cypyyfqf85n3jbd3cty

### DIFF
--- a/webapp/src/components/messages/versionMessage.tsx
+++ b/webapp/src/components/messages/versionMessage.tsx
@@ -18,7 +18,7 @@ import CompassIcon from '../../widgets/icons/compassIcon'
 import TelemetryClient, {TelemetryCategory, TelemetryActions} from '../../telemetry/telemetryClient'
 
 import './versionMessage.scss'
-const helpURL = 'https://docs.mattermost.com/welcome/whats-new-in-v72.html'
+const helpURL = 'https://mattermost.com/pl/whats-new-boards/'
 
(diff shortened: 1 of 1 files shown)


emdecr:
@Pinjasaur Great flag + thanks! We can change to a 302. Didn't gather how frequent we changed some of these links, if ever. Also a good callout for a few legacy `mm.com/pl/inset-slug-here` redirects we have running as well.

sbishel (approved):
No need to CP to 7.2, as 7.2.1 has already been cut and released.

PR #3825, opened 2022-09-10
Adding the new blocks based editor

jespino:
#### Summary
This adds the new blocks based editor behind a feautre flag.

There are still a lot of improvements to do, but for now it supports, drag and drop, different block types (h1, h2, h3, video, attachment, images, dividers, list items, checkboxes and quotes)

To try it you have to restart your mattermost-server with the feature flag that enables it, like this: `env MM_FEATUREFLAGS_BOARDSFEATUREFLAGS=newBoardsEditor make restart-server`

Some pending improvements are:
  - nicer look and feel
  - better handling of arrow keys on multi line markdowns
  - more and better shortcuts
  - add a lot of testing
  - Replace with the new rich text editor (expected to be release after the mattercon)

> **__IMPORTANT NOTE__**: This is a prototype behind a feature flag to keep iterating over it, don't expect the code to be perfect or something like that. This is going to be a bit unstable for now, but that is ok because is behind the feature flag.

#### Ticket Link
Not yet


jespino:
@wiggin77 Here is the new API to move things. Probably we would need to have a better "content" api, something like "Add content" with an "after" or "before" parameter would insert the block and properly possition it.

harshilsharma63:
@jespino whats the plan for this? Its this only an experiment or do we want to switch to the new editor? What are the requirements? What are the expected behaviors? I'm asking because this is a huge PR requiring substantial effort in reviewing and testing.

Also the CI is failing. Can we at lest get that working before review?

jespino:
The intention is to end up merging it as part of the product, for now behind a feature flag. And yes, there are things to fix, like the ci, adding more tests or fixing ToDos, it is not strictly needed a detailed review for all the people in the reviewers list, i added that people to get general opinions, feel free to no review it for now. I'll fix some details today anyway.

jespino:
@harshilsharma63 CI is passing now, still some TODOs to fix, but if you want to can take a look, as I said, most of it is not going to hit production for now, it is just behind a feature flag, so, I would focus on the part that affects the rest of the code (outside the blocksEditor directory).

server/api/blocks.go
```diff
@@ -782,3 +784,182 @@ func (a *API) handleDuplicateBlock(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		a.errorResponse(w, r, err)
+		return
+	}
+
+	contentOrder := card.Fields["contentOrder"].([]interface{})
+	newContentOrder := []interface{}{}
+	foundDst := false
+	foundSrc := false
+	for _, id := range contentOrder {
```
wiggin77, line 900: I think this logic belongs in the app layer.  It can be deferred to a follow-up PR with a ticket.

jespino: Yes, that's true, I think I would love to have an API about Content Blocks, instead of Blocks, and yes, that should have its counterpart in the App layer. Anyway, as I said, this is behind a feature flag and is for iterating over it whenever is finished, I don't want this to be the definitive API.

Anyway, I going to try to move it to the app layer now in a new `content_blocks.go` in the app, and one in the `api`, and add some other APIs.

jespino: Done, PTAL to that part, I changed it to make it more open to grow later.

server/api/content_blocks.go
```diff
@@ -0,0 +1,105 @@
...
+		return
+	}
+
+	if !a.permissions.HasPermissionToBoard(userID, block.BoardID, model.PermissionManageBoardCards) {
+		a.errorResponse(w, r, model.NewErrPermission("access denied to modify board cards"))
+		return
+	}
+
+	auditRec := a.makeAuditRecord(r, "moveBlockAfter", audit.Fail)
+	defer a.audit.LogRecord(audit.LevelRead, auditRec)
```
wiggin77, line 91:
```suggestion
	defer a.audit.LogRecord(audit.LevelModify, auditRec)
```

server/app/content_blocks.go
```diff
@@ -0,0 +1,76 @@
...
+		if block.ID == stringID {
+			foundSrc = true
+			continue
+		}
+
+		newContentOrder = append(newContentOrder, id)
+	}
+
+	if !foundSrc {
+		message := fmt.Sprintf("source block not found")
```
wiggin77, line 53:
```suggestion
		message := fmt.Sprintf("source block %s not found", block.ID)
```

server/app/content_blocks.go
```diff
@@ -0,0 +1,76 @@
...
+		return model.NewErrBadRequest(message)
+	}
+
+	if !foundDst {
+		message := fmt.Sprintf("destination block not found")
```
wiggin77, line 58:
```suggestion
		message := fmt.Sprintf("destination block %s not found", dstBlock.ID)
```

server/api/blocks.go
wiggin77: Looks great.  Thanks @jespino.  I added some suggestions to fix linter.  Good for merge after that.

jespino:
@harshilsharma63 @wiggin77 do you mind to give it a last pass to merge it into main branch? Remember, all this is behind a feature flag, so we shouldn't be too worried about this being perfect.

harshilsharma63 (approved):
Looks fine! I tested without setting the feature flag and there was no issue I could see with the app. If the feature flag is disabled, then the new API isn't called, so that's safe as well.

webapp/src/components/blocksEditor/blockContent.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        editing={null}
+                        setEditing={setEditing}
+                        setAfterBlock={jest.fn()}
+                        onSave={jest.fn()}
+                        onMove={jest.fn()}
+                    />
+                </ReduxProvider>,
+            ))
+        })
+        let item = screen.getByTestId('block-content')
```
wiggin77, line 107:
```suggestion
        const item = screen.getByTestId('block-content')
```

webapp/src/components/blocksEditor/blockContent.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        editing={null}
+                        setEditing={jest.fn()}
+                        setAfterBlock={setAfterBlock}
+                        onSave={jest.fn()}
+                        onMove={jest.fn()}
+                    />
+                </ReduxProvider>,
+            ))
+        })
+        let item = screen.getByTestId('add-action')
```
wiggin77, line 131:
```suggestion
        const item = screen.getByTestId('add-action')
```

webapp/src/components/blocksEditor/blockContent.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        editing={block}
+                        setEditing={jest.fn()}
+                        setAfterBlock={jest.fn()}
+                        onSave={onSave}
+                        onMove={jest.fn()}
+                    />
+                </ReduxProvider>,
+            ))
+        })
+        let input = screen.getByDisplayValue('Title')
```
wiggin77, line 155:
```suggestion
        const input = screen.getByDisplayValue('Title')
```

wiggin77:
webapp/src/components/blocksEditor/blocks/list-item/list-item.scss is an empty file that was added.  Can this be deleted?

wiggin77 (approved):
Wow, a lot of work went into this. Great job.  A new minor notes:
- the feature flag only applies to the editor, not the new Rest APIs; they will be in production soon after merge
- there is a TODO in the root editor (not sure if todo now, or todo later)
- the video type doesn't have any checks for supported formats, meaning not all browsers support the same formats, however there are some formats that work in all our support browsers;  perhaps video should have its own feature flag?

jespino:
> `webapp/src/components/blocksEditor/blocks/list-item/list-item.scss` is an empty file that was added. Can this be deleted? Same with `webapp/src/components/blocksEditor/blocks/text-dev/text.scss`

@wiggin77 this is just a placeholder just in case we add it later. But I don't mind removing it.

jespino:
@wiggin77 Probably video is going to be refactored before we remove the feature flag, I don't see a problem there.

PR #3832, opened 2022-09-12
Update board store to return inserted / patched values without querying the DB after modifying them

mgdelacroix:
#### Summary
What says in the tin


jespino (approved):
LGTM

PR #3839, opened 2022-09-13
Fix broken image in plugin

hmhealey:
I moved this image with the MPA changes, and Webpack is no longer able to find it on its own when running as a plugin

#### Ticket Link
None


Pinjasaur (approved):
Thanks @hmhealey, glad it was a relatively simple fix!

PR #3840, opened 2022-09-13
Fix broken main store.getBoardsInTeamByIds

wiggin77:
#### Summary
This PR is a temporary fix for main.  A recent PR made `store.getBoardsInTeamByIds` more strict in that the number of ids passed in must match the number of boards found.  This API has been returning a mismatched list for a while now but the  error was hidden.

A proper fix may be to simply allow templates to be returned in `store.getBoardsInTeamByIds`.  This PR simply reverts to yesterday's behaviour.  A ticket will be created to revisit this.

#### Ticket Link
NONE


mgdelacroix (approved):
LGTM! Thanks for the quick fix! :rocket:

PR #3841, opened 2022-09-13
fix: display non-explicit members in Share Board modal <Select>

### LINKED ISSUE #3808: Bug: Cannot create explicit membership for implicit member

1. Create a board with a channel association. 
2. In share dialog, attempt to find a member of the channel 
3. Channel members are not displayed, so you cannot choose one.

## Expected behavior

Channel members should be available to create explicit membership.

## Screenshots (optional)
Screen Shot 2022-09-08 at 12 26 10 PM
Screen Shot 2022-09-08 at 12 28 16 PM

## Edition and Platform

 - Version: 7.3.0-rc1
 - Browser and OS: [Chrome on Mac]

Pinjasaur:
#### Summary

Previously, the `<Select>` in the Share Board modal was only filtering on if the user was a member of the board, regardless of explicit or implicit type. This fixes the logic to include implicit members so they could be made explicit if desired.

#### Ticket Link

closes #3808

### DIFF
--- a/webapp/src/components/shareBoard/shareBoard.tsx
+++ b/webapp/src/components/shareBoard/shareBoard.tsx
@@ -360,7 +360,13 @@ export default function ShareBoardDialog(props: Props): JSX.Element {
                             value={selectedUser}
                             className={'userSearchInput'}
                             cacheOptions={true}
-                            filterOption={(o) => !members[o.value]}
+                            filterOption={(o) => {
+                                // render non-explicit members
+                                if (members[o.value])
+                                    return members[o.value].synthetic
+                                // not a member, definitely render
+                                return true
+                            }}
                             loadOptions={async (inputValue: string) => {
                                 const result = []
                                 if (Utils.isFocalboardPlugin()) {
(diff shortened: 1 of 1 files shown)


webapp/src/components/shareBoard/shareBoard.tsx
```diff
@@ -360,7 +360,13 @@ export default function ShareBoardDialog(props: Props): JSX.Element {
                             value={selectedUser}
                             className={'userSearchInput'}
                             cacheOptions={true}
-                            filterOption={(o) => !members[o.value]}
+                            filterOption={(o) => {
+                                // render non-explicit members
+                                if (members[o.value])
+                                    return members[o.value].synthetic
```
sbishel, line 366: This style will cause lint errors, once that PR gets merged. 
```suggestion
                                if (members[o.value]) {
                                    return members[o.value].synthetic
                                }
```

sbishel (approved):
Approving with recommended change which will cause linter errors.

PR #3842, opened 2022-09-14
Check on GetBoardsForUserAndTeam if the board result list is incomplete and continue if that's the case

mgdelacroix:
#### Summary
Follow up to https://github.com/mattermost/focalboard/pull/3840 that moves the responsibility of ignoring the `NotAllFound` error to the caller, so in this case, the `GetBoardsForUserAndTeam` explicitly decides that if the result list is incomplete, it's ok and can be returned.

Store method is left untouched so other callers can decide not to continue if the list is incomplete.

### DIFF
--- a/server/services/store/mattermostauthlayer/mattermostauthlayer.go
+++ b/server/services/store/mattermostauthlayer/mattermostauthlayer.go
@@ -911,12 +911,15 @@ func (s *MattermostAuthLayer) GetBoardsForUserAndTeam(userID, teamID string, inc
 
 	boards, err := s.Store.GetBoardsInTeamByIds(boardIDs, teamID)
 	// ToDo: check if the query is being used appropriately from the
-	//       interface, Wiggin77 thinks we're passing IDs that belong
-	//       to boards and templates, and the store method only
-	//       fetches boards, so we need to accept a partial result as
-	//       valid
-	var naf *model.ErrNotAllFound
-	if err != nil && !errors.As(err, &naf) {
+	//       interface, as we're getting ID sets on request that
+	//       return partial results that seem to be valid
+	if model.IsErrNotFound(err) {
+		if boards == nil {
+			boards = []*model.Board{}
+		}
+		return boards, nil
+	}
+	if err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -910,7 +910,13 @@ func (s *MattermostAuthLayer) GetBoardsForUserAndTeam(userID, teamID string, inc
 	}
 
 	boards, err := s.Store.GetBoardsInTeamByIds(boardIDs, teamID)
-	if err != nil {
+	// ToDo: check if the query is being used appropriately from the
+	//       interface, Wiggin77 thinks we're passing IDs that belong
```
wiggin77, line 914: I tried removing the filter on templates from the store method and got the same mismatched lengths, so I'm not so sure templates are the issue.  Why do we filter out templates from the query anyway?  If a template id is in the list why not return it?

server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -910,7 +910,13 @@ func (s *MattermostAuthLayer) GetBoardsForUserAndTeam(userID, teamID string, inc
...
+	//       to boards and templates, and the store method only
+	//       fetches boards, so we need to accept a partial result as
+	//       valid
+	var naf *model.ErrNotAllFound
```
wiggin77, line 918: Why not use the IsNotFound method?  I think there is a IsNotAllFound method as well.

server/services/store/mattermostauthlayer/mattermostauthlayer.go
mgdelacroix: I don't see a reason to filter out templates in this method, and the `mmauthlayer` method is the only caller, so I'm removing the filter

server/services/store/mattermostauthlayer/mattermostauthlayer.go
mgdelacroix: I wanted to filter out "full" not founds as in those cases we'd return a `nil` struct rather than an empty one, but we should control those instead of returning the error, I'll fix it

mgdelacroix: Fixed on [146f4c6](https://github.com/mattermost/focalboard/pull/3842/commits/146f4c67b207a29b45a2b91488645fda9105ed23)

server/services/store/mattermostauthlayer/mattermostauthlayer.go
mgdelacroix: Fixed on [146f4c6](https://github.com/mattermost/focalboard/pull/3842/commits/146f4c67b207a29b45a2b91488645fda9105ed23)

Issue #3799, opened 2022-09-08
Bug: hiding a board doesn't work the first time


mgdelacroix:
## Steps to reproduce the behavior

1. Hide a board on your RHS
2. It will not hide
3. Hide it again
5. It will work

## Expected behavior

The operation should work at the first attempt

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.3-RC1
 - Browser and OS: Firefox 104.0.2 on Linux

PR #3843, opened 2022-09-14
Update user store to return the updated entities after inserting / patching them

mgdelacroix:
#### Summary
This PR updates the user store in both `sqlstore` and `mmauthlayer` to return the updated entities after inserting / patching them.

As part of this PR I'm removing all references to `User.Props`, as those were migrated to the `preferences` table and are not fetched / updated anymore. A follow up migration that actually removes the column should follow once we assert that everything is working correctly.

This PR also removes the ambiguity between Props and Preferences on the naming of the model structs and the store methods for consistency.

#### Ticket Link
It could potentially fix https://github.com/mattermost/focalboard/issues/3799

### DIFF
--- a/server/api/users.go
+++ b/server/api/users.go
@@ -115,7 +115,6 @@ func (a *API) handleGetMe(w http.ResponseWriter, r *http.Request) {
 			Email:    model.SingleUser,
 			CreateAt: ws.UpdateAt,
 			UpdateAt: now,
-			Props:    map[string]interface{}{},
 		}
 	} else {
 		user, err = a.app.GetUser(userID)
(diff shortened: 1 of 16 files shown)


wiggin77 (approved):
LGTM :+1: Nice work.

harshilsharma63 (approved):

harshilsharma63:
It does indeed solve #3799

mgdelacroix:
Could be the fix as well for https://github.com/mattermost/focalboard/issues/3848, will make sure that's the case

mgdelacroix:
I think that for the cherry-pick of this PR into `v7.3` we should not remove the `Props` references, just to make it safer and taking into account that we're late on the testing cycle. If nobody is opposed to it, I will leave the deprecation of that field outside of the cherry-pick

/cc @sbishel

PR #3858, opened 2022-09-16
Catch exception on failed WS command to avoid interface from breaking

mgdelacroix:
#### Summary
This PR avoids the interface breaking completely on a failed WS message sent and prints the error that caused the failure


jespino (approved):
LGTM

PR #3870, opened 2022-09-19
Fixes app crash in multiperson property if user is removed from board

### LINKED ISSUE #3869: Bug: Multiperson Property - setting a user and removing them from board crashes the app

1. Create a board and a card and create a multiperson property in the card.
3. Set a user in that property.
4. Configure the board to display that property on the board.
5. Now remove that user from the list of board members.
6. Notice the app crashes and error-loops a few times.

## Expected behavior

The app shouldn't crash, like ever.

harshilsharma63:
Fixes #3869


sbishel (approved):
Thanks @harshilsharma63

Issue #3864, opened 2022-09-16
Bug: New board members do not assume minimum board role


wuwinson:
## Steps to reproduce the behavior

1. Create a new board
2. Set the team role to "Viewer"
3. Add an explicit board member
4. See error - member is added as "Editor" and not "Viewer"

## Expected behavior

New board members are automatically assigned the minimum team role.

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/190798954-9c9a7790-4849-44e6-9ea6-3b2147d00a3a.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.3 Bug Bash
 - Browser and OS: Chrome on Mac

Issue #3307, opened 2022-06-30
Bug: bugbash 7.2 Team members not appearing in Assignee property dropdown


wiggin77:
## Steps to reproduce the behavior

1. Create board from RoadMap template (or any template that has Assignee prop or any person property type)
2. Click Share button
3. Set `Everyone at Focalboard Team` to have comment permission
4. Open card
5. Try to set Assignee property
6. Notice only explicit board members are in the dropdown list.

## Expected behavior

Users won't know the difference between implicit and explicit membership.  If they set permissions for all team members they would expect all team members to show up in person lists.

## Edition and Platform

 - Edition: Bugbash 7.2
 - Version: 7.2
 - Browser and OS: FF Linux

wuwinson:
@asaadmahmood Can we get designs for this? Want to show everyone on the team with a divider/indicator for non-explicit board members

asaadmahmood:
@wuwinson We already have a design for that:
https://www.figma.com/file/tBRYmKqR1pUo02EeAHXF2P/Issue-512---User-not-in-Channel%2FBoard?node-id=501%3A24609

wuwinson:
Missed that, thanks @asaadmahmood!

wuwinson:
Spec: https://docs.google.com/document/d/1IiIgtXI_lYtTSpm6XfqPvSe_jD0E0rajqerWSRLC6OQ/edit?usp=sharing

jespino:
I think this is not fixed in main, if you have permissions to manage members of the board you are going to be able to search for users in the team, if not you are not going to be able. Once you choose one of the members, you need to add the member to the board to assign it, if not, you are no going to be able to assign it. Can you confirm that is what you expect @wuwinson ?

wuwinson:
> if you have permissions to manage members of the board you are going to be able to search for users in the team, if not you are not going to be able. Once you choose one of the members, you need to add the member to the board to assign it, if not, you are no going to be able to assign it.

@jespino Yep, this is the expected behavior for a private board. Only board admins will be able to search for users on the team and add them to the board. Everyone else will only see board members on the person property and for @mentions.

For a public board however, an Editor or Commenter can also search for users on the team, and add them to the board as the **default** role. A board Admin on a public board can **specify the role** of the new user when adding them to the board. See the matrix for reference:

[...]

PR #3875, opened 2022-09-19
Fix members

sbishel:
#### Summary
Fixes issues to display the correct team or board members when @mentioning or in a person property. 

Also introduces enum to ensure consistency when storing, converting to correct language only for display.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/3307
  Fixes https://github.com/mattermost/focalboard/issues/3864


webapp/src/components/viewHeader/filterEntry.tsx
```diff
@@ -126,7 +126,7 @@ const FilterEntry = (props: Props): JSX.Element => {
                         <>
                             <Menu.Text
                                 id='is'
-                                name={intl.formatMessage({id: 'Filter.contains', defaultMessage: 'is'})}
+                                name={intl.formatMessage({id: 'Filter.is', defaultMessage: 'is'})}
```
sbishel, line 129: Caused issues running i18n-extract.

webapp/src/mutator.ts
```diff
@@ -1110,7 +1108,7 @@ class Mutator {
         const boardTemplate = createBoard()
         boardTemplate.isTemplate = true
         boardTemplate.teamId = teamId
-        boardTemplate.title = intl.formatMessage({id: 'View.NewTemplateTitle', defaultMessage: 'Untitled Template'})
+        boardTemplate.title = intl.formatMessage({id: 'View.NewTemplateTitle', defaultMessage: 'Untitled'})
```
sbishel, line 1111: Caused conflicts with i18n-extract

wuwinson:
https://user-images.githubusercontent.com/93531870/191125059-c188dd12-f25b-4341-a747-6206b81845fd.mov

https://github.com/mattermost/focalboard/issues/3307 is partially fixed:
1. Create new board
2. Add person property
3. Select a non-board member (fixed)
4. Assign role to non-board member
5. Open Share modal
6. See error - user's role displays as admin, but when clicking on the role dropdown, the proper role is assigned
7. Repeat steps 2-6 with @mention and the same issue occurs

https://github.com/mattermost/focalboard/issues/3864 not fixed:
1. Create a new board
2. Open share modal and set team role to Commenter
3. Add an explicit board member
4. See error - user is still added as "Editor" rather than "Commenter"
5. Additionally, no longer able to set the board back to private.

wuwinson:
https://github.com/mattermost/focalboard/issues/3307 is fixed, but https://github.com/mattermost/focalboard/issues/3864 is still not fixed:

1. Create a new board
2. Open share modal and set team role to "Commenter"
3. Add an explicit board member
4. See error - the user is still incorrectly added as "Editor" rather than "Commenter"
5. Remove the same explicit member from board (step 3)
6. Open a card and add the user via the person property and/or @mentioning
7. Modal appears for admin to assign a role to the board member
8. See error - Able to select "Viewer" role from the dropdown even though "Commenter" was set as the min role (step 2)
9. Open Share modal - User is added as a "Viewer" even though min role is set to "Commenter"

https://user-images.githubusercontent.com/93531870/192323545-570e4175-c0cc-492a-8d35-e55ff36f1a6e.mov

wuwinson:
> 
Steps 1-4 are now fixed, but steps 6-9 are still an issue.

https://user-images.githubusercontent.com/93531870/192592370-9dfe0e67-9e56-48c7-89c7-11f1ccb08a61.mov

sbishel:
I'm still working on this, but actually 6-9, I don't think are an issue. An admin should be able to add a user at any level. They can change it in share board, but why make them go there and update. I'm still working on some of the styling, but I don't think we should restrict the admin.

wuwinson:
Got it, thanks Scott. 6-9 would create an inconsistent experience from share board, because the admin wouldn't be able to assign a role lower than the min role from share boards. It would also create a confusing experience with "min roles", otherwise we could have kept "default roles" like what we have in v7.2.1.

wuwinson:
Steps 6-9 are still occurring, but everything else looks good @sbishel. You can't assign a role lower than the minimum role from the Share boards modal (the role is not available to select), and if you change the minimum role, it would promote everyone on the board to that role, so allowing an admin to set a lower role for an individual user only from the @mention/person property modal can be misleading.

sbishel:
@harshilsharma63 @jespino - Can you take a look at this when you get a chance? There are a couple other bugs dependent on this fix and this one is important to 7.3 release.

harshilsharma63:
@sbishel for #3864 adding a new user to board does follow the min role set for the team, but adding a new channel sets the channel's roles to editor. Should linked boards follow the same team-level min role? cc @wuwinson 


https://user-images.githubusercontent.com/18575143/193204527-4c61c368-1335-43f8-8ff4-72e561101d16.mp4

When adding a new board member by adding them to an assignee property, the font size for some texts in the popup seem off. I think you're already fixing it but still adding here just so its not missed

Screenshot 2022-09-30 at 11 58 00 AM

Everything else looks great!

harshilsharma63 (approved):
LGTM 👍

wuwinson:
> @sbishel for #3864 adding a new user to board does follow the min role set for the team, but adding a new channel sets the channel's roles to editor. Should linked boards follow the same team-level min role? cc @wuwinson
> 
>  min-role-for-channel.mp4 
> ~When adding a new board member by adding them to an assignee property, the font size for some texts in the popup seem off. I think you're already fixing it but still adding here just so it's not missed~ I see you already have aPR for it #3887
> 
> Screenshot 2022-09-30 at 11 58 00 AM
> 
> Everything else looks great!

@harshilsharma63 Thanks for calling that out. We intentionally left out linked channels in this release due to bandwidth. We'd have to update the modal when linking a board on the RHS as well and include the new roles in the dropdown for the linked board modal. Was easier to keep linked channels on the "Editor" role for now, since they'd always be equal or above the min role set for the team.

PR #3877, opened 2022-09-23
Boards as persisted category

### LINKED ISSUE #3638: Bug: 'Boards' category expands/collapses for all teams

1. Go to Team 1, expand/collapse "Boards"
2. Go to Team 2, note that state matches Team 1 - expand/collapse
3. Go to Team 1, state matches Team 2, not how Team 1 was left.

## Expected behavior

The "Boards" category to expand/collapse independantly across teams

## Edition and Platform

 - Version: v7.2.0-rc3

Pretty sure this is due to "Boards" being the default category and same ID for expand collapse is used across teams

harshilsharma63:
#### Summary
Currently, the `Boards` category was a pseudo categroy, as in, it didn't really exist in the DB created in the webapp while rendering the LHS. This caused some issues like inability to persist its expanded/collapsed state, inability to store its order when implementing drag and drop to re-arrange categories. Hence, we're persisting the category.

#### Ticket Link
Fixes #3638

Please note that purpose of this PR is not to just fix #3638 but to lay up the ground work for LHS drag and drop support and allow categories to be rearranged.


server/app/category_boards.go
```diff
@@ -1,9 +1,107 @@
...
+		if !belongsToCategory {
+			if err := a.AddUpdateUserCategoryBoard(teamID, userID, createdCategory.ID, board.ID); err != nil {
+				return nil, fmt.Errorf("createBoardsCategory failed to add category-less board to the default category, defaultCategoryID: %s, error: %w", createdCategory.ID, err)
+			}
+
+			createdCategoryBoards.BoardIDs = append(createdCategoryBoards.BoardIDs, board.ID)
+		}
+	}
+
+	return createdCategoryBoards, nil
```
sbishel, line 104: Couldn't all of this code be eliminated and handle the entire update via SQL.

sbishel (commented):
Curious as to why do a "check" for defaultCategoryBoards every time GetUserCategoryBoards is called rather than handle it via a migration. It seems it would be a simple migration to add the DefaultBoardCategory and then update those board without a category. I think the migration would eliminate quite a bit of the code changes.

server/app/category_boards.go
harshilsharma63: I thought I added a code comment on explaining this, but I might have missed it.

Even if you create the default board categories in SQL migration, you still need the logic to create the categories for new users. Channels does it during the user signup process. We don't have any such thing. So we still need logic in the code to create the default boards category for a new user. Hence, doing everything in code instead of duplicating the logic in code + SQL migration.

sbishel (approved):

Issue #3551, opened 2022-08-04
Bug: Deleting the current group by property generate a javascript error


jespino:
## Steps to reproduce the behavior

1. Go to a board with a kanban board
2. Go to the kanban board
3. Open a card
4. Remove the current group by property in the card (needs to be the last select property available)
5. It fails and redirects you to the error page

## Expected behavior

Delete it without a problem (And show that there is no property for orther this, see #3550 

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: 7.2
 - Browser and OS: Firefox on Linux

cannalee90:
May I take this one?

PR #3900, opened 2022-10-01
fix javascript error when curren group is delete

cannalee90:
#### Summary

I just fixed that described in #3551. It reproduced when there is only one group property which is grouping the card and it removed. In Kanban component, it has early return statement which caused the render-hook error.  I just moved it to before the return statement.

#### Ticket Link

https://github.com/mattermost/focalboard/issues/3551

### DIFF
--- a/webapp/src/components/kanban/kanban.tsx
+++ b/webapp/src/components/kanban/kanban.tsx
@@ -66,12 +66,7 @@ const Kanban = (props: Props) => {
         }
     }, [activeView.fields.defaultTemplateId])
 
-    if (!groupByProperty) {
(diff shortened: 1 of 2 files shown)


harshilsharma63:
@cannalee90 thanks a ton for your contribution! The changes works great. Can you please take a look at the failing tests in CI?

cannalee90:
> @cannalee90 thanks a ton for your contribution! The changes works great. Can you please take a look at the failing tests in CI?

sure I'm finished it

harshilsharma63:
CI passed 🎉

PR #3907, opened 2022-10-03
Fix bug where workspace era uploaded files won't load on team era codebase

### LINKED ISSUE #3879: Bug: Updating to 7.2 replaced all images into 4 empty text block

1. Create a board and card in pre-7.2 versions
2. Add add an image to the card
3. Update to 7.2 through updating mattermost
4. Images get replaced into 4 empty text block

## Expected behavior

The images stay there as how it was before updating

## Screenshots (optional)

image
The red box is where the image should be

## Edition and Platform

 - Version: 7.2.1
 - Browser and OS: Tested on, but not limited to Chrome on Mac, Mattermost Desktop app (windows/mac)

* Sev 1: Affects all past cards with images, while backing up does not help restoring easily
* Sev 2: Not updating is the only workaround for now

harshilsharma63:
The files uploaded in old version resided in a path build via workspace ID. The new code used team ID, which obviously didn't work. So added logic to check for old path if file isn't found at the new path.

Fixes #3879


sbishel:
@harshilsharma63 - is this ready to be reviewed?

harshilsharma63:
@sbishel yes, its ready to review.

sbishel (changes requested):
Don't these files need moved as well? If the user changes the channel association on the board, the files will be lost.

server/app/files.go
```diff
@@ -120,3 +123,18 @@ func (a *App) GetFileReader(teamID, rootID, filename string) (filestore.ReadClos
 
 	return reader, nil
 }
+
+func (a *App) MoveFile(channelID, teamID, boardID, filename string) error {
+	oldPath := filepath.Join(channelID, boardID, filename)
+	newPath := filepath.Join(teamID, boardID, filename)
+	err := a.filesBackend.MoveFile(oldPath, newPath)
+	if err != nil {
+		a.logger.Error("ERROR moving file",
```
Pinjasaur, line 132: would it be useful to do a `logger.Debug` too e.g. the file was successfully moved (`err == nil`)?

sbishel: May-be, I went back and forth and decided not?

sbishel (approved):

Issue #3916, opened 2022-10-03
Bug: Cannot change "type" for existing properties when there are values are assigned


wuwinson:
## Steps to reproduce the behavior

1. Open a card and click on the property name with a value assigned on it
2. Go to the "Type" menu and select a different type for the property
3. See error - Cannot change the property type
4. Works fine with properties with no values assigned

## Expected behavior

A confirmation modal should appear to confirm the property type change.

![Screen Shot 2022-10-03 at 3 04 09 PM](https://user-images.githubusercontent.com/93531870/193658151-f72c8d7c-c963-4e6a-a826-87dcbc06e749.png)

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/193658225-e0ebea10-79e0-403b-a018-7a61b219e7bc.mov

## Edition and Platform

 - Edition:  Mattermost Boards (plugin)
 - Version: v7.3 Bug Bash
 - Browser and OS: Chrome on Mac

PR #3920, opened 2022-10-04
GH-3916 fix property change

sbishel:
#### Summary
Confirm modal is throwing an error, because of object changes. 
Uses the intl display for the property type name.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/3916


Pinjasaur (approved):
Huh, wasn't aware of `displayName`. LGTM.

Issue #3885, opened 2022-09-29
Bug: Role drop-down for Share boards gets pushed down on boards with many members


wuwinson:
## Steps to reproduce the behavior

1. Go to a board with many explicit members
2. Scroll down to the last member
3. Click on the role dropdown
4. See error - dropdown appears below the modal

## Expected behavior

Dropdown appears directly under the role

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/193117891-4fc164b8-1d98-4701-b9fe-3afb5651a2cf.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.3 Bug Bash
 - Browser and OS: Chrome on Mac

PR #3925, opened 2022-10-04
Pins the roles submenu to the user permission row

mgdelacroix:
#### Summary
Adds an option to the `openUp` util to force the alignment to the bottom and uses it to pin the roles submenu to the user permission row.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/3885


harshilsharma63 (approved):
LGTM 👍

Issue #3903, opened 2022-10-01
Bug: Person auto-complete doesn't work in tables


chenilim:
## Steps to reproduce the behavior

1. Open a table view with a person column
2. Type in the person column
3. No auto complete

## Expected behavior

There should be auto-complete like typing in the card.

## Screenshots (optional)

<img width="731" alt="Screen Shot 2022-10-01 at 8 38 27 AM" src="https://user-images.githubusercontent.com/46905241/193417075-b2b3ac21-9c89-4704-a23e-38fd8fe44aaa.png">

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.2.1 (community)
 - Browser and OS: Chrome on Mac

chirag-ghosh:
Can i work on this ?

wuwinson:
@sbishel Looks like we have a regression on this bug as of v7.8.

sbishel:
@wuwinson Not a regression, but the same issue with multiPerson.

PR #3927, opened 2022-10-04
fix: Person autocomplete in table

chirag-ghosh:
#### Summary
Person autocomplete in table view fixed. The dropdown was rendering but because table cell had `overflow: hidden` styling, it was not visible. Overwrote the style for Person cell.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/3903


sbishel (approved):
Awesome, thanks @chirag-ghosh

PR #3943, opened 2022-10-06
Delete children when deleting boards and cards

wiggin77:
#### Summary
This PR ensures child blocks are deleted when a board or card is deleted.  Also handles undelete by bringing the child blocks back.  ~~A migration was added to clean up orphaned blocks.~~

The migration to clean up existing orphans was removed since it may be a long time before we can test that against a community clone.  That part will be a separate PR.

#### Ticket Link
https://community.mattermost.com/boards/team/qghzt68dq7bopgqamcnziq69ao/b4ctmm917q3rzdfi451qtuxjzcw/150d1464-6d59-4c39-810f-ee201439dbaf/ck8s57egkitbmfja47964rkgktw


server/services/store/sqlstore/boards_and_blocks.go
```diff
@@ -102,13 +102,11 @@ func (s *SQLStore) patchBoardsAndBlocks(db sq.BaseRunner, pbab *model.PatchBoard
...
 	for _, boardID := range dbab.Boards {
-		if err := s.deleteBoard(db, boardID, userID); err != nil {
-			return err
-		}
-
 		boardIDMap[boardID] = true
 	}
 
+	// delete the blocks first, since deleting the board will clean up any children and we'll get
+	// not found errors when delting the blocks after.
```
wiggin77, line 109:
```suggestion
	// not found errors when deleting the blocks after.
```

server/services/store/sqlstore/migrate.go
```diff
@@ -236,6 +237,14 @@ func (s *SQLStore) runMigrationSequence(engine *morph.Morph, driver drivers.Driv
 		return fmt.Errorf("error running categoryID migration: %w", mErr)
 	}
 
+	if mErr := s.ensureMigrationsAppliedUpToVersion(engine, driver, moveOrphanedBlocksToHistory); mErr != nil {
+		return mErr
+	}
+
+	if mErr := s.RunMoveOrphanedBlocksToHistoryMigration(); mErr != nil {
```
wiggin77, line 244: Looking for some feedback here.  Since this migration is not required in the same way a schema change would be, it could be run asynchronously and server start-up be allowed to proceed.  It is idempotent as well, in that it could be run multiple times without harm.  

For example, we could let it run async and set a `completed` flag in the database when it is finished. If the server is restarted before completing it would simply pick up where it left off.  Not sure if it is worth the effort.  My main concern is that the migration takes a long time (>15 minutes) with a 26 million block table with 100,000 orphans.

jespino (changes requested):
In general the PR looks good to me, but I don't love the construction of queries using `fmt.Sprintf`, if I readed correctly they are not creating any injection, but can be misleading in the future if somebody decide to modify that query and follow the same pattern.

server/services/store/sqlstore/blocks.go
```diff
@@ -788,3 +802,244 @@ func (s *SQLStore) duplicateBlock(db sq.BaseRunner, boardID string, blockID stri
...
+	sql := `
+		SELECT fb.id from %sblocks AS fb WHERE fb.board_id IN
+		(
+			SELECT bh.id FROM %sboards_history AS bh,
+				(SELECT id, max(insert_at) AS max_insert_at FROM %sboards_history GROUP BY id) AS sub
+			WHERE bh.id=sub.id AND bh.insert_at=sub.max_insert_at AND bh.delete_at > 0
+		)
+		order by fb.insert_at desc limit %d;
+	`
+	sql = fmt.Sprintf(sql, s.tablePrefix, s.tablePrefix, s.tablePrefix, limit)
```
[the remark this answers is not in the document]
wiggin77: > It shouldn't be a problem, because you are concatenating using a `%d` there, but I would prefer to put an `?` and pass the `limit` as a parameter to the `db.Query`

There is no possibility for injection here, but you're right, it's a bad example for other developers. I'll fix.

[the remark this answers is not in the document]
wiggin77: > you said you can't use the multiple from columns in the squirrel select, but looks like you are using it in the previous method

TBH, I think the squirrel version of these particular queries are far less readable and harder to construct than straight SQL.  I used squirrel in the previous query because there were multiple queries needed which I composed as sections with a large column list as well.  I had to stuff a bunch of stuff into the `From(...)` API that doesn't belong there.  Overall the economy of re-using the sub-queries may be worth the less readable code. :man_shrugging: 

> but I would prefer to put an ? and pass the limit as a parameter

Unfortunately we cannot use placeholders for table names, so I still need to compose the string for those. I did modify it to pass the limit.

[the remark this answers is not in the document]
jespino: I'm ok with that for table name with the tablePrefix as input, because that is not an "external" input. But yes, still an example that is far from be perfect.

The reason for that, is to allow on-prem or even allow to reuse your same database adding a prefix for the table name. Probably we can get rid of that in the mid term when we are only the boards product.

server/services/store/sqlstore/blocks.go
```diff
@@ -788,3 +802,264 @@ func (s *SQLStore) duplicateBlock(db sq.BaseRunner, boardID string, blockID stri
...
+	if boardID == "" {
+		return ErrEmptyBoardID{}
+	}
+
+	where := fmt.Sprintf("board_id='%s'", boardID)
+	if parentID != "" {
+		where += fmt.Sprintf(" AND parent_id='%s'", parentID)
+	}
+
+	/*
```
jespino, line 877: Nit: This comments are going to end up being misleading if we don't update them properly, I prefer to avoid them.

server/services/store/sqlstore/blocks.go
```diff
@@ -788,3 +802,264 @@ func (s *SQLStore) duplicateBlock(db sq.BaseRunner, boardID string, blockID stri
...
+	if limit <= 0 {
+		limit = 1000
+	}
+	if limit > 1000 {
+		return nil, ErrLimitExceeded{1000}
+	}
+
+	// Find all orphaned child blocks for deleted boards.
+	// Squirrel can't express multiple tables in FROM, nor can it mix `From` and `FromSelect`.
+	/*
```
jespino, line 973: The query shown here, I think is misleading, the comment on top, on the other hand, I think is right.

server/services/store/sqlstore/blocks.go
```diff
@@ -788,3 +802,264 @@ func (s *SQLStore) duplicateBlock(db sq.BaseRunner, boardID string, blockID stri
...
+	if limit <= 0 {
+		limit = 1000
+	}
+	if limit > 1000 {
+		return nil, ErrLimitExceeded{1000}
+	}
+
+	// Find all orphaned child blocks for deleted cards.
+	// Squirrel can't express multiple tables in FROM, nor can it mix `From` and `FromSelect`.
+	/*
```
jespino, line 1025: same here about the SQL comment, it can end up outdated and leading to confusion, I prefer the textual comment on top showing the intention.

jespino (approved):
Everything looks good to me, only some improvement about adding or not comments including the SQL query, I added my very personal opinion, If you don't agree, feel free to discuss or event merge without discussing it.

sbishel (approved):

wiggin77:
@sbishel I have removed the orphan cleanup migration due to the delays getting a community clone.  I'll add that code into a separate PR.

sbishel (approved):

wiggin77:
@ogi-m since we're no longer doing the expensive data migration that needed to be tested on a community clone, are you comfortable with merging this without a full QA?

ogi-m (approved):
@wiggin77 I've tested deleting cards/boards and undo actions, LGTM

PR #3961, opened 2022-10-06
add setting admin role when add member

### LINKED ISSUE #3957: Bug: Admins cannot add other non-board members as "Admins" via autocomplete list

1. Go to any board as an admin
2. Open a card and click on a person property and select a non-board member to add to the board
3. Confirmation modal appears - select "Admin" from the dropdown list
4. See error - user is not added to the person property, nor are they visible on the Share dialog
5. Refresh browser
6. Open Share dialog - user is now visible as a board member, but assigned the "Viewer" role instead of "Admin" role
7. Open the same card from Step 2
8. See error - user is not assigned on the person property even after a refresh
9. Repeat same steps with @mention in the comments section to see the same error

## Expected behavior

User is added to the board as an "Admin" and is properly assigned on the person property.

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/194408828-076e9509-c610-43ef-87aa-2b4a291ac7e6.mov

## Edition and Platform

 - Version: v7.3 Bug Bash

sbishel:
#### Summary
Renamed variable to avoid confusion and included setting the user as Admin, since admin users can be set by other admins.

#### Ticket Link

  Fixes #3957

### DIFF
--- a/webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
+++ b/webapp/src/components/markdownEditorInput/markdownEditorInput.tsx
@@ -125,9 +125,10 @@ const MarkdownEditorInput = (props: Props): ReactElement => {
             boardId: board.id,
             userId,
             roles: role,
-            schemeEditor: minimumRole === MemberRole.Editor,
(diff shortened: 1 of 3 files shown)


wiggin77 (approved):
LGTM 👍

PR #3964, opened 2022-10-07
DND support for category and boards in LHS

harshilsharma63:
Card link - https://community.mattermost.com/boards/team/qghzt68dq7bopgqamcnziq69ao/b4ctmm917q3rzdfi451qtuxjzcw/vd89yph84ajbkufkbgts9m3sb5a/cmif57z7mt3ypikwtwxtem6dmzo

Pending items-

1. Bulk drag and drop boards [from the spec](https://docs.google.com/document/d/1NkUuKM1gAxOs1UweqzrYJRXOGbJjjmi2Exx45BfWMWk/edit#)
2. Creating new categories design changes [from the spec](https://docs.google.com/document/d/1NkUuKM1gAxOs1UweqzrYJRXOGbJjjmi2Exx45BfWMWk/edit#)
3. Test cases


wuwinson:
@harshilsharma63 This is looking great!

Dragging and dropping categories and boards was very smooth with the exception of your currently opened board. Especially when the current board has many views, then the drag and drop isn't as responsive. This may not be an issue later on if we move the views out of the sidebar. See recording below for more details:

https://user-images.githubusercontent.com/93531870/194577017-cdb357b0-fdee-4d0b-8ca6-19da0f724c6a.mov

server/app/category.go
```diff
@@ -34,10 +36,8 @@ func (a *App) CreateCategory(category *model.Category) (*model.Category, error)
 }
 
 func (a *App) UpdateCategory(category *model.Category) (*model.Category, error) {
-	// set to default category, UI doesn't create with Type
-	if strings.TrimSpace(category.Type) == "" {
```
harshilsharma63, line 38: Note to reviewer: this is now being done in `category.Hydrate()` call.

asaadmahmood (approved):
From a UX perspective, this looks good to me. I see a bunch of errors when dragging items.

server/api/categories.go
```diff
@@ -14,9 +14,11 @@ import (
 func (a *API) registerCategoriesRoutes(r *mux.Router) {
 	// Category APIs
 	r.HandleFunc("/teams/{teamID}/categories", a.sessionRequired(a.handleCreateCategory)).Methods(http.MethodPost)
+	r.HandleFunc("/teams/{teamID}/categories/reorder", a.sessionRequired(a.handleReorderCategories)).Methods(http.MethodPut)
```
jespino, line 17: Probably I would ask @wiggin77 about this new endpoints routes to verify it matches with the approach that we are following for the new API endpoints. I think we can add new endpoints, but if we are consistent upfront, is going to be less painful later.

@wiggin77 can you confirm if this is inline with your new API design, and if not, how would you change the URLs there?

server/api/categories.go
```diff
@@ -362,3 +364,136 @@ func (a *API) handleUpdateCategoryBoard(w http.ResponseWriter, r *http.Request)
...
+
+	var newCategoryOrder []string
+
+	err = json.Unmarshal(requestBody, &newCategoryOrder)
+	if err != nil {
+		a.errorResponse(w, r, err)
+		return
+	}
+
+	auditRec := a.makeAuditRecord(r, "reorderCategories", audit.Fail)
```
jespino, line 413: Should we add more information about what is going on in to the audit record here? (the same for the other API endpoint).

server/app/boards.go
```diff
@@ -327,10 +327,13 @@ func (a *App) addBoardsToDefaultCategory(userID, teamID string, boards []*model.
...
 
+	boardCategoryMapping := map[string]string{}
 	for _, board := range boards {
-		if err := a.AddUpdateUserCategoryBoard(teamID, userID, defaultCategoryID, board.ID); err != nil {
-			return err
-		}
+		boardCategoryMapping[board.ID] = defaultCategoryID
+	}
+
+	if err := a.AddUpdateUserCategoryBoard(teamID, userID, boardCategoryMapping); err != nil {
```
jespino, line 335: Nice, this looks like an improvement in performance (I'll keep digging to see it, but looks better for sure.

server/app/category.go
jespino: In the `Category.Hydrate` is not doing the `string.TrimSpace`, I guess that is ok, but just to let you know that it is not exactly equivalent.

server/app/category.go
```diff
@@ -130,3 +134,109 @@ func (a *App) DeleteCategory(categoryID, userID, teamID string) (*model.Category
...
+
+	var sourceCategoryBoards *model.CategoryBoards
+	defaultCategoryID := ""
+
+	// iterate user's categories to find the source category
+	// and the default category.
+	// We need source category to get the list of its board
+	// and the default category to know its ID to
+	// move source category's boards to.
+	for i := range categoryBoards {
```
jespino, line 154: The way you are looping here is confusing to me, I don't know exactly what you need to use the index instead of a copy to the original data. For me is way more readable something like:

```go
	for _, category := range categoryBoards {
		if category.ID == sourceCategoryID {
			sourceCategoryBoards = &category
		}

		if category.Name == defaultCategoryBoards {
			defaultCategoryID = category.ID
		}

		// if both categories are found, no need to iterate furthur.
		if sourceCategoryBoards != nil && defaultCategoryID != "" {
			break
		}
	}
```

jespino: the other version makes me think you need to work with references to modify the same memory for some reason, so It was confusing to me.

server/integrationtests/sidebar_test.go
```diff
@@ -0,0 +1,54 @@
+package integrationtests
```
jespino, line 1: Not super critical for this part, but I think is needed the permissions tests for the new api endpoints (look at the `server/integrationtests/permissions_test.go`

server/services/store/sqlstore/category_boards.go
```diff
@@ -78,39 +82,50 @@ func (s *SQLStore) addUserCategoryBoard(db sq.BaseRunner, userID, categoryID, bo
...
+
+	if _, err := query.Exec(); err != nil {
 		s.logger.Error("addUserCategoryBoard error", mlog.Err(err))
 		return err
 	}
 	return nil
 }
 
-func (s *SQLStore) deleteUserCategoryBoard(db sq.BaseRunner, userID, boardID string) error {
+func (s *SQLStore) deleteUserCategoryBoard(db sq.BaseRunner, userID string, boardIDs []string) error {
```
jespino, line 110: Probably this should be rename to `deleteUserCategoryBoards` with the `s` at the end.

jespino (approved):
In general looks good to me. I don't see any stopper to merge this PR, but I see some places where we can improve before or after merge. But in general I really like the PR. :)

server/services/store/sqlstore/category_boards.go
harshilsharma63: Done.

server/integrationtests/sidebar_test.go
harshilsharma63: As it's not critical I'll defer for now in favor of getting the feature merged in for release. I'll work on this after merging.

server/app/category.go
harshilsharma63: I do nee to work with references.

```go
if category.ID == sourceCategoryID {
			sourceCategoryBoards = &category
		}
```

Here I need to save the reference of categoryboard while iterating.

server/app/boards.go
harshilsharma63: Yup, it's making the updates in a single query.

server/api/categories.go
harshilsharma63: Done.

harshilsharma63:
@jespino I've addressed your comments. Waiting for @wiggin77 about the endpoint URL.

mgdelacroix (changes requested):
A couple of permission checks required

server/api/categories.go
```diff
@@ -362,3 +364,136 @@ func (a *API) handleUpdateCategoryBoard(w http.ResponseWriter, r *http.Request)
...
+	//     description: internal error
+	//     schema:
+	//       "$ref": "#/definitions/ErrorResponse"
+
+	vars := mux.Vars(r)
+	teamID := vars["teamID"]
+
+	ctx := r.Context()
+	session := ctx.Value(sessionContextKey).(*model.Session)
+	userID := session.UserID
```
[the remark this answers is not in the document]
harshilsharma63: Done.

server/api/categories.go
```diff
@@ -362,3 +364,136 @@ func (a *API) handleUpdateCategoryBoard(w http.ResponseWriter, r *http.Request)
...
+	//     schema:
+	//       "$ref": "#/definitions/ErrorResponse"
+
+	vars := mux.Vars(r)
+	teamID := vars["teamID"]
+	categoryID := vars["categoryID"]
+
+	ctx := r.Context()
+	session := ctx.Value(sessionContextKey).(*model.Session)
+	userID := session.UserID
```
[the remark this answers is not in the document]
harshilsharma63: Done.

server/api/categories.go
wiggin77: This looks good to me.

ogi-m:
I'm seeing the same issue that Winson noted above, I've created a separate ticket for it - https://github.com/mattermost/focalboard/issues/4242

server/api/categories.go
harshilsharma63: Awesome! Thanks!

harshilsharma63:
@ogi-m since the dev reviews are complete, can you please test this out before I merge it to `main`?

mgdelacroix (approved):

Issue #3776, opened 2022-09-04
Feature Idea: export board archive in the side bar context menu


kamre:
## Summary

Currently it is possible to export board archive via board view options menu:
<img width="396" alt="image" src="https://user-images.githubusercontent.com/5587620/188308194-4e3fa01e-4a0a-4ee3-9a29-b9c1242c12b0.png">

This action can also be added to the side bar context menu:
<img width="241" alt="image" src="https://user-images.githubusercontent.com/5587620/188308249-f39e404a-2a52-4986-adaa-b5508338a413.png">

## How important this is to me and why

Importance: Low (nice to have, logically this export is for the board and not just board view)

@asaadmahmood what do you think?

asaadmahmood:
Makes sense @kamre

sbishel:
@wuwinson?  We can open for Hacktoberfest?

wuwinson:
Thanks @kamre @sbishel! Let's add the "Export board archive" option below the "New template from board" option for the ... menu next to each board on the sidebar. Button should behave exactly the same as from the boards view. Asaad is out for a few weeks, but please see my hacked together mockup for reference:

![Screen Shot 2022-09-29 at 5 16 08 PM](https://user-images.githubusercontent.com/93531870/193148681-496fa73a-869c-47d8-9ec0-ab10715251de.png)

Icon attached - will need to be recolored to match the other icons on the menu:
<img width="38" alt="icon-export-variant" src="https://user-images.githubusercontent.com/93531870/193148826-67e63409-39c2-4f3a-b3f7-ea096e4a994c.png">

cannalee90:
Could you assign this issue for me?

kamre:
@wuwinson there is a corresponding Compass icon that can be used like this:
```ts
export default function ExportIcon(): JSX.Element {
    return (
        <CompassIcon
            icon='export-variant'
            className='ExportIcon export-variant'
        />
    )
}
```
Will look like this:
![Screenshot 2022-10-03 at 09 58 31](https://user-images.githubusercontent.com/5587620/193494023-f4713df8-7815-4b5f-ab8c-3ff6298ca045.png)

wuwinson:
Perfect, thanks @kamre! @cannalee90 you got it!

PR #3977, opened 2022-10-09
Add `export board archive` in the side bar context menu

cannalee90:
#### Summary

- add `export board achieve` int the sidebar context menu
- I don't know it was right but I update snapshot by using `npm run updatesnapshot`(make sure that I'm doing right)
- everyone who could use board are authorized to export board
- order of context menus in capture
image

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/3776


sbishel (approved):
@cannalee90 - Looks good! Thank your for your contribution.

PR #3984, opened 2022-10-10
Check-in `templates.boardarchive` file

wiggin77:
#### Summary
Partially reverts [issue-focalboard-2558](https://github.com/mattermost/focalboard/issues/2558) which builds templates archive at compile time instead of having that file checked into the codebase.  The file is imported into the binary via the //go:embed directive, and the build will fail if the file does not exist.

When another application imports a Boards package, such as [client](http://github.com/mattermost/focalboard/tree/main/server/client), the importing app will fail to build due to this missing file.  To work around this, the developer needs to clone the Boards repo, build the templates archive file, and use a go.work file in their own project to pick up the local repo instead of importing via the package cache.  This is a lot of hoops for a developer to jump through and caused (continues to cause) headaches for projects importing Board's client and model.

~~@mgdelacroix had the great idea to check in an empty archive to satisfy the embed when importing without including a 6MB+ binary.  This PR adds that file.  The file is already excluded via `.gitignore`.~~  Unfortunately a small placeholder file checked in will appear as changed after every build.  There appears to be no way to force git to ignore changes to a tracked file.

The full `templates.boardarchive` file has been checked in.  Since it no longer needs to be rebuilt on every compile, that step has been removed from the makefile.  `make templates-archive` can still be used to update the file when the sources change.

#### Ticket Link
NONE


wiggin77:
Unfortunately I cannot find a way to have git ignore future changes to the file.  Once this tiny file is checked it appears as a changed file after every build.

Pinjasaur:
@wiggin77 the only way I could think about doing the placeholder empty file is with [`git update-index --skip-worktree`](https://git-scm.com/docs/git-update-index#_skip_worktree_bit) on the file. This would need to be ran locally upon cloning the repo, similar to how the `go.work` is handled via the Makefile.

I'm slightly more partial to this strategy since the `.boardarchive` file being a binary type we don't get the niceties of seeing diffs in version control as it changes. But, the tradeoff is it adds some complexity in a similar way the `go.work` file did. Regardless, wanted to ping you that it might be a viable option. 🙂

wiggin77:
> [`git update-index --skip-worktree`](https://git-scm.com/docs/git-update-index#_skip_worktree_bit)

Thanks @Pinjasaur.  This only works locally and each developer/contributor would need to know to do this.  Also, the flag it sets is cleared any time the file is changed upstream, and they would have to run the command again.

The file is actually created [from the sources](https://github.com/mattermost/focalboard/tree/main/server/assets/templates-boardarchive) and a history of changes is preserved there.  I think the only downside is the extra 6MB downloaded on first clone.

Pinjasaur (approved):
Thanks @wiggin77. LGTM. Only question is if this would warrant docs changes (cc @justinegeffen)? I'd sort of lean no since I imagine we're the only ones changing builtin template.

Issue #3871, opened 2022-09-19
Bug: Plugin not able to migrate properly after latest Mattermost v7.3 upgrade


N3rdP1um23:
## Steps to reproduce the behavior

1. Upgraded Mattermost server from v7.2 to v7.3
2. After the Mattermost service was restarted, boards within Mattermost were no longer listed as an option (in the Channel, Boards, Playbooks dropdown)
3. Disabled and reenabled the boards plugin following a similar issue from here https://community.mattermost.com/core/pl/rpttbwstwpbbxcrsgg3d68ikwc - nothing worked
4. While disabling and reenabling upgraded the boards plugin from v7.0.1 to v7.2.1, there's a migration error (listed below)
5. Followed the steps from here https://github.com/mattermost/focalboard/issues/3697#issuecomment-1220817563 to adjust some database items, and after trying a disable/re-enable of the plugin, the migration continued to fail

## Expected behavior

Upgrading the Mattermost server should upgrade the internal boards plugin as well. During that time, all migrations should perform correctly and have it upgrade the data to be compatible between the versions.

## Screenshots (optional)

![image](https://user-images.githubusercontent.com/8093806/191032893-c504ac2a-bbd5-411f-a2d2-561fb6b80f24.png)
![image](https://user-images.githubusercontent.com/8093806/191032938-968ff61c-c30b-468f-8079-2ce8b14fbde0.png)

## Edition and Platform

 - Edition:
   - Mattermost Server v7.3
   - Boards Plugin - v7.2.1
 - Self-hosted
 - Browser and OS: 
   - Browser - Microsoft Edge Version 107.0.1379.1
   - OS - Windows 11
   - Server - Ubuntu

## Additional context (optional)

Output of error message
```
This plugin failed to start. error initializing the DB: driver: mysql, message: failed when applying migration, command: apply_migration, originalError: Error 1050: Table 'focalboard_teams' already exists, query:
RENAME TABLE focalboard_workspaces TO focalboard_teams;
ALTER TABLE focalboard_blocks CHANGE workspace_id channel_id VARCHAR(36);
ALTER TABLE focalboard_blocks_history CHANGE workspace_id channel_id VARCHAR(36);
ALTER TABLE focalboard_blocks ADD COLUMN board_id VARCHAR(36);
ALTER TABLE focalboard_blocks_history ADD COLUMN board_id VARCHAR(36);
[...]

PR #4002, opened 2022-10-12
Fix collation and charset after migrations (every run)

wiggin77:
#### Summary
This PR ensures that all FocalBoard tables have the correct collation and charset settings for existing tables and any added in the future.  This is done via a final migration step that applies these settings in an idempotent fashion on every run.  

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/3871

### DIFF
--- a/mattermost-plugin/server/manifest.go
+++ b/mattermost-plugin/server/manifest.go
@@ -45,8 +45,7 @@ const manifestStr = `
         "type": "bool",
         "help_text": "This allows board editors to share boards that can be accessed by anyone with the link.",
         "placeholder": "",
-        "default": false,
-        "hosting": ""
+        "default": false
       }
     ]
   }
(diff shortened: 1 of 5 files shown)


sbishel (approved):

Pinjasaur (approved):
LGTM @wiggin77. 2 questions:

1. is there a specific reason that `Channels` is used for the `table_name`? I'm just curious. If any table had to be the source of truth, it would make sense that would be it (or rather, the metadata for that table 🙂)
2. is CI expected to fail? 🤷🏻

wiggin77:
> 1. is there a specific reason that `Channels` is used for the `table_name`?

The idea is to make all focalboard tables consistent with mm-server tables wrt collation and charset.  `Channels` was chosen arbitrarily as a table we know exists in mm-server.  

> is CI expected to fail?

Working on it.  When running unit tests none of the mm-server tables exist.

PR #4016, opened 2022-10-13
MM-46725 Add build-product target to compile the Boards product

hmhealey:
Straightforward enough. One small change other than adding the new target is a slight change to the webpack config when building as a product because the JS code was being output to the wrong place previously for a product.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46275


wiggin77 (approved):
LGTM :+1:

PR #4025, opened 2022-10-17
Display board statistics

sbishel:
#### Summary
Initial implementation of board statistics to display in the admin console. Required an additional security check for user permissions.

#### Dependent on
https://github.com/mattermost/mattermost-server/pull/21433

#### Ticket Link
https://community.mattermost.com/boards/team/qghzt68dq7bopgqamcnziq69ao/b4ctmm917q3rzdfi451qtuxjzcw/vn3tt1q8ndj87frb4uyu4936yfh/ccitgbkb11bgxbyirdgnun9u7mw


server/services/store/sqlstore/blocks.go
```diff
@@ -521,7 +521,8 @@ func (s *SQLStore) getBoardCount(db sq.BaseRunner) (int64, error) {
 	query := s.getQueryBuilder(db).
 		Select("COUNT(*) AS count").
 		From(s.tablePrefix + "boards").
-		Where(sq.Eq{"delete_at": 0})
+		Where(sq.Eq{"delete_at": 0}).
+		Where(sq.Eq{"is_template": false})
```
sbishel, line 525: @wuwinson We are currently counting templates in board count for telemetry, which is wrong. But this may change some telemetry for boards count when this is released or we count templates, which means we should count template cards.

jespino (approved):
LGTM

mattermost-plugin/product/api_adapter.go
```diff
@@ -131,31 +135,23 @@ func (a *serviceAPIAdapter) HasPermissionToChannel(askingUserID string, channelI
 	return a.api.permissionsService.HasPermissionToChannel(askingUserID, channelID, permission)
 }
 
-//
 // Bot service.
-//
```
wiggin77, line 136: :cry:  Removing the empty comment lines associates the comment directly with the method following when viewing docs in the IDE or auto-generated docs.  These comment are for demarcation of API groups.

sbishel: Reverted change, had to add a blank line to prevent from auto-removing.

Issue #4018, opened 2022-10-14
Bug: Migration is not working (MM 7.4.0)


haivala:
## Steps to reproduce the behavior

1. Upgrade mattermost to from 7.3.0 to 7.4.0
2. Boards are not working

## Expected behavior

Broards should be working

## Edition and Platform

 - Edition:  Mattermost Docker Team edition 
 - Version: 7.4.0

## Additional context 
Log says
```
mattermost-postgres-1    | 2022-10-14 07:24:49.204 UTC [22] LOG:  database system was shut down at 2022-10-14 07:24:44 UTC
mattermost-postgres-1    | 2022-10-14 07:24:49.212 UTC [1] LOG:  database system is ready to accept connections
mattermost-postgres-1    | 2022-10-14 07:24:55.131 UTC [40] ERROR:  column "dirty" does not exist at character 17
mattermost-postgres-1    | 2022-10-14 07:24:55.131 UTC [40] STATEMENT:  SELECT version, dirty FROM "focalboard_schema_migrations" LIMIT 1
mattermost-postgres-1    | 2022-10-14 07:25:00.711 UTC [41] ERROR:  cannot delete from scalar
mattermost-postgres-1    | 2022-10-14 07:25:00.711 UTC [41] STATEMENT:
mattermost-postgres-1    | 	        INSERT INTO Preferences (UserId, Category, Name, Value) SELECT Id, 'focalboard', 'welcomePageViewed', replace((Props->'focalboard_welcomePageViewed')::varchar, '"', '') FROM Users WHERE Props->'focalboard_welcomePageViewed' IS NOT NULL ON CONFLICT DO NOTHING;
mattermost-postgres-1    | 	        INSERT INTO Preferences (UserId, Category, Name, Value) SELECT Id, 'focalboard', 'hiddenBoardIDs', replace(replace(replace((Props->'hiddenBoardIDs')::varchar, '"[', '['), ']"', ']'), '\"', '"') FROM Users WHERE Props->'hiddenBoardIDs' IS NOT NULL ON CONFLICT DO NOTHING;
mattermost-postgres-1    | 	        INSERT INTO Preferences (UserId, Category, Name, Value) SELECT Id, 'focalboard', 'tourCategory', replace((Props->'focalboard_tourCategory')::varchar, '"', '') FROM Users WHERE Props->'focalboard_tourCategory' IS NOT NULL ON CONFLICT DO NOTHING;
[...]

PR #4029, opened 2022-10-18
make sure users.props contains a hash!

oetiker:
#### Summary

Make sure the props field in the postgresql Users table does contain a JSON hash and not some other data structure like a string or an array.

If props contains something other thatn a hash, `props - 'key'` will fail and thus the migration will fail.

#### Ticket Link

This is a partial fix for #4018. 

#### Note

This PR does not address the other issue mentioned, about the `dirty` column. My guess is that this is spurious issue caused by a second/old copy of focalboard in the users installation.


harshilsharma63:
Need more investigation on this. This might not be as straight forward. I don't know if having a actual string in there is correct or not, of if having an array instead of an object in there is correct or not. My initial though is instead of resetting the data for all user rows, we should just selectively update the rows which do contain an object, i.e., 

```sql
UPDATE Users SET props = (props - 'focalboard_welcomePageViewed' - 'hiddenBoardIDs' - 'focalboard_tourCategory' - 'focalboard_onboardingTourStep' - 'focalboard_onboardingTourStarted' - 'focalboard_version72MessageCanceled' - 'focalboard_lastWelcomeVersion') WHERE props::text like '{%';
```

and just leave everything else untouched.

server/services/store/sqlstore/migrations/000027_migrate_user_props_to_preferences.up.sql
```diff
@@ -9,7 +9,8 @@
         INSERT INTO Preferences (UserId, Category, Name, Value) SELECT Id, 'focalboard', 'onboardingTourStarted', replace((Props->'focalboard_onboardingTourStarted')::varchar, '"', '') FROM Users WHERE Props->'focalboard_onboardingTourStarted' IS NOT NULL ON CONFLICT DO NOTHING;
         INSERT INTO Preferences (UserId, Category, Name, Value) SELECT Id, 'focalboard', 'version72MessageCanceled', replace((Props->'focalboard_version72MessageCanceled')::varchar, '"', '') FROM Users WHERE Props->'focalboard_version72MessageCanceled' IS NOT NULL ON CONFLICT DO NOTHING;
         INSERT INTO Preferences (UserId, Category, Name, Value) SELECT Id, 'focalboard', 'lastWelcomeVersion', replace((Props->'focalboard_lastWelcomeVersion')::varchar, '"', '') FROM Users WHERE Props->'focalboard_lastWelcomeVersion' IS NOT NULL ON CONFLICT DO NOTHING;
-
+        -- make sure all props actually contain a json hash and not some other datastructure
+        UPDATE Users SET props = '{}'::json WHERE props::text NOT LIKE '{%';
```
oetiker, line 13:
```suggestion
        UPDATE Users SET props = '{}'::json WHERE jsonb_typeof(props) != 'object';
```

harshilsharma63 (approved):
Awesome! LGTM 👍

harshilsharma63:
Corresponding cherrypick PR - https://github.com/mattermost/focalboard/pull/4040

PR #4026, opened 2022-10-17
MM-47238 Add MM_BOARDS_DEV_SERVER_URL environment variable for MPA

hmhealey:
This makes it so that people who develop MM on a URL other than `http://localhost` can redefine the URL, port, and protocol that the Boards Webpack dev server runs on for MPA development. Without it sets, it still defaults to running the dev server on `http://localhost:9006`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47238

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/21435
https://github.com/mattermost/mattermost-webapp/pull/11381

#### Release Note
```release-note
NONE
```

PR #4045, opened 2022-10-19
MM-47562 Serve images from the correct URL in product mode

hmhealey:
This conflicts with https://github.com/mattermost/focalboard/pull/4026, but I didn't want it to have to depend on that one since this one's needed for 7.5 and the other one isn't.

The gist of this change is that line 119 sets the path for static images in product and non-product modes, and then line 206 overrides that only if we're serving the product for development. I would've liked to not set it and then override it below, but I couldn't come up with a better way to keep the webpack dev server setup stuff all together that I was happy with.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47562

### DIFF
--- a/mattermost-plugin/webapp/webpack.config.js
+++ b/mattermost-plugin/webapp/webpack.config.js
@@ -116,7 +116,7 @@ const config = {
                 type: 'asset/resource',
                 generator: {
                     filename: '[name][ext]',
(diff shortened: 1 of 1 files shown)


wiggin77 (approved):
LGTM :+1:

PR #4055, opened 2022-10-20
MM-46392 - load boards memberships on rhs component mount

pvev:
#### Summary
For the new webapp functionality that allows to create a new channel along with a board, it was noticed that when the right hand side component was opened/toggled, the just created board was not present on the list. The only way to see it was after reloading the browser (See attached videos). This PR add code to load current user boards memberships and update the global state on RHS component mount event, so when pulling the information about my sorted boards the just created board associated with the just created channel is shown on the RHS plugin without needing to reload the page. Also added some logic to prevent the component blink when first opening.

Before:

https://user-images.githubusercontent.com/10082627/196929423-281dbfbf-85a6-47a6-81fb-cb89ff87f951.mp4

After:

Load the board but blinked between the "no boards associated" and the "boards associated" component states

https://user-images.githubusercontent.com/10082627/196929555-fb20d181-462c-4e3d-86e9-fed8d0c7da7d.mp4

Preventing the blink:

https://user-images.githubusercontent.com/10082627/196954975-936826f0-8750-4be1-bf91-b4003fccebfc.mp4

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46392


mattermost-plugin/webapp/src/components/rhsChannelBoards.tsx
```diff
@@ -78,6 +84,11 @@ const RHSChannelBoards = () => {
     if (!currentChannel) {
         return null
     }
+
+    if (!dataLoaded) {
```
pvev, line 88: this prevents the component to blink when changing from zero state to listing the boards state.

jespino (approved):
LGTM

jespino (changes requested):
There are some snapshot testing failing, please take a look and update them accordingly.

mattermost-plugin/webapp/src/components/rhsChannelBoards.tsx
```diff
@@ -41,10 +41,16 @@ const RHSChannelBoards = () => {
     const me = useAppSelector<IUser|null>(getMe)
     const dispatch = useAppDispatch()
     const intl = useIntl()
+    const [dataLoaded, setDataLoaded] = React.useState(false)
 
     useEffect(() => {
-        dispatch(loadBoards())
-        dispatch(fetchMe())
+        async function loadRhsRequiredData() {
+            await dispatch(loadBoards())
```
jespino, line 48: I would prefer here to use something different, something like an `await Promise.all(...)` and then the `setDataLoaded(true)`. Also, I don't love creating an asyncrhonous function just to make it async. you can still use something like:

```ts
useEffect(() => {
    Promise.all([dispatch(loadBoards()), dispatch(loadMyBoardsMemberships()),  dispatch(fetchMe())]).then(() => setDataLoaded(true))
}, [])
```

mattermost-plugin/webapp/src/components/rhsChannelBoards.test.tsx
```diff
@@ -16,6 +16,11 @@ import RHSChannelBoards from './rhsChannelBoards'
 jest.mock('../../../../webapp/src/octoClient')
 const mockedOctoClient = mocked(octoClient, true)
 
+jest.mock('react', () => ({
```
jespino, line 19: I'm not sure about this kind of solution to show the content. My perception is you are mocking here the wrong thing, my expectation is that the mocking of the client should be enough and just using `act` around the rendering should wait until the effect is executed.

mattermost-plugin/webapp/src/components/rhsChannelBoards.tsx
pvev: ok, thanks for the feedback!

mattermost-plugin/webapp/src/components/rhsChannelBoards.test.tsx
pvev: by "mocking of the client" do you mean mocking the dispatched requests? and then with act make sure the state change has been already executed and so the subsequent re-render?

jespino (approved):
now looks good to me. Thanks @pvev :tada:

Issue #4059, opened 2022-10-20
Bug: Panic due to GetMemberForBoard can return a null member.


sbishel:
The Community Mattermost server is encountering several panics in the focalboard plugin. This seems to have started in the  last day or so.

https://mattermost.atlassian.net/browse/MM-47528

PR #4060, opened 2022-10-20
Fix GetMemberForBoard from returning a null member

sbishel:
#### Summary
GetMemberForBoard can return a nil member if the member is not found and it meets no other membership criteria. This is due to the `err` variable being reset within the `if()` clause, but no member being created.

Updated to a more specific error variable, so it will hopefully not be overwritten in the future. The function is fragile and should be rewritten.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/4059
  Fixes https://mattermost.atlassian.net/browse/MM-47528

### DIFF
--- a/server/services/permissions/mmpermissions/mmpermissions.go
+++ b/server/services/permissions/mmpermissions/mmpermissions.go
@@ -24,8 +24,9 @@ type Service struct {
 
 func New(store permissions.Store, api APIInterface, logger mlog.LoggerIFace) *Service {
 	return &Service{
-		store: store,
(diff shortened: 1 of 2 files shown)


Pinjasaur (approved):
I'm pretty sure that's on me. Was fixing a linter complaint: [`ad1faef` (#3786)](https://github.com/mattermost/focalboard/pull/3786/commits/ad1faef7b1d37be53ae15d1f6357cda47c575de5)

wiggin77 (approved):

Issue #3938, opened 2022-10-05
Bug: Creating a board from a custom template adds the board to the same category as the origin board


wuwinson:
## Steps to reproduce the behavior

1. Go to a board under a custom category
2. From the sidebar, open the board's option menu and select `New template from board`
3. Go to the template picker and select the template
4. See error - the board gets created in the same category as the origin board. Only applies to the board creator. The board will get created correctly in the default Boards category for other users.

## Expected behavior

The board is added to the default "Boards" category.

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/194176503-9cffefe0-62ec-42c6-bdc0-d466ea64c25f.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.3 Bug Bash
 - Browser and OS: Chrome on Mac

shadybotros:
@wuwinson I can work on that if no one is already working on it (would be my first contribution)

wuwinson:
Awesome, thanks @shadybotros! I've assigned the bug to you!

PR #4066, opened 2022-10-22
[GH-3938] Fix: Duplicated board as template should be in the default category

shadybotros:
##### What
When a board in a custom category is duplicated as a template (e.g. using the "New template from board" menu option), the new template should be in the default Boards category. Currently, it is added to the same category as the source board.

##### How
* Server: Changed to not assign a category if duplicating as a template.
This change is tested via an integration test. I usually prefer to have unit tests in addition, but the current behaviour is integration-tested (and not unit-tested). I'm not sure what the intended testing strategy is for this project, so please let me know if you think a unit test would be useful here.
* Web app: The web app adds the duplicated board to the same category immediately (i.e. before receiving the update from the server), so I changed it to not do that if duplicating as a template.
The existing behaviour here is not tested and I would need some suggestions on how to do that if you think we should test that change.

I tested the change locally with the Mattermost Boards Plugin and it works as expected.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/3938

### DIFF
--- a/server/app/boards.go
+++ b/server/app/boards.go
@@ -183,9 +183,11 @@ func (a *App) DuplicateBoard(boardID, userID, toTeam string, asTemplate bool) (*
 		a.logger.Error("Could not copy files while duplicating board", mlog.String("BoardID", boardID), mlog.Err(err))
 	}
 
-	for _, board := range bab.Boards {
-		if categoryErr := a.setBoardCategoryFromSource(boardID, board.ID, userID, board.TeamID); categoryErr != nil {
-			return nil, nil, categoryErr
+	if !asTemplate {
+		for _, board := range bab.Boards {
+			if categoryErr := a.setBoardCategoryFromSource(boardID, board.ID, userID, board.TeamID); categoryErr != nil {
+				return nil, nil, categoryErr
+			}
 		}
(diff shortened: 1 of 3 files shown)


harshilsharma63:
> the new template should be in the default Boards category

Whats you're rational for this?

We've intentionally implemented it this way so that a duplicated board is exactly where your attention is - the original board in the sidebar. Previously a duplicated board did end up in the default category and users weren't able to easily detect where it went. We could have added a scroll to the duplicated board but it felt jumpy and a bit wierd.

cc @asaadmahmood

shadybotros:
> > the new template should be in the default Boards category
> 
> Whats you're rational for this?
> 
> We've intentionally implemented it this way so that a duplicated board is exactly where your attention is - the original board in the sidebar. Previously a duplicated board did end up in the default category and users weren't able to easily detect where it went. We could have added a scroll to the duplicated board but it felt jumpy and a bit wierd.

@harshilsharma63 Fair enough, then it's probably simpler to keep it in the same category.

And actually I misunderstood the issue. It's supposed to be about boards created using this new template, and not the template itself. I'll close this PR and create a new one instead.

Issue #4070, opened 2022-10-23
UX: Adding a new table row should focus on the new row's title


shadybotros:
## Summary

Steps to Reproduce:
1. Enter a table view
2. Click on `+ New` at the bottom OR update the last row and press enter
3. A new row is added
4. But: The title of the new row is not focused and you have to click on it to edit

Suggested Behaviour: The title of the new row should be focused so that I can edit right away

## How important this is to me and why

Importance: Medium

Why: When I create a new row, I usually want to edit the title right away and it helps to stay in the flow if I don't have to click on it every time.

shadybotros:
cc @wuwinson

shadybotros:
I'm working on it, should be a simple fix

PR #4071, opened 2022-10-23
[GH-4070] Fix focus on new table row

shadybotros:
##### What
Creating a new table row should focus on the new row's title so that the user can edit it right away.

##### How
The problem is that we reset the `cardIdToFocusOnRender` state too fast, so the new card doesn't see the value anymore when mounted and therefore doesn't focus. That's why I increased the timeout from 100ms to 300ms.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/4070

### DIFF
--- a/webapp/src/components/centerPanel.tsx
+++ b/webapp/src/components/centerPanel.tsx
@@ -197,6 +197,7 @@ const CenterPanel = (props: Props) => {
                     } else {
                         // Focus on this card's title inline on next render
                         setCardIdToFocusOnRender(block.id)
+                        setTimeout(() => setCardIdToFocusOnRender(''), 300)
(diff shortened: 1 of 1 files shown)


sbishel:
@jespino - Hoping you know a better way to handle rather than a setTimeout().

jespino (approved):
This is a bit a patch more than a solution, but yes, this can mitigate the problem until we find the right approach.

harshilsharma63 (approved):
LGTM 👍

harshilsharma63:
@ogi-m do you want to give this a pass?

ogi-m (approved):
LGTM 👍🏻

Issue #3496, opened 2022-08-01
Bug: Link with specific characters are double encoded


roukmoute:
## Steps to reproduce the behavior

1. Add a comment on a ticket and add a link with spaces or another custom character.
2. Try with : `https://example.com?title=August 1 - 2022` or `https://example.com?title=August%201%20-%202022`

For the first link, it is cut and it is not possible to use it entirely event if I try to do some markdown.

For the second link, the link is reinterpreted to `https://example.com?title=August%25201%2520-%25202022`.
![image](https://user-images.githubusercontent.com/2140469/182122318-f0a96551-77a8-4b5e-a07d-6a581b081b44.png)

## Expected behavior

If I click I must be able to redirect on the same link.

## Edition and Platform

 - Edition: Focalboard on Unraid
 - Version: 7.1.0
 - Browser and OS: Firefox

chenilim:
@wuwinson, please investigate the expected behavior. Thanks.

sbishel:
This can be fixed by unencoding, then encoding.

wuwinson:
Expected behavior:
1. `https://example.com?title=August 1 - 2022` should take user to URL prior to any spacing, in this case it should direct user to `https://example.com?title=August`
2. `https://example.com?title=August%201%20-%202022` should not encode the '%', and instead direct user to the URL as is.

roukmoute:
3: In a markdown link, `1.` must be encoded correctly

PR #4079, opened 2022-10-24
Fix double encoding

sbishel:
#### Summary
Fixes issue where links are encoded even thought they may already be encoded. Fix is simply decode prior to encoding. 

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/3496

### DIFF
--- a/webapp/src/utils.test.ts
+++ b/webapp/src/utils.test.ts
@@ -58,6 +58,10 @@ describe('utils', () => {
             expect(Utils.htmlFromMarkdown('[]("xss-attack="true"other="whatever)')).toBe(expectedHtml)
             window.openInNewBrowser = null
         })
+
+        test('should not double encode links href on the webapp', () => {
+            expect(Utils.htmlFromMarkdown('https://example.com?title=August%201%20-%202022')).toBe('<p><a target="_blank" rel="noreferrer" href="https://example.com?title=August%201%20-%202022" title="" onclick="">https://example.com?title=August%201%20-%202022</a></p>')
+        })
     })
 
     describe('countCheckboxesInMarkdown', () => {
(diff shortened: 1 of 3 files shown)


Pinjasaur (approved):
overall LGTM @sbishel, could you update the test or add a new test to assert the `title` too?

sbishel:
> overall LGTM @sbishel, could you update the test or add a new test to assert the `title` too?

Nope, it looks like that line can actually be removed. We have no way of setting it with this function. My test data just happens to use "title" as a query parameter, but they are unrelated.

Pinjasaur:
@sbishel, sorry didn't see the collision of the `title` query parameter and the `title` HTML attribute. I was talking about the attribute here. I _think_ it can be set using this syntax: https://www.markdownguide.org/basic-syntax/#adding-titles

sbishel:
@Pinjasaur - Nice catch. By encoding the title we end up with encoded values in the tooltip. Actually appears that "title" does not need to be encoded/decoded at all, fixed and added additional tests.

sbishel:
@Pinjasaur - I know you approved earlier, but please review again for latest changes.

webapp/src/utils.test.ts
```diff
@@ -58,6 +58,16 @@ describe('utils', () => {
             expect(Utils.htmlFromMarkdown('[]("xss-attack="true"other="whatever)')).toBe(expectedHtml)
             window.openInNewBrowser = null
         })
+
+        test('should encode title', () => {
```
Pinjasaur, line 62:
```suggestion
        test('should encode links', () => {
```

Pinjasaur (approved):
LGTM. The name of the first test confused me a bit; I provided an alternative but 0/5 if you take it or not.

Issue #4067, opened 2022-10-23
Bug: Fail to upgrade from v7.4.2 to v7.4.3 on focalboard server w/ MySQL


zhzy0077:
## Steps to reproduce the behavior

1. A working focalboard server on v7.4.2 with MySQL data store via Docker.
2. docker run v7.4.3 version of focalboard server

## Expected behavior
The server is able to start.

## Screenshots (optional)
```
error [2022-10-23 05:05:44.632 Z] Table creation / migration failed        caller="sqlstore/sqlstore.go:66" error="error running fix collations and charsets migration: error fetching collation: sql: no rows in result set"
fatal [2022-10-23 05:05:44.632 Z] server.NewStore ERROR                    caller="main/main.go:136" error="error running fix collations and charsets migration: error fetching collation: sql: no rows in result set"
  main.main
      /go/src/focalboard/server/main/main.go:136
  runtime.main
      /usr/local/go/src/runtime/proc.go:250
  runtime.goexit
      /usr/local/go/src/runtime/asm_amd64.s:1571
```
## Edition and Platform

 - Edition: Personal Server
 - Version: v7.4.3
 - Browser and OS: Docker on Linux

## Additional context (optional)

The failure seems coming from this PR https://github.com/mattermost/focalboard/pull/4002. By looking at the code, it checks `channel` table in MySQL database by
```
select table_collation from information_schema.tables
where table_name = 'Channels'
and table_schema=(SELECT DATABASE())
```
However I don't have `Channels` table in my MySQL database at all. Not sure if it's from mattermost servers.

cc @wiggin77

wiggin77:
TBH I didn't know personal server could use MySQL.  I thought it was Sqlite only.  I'll fix today.

PR #4080, opened 2022-10-24
Fix collation for personal server

wiggin77:
#### Summary
As of 7.4.3 the plugin fixes up any inconsistent collation or charset settings in the database (MySQL only).  It uses the Channels table to determine the current collation setting.  In the case of personal server there is no Channels table so this would fail. TBH I did not know personal server could use MySQL - I thought it was Sqlite only.  

The fix is to use a sensible default for personal server.

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/4067

### DIFF
--- a/server/services/store/sqlstore/data_migrations.go
+++ b/server/services/store/sqlstore/data_migrations.go
@@ -655,11 +655,11 @@ func (s *SQLStore) RunFixCollationsAndCharsetsMigration() error {
 	}
 
 	// get collation and charSet setting that Channels is using.
-	// when unit testing, no channels tables exist so just set to a default.
+	// when personal server or unit testing, no channels tables exist so just set to a default.
 	var collation string
 	var charSet string
 	var err error
-	if os.Getenv("FOCALBOARD_UNIT_TESTING") == "1" {
+	if !s.isPlugin || os.Getenv("FOCALBOARD_UNIT_TESTING") == "1" {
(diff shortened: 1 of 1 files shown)


wiggin77:
@wuwinson do we want to cherry pick this into a dot release?

sbishel (approved):

sbishel:
@wiggin77 - Is there a work around?

wiggin77:
> @wiggin77 - Is there a work around?

Unfortunately, no.  Personal server startup will error out when using MySQL on upgrade.

wuwinson:
@wiggin77 Is it possible to downgrade to the previous version?

wiggin77:
@wuwinson  We don't actually support downgrading in the migration scripts.  What will happen is, after the upgrade, most of the migrations will run but the last one will fail.  Downgrading the binary won't try to run the migrations again, but if the partial migration had anything incompatible with the old binary things won't work.

sbishel:
@wiggin77, @wuwinson - Sounds like we should cherry-pick to dot release?

wuwinson:
@sbishel @wiggin77 Yea agreed, otherwise Boards will be broken for them until the next release.

wiggin77:
Automated cherry-pick failed.  https://github.com/mattermost/focalboard/pull/4097 does it manually.

PR #4089, opened 2022-10-26
Disable patches on unmount on readOnly mode

jespino:
#### Summary
Disable patches on unmount on readOnly mode

#### Ticket Link
Fixes #4081
Fixes #4074

PR #4098, opened 2022-10-27
fix: convert `multiSelect` prop value to [] if required

### LINKED ISSUE #4092: Bug: search throwing runtime error
### Continued Search Problems ###
Hey everyone, not sure what the cause is this time, but anytime i try to use the search it just crashes and resets.. any help to getting it fixed would be appreciated... 
https://community.mattermost.com/boards/team/rcgiyftm7jyrxnma1osd8zswby/bz5upto7c6ty8bn4j57zjf4s4ch/v9t865zgc5p8s8rfa1ht8t5puba

_Issue created from a [Mattermost message](https://community.mattermost.com/_redirect/pl/j3ze9genr7fhmxzwbk4grchzpe) *by @michaelgamble*._

Pinjasaur:
#### Summary

@michaelgamble reported a bug where searching in a specific board was crashing the frontend. There was an instance of trying to call `.map` on a _string_ for the property value of a `multiSelect` type and it was expecting an array of strings.

My concern is I'm not sure _why_ the property value was a string and not an array of strings in the first place. I'd rather fix the root cause and not a symptom.

#### Ticket Link

fixes #4092

### DIFF
--- a/webapp/src/store/cards.ts
+++ b/webapp/src/store/cards.ts
@@ -354,7 +354,7 @@ function searchFilterCards(cards: Card[], board: Board, searchTextRaw: string): 
                     }
                 } else if (propertyTemplate.type === 'multiSelect') {
                     // Look up the value of the select option
-                    const options = (propertyValue as string[]).map((value) => propertyTemplate.options.find((o) => o.id === value)?.value.toLowerCase())
+                    const options = (Array.isArray(propertyValue) ? propertyValue : [propertyValue]).map((value) => propertyTemplate.options.find((o) => o.id === value)?.value.toLowerCase())
                     if (options?.includes(searchText)) {
                         return true
                     }


Pinjasaur:
@sbishel similar to #4089 I think this would be an excellent candidate to cherrypick to a patch release.

Rajat-Dabade (approved):
LGTM 👍

sbishel:
@Pinjasaur Is this ready to merge and cherry-pick?

Pinjasaur:
> Is this ready to merge and cherry-pick?

@sbishel should be good—pending reviews—only concern being:

> My concern is I'm not sure why the property value was a string and not an array of strings in the first place. I'd rather fix the root cause and not a symptom.

sbishel:
> > Is this ready to merge and cherry-pick?
> 
> @sbishel should be good—pending reviews—only concern being:
> 
> > My concern is I'm not sure why the property value was a string and not an array of strings in the first place. I'd rather fix the root cause and not a symptom.

Ok, let me know when you're satisfied. Thoughts off the top of my head would be a converted property. (ie. Select -> MultiSelect) I don't think our conversion code is very thorough. Most of the time we don't do anything, Select and MultiSelect being the exception. Also, I'm guessing this board has been around for a while, so there may be an old bug involved as well.

sbishel (approved):

Pinjasaur:
@sbishel let's merge it then, would rather get community fixed ASAP

can you handle the cherry pick? I'm not sure when you wanted to cut the patch (and assuming that's something we're still interested in doing)

Issue #2439, opened 2022-03-02
UX: Text box in table cell does not extend to fill the cell


chenilim:
## Steps to reproduce the behavior

1. Open a table with cards that have a text property
2. Make the text property column wider
3. Hover over a cell
4. The text input box has a limited width

## Expected behavior

The text input box should expand to fill the width of the cell.
Clicking anywhere in the cell should enter edit mode for the text

## Screenshots

<img width="351" alt="Screen Shot 2022-03-02 at 2 53 16 PM" src="https://user-images.githubusercontent.com/46905241/156463516-67cefd52-d4b4-4b2b-97d0-053b1b1cce5d.png">

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.14.0
 - Browser and OS: Chrome on Mac

adithyaakrishna:
Can I take this up?

PR #4099, opened 2022-10-28
Fix not filling cell

redhoyasa:
#### Summary

Making the text and URL fields stretch to the table cell in the table view.

Test steps:
1. Open a table with cards that have text and URL fields
2. Make the text field column wider
3. Make the URL field column wider

**Before**
Screenshot 2022-10-28 at 3 01 31 PM

Screenshot 2022-10-28 at 3 01 56 PM

**After**

https://user-images.githubusercontent.com/1910192/198535538-306259f4-55b4-4438-8893-0e6513d20dfc.mov

#### Ticket Link

Fix https://github.com/mattermost/focalboard/issues/2439


sbishel (approved):
Thanks @redhoyasa

webapp/src/widgets/editable.tsx
```diff
@@ -116,23 +116,14 @@ export function useEditable(
...
     const elementRef = useRef<HTMLInputElement>(null)
     const elementProps = useEditable(props, ref, elementRef)
 
     useLayoutEffect(() => {
         if (props.autoExpand && elementRef.current) {
             const input = elementRef.current
-            const computed = getComputedStyle(input)
-            input.style.width = 'auto'
-            input.style.width = `${input.scrollWidth + borderWidth(computed) + 1}px`
+            input.style.width = '100%'
```
kamre, line 126: @redhoyasa @sbishel this is really strange change: the whole point of using layout effect was that we need to compute the required width for input element. Just setting width to `100%` can be easily done via CSS.

Also this change leads to regression in the card dialog because the text properties including URL now have very wide input element:
image

Issue #3246, opened 2022-06-17
Bug: kanban card property tooltip is cut off


kamre:
## Steps to reproduce the behavior

1. Create `Project Tasks` board from template
2. Move mouse to `Estimated Hours` number property on the first card in the first column
3. Look at the tooltip that shows the property name
4. Tooltip is not fully visible

## Expected behavior
Tooltip is fully visible

## Screenshots (optional)
<img width="556" alt="image" src="https://user-images.githubusercontent.com/5587620/174343694-216ebb5b-a964-4ad8-9393-df063c8c4ef4.png">

## Edition and Platform

 - Edition: Personal Server
 - Version: `main` branch
 - Browser and OS: Chrome on Mac

## Additional context (optional)
Probably tooltip should be positioned in a more accurate way, for example main part with the text can be moved to the right.

* Sev 3: Affects minor, non-critical functionality

redhoyasa:
I'll work on this.

PR #4100, opened 2022-10-28
Fix tooltip cutoff

redhoyasa:
#### Summary

Fix the kanban card tooltip that cutting off. I consider the both tooltip with short text and long text.

<table>
<tr>
	<td>Before</td>
	<td>After</td>
<tr>
	<td>
	    Screen Shot 2021-10-23 at 13 14 38
	</td>
        <td>
            Screen Shot 2021-10-23 at 13 15 58
        </td>    
</table>

#### Ticket Link

Fix https://github.com/mattermost/focalboard/issues/3246.

### DIFF
--- a/webapp/src/widgets/tooltip.scss
+++ b/webapp/src/widgets/tooltip.scss
@@ -54,15 +54,14 @@ $tooltop-vertical-offset: 2px;
 
     // Top tooltip arrow style
     &.tooltip-top:hover::before {
-        left: 50%;
+        left: 0;
         bottom: calc(100% - 2px);
-        transform: translate(-50%, #{$tooltop-horizontal-offset});
     }
     // Top tooltip body style
     &.tooltip-top:hover::after {
         bottom: calc(100% + 10px);
-        left: 50%;
-        transform: translate(-10%, #{$tooltop-horizontal-offset});
+        left: -20%;
+        transform: translate(0, #{$tooltop-horizontal-offset});
     }
 
     // Right tooltip arrow style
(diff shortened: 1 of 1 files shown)


kamre:
Hi @redhoyasa, it seems that proposed fix changes tooltip for all cards, even in the second column of kanban board:

before applying the fix it looks like this:

redhoyasa:
@kamre thanks for the feedback. I try to make some changes. Does it look good to you?

Screenshot 2022-10-31 at 4 33 42 PM

kamre:
@redhoyasa this definitely looks better.

@asaadmahmood what do you think about such changes to tooltip?

asaadmahmood:
@kamre @redhoyasa I think those look good, but it wouldn't make sense to always have the tip on the left. If we can make the tip position dynamic, like center, left, or right depending on the space available, that would be much better.

Otherwise, I don't want to change all tooltips to have the tip on the bottom left.

redhoyasa:
@asaadmahmood @kamre that makes sense. Depending on available space could work but I'd like to propose something like this:
1. For card within the left most list, the position would be on left
2. For card within the right most list, the position would be on right
3. Otherwise, the tooltip will be on the center

I think that way is easier to reason, what do you think?

asaadmahmood:
@redhoyasa That can be done though important to note that it's not always about the left most or right most list.
It can be any list, as we allow scrolling.

So imagine you have 12 lists, and you hoirzontally scroll, so now the 3rd or 4th list can be the first list (visually speaking).

So even if we do it, it should focus more on the list position visually rather than its order.

asaadmahmood (changes requested):
Stale

Issue #2068, opened 2022-01-07
Bug: Description doesn't directly enter edit mode when adding


sbishel:
## Steps to reproduce the behavior

1. Go to a Card
2. Select  'Add Content' in the description field
2a, Select 'text'
3. Adds a line with 'Edit text'
4. Must click again to begin editing.

## Expected behavior

Expect text to be editable and cursor at beginning of line. 
What happens after you hit "Edit text"

## Screenshots

If applicable, add screenshots to elaborate on the problem.

## Edition and Platform

 - Edition: Mattermost plugin
 - Version: v0.12.0
 - Browser and OS: [Chrome on macOS]

sridhar02:
@chenilim can I work on this?

michkrej:
Hi! Can I work on this issue?

sbishel:
@michkrej - Sure, all yours.  Thanks.

michkrej:
Wasn't sure if I should make a pull request against my forked main-branch or against the mattermost-one. But there is a pull request up now!

Adilahmed03:
Hi, I’d like to work on this issue if it’s available. Could you please assign it to me?

PR #4102, opened 2022-10-28
[GH-2068] [text] doesn't directly enter edit mode when adding

michkrej:
#### Summary

When adding a text block in a card it enters edit-mode straight away instead of having to click on it.

#### Ticket Link

 Fixes https://github.com/mattermost/focalboard/issues/2068

### DIFF
--- a/webapp/src/components/__snapshots__/addContentMenuItem.test.tsx.snap
+++ b/webapp/src/components/__snapshots__/addContentMenuItem.test.tsx.snap
@@ -40,6 +40,46 @@ exports[`components/addContentMenuItem return a checkbox menu item 1`] = `
 </div>
 `;
 
+exports[`components/addContentMenuItem return a checkbox menu item 2`] = `
+<div>
+  <div
+    aria-label="checkbox"
+    class="MenuOption TextOption menu-option"
+    role="button"
+  >
+    <div
+      class="d-flex"
+    >
+      <div
+        class="menu-option__icon"
+      >
+        <svg
+          class="CheckIcon Icon"
+          viewBox="0 0 100 100"
+          xmlns="http://www.w3.org/2000/svg"
+        >
+          <polyline
+            points="20,60 40,80 80,40"
+          />
+        </svg>
+      </div>
+    </div>
+    <div
+      class="menu-option__content"
+    >
+      <div
+        class="menu-name"
+      >
+        checkbox
+      </div>
+    </div>
+    <div
+      class="noicon"
+    />
+  </div>
+</div>
+`;
+
 exports[`components/addContentMenuItem return a divider menu item 1`] = `
 <div>
   <div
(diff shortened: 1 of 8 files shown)


michkrej:
@sbishel haven't worked with snapshots before, am I allowed to update the snapshots for the test to pass?

kamre:
Hi @michkrej, you can update the snapshots via running: `npm run updatesnapshot`,  after this you can inspect the changes in snapshots and confirm that they are expected ones (i.e. test for adding text actually enters the edit mode and produces corresponding dom tree).

michkrej:
Thanks @kamre! Will take a look at this right away

michkrej:
@sbishel everything works when I test I locally now!

sbishel:
Hi @michkrej - I couldn't get this to work locally. Am I missing something?

Uploading Screen Recording 2022-11-04 at 9.20.47 AM.mov…

michkrej:
Hi @sbishel! 

Can't see your video for some reason, but this is how it looks for me. I don't think that I've done anything special.

Running the frontend trough `npm run pack` (tried watchdev too, works) and the backend trough  `./bin/focalboard-server`
as described in the setup guide. 

https://user-images.githubusercontent.com/25232126/200039763-ccbd2900-456c-45eb-b853-fcf51286b53f.mov

sbishel:
Apparently there is a difference between running as a plugin vs Personal Server. It does work when I run it as personal server, but not as Plugin. Not sure why. I don't have time investigate right now. 

But when running as Personal Server, it only works when adding as the last text block. If you attempt to insert in the middle of the comments, it either doesn't work or will activate the last block for edit rather than the added block.

michkrej:
That is strange, I'll look at the second part first since I know where to get started. Won't be done until sometime next week though!

kamre:
Hi @michkrej @sbishel,  regarding this:
> If you attempt to insert in the middle of the comments, it either doesn't work or will activate the last block for edit rather than the added block.

it looks like [adding new content blocks inside AddContentMenuItem](https://github.com/mattermost/focalboard/blob/a30c1692ac8d4dd447a54930a6f4a198af0bf51e/webapp/src/components/addContentMenuItem.tsx#L39-L59) doesn't go through the same code path as [adding inside CardDetailContentsMenu](https://github.com/mattermost/focalboard/blob/a30c1692ac8d4dd447a54930a6f4a198af0bf51e/webapp/src/components/cardDetail/cardDetailContentsMenu.tsx#L22-L27)

This can be fixed reusing the code from `CardDetailContentsMenu` like this:
```ts
    const cardDetail = useCardDetailContext()
    const addElement = useCallback(() => {
        cardDetail.addBlock(handler, index, false)
    }, [cardDetail, handler])

    return (
        <Menu.Text
            key={type}
            id={type}
            name={handler.getDisplayText(intl)}
            icon={handler.getIcon()}
            onClick={addElement}
        />
    )
```

michkrej:
Thanks for figuring out this for me @kamre! I'll look at implementing this now :)

michkrej:
@sbishel managed to fix the issue with it only working when adding it as the last block. As @kamre suggested I switched out the function in onClick on AddContentMenuItem to addElement instead, just as in CardDetailContentsMenu. From my own testing it everything seems to work fine, but I'm not sure if this change could have any other consequences I didn't notice. 

Regarding why it doesn't work when running the app as a plugin I can't seem to build the plugin, so I have no idea why it doesn't work there.

michkrej:
@sbishel still can't setup the project as a plugin. Any tips on this?

michkrej:
@sbishel still not able to run the plugin version. Tried following the instructions in the documentation but I only encounter errors. Won't be able to proceed with this currently.

PR #4110, opened 2022-10-31
Fix to respect System Console settings

sbishel:
#### Summary
Fix to respect full name and email configuration settings. We currently don't return email in the user json, but implemented it as well when using plugin.

Did not implement on the Personal server side as we don't set/retrieve `LastName` and `FirstName` and as stated above, we don't return the `Email` in the user json.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48009

### DIFF
--- a/mattermost-plugin/server/boards/boardsapp_test.go
+++ b/mattermost-plugin/server/boards/boardsapp_test.go
@@ -46,6 +46,12 @@ func TestSetConfiguration(t *testing.T) {
 		TeammateNameDisplay: &usernameRef,
 	}
 
+	falseRef := false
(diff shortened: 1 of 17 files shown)


wiggin77 (approved):
LGTM :+1:   My only nit is that no tests have the true, true values.

sbishel:
> LGTM 👍 My only nit is that no tests have the true, true values.

Agreed, but because it wasn't implemented in Personal Server. Setting the tests to true, true values would have no effect. And we, unfortunately, have any tests for `mattermostauthlayer` - (which I just created a card for)

PR #4116, opened 2022-11-01
MM-47793 Change product dev server to not use deprecated options

hmhealey:
Webpack complains about using the `https` option instead of the `server` one, so I changed that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47793


wiggin77 (approved):
LGTM :+1:

PR #4117, opened 2022-11-01
fix: don't send multiple `viewBoard` telemetries

### LINKED ISSUE #4111: Bug: Loading a board triggers multiple "viewBoard" Telemetry events

1. Open developer console and go to the "Network" tab
2. With the console open, load any board
3. See error - multiple "viewBoard" events are triggered for any single board

## Expected behavior

Only 1 "viewBoard" event is triggered each time a board is loaded

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/199080285-10a19f3d-c8b7-4d8d-a455-2d6f1d265308.mp4

## Edition and Platform

 - Version: v7.4.3

Pinjasaur:
#### Summary

There was 2 instances of the `viewBoard` telemetry event being emitted. One of them was _also_ in a `useEffect` with _no_ dependency array so it was being called upon mounting _plus_ after every render.

This fixes the behavior so the telemetry event is only emitted _once_ per viewing the board.

#### Ticket Link

fixes #4111

### DIFF
--- a/webapp/src/components/centerPanel.tsx
+++ b/webapp/src/components/centerPanel.tsx
@@ -86,6 +86,8 @@ const CenterPanel = (props: Props) => {
     const currentCard = useAppSelector(getCurrentCard)
     const dispatch = useAppDispatch()
 
+    // empty dependency array yields behavior like `componentDidMount`, it only runs _once_
+    // https://stackoverflow.com/a/58579462
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


harshilsharma63 (approved):
LGTM 👍

harshilsharma63:
@Pinjasaur @wuwinson @sbishel do we wanna pick this up for the patch release?

wuwinson:
> @Pinjasaur @wuwinson @sbishel do we wanna pick this up for the patch release?

If there's still time to get it in the patch release, then yea that would be great, but otherwise we can hold off until v7.5

wuwinson:
cc @sri-byte @henry-shxu

Issue #3881, opened 2022-09-29
Bug: Person property edit cursor behind avatar


wiggin77:
## Steps to reproduce the behavior

1. Create card with a `Person` property.  
2. Select a person to populate it
3. Edit the Person property by clicking on it
4. See the cursor appearing behind the avatar of the person selected in step 2

## Expected behavior

The fact that people who are not members of the board do not appear in the auto-complete drop-down is confusing IMO.  With the cursor hidden there is no way for a user to know how to add a team member who is not a member of the board.

## Screenshots (optional)
![person-prop-bugs](https://user-images.githubusercontent.com/5024605/193115733-246a251c-6de8-4f05-a4b0-a9fda4b14397.gif)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: 7.3 bugbash 
 - Browser and OS: Linux FF

## Additional context (optional)

Add any other context about the problem here, and any notes about the severity:
* Sev 1: Affects critical functionality without a workaround
* Sev 2: Affects major functionality with a difficult or non-obvious workaround
* Sev 3: Affects minor, non-critical functionality

Pinjasaur:
Repros on Mac FF too.

![image](https://user-images.githubusercontent.com/6335792/193120103-538d7bcd-b3de-4e8d-8b61-5db37ad5a2bf.png)

sbishel:
@wiggin77 - Could you retest on Linux and see if still an issue. I tried on https://boards-bugbash74.test.mattermost.cloud/ and could no longer duplicate either on MAC (chrome nor FF).

wiggin77:
> @wiggin77 - Could you retest on Linux and see if still an issue. I tried on https://boards-bugbash74.test.mattermost.cloud/ and could no longer duplicate either on MAC (chrome nor FF).

Still happening for me on bugbash74.  FF 106.0.1 Linux

![caret-bug](https://user-images.githubusercontent.com/5024605/197208203-986243d0-d1e3-4550-a9f0-45f527bde4ba.gif)

sbishel:
Thanks.  Also occurs on the normal "Select" property.

PR #4120, opened 2022-11-02
Minor CSS fix

Rajat-Dabade:
#### Summary

Minor UI fixes

Before:

The focus was going behind the person's avatar.

https://user-images.githubusercontent.com/26198062/199473719-f2fe9489-7e0b-4461-a68c-4d27861ef524.mov

After: 

Added some margin for the person's avatar and name.

https://user-images.githubusercontent.com/26198062/199473838-1bea02d9-3113-4a11-be31-a16c85da8cef.mov

#### Reason
The default behaviour for component `Select` in `react-select` is like below. The focus by default will be at the beginning of the `input` field.

https://user-images.githubusercontent.com/26198062/199474499-e7a19729-84ca-48d5-b298-c131fce75557.mov

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/3881

### DIFF
--- a/webapp/src/properties/person/person.scss
+++ b/webapp/src/properties/person/person.scss
@@ -14,6 +14,7 @@
         display: flex;
         align-items: center;
         color: rgba(var(--center-channel-color-rgb), 1);
+        margin-left: 10px;
 
         img {
             border-radius: 50px;


Rajat-Dabade:
cc @sbishel @wiggin77

wiggin77:
Shouldn't the caret be placed after the existing people when editing?

asaadmahmood:
@Rajat-Dabade The hover state and the default state alignment is not correct. Even if we're adding this, we need the profile icon to start from the same place as the other items in the list.
Screenshot 2022-11-02 at 5 16 46 PM

Rajat-Dabade:
@wiggin77, when we start typing the existing selected value will disappear. If we keep the caret at the end of the already selected value, it would not look that great because of the disappearing nature (The caret will go at the beginning). If we take the example of the multi-select, the caret is always kept at the end so that we can add more value. It's just to make a difference between the `select` and `multi-select` fields. Thought?

Rajat-Dabade:
@asaadmahmood, after suggested changes


https://user-images.githubusercontent.com/26198062/199977605-5f814cd8-8378-4df1-a428-5f64132aed24.mov

Screenshot 2022-11-04 at 6 24 48 PM

asaadmahmood (approved):
LGTM! Nice work.

PR #4133, opened 2022-11-03
Fix a bug where board didn;t show up in LHS when someone added me on it

### LINKED ISSUE #4130: Bug: Shared Boards are not added to the user's sidebar

1. As User A create a board
2. Click "Share" and add User B to the board
3. As User B observe the sidebar
4. The board is not added to the user's sidebar (even after a refresh)

## Expected behavior

Boards that are shared with the user should appear in the sidebar

## Edition and Platform

 - Version: v7.5
 - Browser and OS: Chrome

harshilsharma63:
Fixes #4130

### DIFF
--- a/server/app/boards.go
+++ b/server/app/boards.go
@@ -519,6 +519,10 @@ func (a *App) AddMemberToBoard(member *model.BoardMember) (*model.BoardMember, e
 		return nil, err
 	}
 
+	if err = a.addBoardsToDefaultCategory(member.UserID, board.TeamID, []*model.Board{board}); err != nil {
+		return nil, err
+	}
+
 	a.blockChangeNotifier.Enqueue(func() error {
 		a.wsAdapter.BroadcastMemberChange(board.TeamID, member.BoardID, member)
 		return nil
--- a/server/app/boards_test.go
+++ b/server/app/boards_test.go
@@ -27,6 +27,7 @@ func TestAddMemberToBoard(t *testing.T) {
 		}
 
 		th.Store.EXPECT().GetBoard(boardID).Return(&model.Board{
+			ID:     "board_id_1",
 			TeamID: "team_id_1",
 		}, nil)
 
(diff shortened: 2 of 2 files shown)


jespino:
This looks like error prone to me, there is any reason to not have the "default category" applied to any board without category?

sbishel:
Please...let's add a summary to PRs.

harshilsharma63:
> This looks like error prone to me, there is any reason to not have the "default category" applied to any board without category?

@jespino to maintain a system invariant that all boards belong to a category. We always need boards to belong to a category to make re-arranging boards in a category simple.Also trying to replicate the channel's behavior. I'll add an "unknown" category on the client side for any boards that exist without any category but that would only be for the porpose of us knowing that we're missing assigning a category to a board somewhere.

harshilsharma63:
On second thought, I think we should not show category-less boards in LHS. It's not like the user loses access to the board or something. They can search for the board, have a bookmarked link, or a dozen other ways to access the boards. Even with the "unknown" category idea, when (and if) someone reports that as a bug, we'll anyways be asking them how they they gain access to the board, as in, did someone add you to it or something else and then figuring out where we missed adding the board to category. Either way, we'll flush out the bugs and either way user doesn't loose access to the board.

Thoughts @sbishel @asaadmahmood @wuwinson ?

jespino:
My concern is that you are saying that this is an invariant, but at the same time, you are accepting that invariant can be violated, and we have a way in the database to represent that invalid state. So... In my opinion the invariant has to be something different. Something like "Any board that doesn't belongs to any board have the meaning X", that is you are defining in your last comment. But I disagree that the definition that you propose is the expected one by the user. I think we should show everything, that you have explicit access to it, unless you hide it. The thing is, you are getting that weird state as something that can happen and is uncommon, but the reality is that in case that that happens is going to fail sinlently instead on make it clear something is going wrong.

For example, now we have fixed one of the corner cases with this, but is going to be really hard to detect other corner cases because we are not seeing it happening, and people are going to not notice it, or the fact that they were added wrongly for some way.

For me this approach is error prone and leads to fail silently instead of fail in a way that we can detect and correct.

harshilsharma63:
@jespino so do you recommend we don't consider this as an invariant and allow boards to exist both in and out of categories?

jespino (approved):
LGTM

Issue #3541, opened 2022-08-03
Bug: Docker-compose gives permission denied on `data` directory


wiggin77:
## Steps to reproduce the behavior

1. Go to `docker` folder in `focalboard` repo
2. Type `docker-compose -f docker-compose-db-nginx.yml up`
4. Boards is not able to write images to the `data` directory created. See error:

```
focalboard       | error [2022-08-03 22:37:05.318 Z] InitializeTemplates failed               caller="app/initialize.go:13" error="cannot initialize global templates for team 0: cannot import file 76fwrj36hptg6dywka4k5mt3sph.png for board bbn1888mprfrm5fjw9f1je9x3xo: unable to create the directory /opt/focalboard/files/0/b1gx7eisg4jdtfmkcwbh1qmpche for the file files/0/b1gx7eisg4jdtfmkcwbh1qmpche/76fwrj36hptg6dywka4k5mt3sph.png: mkdir files: permission denied"
```

## Expected behavior

Boards is able to write images to local storage.

## Edition and Platform

 - Edition: Personal Server (docker)
 - Version: 7.3 (main)
 - Browser and OS: Linux

wiggin77:
This was broken by https://github.com/mattermost/focalboard/pull/552

PR #4148, opened 2022-11-04
GH-3541 docker data permissions & cleanup

wiggin77:
#### Summary
This PR simplifies the different Docker configurations while maintaining best practices. The README is also updated to reflect changes.  The goal is to have the Docker configurations just work, without necessarily reading the README. 

- continue using user `nobody`

- switch from bind mounts to named volumes
	- ["Volumes are the preferred way to persist data in Docker containers and services."](https://docs.docker.com/storage/#good-use-cases-for-volumes)
	- when combined with user `nobody` bind mounts required users to pre-create a data directory and set permissions for `nobody`, otherwise a `permission denied` error resulted on start-up

- switch OS image from `distroless/base-debian` to `buster-slim`
	- the previous OS was very small, but didn't even contain a shell which is needed for things like mkdir
	- `Buster-slim` is also quite small but includes a shell. This is what Mattermost server uses.

- fix data paths in app config

Tested for these configurations: 

- `docker build ...` , `docker run ...`
- `docker-compose up`
- `docker-compose -f docker-compose-db-nginx.yml up`

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/3541


jespino (approved):
LGTM

Issue #4128, opened 2022-11-03
Bug: Text filter is capital sensitive


mgdelacroix:
## Steps to reproduce the behavior

1. Create a card with title "Test"
2. Add a filter for titles that include "test"
3. See how your card is not included
4. Change the filter for titles that include "Test"
5. See how now the card is included

## Expected behavior

The text filter should be case insensitive.

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.5-RC
 - Browser and OS: Firefox 105.0.3 on Linux

wuwinson:
Still seeing this on https://boards-master.test.mattermost.cloud/ with the v7.5 release branch. Re-opening.

Rajat-Dabade:
I can able to see the fix @wuwinson. 

https://user-images.githubusercontent.com/26198062/200754524-493c59b4-ab69-44da-a171-994e34ca784e.mov

Can you tell me by which data you are testing? Thanks.

wuwinson:
Thanks for checking @Rajat-Dabade! I'm testing filters by any of the following property types:
- Text
- Phone
- Email
- URL

See screen recording for reference:
https://user-images.githubusercontent.com/93531870/200871515-eb36f9dd-b497-4b9e-a6bd-d0c1baaa05fe.mov

Rajat-Dabade:
@wuwinson, I guess this is another issue, as the filter with type `text` should not contain `include`, it should have `is`, `contains`, `doesn't contain` etc. This happens only when you change from another property to `text` property. 


https://user-images.githubusercontent.com/26198062/201852172-8c2d6e47-216e-4b3b-8eda-24be41cfea2f.mov

wuwinson:
Separate bug filed for https://github.com/mattermost/focalboard/issues/4206, closing this out.

PR #4153, opened 2022-11-07
Case-insensitive filter for text

Rajat-Dabade:
#### Summary
This PR will add the ability to filter the text case-insensitive way.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/4128


webapp/src/cardFilter.ts
```diff
@@ -77,43 +81,43 @@ class CardFilter {
             if (filter.values.length === 0) {
                 return true
             }
-            return filter.values[0] === value
+            return filterValue === value
```
mgdelacroix, line 84: nit: we can take advantage of TS here and use directly:
```suggestion
            return filter.values?[0].toLowerCase() === value
```

mgdelacroix (approved):
LGTM! :+1: Left one nonblocking nit

webapp/src/cardFilter.ts
Rajat-Dabade: Sure, we can directly use `filter.values[0].toLowerCase()` because the above `if` is checking if there is at least one element in the array.

webapp/src/cardFilter.ts
```diff
@@ -77,43 +77,43 @@ class CardFilter {
             if (filter.values.length === 0) {
                 return true
             }
-            return filter.values[0] === value
+            return filter.values[0].toLowerCase() === value
```
mgdelacroix, line 80: We should add the safety operator if we access the values directly:

```suggestion
            return filter.values?[0].toLowerCase() === value
```

Same for all the other accesses

mgdelacroix (changes requested):

webapp/src/cardFilter.ts
Rajat-Dabade: Okay

mgdelacroix (approved):

Issue #2932, opened 2022-04-26
Feature Idea: Disable view dropdown selector on public "read-only" views


wuwinson:
## Summary
The view dropdown selector should be disabled on public "read-only" views so users can't switch views.

## How important this is to me and why

Importance: Medium

Use cases:
Some users would like to share not only a board but also a specific view. Think about an internal & public roadmap. There are some tasks that a user may not want to show up on the public roadmap (like internal tools) etc. that can be filtered out on another view, but if the user shares the board, the public can switch between both views and also see the internal tasks.

wuwinson:
See #2938 for reference.

wuwinson:
Additional improvements to view-only public boards: Option to prevent viewers from opening cards.

See this [thread](https://community.mattermost.com/core/pl/n9siwxekm3dimnmgorqr1yufra) for reference.

redhoyasa:
I have some questions:
1. Since the permission is attached to the board (not the view), any user that has the URL for other views can still open the other views using the URL directly right?
2. What if the board owner wants to share all the views with "read-only"? In this case, do you expect the dropdown selector still be disabled?

wuwinson:
@redhoyasa Thanks for your questions! Answers below:

1. Correct!
2. The common use case for sharing public read-only boards is to communicate specific information to an external audience, and not all the information from the board, thus the reason for wanting to restrict access to a single view. If the board owner wants to share all views publicly, I'd expect the dropdown to still be disabled and to have all views shared as individual links. In the case where a board owner wants to share all the views to internal collaborators with "read-only" access, they can add the user to the board and assign them the "Viewer" role.

Hope that clears things up, but let us know if you have any other questions!

redhoyasa:
@wuwinson thanks for the answer!

To summarize:
1. in the public "read-only" view, the dropdown is disabled (or hidden? which one do you prefer?)
2. if the user is directly assigned the "Viewer" role, the dropdown is available for that user

Is that correct?

wuwinson:
@redhoyasa Yep, that's right!

1. Dropdown should be disabled but not hidden, so the viewers can still see the name of the view
2. Dropdown is still available for users assigned the "Viewer" role (no changes from current behavior)

LMK if you'd like to pick up this ticket, and I'll assign to you. Thanks!

redhoyasa:
Yes I would like to pick this one.

wuwinson:
Awesome, thanks @redhoyasa!

[...]

PR #4154, opened 2022-11-07
Disable the view dropdown selector on public "read-only" 

redhoyasa:
#### Summary

Disable the view dropdown selector on public "read-only" so the complete list of views won't be shown.

**Before**
Screenshot 2022-11-07 at 10 10 03 PM

**After**
Screenshot 2022-12-09 at 4 47 46 PM

#### Ticket Link

Fix https://github.com/mattermost/focalboard/issues/2932


Rajat-Dabade:
Changes look good to me @redhoyasa, Nice work 🚀 .

Though I think the dropdown button to change the view is still clickable, which should not be. Can you please make it disabled (`cursor: not-allowed` maybe)?

Also, I can see we have some test-case failures, mostly snapshots because of your changes.  You can run 
`npm run test -- -u` in the `webapp` directory to fix it. 

Thanks

wuwinson:
Awesome work @redhoyasa! Would it be possible to hide the dropdown button altogether, but leave the view name visible? It occurred to me that users might think they are encountering a bug if they see a dropdown button, but won't be able to click on it. If not, then disabling the button works too. Thanks!

redhoyasa:
Hi @wuwinson, sorry for waiting for a long time. I already removed the dropdown icon. Could you please review it again?

Thanks!

wuwinson:
Awesome work @redhoyasa! Looks great to me! @Rajat-Dabade Can you review and approve if everything looks good on your end? Thanks all!

Screen Shot 2022-12-09 at 5 51 03 PM

wuwinson:
Thanks again @redhoyasa @Rajat-Dabade!

Issue #4136, opened 2022-11-03
Bug: Cannot rename sidebar categories


wuwinson:
## Steps to reproduce the behavior

1. Go to **Option Menu (...)** on any custom category on the sidebar
2. Select "Rename Category"
3. Enter a new name
4. Click "Update"
5. See error - category is not renamed

## Expected behavior

Category is renamed

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/199806362-d0cd45e6-96ba-493e-869b-c5c86da8fe1b.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.5
 - Browser and OS: Chrome on Mac

PR #4161, opened 2022-11-08
Fix for category not being set by UI, set to existing Category.

sbishel:
#### Summary
The category type does not get set on an Update and therefore fails the `IsValid()` check. 

Set the type to default to pass the `IsValid()` check, then set to existing category.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/4136

### DIFF
--- a/server/app/category.go
+++ b/server/app/category.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"errors"
+	"strings"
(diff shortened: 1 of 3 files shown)


Rajat-Dabade (approved):
LGTM 👍

PR #4186, opened 2022-11-09
Set down migrations as noops and add CI job

mgdelacroix:
#### Summary
This PR makes down migrations consistently noops, so it's explicit that boards is not currently supporting them and adds a CI job to ensure future migrations are added with a noop down script.


.github/workflows/lint-server.yml
```diff
@@ -8,10 +8,22 @@ on:
...
+  down-migrations:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          path: "focalboard"
+      - name: assert that down migrations are SELECT 1 scripts
+        run: |
+          cd focalboard
+          echo 'SELECT 1;' > downmigration
```
Pinjasaur, line 24: @mgdelacroix should the `downmigration` file be `rm`-ed afterwards? or is the filesystem not persisted between jobs? my concern would just be keeping the working directory clean

jespino (approved):
LGTM

wiggin77 (approved):
This looks great!  I like the CI check for correct file contents (ignoring whitespace).

wiggin77:
Weird that the PR is marked "Do Not Merge" yet the UI allows merge.

.github/workflows/lint-server.yml
mgdelacroix: Each job is fully isolated AFAIK, which would be consistent with each job doing its own `checkout`

PR #4207, opened 2022-11-15
Optimised Sorting of comments inside cards

Rajat-Dabade:
#### Summary
The PR will optimise the sorting of the comments inside the card.


jespino (approved):
Makes sense. Good catch.

PR #4211, opened 2022-11-15
fix: `toString` `propertyValue` for lowercasing

### LINKED ISSUE #4150: Search failed with a new item
### Discussed in https://github.com/mattermost/focalboard/discussions/4149

<div type='discussions-op-text'>

<sup>Originally posted by **Alderaaan** November  5, 2022</sup>
Hi,
I created a new table, and when I search some words, the page appear to be white.
When I use a default template, It's work.

What I missed ?
Tks

Focalboard on docker.
**Docker install :** mattermost/focalboard:edge

[screen-recording(1).webm](https://user-images.githubusercontent.com/21173299/200119841-358ecd87-3bd5-4939-b282-3dd39b889336.webm)
</div>

Pinjasaur:
#### Summary

Near identical to #4092, just slightly farther down the code path. `.toLowerCase()` was the culprit this time, so I updated the code to call `.toString()` which would effectively be idempotent on a string or array so no need to check the type at runtime to conditionally call it.

#### Ticket Link

closes #4150


Pinjasaur:
I'm not sure if we're considering any patch releases but I think this would be a good candidate.

wiggin77 (approved):

wiggin77:
@wuwinson do you want this in a dot release?

wuwinson:
@wiggin77 @Pinjasaur Agreed, this one looks like a good candidate for a dot release. cc @sbishel

jespino (approved):
LGTM

Issue #4210, opened 2022-11-15
Bug: Servers are no longer supported on Ubuntu 18.04


amyblais:
See this thread: https://community.mattermost.com/private-core/pl/48an6do9rbyh3kggbh7iqntmpy  Ubuntu 18.04 no longer seems to be supported due to recent MPA changes. Instead of bumping our supported OS to Ubuntu 20.04, we want to fix this issue so that Ubuntu 18.04 continues to be supported.

amyblais:
The bug fix has been merged. The next step is for Focalboard team to cut a new release and prepackage it with the server.

PR #4213, opened 2022-11-15
Require build tag to include sqlite

wiggin77:
#### Summary
This PR ensures that the `go-sqlite3` package is only included when the `sqlite3` build tag is specified.  This tag is automatically included for all Focalboard makefile targets, but not included when building in product mode from mm-server.

This fixes an issue with running mm-server on Ubuntu 18.04 which includes an older version of glibc (2.27) while `go-sqlite3` requires 2.28 or higher.  mm-server does not support sqlite thus there is no point including it as a dependency.

go.mod files are not impacted as they always include all packages as if all build tags are active.

See https://community.mattermost.com/private-core/pl/5z4etm5883n6ie88sas9jh1owr for discussion.

Tested with personal server (Linux & Windows). 

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/4210

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ ifeq ($(BUILD_NUMBER),)
 	BUILD_DATE := n/a
 endif
 
-BUILD_TAGS += json1
+BUILD_TAGS += json1 sqlite3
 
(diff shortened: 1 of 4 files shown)


wiggin77:
> it doesn't seem like there were any custom code after all. 

Yes, I really thought there was. 🤷‍♂️

PR #4218, opened 2022-11-16
Add sqlite3 tag to modd files and linux app build

mgdelacroix:
#### Summary
This PR adds the `sqlite3` tag on both the `modd` files and the linux `Makefile` and updates the `go mod` dependencies for the latter


wiggin77 (approved):
LGTM.  I guess the Squirrel version bump is harmless.

mgdelacroix:
@wiggin77 same thinking here, and it's done to match the one on the `server/go.mod` file, as the build would fail if they mismatched. Will keep this in mind while cherrypicking tho

Issue #4206, opened 2022-11-15
Bug: When changing filter from Status to any other text contained property the `include` should change to `is`


Rajat-Dabade:
## Steps to reproduce the behaviour

1. Go to any board with cards containing some text properties.
2. Click on the filter button and add a filter.
3. Change the filter from Status to any other text-contained properties.
4. See the error the include hasn't changed, it should change to `is` according to the dropdown.

## Expected behaviour

Whenever text-contained properties are selected for filter the first filter to be chosen should be `is` according to the dropdown

## Screenshots (optional)

https://user-images.githubusercontent.com/26198062/201853735-37d86dfd-44b8-4533-b372-1778ae8aba76.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: [e.g. v7.6]
 - Browser and OS: Chrome on Mac

PR #4221, opened 2022-11-16
Set the default select value for filterOptions

Rajat-Dabade:
#### Summary
This PR will set the default value for the filter option in boards. 
Example: For the `text` field the default will be `is` and for the `options` field it will be `includes`

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/4206


webapp/src/components/viewHeader/__snapshots__/filterComponent.test.tsx.snap
```diff
@@ -1013,7 +1013,7 @@ exports[`components/viewHeader/filterComponent return filterComponent and filter
             type="button"
           >
             <span>
-              includes
+              (unknown)
```
Pinjasaur, line 1016: @Rajat-Dabade is this an expected side effect of the change?

Rajat-Dabade: @Pinjasaur, Yes this is expected.

Pinjasaur (approved):
LGTM 👍🏻

harshilsharma63 (approved):
LGTM 👍 Nice UX improvement!

PR #4232, opened 2022-11-22
Manually de-duplicate result set for SearchBoardsForUserInTeam

wiggin77:
#### Summary
Manually de-duplicate result set for SearchBoardsForUserInTeam since adding `distinct` to the query increased cost by 15X.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48549


sbishel (approved):
Do you even need/want the boolean `removeDuplicates`?

Other than that LGTM.

wiggin77:
> Do you even need/want the boolean `removeDuplicates`?

I did that because it is an unexpected side-effect.  Currently only two methods use the baordsFromRows but when we add more, we don't want to manually-duplicate when not needed for performance or logic reasons.  i.e.  a query that produces a large result set that is already unique (no left joins)

PR #4244, opened 2022-11-24
Adding only boards with explicit access to default category

### LINKED ISSUE #4230: Bug: All synthetic-access boards end up displaying in sidebar

1. Create a channel and link multiple boards to it.
2. Signup as a new user and navigate to the channel in step 1.
3. Open any board from the boards RHS in channels.
4. Notice that all boards which were linked to the channel in step 1 end up being displayed in Boards LHS.

## Expected behavior

Only the board opened from channels RHS should show up in the Boards LHS.

## Edition and Platform

 - Edition: Mattermost Boards
 - Version: `main`

harshilsharma63:
Fixes #4230

The issue was, we were including all board memberships for the user - implicit and explicit. This caused all linked boards of all channels the user was part of when opening Boards for the first time. The fix is to include only boards with explicit memberships.


sbishel (approved):
I like the additional tests....

harshilsharma63:
@ogi-m can you please test this out?

ogi-m (approved):
LGTM 👍🏻

PR #4268, opened 2022-11-30
Disable cloud limits

mgdelacroix:
#### Summary
This PR updates the Cloud Limits checks to effectively disable the limits.


harshilsharma63 (approved):
LGTM 👍

sbishel (approved):

ogi-m (commented):
@mgdelacroix I've tested this build against a cloud test server with the cloud free license, and the limits are not being applied 👍🏻 

However, there are certain issues with the test setup - when running a build from main or 7.5.2 on the same server I tested the PR build, I could not add views to a board at all. As soon as I would remove the cloud licence, boards would work fine. So, this doesn't give me confidence that the cloud limits were being applied properly on the server in the first place.

This is only an issue on cloud test, I checked on cloud production and it's working fine - user can add views and trying to add more than 5 will display the expected error.

I don't want to block the PR any longer, so if it was tested locally and it works I am for merging this.

mgdelacroix:
Thanks @ogi-m! I'm updating the branch with the current `main` and retesting for good measure, will let you know how it goes before merging

mgdelacroix:
I couldn't find any blockage with the views, but I've added 15c132108 to disable and shortcircuit a couple more places in the codebase. Neither should have an effect, so I don't think it's necessary to review the PR again, but pinging @sbishel and @harshilsharma63 in case you want to take a look at that commit

As per testing, I've talked with @ogi-m and we'll wait for a test round before merging the PR

harshilsharma63:
@ogi-m could this be an issue with the CWS feature flag that we have to enable to allow the whole limits thing to work?

harshilsharma63:
@mgdelacroix the commit looks good 👍

PR #4270, opened 2022-11-30
Update version to v7.7.0

sbishel:
#### Summary
Update version to v7.7.0

### DIFF
--- a/webapp/src/components/sidebar/__snapshots__/sidebar.test.tsx.snap
+++ b/webapp/src/components/sidebar/__snapshots__/sidebar.test.tsx.snap
@@ -51,9 +51,9 @@ exports[`components/sidebarSidebar dont show hidden boards 1`] = `
                   >
                     <div
                       class="version"
-                      title="v7.6.0"
+                      title="v7.7.0"
                     >
-                      v7.6.0
+                      v7.7.0
                     </div>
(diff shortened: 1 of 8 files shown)


webapp/src/components/sidebar/__snapshots__/sidebar.test.tsx.snap
```diff
@@ -51,9 +51,9 @@ exports[`components/sidebarSidebar dont show hidden boards 1`] = `
                   >
                     <div
                       class="version"
-                      title="v7.6.0"
```
wiggin77, line 54: If these version strings in typescript files started using the string in the constants file it would require a lot less file edits.

wiggin77 (approved):

webapp/src/components/sidebar/__snapshots__/sidebar.test.tsx.snap
sbishel: Can you do that with snapshot files? Because they get built by running the tests.

wiggin77: Good point.  I realize now those aren't changing by manually editing them - they are changing by updating the snapshots.  My mistake.

PR #4284, opened 2022-12-02
Fixed bug where boards would move to category of a different team

### LINKED ISSUE #4283: Bug: Boards get linked to category of incorrect team when default boards category gets created

With a new user-

1. Create multiple teams in channels.
2. Open Boards in any one team.
3. Create a few boards in this team.
4. Now in boards, switch to a different team.
5. Now switch back to the previous team. Notice the boards you just created no longer show up.

With existing user-

1. create some boards in a team if you don't already have existing boards.
2. Now create a new team from channels.
3. In boards, switch to this new team.
5. Now switch back to the previous team. Notice the boards you just created no longer show up.

## Expected behavior

Boards should show up correctly.

## Edition and Platform

 - Version: 7.6

Caused by #4244

harshilsharma63:
Fixes #4283

## Explanation

In PR #4283 I was using the method `GetMembersForUser(<userID>)` to fetch the list of all boards for a user in a team. However, I missed the fact that that method is NOT scoped to a specific team, as you can see from its parameters.
What this caused was for all boards in all teams to be considered as candidate boards to be moved into the current team's default `Boards` category. This ended the system in a state where the board belongs to Team A but is associated with a category of Team B, so it didn't show up anywhere.

I've now fixed this by using the method `GetBoardsForUserAndTeam(<userID,> <teamID>, <includePublicBoards>` to gets user's boards for the current team.

### DIFF
--- a/server/app/category_boards.go
+++ b/server/app/category_boards.go
@@ -11,6 +11,7 @@ const defaultCategoryBoards = "Boards"
 
 var errCategoryBoardsLengthMismatch = errors.New("cannot update category boards order, passed list of categories boards different size than in database")
 var errBoardNotFoundInCategory = errors.New("specified board ID not found in specified category ID")
+var errBoardMembershipNotFound = errors.New("board membership not found for user's board")
 
 func (a *App) GetUserCategoryBoards(userID, teamID string) ([]model.CategoryBoards, error) {
 	categoryBoards, err := a.store.GetUserCategoryBoards(userID, teamID)
@@ -66,9 +67,15 @@ func (a *App) createBoardsCategory(userID, teamID string, existingCategoryBoards
 
 	// once the category is created, we need to move all boards which do not
 	// belong to any category, into this category.
+
 	boardMembers, err := a.GetMembersForUser(userID)
 	if err != nil {
 		return nil, fmt.Errorf("createBoardsCategory error fetching user's board memberships: %w", err)
+	}
+
+	boardMemberByBoardID := map[string]*model.BoardMember{}
+	for _, boardMember := range boardMembers {
+		boardMemberByBoardID[boardMember.BoardID] = boardMember
 	}
 
 	createdCategoryBoards := &model.CategoryBoards{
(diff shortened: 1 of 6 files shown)


harshilsharma63:
cc @ogi-m

ogi-m (approved):
LGTM, I can't repro anymore 👍🏻

server/app/category_boards.go
```diff
@@ -66,31 +67,48 @@ func (a *App) createBoardsCategory(userID, teamID string, existingCategoryBoards
...
+	// get user's current team's baords
+	userTeamBoards, err := a.GetBoardsForUserAndTeam(userID, teamID, false)
+	if err != nil {
+		return nil, fmt.Errorf("createBoardsCategory error fetching user's team's boards: %w", err)
+	}
+
+	for _, board := range userTeamBoards {
+		boardMembership, ok := boardMemberByBoardID[board.ID]
+		if !ok {
+			return nil, fmt.Errorf("createBoardsCategory: %w", errBoardMembershipNotFound)
```
sbishel, line 95: Should we fail here, or log and continue?

sbishel (approved):
@harshilsharma63 - How often is this function called. It looks like it could have some performance issues against large data. And would end up with lots of loops to find minimal data. Would it be possible to increase performance of this function by creating a special query to just find those boards necessary to act upon. (Boards with no entry in CategoryBoards table)

If you think it's worth exploring, let's add a card to investigate. Approving but we may want to reevaluate how this is done.

server/app/category_boards.go
```diff
@@ -66,31 +67,48 @@ func (a *App) createBoardsCategory(userID, teamID string, existingCategoryBoards
...
+		boardMemberByBoardID[boardMember.BoardID] = boardMember
+	}
+
 	createdCategoryBoards := &model.CategoryBoards{
 		Category: *createdCategory,
 		BoardIDs: []string{},
 	}
 
-	for _, bm := range boardMembers {
+	// get user's current team's baords
```
mgdelacroix, line 86:
```suggestion
	// get user's current team's boards
```

mgdelacroix (approved):

mgdelacroix:
@harshilsharma63 @sbishel should we cherry-pick this to cloud as well?

sbishel:
> should we cherry-pick this to cloud as well?
I don't think that is necessary, as cloud still runs as a plugin. Also, cloud get fast forwarded to main every Thursday before release.

PR #4282, opened 2022-12-02
Allow empty dates returned from date property

sbishel:
#### Summary
Currently if a date property does not contain a value, the "createAt" property is returned. Dates are not allowed to return undefined. This results in calendar items being displayed by their "CreateAt" property if the Calendar's "Date" Property is not defined.

This PR updates properties to allow dates to be returned as undefined.

#### Ticket Link
  Fixes #4181

PR #4287, opened 2022-12-02
Implement Date property type

sbishel:
#### Summary
While working on https://github.com/mattermost/focalboard/pull/4282, it became apparent that we should treat date properties separately from the basic Property types. 

This PR creates another abstract class `DatePropertyType`. The date specific property functions are then moved to this class.


webapp/src/components/calendar/fullCalendar.tsx
```diff
@@ -96,11 +98,21 @@ const CalendarFullView = (props: Props): JSX.Element|null => {
...
             const property = propsRegistry.get(dateDisplayProperty?.type || 'unknown')
+
             let dateFrom = new Date(card.createAt || 0)
             let dateTo = new Date(card.createAt || 0)
-            if (property.isDate && property.getDateFrom && property.getDateTo) {
-                dateFrom = property.getDateFrom(card.fields.properties[dateDisplayProperty?.id || ''], card)
-                dateTo = property.getDateTo(card.fields.properties[dateDisplayProperty?.id || ''], card)
+            if (property instanceof DatePropertyType) {
+                const dateFromValue = property.getDateFrom(card.fields.properties[dateDisplayProperty?.id || ''], card)
+                if (!dateFromValue) {
```
jespino, line 106: Just to confirm that, is there the possibility of having a date that has a from and not a to value? I mean. Something like "not-range-date". In that case you have both anyway? If that is the case, it is right, if not, we can end up filtering here cases where you don't have ranges.

webapp/src/properties/types.tsx
```diff
@@ -75,3 +71,22 @@ export abstract class PropertyType {
     abstract type: PropertyTypeEnum
     abstract displayName: (intl: IntlShape) => string
 }
+
+export abstract class DatePropertyType extends PropertyType {
+    isDate = true
+    canGroup = false
```
jespino, line 77: There is a lot of extra fields that are set by default for the DatePropertyType, that could be inherited from PropertyType. Everything except isData, getDateFrom and getDateTo, is the same in PropertyType (name and value).

webapp/src/components/calendar/fullCalendar.tsx
sbishel: It was correct because `property.getDateTo()` always returned value. If not a range it created one. `getDateTo()` also incremented the date by 1 day. this is because FullCalendar's "end" date is exclusive. So if you set end date on 22nd. It will cover the 21st, but not the 22nd. 

Because this is a full calendar issue, I moved the date increment to full calendar and `getDateTo()` returns the actual `To` date or undefined if there is no range.

sbishel: And...this is currently the only caller of `getDateFrom` and `getDateTo`.

webapp/src/properties/date/property.tsx
```diff
@@ -65,14 +62,11 @@ export default class DateProperty extends PropertyType {
...
             return undefined
         }
-        const dateFrom = dateProperty.from ? new Date(dateProperty.from + (dateProperty.includeTime ? 0 : timeZoneOffset(dateProperty.from))) : new Date()
-        dateFrom.setHours(0, 0, 0, 0)
-
-        const dateToNumber = dateProperty.to ? dateProperty.to + (dateProperty.includeTime ? 0 : timeZoneOffset(dateProperty.to)) : dateFrom.getTime()
-        const dateTo = new Date(dateToNumber + oneDay) // Add one day.
+        const dateToNumber = dateProperty.to + (dateProperty.includeTime ? 0 : timeZoneOffset(dateProperty.to))
+        const dateTo = new Date(dateToNumber)
         dateTo.setHours(0, 0, 0, 0)
```
sbishel, line 70: Moved incrementing a day to FullCalendar, since it is for that functionality. Also, don't default to `from`. Return undefined, if not a range.

jespino (approved):
LGTM

sbishel:
@ogi-m - https://github.com/mattermost/focalboard/pull/4412 is built on top of this PR. So you could just test that one.

PR #4291, opened 2022-12-02
Search boards by prop

wiggin77:
#### Summary
The cloud team has requested the ability to search for boards based on a property name.  This was also a requirement for the Playbooks/Agenda/Call integration plugin from Mattercon Contribute.

This has been implemented in a fashion that allows other search fields to be added (e.g. property value).

#### Ticket Link
https://community.mattermost.com/boards/team/qghzt68dq7bopgqamcnziq69ao/b4ctmm917q3rzdfi451qtuxjzcw/150d1464-6d59-4c39-810f-ee201439dbaf/cg7cp6poogtfc5y8xszixd6qzrw


server/services/store/mattermostauthlayer/mattermostauthlayer.go
```diff
@@ -716,19 +716,26 @@
...
-		// to break words.
-
-		conditions := sq.And{}
-
-		for _, word := range strings.Split(strings.TrimSpace(term), " ") {
-			conditions = append(conditions, sq.Like{"lower(b.title)": "%" + strings.ToLower(word) + "%"})
+		if searchField == model.BoardSearchFieldPropertyName {
+			if s.dbType == model.PostgresDBType {
+				where := fmt.Sprintf("b.properties->>'%s' is not null", term)
+				query = query.Where(where)
```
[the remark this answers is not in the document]
wiggin77: Ugh. That's what I get for trying to use my Mattercon contribute code in a production PR.  I'll fix.  Of course Squirrel has no easy to express this.

server/services/store/sqlstore/board.go
```diff
@@ -680,21 +680,35 @@ func (s *SQLStore) searchBoardsForUser(db sq.BaseRunner, term, userID string, in
...
+				conditions = append(conditions, sq.Like{"lower(b.title)": "%" + strings.ToLower(word) + "%"})
+			}
+			query = query.Where(conditions)
 		}
-
-		query = query.Where(conditions)
 	}
 
+	sql := sq.DebugSqlizer(query)
+	s.logger.Debug("!!!!! searchBoardsForUser", mlog.String("sql", sql))
```
mgdelacroix, line 710: Probably debug statements left behind

server/services/store/sqlstore/board.go
```diff
@@ -680,21 +680,35 @@ func (s *SQLStore) searchBoardsForUser(db sq.BaseRunner, term, userID string, in
...
+			switch s.dbType {
+			case model.PostgresDBType:
+				where := "b.properties->? is not null"
+				query = query.Where(where, term)
+			case model.MysqlDBType:
+				where := "JSON_EXTRACT(b.properties, ?) IS NOT NULL"
+				query = query.Where(where, "$."+term)
+			default:
+				where := "b.properties LIKE ?"
+				query = query.Where(where, "%\""+term+"\"%")
```
mgdelacroix, line 693: Shouldn't we be able to use something like this here for SQLite?

```suggestion
				where := "json_extract(b.properties, '$.?') IS NOT NULL"
				query = query.Where(where, term)
```

wiggin77: Thanks @mgdelacroix.  I really thought `json_extract` wasn't available until a later version of Sqlite than we support as minimum version.  Now I think I was looking at the [-> and ->> operators](https://www.sqlite.org/json1.html#jex) that require 3.38.0.   My distro comes with 3.31.1

server/services/store/sqlstore/board.go
wiggin77: Thanks. Fixed.

mgdelacroix (approved):
LGTM! :+1:

sbishel (approved):
LGTM - Does PM know about this feature? I'm thinking we should probably add a feature to the FE to take advantage of this.

Also, add CP tag and version if wanted for v7.7.0

wiggin77:
> we should probably add a feature to the FE to take advantage of this.

@sbishel  I think we should add searching on property values first.  Also, we may want to allow the searchFields to be combined, ie search title and property values.

PR #4325, opened 2022-12-13
Updates migrations to check collation/charset before trying to fix it

mgdelacroix:
#### Summary
This PR obtains the collation and charset of the `Boards` tables and only fixes them if necessary. Currently it maintains the approach of fixing the issue automatically, but I'm happy to update it to notify the admin instead if that's the consensus.


sbishel (approved):

wiggin77 (approved):
LGTM :+1:

Issue #4326, opened 2022-12-13
Bug: latest `main` fails to fetch team boards in plugin mode


harshilsharma63:
## Steps to reproduce the behavior

1. Build from the latest `main`.
2. Deploy in plugin mode.
3. Open boards. Note in network tab of browser that the api call `/teams/<team-id>/boards` API call fails with the error-

```
{"timestamp":"2022-12-13 14:12:19.219 +05:30","level":"error","msg":"searchBoardsForUser ERROR","caller":"app/plugin_api.go:976","plugin_id":"focalboard","error":"\"pq: got 9 parameters but the statement requires 3\""}
{"timestamp":"2022-12-13 14:12:19.219 +05:30","level":"error","msg":"pq: got 9 parameters but the statement requires 3","caller":"app/plugin_api.go:976","plugin_id":"focalboard"}
{"timestamp":"2022-12-13 14:12:19.219 +05:30","level":"error","msg":"API ERROR","caller":"app/plugin_api.go:976","plugin_id":"focalboard","code":"500","error":"\"pq: got 9 parameters but the statement requires 3\"","api":"/api/v2/teams/hyk1pjobobbf7cpea9zd4iaxwe/boards"}
```

## Expected behavior

API call shouldn't fail

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: current `main`
 - Browser and OS: Firefox, on MacOS on Postgres 14.2

PR #4327, opened 2022-12-13
Fixes positional arguments in UNION queries

mgdelacroix:
#### Summary
This PR makes the UNION queries use always the question mark positional argument (that is not numbered) and replaces it after the queries have been merged with a positional argument if the database requires it.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/4326


mgdelacroix:
We should include this fix on `v7.5.x` if we're cherrypicking the `UNION` queries as well

sbishel (approved):

wiggin77 (approved):
LGTM :+1:

Issue #4333, opened 2022-12-13
Bug: Able to see private boards through board switcher


wuwinson:
## Steps to reproduce the behavior

1. Login with user A
2. Create a private board for user A
3. Login with user B
4. Go to board switcher
5. Search for board created in step 2
6. See error - board appears in search results
7. Clicking on the board leads to an error page

## Expected behavior

Board should not appear in results. User should only see results for boards they have access to.

## Screenshots (optional)

![Screen Shot 2022-12-13 at 4 48 19 PM](https://user-images.githubusercontent.com/93531870/207451682-7569b04c-f3da-4b95-93dd-acf93f4c2f6b.png)

![Screen Shot 2022-12-13 at 4 48 27 PM](https://user-images.githubusercontent.com/93531870/207451444-48d24ac5-0cbf-48fd-8cd6-d1252e240965.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.5.4
 - Browser and OS: Chrome on Mac
 - [mattermost-server master with MySQL](https://sbishel-master.test.mattermost.cloud/boards/team/5sc6ey6j7f8n8dozkcaq8j9q3h/bhf3sdcjmsjn3ifkythozc4hoxr/vk5guubgrqjnkidphcqg3ypwknr)

wuwinson:
Also occurring on https://sbishel-75.test.mattermost.cloud/

sbishel:
Yep...We have a fix for it...

PR #4335, opened 2022-12-14
Update query to only allow open boards for team members

sbishel:
#### Summary
This PR fixes the issue, by updating the `team_member` query to only include Open Boards. Because the `team_member` query should be used to allow access to Open Boards to team members. Also the `channel_members` and `board_members` queries should NOT return Open Boards.

The rest of the PR was for clean-up, simplification. Since we now have simple queries. `includePublicBoards` means include the `team_members` query. If the User is a guest. Don't include the `channel_members` query. 

`includePublicBoards` and `user.IsGuest` are handled separately. However, currently, both are always either true or false. 

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/4333

### DIFF
--- a/server/services/store/mattermostauthlayer/mattermostauthlayer.go
+++ b/server/services/store/mattermostauthlayer/mattermostauthlayer.go
@@ -769,11 +769,6 @@ func (s *MattermostAuthLayer) SearchBoardsForUser(term, userID string, includePu
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


mgdelacroix (approved):
LGTM :+1:

wiggin77 (approved):

Issue #4316, opened 2022-12-09
Bug: Linked Board message displays Template Name, not board name. 


sbishel:
## Steps to reproduce the behavior

1. Go to channel and link a board.
2. Select  "+Add"
3. Select "Create a board"
4. Create board from template...rename.
5. See message in channel. 
Message refers to the board as the Template Name, not the board name.

## Expected behavior

Should display the board name.

## Screenshots (optional)
![Screen Shot 2022-12-09 at 10 02 21 AM](https://user-images.githubusercontent.com/12704875/206754275-ee4a827d-0724-4c00-875c-e669e138064b.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: [7.5.3 - community]
 - Browser and OS: [Chrome on Mac]

## Additional context (optional)

Add any other context about the problem here, and any notes about the severity:
* Sev 1: Affects critical functionality without a workaround
* Sev 2: Affects major functionality with a difficult or non-obvious workaround
* Sev 3: Affects minor, non-critical functionality

PR #4353, opened 2022-12-15
GH-4316 - if no title, use 'untitled board' in message

sbishel:
#### Summary
In message when a board is linked, if there is no title in the board, set to "Untitled board", same as UI does. 

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/4316
Screen Shot 2022-12-15 at 4 25 58 PM


harshilsharma63 (approved):
LGTM 👍

PR #4367, opened 2022-12-19
Rename migration 24 back to original name

sbishel:
#### Summary
This migration was renamed as part of the idempotent PR. However, it appears this causes the migration to be run again. The migration then fails when attempting the update the Schema table because Primary key 24 already exists.

The PR simply renames the migration back to the original name as the quickest solution.

```
{"timestamp":"2022-12-19 22:33:34.311 Z","level":"error","msg":"Unable to restart plugin on upgrade.","caller":"web/context.go:117","path":"/api/v4/plugins","request_id":"cxi5xcmni7nkf8mygp65humozw","ip_addr":"100.121.135.192","user_id":"mh37ihwuaprm8nsc7xkx9f39oo","method":"POST","err_where":"installExtractedPlugin","http_code":500,"error":"installExtractedPlugin: Unable to restart plugin on upgrade., cannot activate plugin: error initializing the DB: driver: mysql, message: failed when updating migrations table with the new version, command: update_version, originalError: Error 1062: Duplicate entry '24' for key 'PRIMARY', query: \n\nINSERT INTO focalboard_schema_migrations (Version, Name) VALUES (24, 'mark_existing_categories_collapsed')\n"}
```


wiggin77 (approved):

sbishel:
Does NOT need cherry-picked.

PR #4377, opened 2022-12-20
Update text, used to set the button text

sbishel:
#### Summary
In making the  call to 'registerChannelIntroButtonAction' generic. The text being passed in should be used to label the button. However, the "non-generic" implementation, set the button text on the channels side.

This PR, updates the text being passed, to be the button text currently hard-coded.


mattermost-plugin/webapp/src/index.tsx
```diff
@@ -329,7 +329,7 @@ export default class Plugin {
             }
 
             if (registry.registerChannelIntroButtonAction) {
-                this.channelHeaderButtonId = registry.registerChannelIntroButtonAction(<FocalboardIcon/>, goToFocalboardTemplate, 'Boards')
+                this.channelHeaderButtonId = registry.registerChannelIntroButtonAction(<FocalboardIcon/>, goToFocalboardTemplate, intl.formatMessage({id: 'ChannelIntro.CreateBoard', defaultMessage: 'Create a board'}))
```
hmhealey, line 332: A couple thought here:
1. I actually added the new text prop as a fourth parameter to prevent us from showing "Boards" as that text if someone has an old version of Boards installed on a newer server. Do you think that's a case we should worry about, or should I just reuse this parameter in https://github.com/mattermost/mattermost-webapp/pull/11813? I don't know how likely it is that people are using old Boards with a new web app.

    We could also go with the slightly cleaner approach on the web app side of just adding a special case where the string "Boards" always gets rendered as "Create a board" to keep that compatibility without leaving that unused parameter on the web app side.
2. I'm realizing that having someone pass in a localized string like you did here means that this won't be updated live when the locale changes. We can leave this as-is since the other pluggables have this issue, but in the longer term, we'll probably have to change all of these regsitry methods to accept an object with an `id` and `defaultMessage` at some point

sbishel: @hmhealey - I don't think I would worry about people running an old boards, using prebuilt plugins will minimize this. Also even if someone did. The header button would say 'Boards' rather than 'Create a board'. To me, minimum risk, minimum consequences, but much cleaner code. But ultimately your decision. I just wanted to give the option to update boards so you can do it cleaner if desired. 

Regarding the second point, as you said, probably a larger issue than this PR. Do we support this in any aspect of the plugin framework?

hmhealey: 1. That's fair. As you said, worst case scenario, the text is slightly wrong, so it's not worth adding that extra logic.

2. I'm not sure. I'm actually unclear on how well we support i18n with plugins since it wasn't supported at all last time I really worked with plugins. I'd need to look more into how i18n is done for plugins like, for example, if the web app can even call `intl.formatMessage(...)` with a message descriptor provided by the plugin.

sbishel:
@hmhealey - Let me know if you want me to go ahead and merge this...

hmhealey (approved):
Right, sorry about that. This looks good to me

PR #4386, opened 2022-12-21
Don't set cattegory for board template

### LINKED ISSUE #3938: Bug: Creating a board from a custom template adds the board to the same category as the origin board

harshilsharma63:
#### Summary
When duplicating a board, whether from a template or not, we move the new board to the old board's category. When creating a template from a board, we were assigned a category to the template, which later was assigned to the board created from this template. Not assigning a category to the template stops the chain and causes the new board to be moved to the default category.

#### Ticket Link
Fixes #3938

### DIFF
--- a/server/app/boards.go
+++ b/server/app/boards.go
@@ -189,9 +189,11 @@ func (a *App) DuplicateBoard(boardID, userID, toTeam string, asTemplate bool) (*
 		a.logger.Error("Could not copy files while duplicating board", mlog.String("BoardID", boardID), mlog.Err(err))
 	}
 
-	for _, board := range bab.Boards {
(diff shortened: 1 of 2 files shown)


wiggin77 (approved):
LGTM 👍

sbishel (approved):

PR #2881, opened 2022-04-21
Marketing/website redesign

emdecr:
#### Summary
This branch encompasses the redesign of the focalboard.com website.

#### Ticket Link
On Marketing Asana Board:
https://app.asana.com/0/1201867728201087/1201774448360374/f

@chenilim A few questions:

- I vaguely remember you mentioning you'd prefer images hosted outside of the repo. What's the protocol around that again? 
- We don't have staging sites for preview purposes, right? Asking because I'd like to send JuliAnn a link to review.
- Also let me know if there's a preferred way to go about PRs like this. 

Thanks!

PR #4391, opened 2022-12-22
Focalboard Website Update V2

asaadmahmood:
#### Summary
Focalboard Website Update V2


emdecr (commented):
This looks wicked! Animations are a nice touch 😎 

Only things I have to mention are:
- There's a dead link under the roadmap section (made a comment)
- I think there's a preference to offload larger assets to be hosted via GitHub discussions. It's historically been a bit of a grey area, so I'll defer to @chenilim if this is necessary for merging. We had a previous conversation [here](https://github.com/mattermost/focalboard/pull/2881#issuecomment-1105404499). An example of a discussion I had to make to host images is [here](https://github.com/mattermost/focalboard/discussions/2883).

Otherwise, 🎉

emdecr (approved):
Looks good to me! 

Was chatting with @asaadmahmood re: off-loading images and vs being a blocker think that could be a followup PR cc: @chenilim 

Approving otherwise~

PR #4396, opened 2022-12-23
Pages POC

jespino:
#### Summary
Pages POC


harshilsharma63 (changes requested):

sbishel (commented):
Wow...huge! I didn't see anything egregious, but kept loosing myself. One thing not clear to me from the code is 
category -> board -> page(s)
Seems like an extra layer in there. But probably just not clear.

Also, before merging I think we should bug bash both with the feature flag off and on. Let's get a couple spinwicks up and try to bug bash (ff off) next week.

asaadmahmood (changes requested):
Not sure if and when we'll be doing this, so taking this off of my PR to review list by requesting a review.

PR #4415, opened 2022-12-31
Fix Chinese lang code

zhsj:
zh-cn is for Simplified Chinese, and zh-tw is for Traditional Chinese. And remove invalid zh-tx code.

See wikipedia link example,

https://zh.wikipedia.org/zh-cn/Wikipedia:%E9%A6%96%E9%A1%B5
https://zh.wikipedia.org/zh-tw/Wikipedia:%E9%A6%96%E9%A1%B5


sbishel (approved):
@justinegeffen These changes look right to me. But I would like to extra eyes to make sure our names are consistent with Channels. cc: @wuwinson

wuwinson (approved):
LGTM!

PR #4418, opened 2023-01-03
Update snapshots for date format

sbishel:
#### Summary
Date format changed because these dates are now "last year", the years is now displayed.

Update snapshots to fix.


wiggin77 (approved):
Nice. :+1:

Issue #4126, opened 2022-11-03
Bug: any card with due date field is going to appear in search results for `from`


jespino:
## Steps to reproduce the behavior

1. Go to any board
2. Select  a card and set the due date
3. Search for from
4. You will see that the card matches even there is not any `from` text in the card

## Expected behavior

Only matches in search results when it contains that text (specially in the interface).

## Suggested solution

To solve this bug I would render the date before matching the search, to be able to find things like `June`, and not match always something like `from`.

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: 7.5 (master)
 - Browser and OS: Firefox

## Additional context (optional)

* Sev 3: Affects minor, non-critical functionality

PR #4419, opened 2023-01-03
don't check date fields for text search

sbishel:
#### Summary
Because a date field is a json object, searching on the name of one of the fields results in the card being found. Don't search date fields for text search.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/4126


Rajat-Dabade (approved):
LGTM 👍

Issue #4275, opened 2022-12-01
Bug: Export/Import - Memberships not retained


sbishel:
## Steps to reproduce the behavior

1. Go to Board, add members to that board.
2. Select  Export to Board Archive
3. Select Import Archive
4. Once imported, open share dialog.
5. Note no memberships are retained.

If that person is set in a person property, the property will appear blank, so also affects card data (display, data is still there.)

## Expected behavior
Memberships should be retained

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: [7.6.0-rc1]
 - Browser and OS: [Chrome on Mac]

## Additional context (optional)

Add any other context about the problem here, and any notes about the severity:
* Sev 1: Affects critical functionality without a workaround
* Sev 2: Affects major functionality with a difficult or non-obvious workaround
* Sev 3: Affects minor, non-critical functionality

wuwinson:
When importing to a different team, and a member is not on the team, display an "unknown" user or their userID instead of not importing anyone to the board.

PR #4422, opened 2023-01-04
Export Import Memberships is retained

Rajat-Dabade:
#### Summary
This PR will solve the issue of retaining the membership of the user when imported from the board archive 

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/4275

### DIFF
--- a/server/app/import.go
+++ b/server/app/import.go
@@ -202,7 +202,6 @@ func (a *App) ImportBoardJSONL(r io.Reader, opt model.ImportArchiveOptions) (str
 					if err2 := json.Unmarshal(archiveLine.Data, &boardMember); err2 != nil {
 						return "", fmt.Errorf("invalid board Member in archive line %d: %w", lineNum, err2)
 					}
-
 					boardMembers = append(boardMembers, boardMember)
 				default:
 					return "", model.NewErrUnsupportedArchiveLineType(lineNum, archiveLine.Type)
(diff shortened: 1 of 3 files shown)


server/app/import.go
```diff
@@ -239,6 +247,27 @@ func (a *App) ImportBoardJSONL(r io.Reader, opt model.ImportArchiveOptions) (str
 		}
 	}
 
+	if len(boardMembers) > 0 {
+		for _, board := range boardsAndBlocks.Boards {
```
sbishel, line 251: Shouldn't there be a different set of boardMembers for each board. Is there only one board in this range?

server/app/import.go
```diff
@@ -239,6 +247,27 @@ func (a *App) ImportBoardJSONL(r io.Reader, opt model.ImportArchiveOptions) (str
...
+					UserID:          boardMember.UserID,
+					Roles:           boardMember.Roles,
+					MinimumRole:     boardMember.MinimumRole,
+					SchemeAdmin:     boardMember.SchemeAdmin,
+					SchemeEditor:    boardMember.SchemeEditor,
+					SchemeCommenter: boardMember.SchemeCommenter,
+					SchemeViewer:    boardMember.SchemeViewer,
+					Synthetic:       boardMember.Synthetic,
+				}
+				if _, err2 := a.AddMemberToBoard(bm); err2 != nil {
```
sbishel, line 264: Will this work if same user is importing. We add the default admin above.

sbishel (changes requested):
Looks good Rajat. Let's add some unit tests.

sbishel:
I don't think this is necessary for v7.7.0.

harshilsharma63:
@Rajat-Dabade When importing the archive in a teamthe  different from original team, it adds the board memberships even if the users aren't part of the current team. This isn't right as we scope everything to the team and we can't add members from other teams. IMO we shouldn't add board memberships if the user doesn't belong to the current team.

Also, it seems to create board memberships in the database irrespective of whether the corresponding user exists or not. While this doesn't do any harm, we're knowingly inserting incorrect data. IMO we shouldn't add board membership entries in the database in this case.

In any case, the end user will have no clue what happened, or rather what didn't happen. We don't show any warning or error message before or after the import, so not sure how we communicate what's going on to the user. I love how Jira does bulk import - it runs a dry-import first without actually committing any data, and then asks the user to map data in the import file to data in the system. For example, map the user IDs in the board archive file to the users existing in the current team. But that's a big, long enhancement on its own.

cc @sbishel @wuwinson

harshilsharma63 (approved):
Looks good to me other than missing new tests.

server/app/import.go
harshilsharma63: @sbishel I think now we can only export one board at a time instead of the whole workspace in the workspace era. So I believe an import will consist of only one board. However, I'm not sure if we're still supporting import of the original workspace format imports. If we are, that can definitely contain multiple boards, but then it won't have board members.

sbishel:
I created the following [card](https://community.mattermost.com/boards/team/qghzt68dq7bopgqamcnziq69ao/b4ctmm917q3rzdfi451qtuxjzcw/vuswsftt1rb873datosmmbtwkpa/cntdrb8yxttnw8n6n7pjtnecfza) to track some of Harshil's concerns. Because I do believe they are bigger than this PR.

However, I do agree that we shouldn't be importing bad data. That is...if the user doesn't exist in the system at all we should not import. If the user doesn't exist in current team, we could either handle the same way (don't import) or handle the same way as a user who leaves a team. We can discuss when @wuwinson is back.

sbishel (approved):

sbishel:
@Rajat-Dabade - Before merging, did you fix ignoring users on import if they didn't exist. (ie, if imported into a completely different system)?

wuwinson:
> I created the following [card](https://community.mattermost.com/boards/team/qghzt68dq7bopgqamcnziq69ao/b4ctmm917q3rzdfi451qtuxjzcw/vuswsftt1rb873datosmmbtwkpa/cntdrb8yxttnw8n6n7pjtnecfza) to track some of Harshil's concerns. Because I do believe they are bigger than this PR.
> 
> However, I do agree that we shouldn't be importing bad data. That is...if the user doesn't exist in the system at all we should not import. If the user doesn't exist in current team, we could either handle the same way (don't import) or handle the same way as a user who leaves a team. We can discuss when @wuwinson is back.

Agree with @sbishel and @harshilsharma63. We shouldn't import users to other teams if they are not already part of the team. However, we would need a way to communicate this to the importer under this use case. Not needed for this PR, I think this can go as-is for now and we add the enhancement as a follow-up.

@asaadmahmood Thoughts on a way to address this? Maybe a simple modal with a list of board members who won't be imported similar to what you have [here](https://www.figma.com/file/WTIZstL9TuXEuvNc7vcVuW/Moving-Cards?node-id=18%3A28354&t=ogmgEpq3WFoN9WB9-4) for moving cards across boards.

Rajat-Dabade:
@sbishel, added the functionality to ignore the user if they are not present in a different system. Please review it. Thanks.

sbishel (approved):

Issue #4380, opened 2022-12-20
Bug: Duplicated attachments from templates with pre-attached files


wuwinson:
## Steps to reproduce the behavior

1. Open a card and attach files
2. On the sidebar options menu from the current board, select "New template from board"
3. Go to template picker and create a board from the new template
4. Open the card with attached files in Step 1
5. See error - File is duplicated on the card

## Expected behavior

Card should only display attachments if there are any, and attachment should not be duplicated.

## Screenshots (optional)

![Screen Shot 2022-12-20 at 1 41 34 PM](https://user-images.githubusercontent.com/93531870/208742542-400a0bad-6855-4ee0-9b50-fcf89f1fb80c.png)

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.7 Bug Bash
 - Browser and OS: Chrome on Mac

wuwinson:
@harshilsharma63 I logged this bug already.

harshilsharma63:
@wuwinson BTW do we want to fix duplicate file bug or disallow attachment in board templates?

wuwinson:
@harshilsharma63 We should fix the duplicate file bug here, but there may be some use cases where file attachments would make sense in a template, such as onboarding new hires and having them fill out docs. Would allow attachments in the templates unless we hear otherwise.

PR #4444, opened 2023-01-10
Fixed duplicate attachment in board template

Rajat-Dabade:
#### Summary
This PR will fix the issue of duplicating the file attachment inside the card when the new template is created from the board containing the file attachment inside the card.

#### Ticket Link
Fixes https://github.com/mattermost/focalboard/issues/4380


server/app/boards.go
```diff
@@ -210,6 +210,14 @@ func (a *App) DuplicateBoard(boardID, userID, toTeam string, asTemplate bool) (*
...
 			})
 		}
+		if attachmentID, ok := block.Fields["attachmentId"]; ok {
+			blockIDs = append(blockIDs, block.ID)
+			blockPatches = append(blockPatches, model.BlockPatch{
+				UpdatedFields: map[string]interface{}{
+					"attachmentId": attachmentID,
+				},
+			})
+		}
```
sbishel, line 220: Rather than duplicate the code. I would set the FieldID based on block type.
```
		fieldName := ""
		if block.Type == model.TypeImage {
			fieldName = "fileId"
		} else if block.Type == model.TypeAttachment {
			fieldName = "attachmentID"
		}
		if fieldID, ok := block.Fields[fieldName]; ok {
		...
```

webapp/src/pages/boardPage/boardPage.tsx
```diff
@@ -215,13 +215,15 @@ const BoardPage = (props: Props): JSX.Element => {
...
-            }
         }
     }, [teamId, match.params.boardId, viewId, me?.id])
 
+    useEffect(() => {
+        if (match.params.boardId && !props.readonly && me) {
+            loadOrJoinBoard(me.id, teamId, match.params.boardId)
+        }
+    }, [teamId, match.params.boardId, me?.id])
+
```
sbishel, line 226: Could you explain the need for this change?

sbishel (changes requested):

server/app/blocks.go
```diff
@@ -309,10 +310,15 @@ func (a *App) CopyCardFiles(sourceBoardID string, copiedBlocks []*model.Block) e
...
 		fileName, ok := block.Fields["fileId"]
 		if !ok || fileName == "" {
-			continue // doesn't have a file attachment
+			fileName, ok = block.Fields["attachmentId"]
+			if !ok || fileName == "" {
+				continue // doesn't have a file attachment
+			} else {
+				attachmentPresent = true
+			}
 		}
```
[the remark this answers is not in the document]
Rajat-Dabade: Ah, True. Will make the changes. Thanks @sbishel

webapp/src/pages/boardPage/boardPage.tsx
Rajat-Dabade: This piece of code should only execute when `teamId`, `boardId` and `me.Id` changes, as it is dependent on it. But it should not execute whenever `viewId` changes.  That's the reason to separate it.

server/app/boards.go
Rajat-Dabade: Okay, will change it

server/app/blocks.go
```diff
@@ -345,7 +351,21 @@ func (a *App) CopyCardFiles(sourceBoardID string, copiedBlocks []*model.Block) e
...
+			block.Fields["attachmentId"] = destFilename
+			parts := strings.Split(fileName.(string), ".")
+			fileInfoID := parts[0][1:]
+			fileInfo, _ := a.store.GetFileInfo(fileInfoID)
+			newParts := strings.Split(destFilename, ".")
+			newFileID := newParts[0][1:]
+			fileInfo.Id = newFileID
+			err := a.store.SaveFileInfo(fileInfo)
+			if err != nil {
+				return fmt.Errorf("cannot create file info for CopyCardFiles: %w", err)
```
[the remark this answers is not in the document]
Rajat-Dabade: We should, I think so, but then this function return's to `server/app/boards.go` https://github.com/mattermost/focalboard/blob/c1c94e9794fb27bbd45c9e340880d9d9f9c07601/server/app/boards.go#L188
which is just logging the error. Will it be okay to return `nil, nil, err` from there which I guess will cancel the whole process of duplicating the board?

server/app/blocks.go
```diff
@@ -309,14 +310,24 @@ func (a *App) CopyCardFiles(sourceBoardID string, copiedBlocks []*model.Block) e
...
 	for i := range copiedBlocks {
 		block := copiedBlocks[i]
-
-		fileName, ok := block.Fields["fileId"]
-		if !ok || fileName == "" {
-			continue // doesn't have a file attachment
+		fileName := ""
+		switch block.Type {
+		case model.TypeImage:
+			fileName, _ = block.Fields["fileId"].(string)
```
harshilsharma63, line 316: Can we check the ignored type assertion error?

server/app/blocks.go
```diff
@@ -309,14 +310,24 @@ func (a *App) CopyCardFiles(sourceBoardID string, copiedBlocks []*model.Block) e
...
+			if fileName == "" {
+				continue
+			}
+		case model.TypeAttachment:
+			fileName, _ = block.Fields["attachmentId"].(string)
```
harshilsharma63, line 321: Same here for the error.

harshilsharma63 (changes requested):

harshilsharma63:
Just one minor comment. Else looks good.

sbishel (approved):

sbishel:
@harshilsharma63 - I updated for your type assertion check, and added error handling in another spot. Also fixed an issue when duplicating a card with multiple attachments. 

Please review, test and merge. As we wanted this for v7.8.  Thanks.

harshilsharma63 (approved):

PR #4449, opened 2023-01-10
MM48320 - refactor create channel with board from webapp to pluggable

pvev:
#### Summary
This PR removes the code placed in the Webapp for creating a board along with the channel creation and moves that logic to boards as a pluggable. So basically the functionality remains exactly the same but this is now a pluggable.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48320

#### Related Pull Requests
https://github.com/mattermost/mattermost-webapp/pull/11958

#### Screenshots

https://user-images.githubusercontent.com/10082627/211880411-26a084e8-453c-4f26-ba62-5d3e826071b9.mp4

https://user-images.githubusercontent.com/10082627/211880425-5001e7bb-22b3-4ebf-83c8-a73a53578f58.mp4

#### Release Note
```release-note
NONE
```


mattermost-plugin/webapp/src/components/createBoardFromTemplate.tsx
```diff
@@ -0,0 +1,240 @@
...
+                    <span>
+                        {formatMessage({id: 'new_channel_modal.create_board.title', defaultMessage: 'Create a board for this channel'})}
+                    </span>
+                    {props.newBoardInfoIcon}
+                </label>
+                {addBoard && <div className='templates-selector'>
+                    <Select
+                        classNamePrefix={'CreateBoardFromTemplate--templates-selector'}
+                        placeholder={formatMessage({id: 'new_channel_modal.create_board.select_template_placeholder', defaultMessage: 'Select a template'})}
+                        onChange={(item: SingleValue<ReactSelectItem>) => {
```
hmhealey, line 208: 0/5 - We might want to wrap these methods as well as some of the named ones like `loadOptions` in `useCallback`, otherwise the `Select` will always re-render whenever `CreateBoardFromTemplate` does. That said, I don't know how much that actually matters in terms of actual performance or if it's just theoretically better to reduce re-rendering that way

webapp/src/hooks/useGetAllTemplates.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    useEffect(() => {
+        if (octoClient.teamId !== Constants.globalTeamId && globalTemplates.length === 0) {
+            dispatch(fetchGlobalTemplates())
+        }
+    }, [octoClient.teamId])
+
+    const unsortedTemplates = useAppSelector(getTemplates)
+    const templates = useMemo(() => Object.values(unsortedTemplates).sort((a: Board, b: Board) => a.createAt - b.createAt), [unsortedTemplates])
+    const allTemplates = globalTemplates.concat(templates)
```
hmhealey, line 28: I think this also needs to be wrapped in a `useMemo` for this to not return a new array each time it's called

hmhealey (changes requested):

mattermost-plugin/webapp/src/components/createBoardFromTemplate.tsx
pvev: ok

webapp/src/hooks/useGetAllTemplates.ts
pvev: done

mattermost-plugin/server/manifest.go
```diff
@@ -45,7 +45,8 @@ const manifestStr = `
         "type": "bool",
         "help_text": "This allows board editors to share boards that can be accessed by anyone with the link.",
         "placeholder": "",
-        "default": false
+        "default": false,
+        "hosting": ""
```
pvev, line 49: this change was generated automatically while setting up the plugin in my env, not sure if it needs to be version controlled

hmhealey: The same thing happens for me as well. I've just been ignoring it, but I don't know if it's something that should actually be checked in

hmhealey (approved):
Thanks for changing this. This looks good to me!

sbishel (approved):
I'm good with it...adding @ogi-m to test.

mattermost-plugin/server/manifest.go
sbishel: Not sure what causes the back and forth on this line...but don't worry about it.

pvev:
@ogi-m friendly reminder on this one :)

ogi-m (approved):
@pvev Let's do that since plugin upload is not possible on spinwicks

PR #4453, opened 2023-01-11
Update en.json

justinegeffen:
- Some minor casing changes.
- Rewrote some success messages for readability.
- Removed exclamation marks from some confirmation dialogs.
- Removed references to unlimited views and cards for 7.7 changes.


justinegeffen:
@sbishel, I see some failing checks but the PR is mergable. Just wanted to check if I need to do anything or whether I can merge now?

webapp/i18n/en.json
```diff
@@ -27,15 +27,15 @@
   "BoardPage.newVersion": "A new version of Boards is available, click here to reload.",
   "BoardPage.syncFailed": "Board may be deleted or access revoked.",
   "BoardTemplateSelector.add-template": "Create new template",
-  "BoardTemplateSelector.create-empty-board": "Create empty board",
+  "BoardTemplateSelector.create-empty-board": "Create an empty board",
```
sbishel, line 30: This change is causing a cypress failure...

sbishel (commented):
@justinegeffen - You have a change that is causing a cypress test to fail. I can update the test, but should these changes come through Weblate? (or is that just non en.json)

webapp/i18n/en.json
justinegeffen: That's really weird. I wonder why. I can revert the change if it'll help. 

The weblate changes are for non-en.json (@cwarnermm, just checking if that's correct?)

cwarnermm: All string changes should be made _only_ in the en.json file. All other language resource files for supported languages are populated through Weblate via translation contributions.

cwarnermm: That makes sense that a Cypress test could fail as a result of a string changing. The Cypress tests look for whatever it thinks the string is and if it doesn't find it, it throws an error.

When Docs modifies an existing product string, QA recommends we include a QA team member to review and ensure that associated tests are updated accordingly. Docs isn't required to maintain the Cypress tests, and QA very much appreciates a heads-up that a product change may impact their testing systems.

justinegeffen: Brilliant, thanks Carrie! Learned something new. :) I've added QA.

sbishel (approved):

ogi-m (approved):

Issue #4445, opened 2023-01-10
Bug: Date filter > 'is' option doesn't work initially


ogi-m:
## Steps to reproduce the behavior

1. Open a board and set the date property on a card
2. Apply the date filter with the 'is' option and select the date that was set on the card in step 1
3. Note the card is filtered out even though it should be shown
4. Change the date filter to any other option (e.g. 'is before')
5. Change it back to 'is'
6. Card is now shown as expected, as it should have after step 2

## Screenshots (optional)

https://user-images.githubusercontent.com/15314183/211534368-80988ee0-e467-45e1-807c-ddffcaeb8a9b.mov

## Edition and Platform

 - Edition: Mattermost Boards (plugin)
 - Version: v7.8
 - Browser and OS: Chrome on macOS

PR #4455, opened 2023-01-12
Update the filter condition if property changes and not supported

sbishel:
#### Summary
When a user selects a new property, the condition does not change. This results in the actual filter clause not being what is expected. This PR will update the condition to a default if the existing condition is not valid for that property.

#### Ticket Link

Fixes https://github.com/mattermost/focalboard/issues/4445


harshilsharma63 (approved):
:LGTM 👍

PR #4463, opened 2023-01-12
MM-49637 Make chunk filenames generated for MPA unique

hmhealey:
This is fixes the caching issue that Christopher reported earlier and that some others have reported over the last while as well. Basically, the product version of Boards was generating chunks with non-unique file names causing some people to get weeks old versions of Boards on community. This was never a problem with the plugin version since, even though it just generates a consistently-named `bundle.js`, the MM server would give it a unique name instead.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49637


wiggin77 (approved):
LGTM :+1:

Issue #4441, opened 2023-01-09
Bug: Difficulty accessing the board I just created


amyblais:
Original report: https://mattermost.atlassian.net/browse/MM-49532

When i click on the linked board in the channel i get this 404 message in a black popout window

<img width="1019" alt="Screen Shot 2023-01-06 at 2 36 01 PM" src="https://user-images.githubusercontent.com/29708087/211336836-b9116d42-2e38-40c6-ae85-14095d493090.png">

sbishel:
There are two issues here that need fixed. This is due to running in ProductMode. Since we are running in Product mode the plugin path no longer exists (is being created.)

`http://localhost:8065/plugins/focalboard/team/w6br7eb94jgrxpg7iobhrg56jy/b9oe5wbwjni8rud6qw1tgrg9dae`

There should be two fixes made - 
Focalboard, should build these links using the "boards" product path [here.](https://github.com/mattermost/focalboard/blob/main/mattermost-plugin/server/boards/boardsapp_util.go#L84)

Also, I believe the Channels Product should redirect "/plugins/focalboard" to "/boards", since these paths have been used in the past for some time.

sbishel:
The first fix was made, but apparently the second fix isn't necessary per @wiggin77.

PR #4467, opened 2023-01-12
Change board serverRoot from '/plugins/focalboard' to '/boards'

sbishel:
#### Summary
Changes the base URL for the server from '/plugins/focalboard' to '/boards'
This fixes the "Page not found" error in links when linking/unlinking a board. But does NOT fix the issue with Shared Boards.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/4441

### DIFF
--- a/mattermost-plugin/server/boards/boardsapp_util.go
+++ b/mattermost-plugin/server/boards/boardsapp_util.go
@@ -80,8 +80,12 @@ func createBoardsConfig(mmconfig mm_model.Config, baseURL string, serverID strin
 		showFullName = *mmconfig.PrivacySettings.ShowFullName
 	}
 
+	serverRoot := baseURL + "plugins/focalboard"
+	if mmconfig.FeatureFlags.BoardsProduct {
+		serverRoot = baseURL + "/boards"
+	}
 	return &config.Configuration{
-		ServerRoot:               baseURL + "/boards",
+		ServerRoot:               serverRoot,
 		Port:                     -1,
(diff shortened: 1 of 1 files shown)


wiggin77 (approved):

ogi-m:
@sbishel I don't think I can test this in product mode before merge, so let's do that first.

sbishel:
@mgdelacroix - Do you see any issues with this change? I don't think this overlaps with the non-authenticated urls issue, but want to make sure prior to merging.

mattermost-plugin/server/boards/boardsapp_util.go
```diff
@@ -80,8 +80,12 @@ func createBoardsConfig(mmconfig mm_model.Config, baseURL string, serverID strin
 		showFullName = *mmconfig.PrivacySettings.ShowFullName
 	}
 
+	serverRoot := baseURL + "plugins/focalboard"
```
mgdelacroix, line 83:
```suggestion
	serverRoot := baseURL + "/plugins/focalboard"
```

mgdelacroix (changes requested):
I don't think it should affect, as we're not serving the alternative webapp in product mode. Blocking on the slash comment as it breaks the plugin mode

mattermost-plugin/server/boards/boardsapp_util.go
sbishel: I knew I wanted you to look for a reason...nice catch.

mgdelacroix (approved):
LGTM! :+1:

Issue #4438, opened 2023-01-08
Bug: 


DinerGuru17:
## Steps to reproduce the behavior
I am trying to set up a focalboard on my linux server 22.04. I cloned your whole github repo to a folder on my server and ran the `docker compose up` command from the docker folder. 

The docker compose failed with the following error: 
`ERROR [gobuild 4/4] RUN EXCLUDE_PLUGIN=true EXCLUDE_SERVER=true EXCLUDE_ENTERPRISE=true make server-docker os=linux arch=amd64`

Below is the full commandline readout:

` docker compose up
[+] Building 517.1s (14/22)                                                                                                                                                           [+] Building 517.2s (14/22)
 => [nodebuild 1/4] FROM docker.io/library/node:16.3.0@sha256:ca6daf1543242acb0ca59ff425509eab7defb9452f6ae07c156893db06c7a9a4                                                 516.9ss
 => => sha256:612a5de913f35f1f4648862a6adc5847a922ae6a45de6be602cd19fef640285a 4.20kB / 4.20kB                                                                                 145.8ss
 => => sha256:b86d81a99d4163dc88ac7c8a0a8d5bc8fab559051eda227a58bf5e2406228dd3 34.46MB / 34.46MB                                                                               183.2ss
 => => sha256:5dd61d4ad9e84bebb71328a641516ac0ac01b15849f3cf4771f54493a0fbfe55 2.29MB / 2.29MB                                                                                 149.0ss
 => => sha256:7aae823459655819b0bcb791763166b6a9643b7b4ea96ef9ffd0c614b8294b31 282B / 282B                                                                                     149.4ss
 => => extracting sha256:14feb89c4a52c44f0a7e23a46ebb76e1c25c4b2915a39656a3ab634c74bcda46                                                                                        5.6ss
 => => extracting sha256:612a5de913f35f1f4648862a6adc5847a922ae6a45de6be602cd19fef640285a                                                                                        0.0ss
 => => extracting sha256:b86d81a99d4163dc88ac7c8a0a8d5bc8fab559051eda227a58bf5e2406228dd3                                                                                        1.1ss
 => => extracting sha256:5dd61d4ad9e84bebb71328a641516ac0ac01b15849f3cf4771f54493a0fbfe55                                                                                        0.1ss
 => => extracting sha256:7aae823459655819b0bcb791763166b6a9643b7b4ea96ef9ffd0c614b8294b31                                                                                        0.0ss
[...]

PR #4473, opened 2023-01-17
Bump mm-server version to v6.0.0-20230116174708-240304ad0728

wiggin77:
#### Summary
This PR updates Focalboard server's go.mod to require mattermost-server v6.0.0-20230116174708-240304ad0728

While the go.mod file is not needed for CI builds or local development work, it is used by local docker builds.  This version bump fixes an incompatibility with the older mm-server.

#### Ticket Link
https://github.com/mattermost/focalboard/issues/4438


sbishel (approved):
Any way we can automate this? Or make sure we do it when necessary. 
If you don't already, add this as an issue for mono-repo.

wiggin77:
> Any way we can automate this? Or make sure we do it when necessary.

@sbishel we should do it as part of every release.  Not sure how to automate - it's just a couple of `go get` calls.

PR #4484, opened 2023-01-18
Update configuration to use 'ProductSettings.EnablePublicSharedBoards'

### LINKED ISSUE #4481: Bug: "Enable Publicly-Shared Boards" config not working in product mode

1. Go to a server with product mode enabled
2. Sign in as a sysadmin
4. Open System Console
5. Go to Products > Boards > set "Enable Publicly-Shared Boards" to `true`
6. Go to a board where you are a board admin
7. Open the Share dialog
8. See error - Publish tab on the bottom is not visible

## Expected behavior

Publish tab is visible and you can generate a publicly shared board link. Note: This config was previously set under Plugins > Plugin Management > Installed Plugins > Boards > Settings.

## Screenshots (optional)

Screen Shot 2023-01-17 at 4 16 33 PM

Screen Shot 2023-01-17 at 4 17 53 PM

## Edition and Platform

 - Version: v7.8

sbishel:
#### Summary
Check the ProductSettings.EnablePublicSharedBoards config setting.

Did not remove original check, so not feature flag check is needed.

#### Ticket Link

  Fixes #4481

### DIFF
--- a/mattermost-plugin/server/boards/configuration.go
+++ b/mattermost-plugin/server/boards/configuration.go
@@ -80,6 +80,9 @@ func (b *BoardsApp) OnConfigurationChange() error {
 	if mmconfig.PluginSettings.Plugins[PluginName][SharedBoardsName] == true {
 		enableShareBoards = true
 	}
+	if mmconfig.ProductSettings.EnablePublicSharedBoards != nil {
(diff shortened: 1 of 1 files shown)


wiggin77 (approved):
If we're confident the plugin does not need to be supported going forward this will work. Otherwise we would need to check if in plugin mode and fetch the setting from the correct place.

PR #4485, opened 2023-01-18
Gh 4482 fix empty filters

### LINKED ISSUE #4482: Bug: Date filter "is before" and "is after" modifiers do not default to no filters applied in empty state

1. Go to any board with a date property
2. Go to filters and select the date property to filter by
3. Click through each modifier without specifying any dates
4. See error when selecting `is before` and `is after` - cards are hidden as if filters are being applied, even though the default behavior should be as if there are no filters applied.

## Expected behavior

Cards are not hidden unless a value is specified for those modifiers.

## Screenshots (optional)

https://user-images.githubusercontent.com/93531870/213018234-b80f7168-0b29-4219-8b9a-becc27cec73a.mov

## Edition and Platform

 - Version: v7.8

sbishel:
#### Summary

Fix for if the filter.Value is not set, don't filter.

#### Ticket Link

  Fixes #4482

### DIFF
--- a/webapp/src/cardFilter.test.ts
+++ b/webapp/src/cardFilter.test.ts
@@ -280,6 +280,10 @@ describe('src/cardFilter', () => {
         }
 
         test('verify isBefore clause', () => {
+            const filterClauseIsBeforeEmpty = createFilterClause({propertyId: 'datePropertyID', condition: 'isBefore', values: []})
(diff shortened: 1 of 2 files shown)


ogi-m (approved):
LGTM!

harshilsharma63 (approved):
LGTM 👍

Issue #3846, opened 2022-09-15
Bug: Multiple console errors when on the first tour point


asaadmahmood:
## Steps to reproduce the behavior

1. Create a new user
2. Go take a tour
3. Notice bunch of console errors when the board loads with the first tour point

## Expected behavior
No console errors

## Screenshots (optional)
<img width="1045" alt="Screenshot 2022-09-15 at 1 27 06 PM" src="https://user-images.githubusercontent.com/11034289/190355232-b38ea233-264c-427a-b9d5-011aa0fd5bfe.png">

## Edition and Platform
 - Edition: Mattermost Boards (plugin)
 - Version: 7.3
 - Browser and OS: Chrome

## Additional context (optional)

Add any other context about the problem here, and any notes about the severity:
* Sev 3: Affects minor, non-critical functionality

sbishel:
This should get fixed when we update templates.

PR #4486, opened 2023-01-18
Suppress error message on unfound option id

sbishel:
#### Summary
Several of our templates have been created with different options over time. That has resulted in some invalid date in the templates and missing option ids. Currently, we display an error when an option id isn't found. However, there is no action to take and this is a benign error, this PR removes the error.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/3846


wiggin77 (approved):
LGTM

PR #4496, opened 2023-01-19
Use GetDirectChannelOrCreate when notifying block chg in MPA mode

wiggin77:
#### Summary
There is a bug in Boards notification system in MPA mode. When a notification is sent, it first gets or creates a channel between the Boards bot and the user to be notified. Previously this used the GetDirectChannel method in the plugin API. That method gets or creates the channel. The product API version of that call only gets the channel and will not create it if not found. 

This PR fixes the issue by calling the newly added GetDirectChannelOrCreate product API.  Requires the [matching mm-server PR](https://github.com/mattermost/mattermost-server/pull/22112) to be merged first.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49752


sbishel (approved):
LGTM

Issue #4488, opened 2023-01-19
Bug: Bot notification on deleting attachment


Rajat-Dabade:
## Steps to reproduce the behaviour

1. Go to the board and open a card
2. Add an attachment to the card
3. Now delete the attachment added to the card
4. See an error, no bot notification is sent in channels

Also applies to text blocks, checklist items, and images.

## Expected behaviour
Bot notifications should be sent after removing the attachment from the card.

## Edition and Platform

 - Edition:  Mattermost Boards (product mode)
 - Version: [latest-main]
 - Browser and OS: [e.g. Chrome on Mac, Edge on Windows]

Rajat-Dabade:
cc @wuwinson @wiggin77

PR #4508, opened 2023-01-24
Fix notifications for deleted card content blocks

wiggin77:
#### Summary
When the blocks table started hard deleting blocks, it broke notifications for content block deletions.  This PR fixes that by:

1.  adding new store method to fetch the current version of all card content blocks from history
2. refactor notification code to handle the missing block records in blocks table
3. handle image and file attachment deletion with new text

#### Ticket Link
fixes https://github.com/mattermost/focalboard/issues/4488


server/model/blocktype.go
```diff
@@ -14,13 +14,16 @@ import (
...
+	TypeUnknown    = "unknown"
+	TypeBoard      = "board"
+	TypeCard       = "card"
+	TypeView       = "view"
+	TypeText       = "text"
+	TypeCheckbox   = "checkbox"
+	TypeComment    = "comment"
+	TypeImage      = "image"
+	TypeAttachment = "attachment"
+	TypeDivider    = "divider"
```
mgdelacroix, line 26: <3

server/model/blocktype.go
```diff
@@ -42,6 +45,10 @@ func BlockTypeFromString(s string) (BlockType, error) {
 		return TypeComment, nil
 	case "image":
 		return TypeImage, nil
+	case "attachment":
+		return TypeAttachment, nil
+	case "divider":
+		return TypeDivider, nil
```
mgdelacroix, line 51: I think we're missing `checkbox` here

server/model/blocktype.go
```diff
@@ -57,6 +64,8 @@ func BlockType2IDType(blockType BlockType) utils.IDType {
 		return utils.IDTypeView
 	case TypeText, TypeComment:
 		return utils.IDTypeBlock
+	case TypeImage, TypeAttachment:
```
mgdelacroix, line 67: Same here

server/services/store/sqlstore/blocks.go
```diff
@@ -42,27 +43,31 @@ func (s *SQLStore) timestampToCharField(name string, as string) string {
 	}
 }
 
-func (s *SQLStore) blockFields() []string {
+func (s *SQLStore) blockFields(alias string) []string {
```
mgdelacroix, line 46: nit: we have a similar but slightly different mechanism on the `boards` store, I'd suggest to pick one and use it on everywhere for consistency.

As for which one to choose, I'm leaning more towards the one proposed on this PR

wiggin77: Thanks for reviewing @mgdelacroix.  I modified the boards store to be consistent.

server/model/blocktype.go
wiggin77: Fixed

server/model/blocktype.go
wiggin77: Fixed

Rajat-Dabade (approved):
LGTM 👍 , Nice work @wiggin77

mgdelacroix (approved):
Awesome, thanks @wiggin77 for taking the time to apply the changes!

PR #4518, opened 2023-01-27
Status property edit dialog

harshilsharma63:
#### Summary
Added the frontend implementation of generic status property configuration dialogue.

For testing purposes, you can open the dialogue by pressing `Cmd/Ctrl + Shift + I`. This key binding will be removed once this PR is tested and merged.

Change summary-
1. Implemented status property dialogue component `EditStatusPropertyDialog`.
2. Added types for status property configuration. These are pretty basic and can be easily changed as per the actual implementation of the feature if need be.
3. Created an `ActionDialog` component which is the already existing dialogue with configurable `Cancel` and `Save` buttons whose action and text are customizable. I created this component as we use this pattern in many places and add the UI for the buttons each time.
4. Refactored the existing `Dialog` component to be an `ActionDialog` with its `Cancel` and `Save` button (footer are) hidden. 

#### Ticket Link
https://community.mattermost.com/boards/team/qghzt68dq7bopgqamcnziq69ao/b4ctmm917q3rzdfi451qtuxjzcw/vuswsftt1rb873datosmmbtwkpa/c3fjo9t5e9pygxyy78psym3a3za

#### Pending Tasks
1. The tooltip in header of each column (todo, in progress and completed) is not visible as our tooltip mechanism doesn't work with `overflow: hidden` elements if the tooltip falls in the overflow zone. Need to experiment with `position: fixed` or a portal to render and position the tooltip. Also, the tooltip doesn't allow for long texts to be overflown automatically, so that's something to be added as well.
2. Test on dark mode theme.
3. Add tests.

All pending tasks completed.


webapp/src/widgets/buttons/button.tsx
```diff
@@ -5,6 +5,9 @@ import React from 'react'
 import './button.scss'
 import {Utils} from '../../utils'
 
+export type ButtonSize = 'xsmall' | 'small' | 'medium' | 'large'
+export type Emphasis = 'primary' | 'secondary' | 'tertiary' | 'quaternary' | 'gray' | 'danger' | 'default' | 'link'
```
harshilsharma63, line 9: This is not needed for this PR and is parely an enhancement. By creating a type of allowed values, the IDE can directly show allowed properties in autocomplete suggestions.

harshilsharma63:
Please merge #4580 into this before merging this into `main`.

asaadmahmood (changes requested):
Not sure if and when we'll be doing this, so taking this off of my PR to review list by requesting a review.

PR #4533, opened 2023-01-31
Improves startup logs

mgdelacroix:
#### Summary
This PR updates the `boardsapp.Start` method to receive the mode of the starting backend, and logs it on start. In the case of the product, logic has been added to detect why the product is starting, if the environment variable is enabled or the startup process was triggered by the feature flag only.


wiggin77 (changes requested):
I'll wait until the small changes we discussed are ready.

mgdelacroix:
@sbishel re-requesting review as the approach has changed

wiggin77 (approved):
LGTM 👍

sbishel (approved):

Issue #4475, opened 2023-01-17
Bug: Cannot set date easily from table view


harshilsharma63:
## Steps to reproduce the behavior

1. Create a board with a table view and add a date property to a card.
2. In the table view, try to set the date range on the date property from the table itself, i.d., without opening the card.
3. Notice how the date picker closes as soon as you select the start date, or toggle the range toggle.

Note that this works fine from inside the card.

## Expected behaviour

The date picker should stay open until both the start and end dates have been selected.

## Screenshots (optional)

[If applicable, add screenshots or a screen recording to elaborate on the problem.](https://user-images.githubusercontent.com/18575143/212867246-35521386-3131-451c-ba56-a26b43e26ffa.mp4)

## Edition and Platform

 - Edition: Mattermost Boards
 - Version: v7.8.0
 - Browser and OS: Firefox on MacOS

wuwinson:
Should be consistent with card view - date picker modal does not close until you click away from it.

PR #4535, opened 2023-01-31
GH-4475 - Save date change to database when closing form

sbishel:
#### Summary
When updating the date, the caller's <Table> object gets recreated causing the modal to close.

Rather than update the property value with every change, only update once the modal window closes. This keeps the modal open until date changes are complete. 

This works similar to a text field which doesn't update until the changes are complete, it doesn't update with each keystroke. 

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/4475


harshilsharma63 (approved):
Works great now! LGTM 👍

PR #4536, opened 2023-01-31
Moved `mattermost-plugin-api` dependency from `model` to `utils`

levb:
#### Summary
Moved `mattermost-plugin-api` dependency from `model` to `utils` to resolve a cyclical dependency in `mattermost-server/plugin`

#### Ticket Link
None

### DIFF
--- a/server/model/error.go
+++ b/server/model/error.go
@@ -7,9 +7,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/mattermost/focalboard/server/utils"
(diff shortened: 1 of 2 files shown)


wiggin77 (approved):
LGTM :+1:

levb:
NM, it does not resolve my issue, merely pushes it down a level. Will think of a different solution - is there a better low-level package to host `IsErrNotFound` so `mattermost-server` does not depend on `pluginapi`?

PR #4539, opened 2023-02-01
MM-48246 - ab test open RHS with linked board; add telemetry info

pvev:
#### Summary
Add telemetry information everytime a board card is clicked from the channels RHS boards list.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48246

### DIFF
--- a/mattermost-plugin/webapp/src/components/rhsChannelBoardItem.tsx
+++ b/mattermost-plugin/webapp/src/components/rhsChannelBoardItem.tsx
@@ -6,7 +6,6 @@ import {FormattedMessage, useIntl} from 'react-intl'
 import mutator from '../../../../webapp/src/mutator'
 import {Utils} from '../../../../webapp/src/utils'
 import {getCurrentTeam} from '../../../../webapp/src/store/teams'
-import {getCurrentChannel} from '../../../../webapp/src/store/channels'
 import {createBoard, Board} from '../../../../webapp/src/blocks/board'
 import {useAppSelector} from '../../../../webapp/src/store/hooks'
 import IconButton from '../../../../webapp/src/widgets/buttons/iconButton'
@@ -38,14 +37,9 @@ const RHSChannelBoardItem = (props: Props) => {
         return null
     }
 
-    const currentChannel = useAppSelector(getCurrentChannel)
-    if (!currentChannel) {
-        return null
-    }
-
     const handleBoardClicked = (boardID: string) => {
         // send the telemetry information for the clicked board
-        const extraData = {teamID: team.id, channelID: currentChannel.id, board: boardID}
+        const extraData = {teamID: team.id, board: boardID}
         TelemetryClient.trackEvent(TelemetryCategory, TelemetryActions.ClickChannelsRHSBoard, extraData)
 
         window.open(`${windowAny.frontendBaseURL}/team/${team.id}/${boardID}`, '_blank', 'noopener')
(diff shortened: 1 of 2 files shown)


pvev:
@dontoisme @sbishel this is still a draft pr just to get early feedback of the intention to add telemetry information for every board card click from the channels RHS. @dontoisme would this telemetry data fulfill what you would expect. Thanks!

sbishel (approved):

webapp/src/telemetry/telemetryClient.ts
```diff
@@ -50,6 +50,7 @@ export const TelemetryActions = {
     LimitCardLimitReached: 'limit_cardLimitReached',
     LimitCardLimitLinkOpen: 'limit_cardLimitLinkOpen',
     VersionMoreInfo: 'version_more_info',
+    ClickChannelsRHSBoard: 'click_channels_RHS_board',
```
catalintomai, line 53: see https://docs.google.com/document/d/1CJeMFxnwUq8njKd3S3zxnfqrwUz8zNDsTmUAYqgy_uo/edit#heading=h.gmxxmyi0tggz for event naming recommendations - sometimes useful to add "result of action" if one would have several actions on the same object.

JulienTant (approved):

webapp/src/telemetry/telemetryClient.ts
pvev: @dontoisme do you have any suggestion for the event name?

pvev: @marianunez @esethna this is the telemetry we want to add for every time a user clicks a board card in the RHS panel (aka boards plugin). Do you have any suggestions for the name or the extra data information we should capture here?

mattermost-plugin/webapp/src/components/rhsChannelBoardItem.tsx
```diff
@@ -36,6 +38,10 @@ const RHSChannelBoardItem = (props: Props) => {
     }
 
     const handleBoardClicked = (boardID: string) => {
+        // send the telemetry information for the clicked board
+        const extraData = {teamID: team.id, board: boardID}
+        TelemetryClient.trackEvent(TelemetryCategory, TelemetryActions.ClickChannelsRHSBoard, extraData)
```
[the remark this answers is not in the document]
pvev: I am not aware of it. Maybe @sbishel or @harshilsharma63 might be able to help with this info.

webapp/src/telemetry/telemetryClient.ts
marianunez: I am not sure if there is always a clear result of action that can be added to the telemetry as the action could be a UI interaction that can cause several effects (and could change with the iteration of the product, i.e. in this case the click of the board in the RHS in channels opens the board in Boards but in the future can open some other view in the RHS)

I would suggest a slight variation on this just to make it more readable but don't have a good result of action to suggest: 
```suggestion
    ClickChannelsRHSBoard: 'click_board_in_channels_RHS',
```

mattermost-plugin/webapp/src/components/rhsChannelBoardItem.tsx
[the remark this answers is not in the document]
pvev: @ifoukarakis maybe this is what you mean? https://github.com/mattermost/focalboard/blob/b1fd48faacff27160d6e75fb965a75c9a8c27ca8/mattermost-plugin/webapp/src/index.tsx#L393

JulienTant (approved):

Issue #4532, opened 2023-01-31
Bug: import from trello


lexoyo:
## Steps to reproduce the behavior

1. follow the steps from the docs: https://docs.mattermost.com/boards/migrate-to-boards.html#import-from-trello
1. error: 

```sh
$ npx ts-node importTrello.ts -i ~/Downloads/trello-i2k.json -o ./archive.boardarchive              

/home/lexoyo/Documents/tests/focalboard/import/trello/node_modules/ts-node/src/index.ts:513
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
../../webapp/src/blocks/board.ts:4:24 - error TS2307: Cannot find module 'lodash/difference' or its corresponding type declarations.

4 import difference from 'lodash/difference'
                         ~~~~~~~~~~~~~~~~~~~

    at createTSError (/home/lexoyo/Documents/tests/focalboard/import/trello/node_modules/ts-node/src/index.ts:513:12)
    at reportTSError (/home/lexoyo/Documents/tests/focalboard/import/trello/node_modules/ts-node/src/index.ts:517:19)
    at getOutput (/home/lexoyo/Documents/tests/focalboard/import/trello/node_modules/ts-node/src/index.ts:752:36)
    at Object.compile (/home/lexoyo/Documents/tests/focalboard/import/trello/node_modules/ts-node/src/index.ts:968:32)
    at Module.m._compile (/home/lexoyo/Documents/tests/focalboard/import/trello/node_modules/ts-node/src/index.ts:1056:42)
    at Module._extensions..js (node:internal/modules/cjs/loader:1272:10)
    at Object.require.extensions.<computed> [as .ts] (/home/lexoyo/Documents/tests/focalboard/import/trello/node_modules/ts-node/src/index.ts:1059:12)
    at Module.load (node:internal/modules/cjs/loader:1081:32)
    at Function.Module._load (node:internal/modules/cjs/loader:922:12)
    at Module.require (node:internal/modules/cjs/loader:1105:19)
```

## Expected behavior

Create ./archive.boardarchive

## Edition and Platform

 - Edition: Main branch of focalboard, local fedora
 - Version: git hash 08ce9a753d77eca18ae385ac09dceb246f955254
 - Browser and OS: nodejs v18.13.0 (tried also v16.13.1) on fedora

## Additional context (optional)

I tried with a trello board I just created now and it does the same as with a non-empty/old board

Here is the json: 
[...]

Issue #4524, opened 2023-01-28
Bug: Import Notion


atah:
Hi, 

I use : 

1. From a Notion Board, open the ... menu at the top right
2. Select `Export`, pick `Markdown & CSV` as the export format, select true to include subpages.
3. Save it locally, and unzip the folder e.g. to `notion-export`
4. Run `npm install` from within `focalboard/webapp`
5. Run `npm install` from within `focalboard/import/notion`
6. Run `npx ts-node importNotion.ts -i <path to the notion-export folder> -o archive.boardarchive`
7. In Focalboard, click `Settings`, then `Import archive` and select `archive.boardarchive`

But I have a error : 

Error: EISDIR: illegal operation on a directory, read
    at Object.readSync (node:fs:737:3)
    at tryReadSync (node:fs:432:20)
    at Object.readFileSync (node:fs:478:19)
    at getMarkdown (/Users/axelpichon/Downloads/focalboard-main/import/notion/importNotion.ts:105:29)
    at /Users/axelpichon/Downloads/focalboard-main/import/notion/importNotion.ts:200:26
    at Array.forEach (<anonymous>)
    at convert (/Users/axelpichon/Downloads/focalboard-main/import/notion/importNotion.ts:155:11)
    at main (/Users/axelpichon/Downloads/focalboard-main/import/notion/importNotion.ts:74:30)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)

Can you help me please ?

wuwinson:
@ogi-m Can you try to reproduce this issue? Thanks! cc @sbishel

sbishel:
So, I just tried it myself and it took me a couple tries to get it right. 
The command expects the `-i` path to be to a directory containing the csv file. 

`npx ts-node importNotion.ts -i ~/Downloads/notion -o archive2.boardarchive`

<img width="652" alt="Screen Shot 2023-02-01 at 1 29 07 PM" src="https://user-images.githubusercontent.com/12704875/216156251-0102b36c-e9f7-420c-b294-7efea49736e6.png">

wuwinson:
@atah Can you confirm if this is still happening after trying @sbishel suggestions? Thanks!

sbishel:
@atah - closing, please reopen if still an issue.

PR #4540, opened 2023-02-01
add admin user on import

sbishel:
#### Summary
Only users in the import file are being created. The importing user should always be added as an admin to the imported boards.

#### Ticket Link
Found while investigating...
#4532 and #4524


Rajat-Dabade (approved):
LGTM 👍, Nice work 🚀

PR #4541, opened 2023-02-02
Add a test to help detect potential sync issues with work template

JulienTant:
#### Summary
I didn't really want to do this PR but this is how it is 😬 Because this is kind of a "first", I opened a thread here: https://community.mattermost.com/core/pl/t7d16sx83tnw7phhh8bp9h844a

Work templates rely on the fact that certain boards templates exist, but we have no way to guarantee this on the mattermost-server/webapp side. This test acts as a warning that something might break in work templates because a templates is not reachable.

#### Ticket Link
N/A

### DIFF
--- a/server/integrationtests/work_template_test.go
+++ b/server/integrationtests/work_template_test.go
@@ -8,7 +8,7 @@ import (
 
 // This test is there to guarantee that the board templates needed for
 // the work template are present in the default templates.
-// If this fails, you might need to sync with the channls team.
+// If this fails, you might need to sync with the channels team.
 func TestGetTemplatesForWorkTemplate(t *testing.T) {
 	// map[name]trackingTemplateId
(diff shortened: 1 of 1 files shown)


server/integrationtests/work_template_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// This test is there to guarantee that the board templates needed for
+// the work template are present in the default templates.
+// If this fails, you might need to sync with the channls team.
```
wiggin77, line 11:
```suggestion
// If this fails, you might need to sync with the channels team.
```

wiggin77 (approved):
LGTM :+1:   The trackingTemplateId should be stable (that's what it is for).

PR #4557, opened 2023-02-07
Use `name` column to determine if schema_migrations table is correct format

wiggin77:
#### Summary
A number of support tickets have been due to full migrations being run on existing Sqlite databases when upgrading.  There are only a few causes for the server to think a complete set of migrations is needed.  

1.  The schema_migrations table is empty
2.  The schema_migrations table has a column named `dirty`, forcing a schema migration for Morph 
3.  The schema_migrations table is missing

Point 3 would result in an error being logged.  

The most likely cause is item 2 since we have observed issues deleting columns in Sqlite.  Instead of checking for a column named `dirty` to ensure it was deleted, this PR modifies the check to look for a column named `name` to determine if a schema migration is needed on the schema_migrations table.  The `name` column was added for Morph, and is more reliable than ensuring a column was deleted.

#### Ticket Link
various

### DIFF
--- a/mattermost-plugin/server/manifest.go
+++ b/mattermost-plugin/server/manifest.go
@@ -45,8 +45,7 @@ const manifestStr = `
         "type": "bool",
         "help_text": "This allows board editors to share boards that can be accessed by anyone with the link.",
         "placeholder": "",
-        "default": false,
-        "hosting": ""
+        "default": false
       }
     ]
   }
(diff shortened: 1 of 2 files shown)


wiggin77:
@mgdelacroix ping

mgdelacroix (approved):
LGTM! :+1:

PR #4411, opened 2022-12-29
[MM-49159] Fix status dropdown menu avatar on click

kostaspt:
#### Summary
Removed the code reverting the position and excluded the status dropdown menu altogether as it's the only `.Menu` that could be affected inside `.focalboard-body`, but also the only selectors targeting `.Menu` globally in general. This may be due to a recent unrelated change, but the dropdown works as expected without the boards' plugin loaded, so it looks like this is affecting it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49159

PR #4570, opened 2023-02-10
use specific class, rather than NOT an unrelated ID

sbishel:
#### Summary
A css change for another issue(https://github.com/mattermost/focalboard/pull/4411) started using an ID in a NOT() operation which caused the css specificity to change. Because of that change the "shareboard.css" gets overwritten rather than being the final css.

This PR changes to use the specific classes used by menu.tsx, which allows the specificity to remain the same. 

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/pull/4411

### DIFF
--- a/webapp/src/widgets/menu/menu.scss
+++ b/webapp/src/widgets/menu/menu.scss
@@ -1,6 +1,6 @@
 @import '../../styles/z-index';
 
-.Menu:not(#statusDropdownMenu) {
+.Menu.noselect {
     @include z-index(menu);
     display: flex;
     flex-direction: column;
@@ -173,7 +173,7 @@
     }
 }
 
-.Menu:not(#statusDropdownMenu),
+.Menu.noselect,
 .SubMenuOption .SubMenu {
     @media screen and (max-width: 430px) {
         position: fixed;


sbishel:
@asaadmahmood - Please let me know if you have a better fix, but want to fix for v7.8, which is already ready for release. Thanks.

asaadmahmood (approved):
@sbishel this solution is better than using IDs, but I haven't tested out all the menus in the app that use the `noselect` class, so if it's working fine, I don't have a problem with this.

Previously the code was targetting all menus except `#statusDropdownMenu`

Now it's selecting every menu that has a class `noselect`.

So ideally to every menu that's not `#statusDropdownMenu` should have this class, I'm not sure if that's the case.

sbishel:
@asaadmahmood  - Thanks.. that's exactly why I wanted you to take a look. Looking at the menu.tsx file [here](https://github.com/mattermost/focalboard/blob/main/webapp/src/widgets/menu/menu.tsx#L56) it appears all of these <Menu> objects receive both classes, which is why I went this way.

I also searched to see if we use 'menu' as a class name in any other place. We don't, that I could find.  

I hate making css changes this late, but this fixes a pretty ugly issue.

PR #4577, opened 2023-02-14
Board teamless file path

harshilsharma63:
#### Summary
Currently, we're storing files in the path `/<teamID>/<boardID>/<filename>.<extension>`. This would be an issue when we decide to add the ability to move cards across boards or boards across teams since the path would become invalid on the change of `boardID` or `teamID`.
This PR updates the file path to `/boards/<YYYYMMDD>/<filename>.<extenstion>` (the date format is derived from Mattermost server's file storage dir path convention.
I'm storing the path in `FileInfo.Path` when uploading the file. WHen retrieving a file, I check if `FileInfo.Path` is available, and if so, use that path to fetch the file from filesystem, else fallback to the old `/<teamID>/<boardID>` path.
I'm not moving all existing files to the new location as it would be a long operation due to disk IO being involved for local file storage, and the network is involved in the case of AWS S3 storage, which could negatively impact server startup time.

#### Ticket Link
https://community.mattermost.com/boards/team/qghzt68dq7bopgqamcnziq69ao/b4ctmm917q3rzdfi451qtuxjzcw/vuswsftt1rb873datosmmbtwkpa/c3gyhb8ard7d37mwqw6gbxj1emc


harshilsharma63:
@sbishel @mgdelacroix this is now ready to review.

server/api/files.go
```diff
@@ -123,37 +123,12 @@ func (a *API) handleServeFile(w http.ResponseWriter, r *http.Request) {
...
 	auditRec.AddMeta("filename", filename)
 
-	fileInfo, err := a.app.GetFileInfo(filename)
+	fileInfo, fileReader, err := a.app.GetFile(board.TeamID, boardID, filename)
 	if err != nil && !model.IsErrNotFound(err) {
 		a.errorResponse(w, r, err)
 		return
 	}
 
-	if fileInfo != nil && fileInfo.Archived {
```
harshilsharma63, line 132: Note to the reviewer- we no longer need this if block as this handled the file limits and we no longer have limits.

server/client/client.go
```diff
@@ -380,6 +380,8 @@ func (c *Client) GetCards(boardID string, page int, perPage int) ([]*model.Card,
 		return nil, BuildErrorResponse(r, err)
 	}
 
+	defer closeBody(r)
```
harshilsharma63, line 383: Note to reviewer - these are added to fix linter complaint.

sbishel (approved):

PR #4596, opened 2023-02-22
Remove old desktop apps

mgdelacroix:
#### Summary
Removes source code folders for personal server and personal desktop and the related make targets

PR #4599, opened 2023-02-23
Remove sqlite

mgdelacroix:
#### Summary
Removes all source references to SQLite, its related dependencies, build flags and targets and CI related jobs. It updates the dependencies to point to a newer MM version as well, which will not be necessary for the monorepo but was required to be able to tidy the `go.mod`

The PR branch sits on top of https://github.com/mattermost/focalboard/pull/4596, so it should be merged after it


wiggin77 (approved):
:+1: :+1: :+1:

PR #4612, opened 2023-03-02
[GH-4476] Bug: No card delete confirmation status for some cards 

### LINKED ISSUE #4476: Bug: No card delete confirmation status for some cards

1. Create a new card. Add some properties to it and set their values. Add some comments in there as well. Do not set the card title.
2. Go back to kanban view and try deleting the card.
3. Notice the UI didn't show any delete confirmation for this card.
4. Create a new card and repeat step 1 but this time also set a card title.
5. Go back to kanban view and try deleting the card. Notice that this time you see a delete confirmation.

## Expected behavior

Card delete confirmation should show up if there is any data in the card.

## Edition and Platform

 - Version: v7.8.0
 - Browser and OS: Firefox on MacOS

## Additional context

The log in webapp is probably if the card has no title and no content, don't show the delete confirmation dialogue. It does not take into consideration the properties and comments in the card.

JesusMurguia:
#### Summary

I added checks for any content on the card (title, properties, comments, attachments, description).

#### Ticket Link

Fixes #4476

### DIFF
--- a/webapp/src/components/cardDialog.tsx
+++ b/webapp/src/components/cardDialog.tsx
@@ -109,7 +109,7 @@ const CardDialog = (props: Props): JSX.Element => {
         // use may be renaming a card title
         // and accidently delete the card
         // so adding des
-        if (card?.title === '' && card?.fields.contentOrder.length === 0) {
+        if (card?.title === '' && card?.fields.contentOrder.length === 0 && Object.keys(card?.fields?.properties).length === 0 && attachments?.length === 0 && comments?.length === 0) {
             handleDeleteCard()
             return
         }
--- a/webapp/src/components/kanban/kanbanCard.tsx
+++ b/webapp/src/components/kanban/kanbanCard.tsx
@@ -20,6 +20,9 @@ import OpenCardTourStep from '../onboardingTour/openCard/open_card'
 import CopyLinkTourStep from '../onboardingTour/copyLink/copy_link'
 import CardActionsMenu from '../cardActionsMenu/cardActionsMenu'
 import CardActionsMenuIcon from '../cardActionsMenu/cardActionsMenuIcon'
+import {getCardAttachments} from '../../store/attachments'
+import {getCardComments} from '../../store/comments'
+import {useAppSelector} from '../../store/hooks'
 
 export const OnboardingCardClassName = 'onboardingCard'
 
@@ -38,6 +41,8 @@ type Props = {
 
 const KanbanCard = (props: Props) => {
     const {card, board} = props
+    const comments = useAppSelector(getCardComments(card?.id))
+    const attachments = useAppSelector(getCardAttachments(card?.id))
     const intl = useIntl()
     const [isDragging, isOver, cardRef] = useSortable('card', card, !props.readonly, props.onDrop)
     const visiblePropertyTemplates = props.visiblePropertyTemplates || []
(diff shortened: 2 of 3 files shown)


sbishel (changes requested):
Thanks, @JesusMurguia Let's create a common function for these to call to ensure consistency. It appears that the same changes are also necessary in `galleryCard.tsx` and `fullCalendar.tsx`

JesusMurguia:
Hello @sbishel, I noticed the `fullCalendar.tsx` logic is really different from the other files and I can't find a way to get the comments and attachments for the card because of the rules of hooks and stuff, I am afraid i have much experience with that, could you give me some advice on that?

sbishel:
> Hello @sbishel, I noticed the `fullCalendar.tsx` logic is really different from the other files and I can't find a way to get the comments and attachments for the card because of the rules of hooks and stuff, I am afraid i don’t have much experience with that, could you give me some advice on that?

Hi @JesusMurguia, sure I can try to provide some guidance. If you dig in to why the `fullCalendar` is different. It is essentially because the other uses (`kanbanCard` / `galleryCard`) are dealing with a single card. `fullCalendar` contains all the cards with logic for displaying individual cards. 

I think the best (and probably easiest) way to handle this would be to create a similar component to `kanbanCard`...may-be `calendarCard`. Then the implementations would be able to be very similar to how it is handled in other places.

A further enhancement would then be to create a common `card` component that all the components could share. Not expecting you to do this, just pointing out that all three of these components will be very similar and could be handled by a common component.

webapp/src/store/cards.ts
```diff
@@ -168,6 +168,12 @@ export function getCard(cardId: string): (state: RootState) => Card|undefined {
     }
 }
 
+export function isCardEmpty(cardId: string): (state: RootState) => boolean {
+    return (state: RootState): boolean => {
+        return getCards(state)[cardId]?.title === '' && getCards(state)[cardId]?.fields.contentOrder.length === 0 && Object.keys(getCards(state)[cardId]?.fields?.properties).length === 0 && !state.comments.commentsByCard[cardId] && !state.attachments.attachmentsByCard[cardId]
+    }
+}
+
```
sbishel, line 176: Suggest using existing selectors to retrieve data vs accessing state directly.

JesusMurguia:
Got it, thank you. I will work on this

webapp/src/store/cards.ts
JesusMurguia: I moved the `isCardEmpty` function into a hook and used the selectors there in my latest commit, please let me know if this is okay

sbishel (commented):
Very nice...code looks good. Can you add unit tests where appropriate?

JesusMurguia:
I added the unit tests for the new component

lindy65:
Hi @JesusMurguia @sbishel 

Please create a PR in mattermost server repo so I can create a test server to test this on?

Thank you :)

sbishel (approved):
Thanks @JesusMurguia.  Nice work.

sbishel:
@lindy65 - I created a cloud server and uploaded the plugin built here (since we are back in plugin mode.) Normal cloud logins.

https://sbishel-4476.test.mattermost.cloud/

lindy65 (approved):
Thanks @JesusMurguia - this issue is fixed. I tested on the server Scott spun up and tested according to steps on https://github.com/mattermost/focalboard/issues/4476.  Delete confirmation modal is present.
cc @sbishel

Issue #4637, opened 2023-03-13
Bug: Docker Build v7.8.2 Fails


johncadengo:
## Steps to reproduce the behavior

1. Git clone
2. Check out v7.8.2 by tag
3. Run `docker build -f docker/Dockerfile .`
4. Error on step 10/23

```
Step 10/23 : RUN EXCLUDE_PLUGIN=true EXCLUDE_SERVER=true EXCLUDE_ENTERPRISE=true make server-docker os=${TARGETOS} arch=${TARGETARCH}
 ---> Running in b93f028f5526
go run ./build/gowork/main.go
go.work written successfully.
go 1.18

use ./server

mkdir -p bin/docker
cd server; env GOOS= GOARCH= go build -ldflags '-X "github.com/mattermost/focalboard/server/model.BuildNumber=dev" -X "github.com/mattermost/focalboard/server/model.BuildDate=n/a" -X "github.com/mattermost/focalboard/server/model.BuildHash=1e35878351f9ab75fe6fd9c7562fb4cfdc566027" -X "github.com/mattermost/focalboard/server/model.Edition=linux"' -tags 'json1 sqlite3' -o ../bin/docker/focalboard-server ./main
go: downloading github.com/mattermost/mattermost-server/v6 v6.0.0-20230116174708-240304ad0728
go: downloading github.com/pkg/errors v0.9.1
go: downloading github.com/oklog/run v1.1.0
go: downloading github.com/gorilla/mux v1.8.0
go: downloading github.com/spf13/viper v1.10.1
go: downloading github.com/stretchr/testify v1.8.1
go: downloading github.com/rivo/uniseg v0.4.3
go: downloading github.com/mattermost/mattermost-plugin-api v0.1.1
go: downloading github.com/rudderlabs/analytics-go v3.3.3+incompatible
go: downloading github.com/wiggin77/merror v1.0.4
go: downloading golang.org/x/crypto v0.5.0
go: downloading github.com/gorilla/websocket v1.5.0
go: downloading github.com/prometheus/client_golang v1.12.1
go: downloading github.com/mattermost/morph v1.0.5-0.20221115094356-4c18a75b1f5e
go: downloading github.com/Masterminds/squirrel v1.5.3
go: downloading github.com/lib/pq v1.10.7
go: downloading github.com/mattn/go-sqlite3 v2.0.3+incompatible
go: downloading github.com/krolaw/zipstream v0.0.0-20180621105154-0a2661891f94
[...]

PR #4654, opened 2023-03-21
copy safe writeFileResponse into personal server/desktop

sbishel:
#### Summary
Copy writeFileResponse code from server back to boards code base.
This code is safer than previous implementation, but we don't want to depend on the server code for this, this change is not needed in plugin implementation. 
Also fixes an issue displaying images from templates. 

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/4637


wiggin77 (approved):
LGTM :+1:

harshilsharma63 (approved):
LGTM 👍

PR #4671, opened 2023-03-28
update ubuntu to latest

sbishel:
#### Summary
Github deprecating ubuntu 18.04, use latest.

https://github.com/actions/runner-images/issues/6002
 > Workflows using the ubuntu-18.04 image label should be updated to ubuntu-latest, ubuntu-20.04, or ubuntu-22.04

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ env:
 jobs:
 
   ci-ubuntu-server:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
 
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


sbishel:
Not sure if should use latest or next available version to minimize potential issues. Tests run ok.

wiggin77:
> Not sure if should use latest or next available version to minimize potential issues. Tests run ok.

I believe `lts` works here, which will bring down the latest LTS, which is what customers would be using versus the interim versions.

wiggin77 (approved):
LGTM :+1:

mgdelacroix (approved):
I'm happy using latest, LGTM! :+1:

Issue #4644, opened 2023-03-14
Bug: Sidebar menu items jittering


johncadengo:
## Steps to reproduce the behavior

1. Log in to dashboard
2. Second user logs into dashboard

## Expected behavior

I expect to be able to see the sidebar menu items

## Screenshots (optional)

https://user-images.githubusercontent.com/561587/225161299-1b7404c2-730d-4fa4-99c3-4cf498832cf8.mov

## Edition and Platform

 - Edition: Personal Server
 - Version: v7.8.0
 - Browser and OS: Tried multiple browsers (Chrome, Firefox, Safari, on Macs)

## Additional context (optional)

When the other users log out, it stops occurring. If only one user is logged in and viewing the page, it's fine. When multiple users are viewing the page, this occurs.

We are on a personal server, using Docker.

johncadengo:
This is what the inspector console looks like on Safari.

https://user-images.githubusercontent.com/561587/225161874-904517c4-a454-4903-8f62-ed1515f470c6.mov

The line of code:

<img width="1435" alt="Screen Shot 2023-03-14 at 4 08 35 PM" src="https://user-images.githubusercontent.com/561587/225161989-a6d6bcca-76f7-4814-873d-20a5698b40fc.png">

johncadengo:
The docker compose yaml is very simple:

```
version: "3"
services:
  app:
    image: mattermost/focalboard:latest
    container_name: focalboard
    volumes:
      - fbdata:/opt/focalboard/data
    ports:
      - 8000:8000
    environment:
      - VIRTUAL_HOST=focalboard.local
      - VIRTUAL_PORT=8000
    restart: always

volumes:
  fbdata:
```

johncadengo:
I built the latest docker image and updated to the latest commit on main: https://github.com/mattermost/focalboard/commit/d8e1fb4832953fb31c54d4bc3542e2e4d4fe0fb7

It says v7.10.0 now. It still has the same behavior and this bug has not been fixed.

johncadengo:
As a workaround, I've disabled websockets support, and it stops jittering.

I'm behind an Nginx Proxy, and disabled websockets on the nginx level. I'm not sure if this will cause unintentional errors though. Is that the case?

johncadengo:
Ok, just tested it without web sockets and it essentially breaks all the features. So I have to have web sockets enabled. But it jitters which makes it impossible to click anything in the sidebar.

johncadengo:
[...]

PR #4672, opened 2023-03-28
only send category updates to user

sbishel:
#### Summary
Fixes issue where an update from one user causes cascading updates back and forth. Fix is to only send Category/CategoryBoards update to user. Categories are now individual, not teams.

#### Ticket Link
  Fixes https://github.com/mattermost/focalboard/issues/4644


mgdelacroix (approved):
LGTM! :+1:

Issue #4719, opened 2023-04-22
Bug: When moving a card to a different status, its properties are deleted


StownCraft:
## Steps to reproduce the behavior

1. Create a card
2. Specify any property
3. Move the card to a different status (the bug is reproduced if the card is moved to another card)
4. Properties have been deleted

## Expected behavior

Properties should not be deleted

## Screenshots (optional)

I have made a screen recording, how can I reproduce the problem:

**Mattermost Boards**:

https://user-images.githubusercontent.com/13101480/233806701-7db8e805-e617-4a05-a5d6-5ca872392768.mov

**Focalboard**

https://user-images.githubusercontent.com/13101480/233806861-105c2302-da78-4cf4-873f-06c1533cb8dd.mov

## Edition and Platform

I tested the problem on the following software (probably the error occurs on other versions as well):
 - Edition: Focalboard Server / Mattermost Boards (plugin) self-hosted
 - Version: Focalboard v7.9.2, Mattermost Boards 7.10.0
 - Browser and OS: Chrome on Mac

PR #4725, opened 2023-04-27
fix so dropping on card has latest cards objects

sbishel:
#### Summary
Fixes issue so that dropping on cards has latest card data and doesn't lose recent changed.

#### Ticket Link

  Fixes https://github.com/mattermost/focalboard/issues/4719

### DIFF
--- a/webapp/src/components/kanban/kanban.tsx
+++ b/webapp/src/components/kanban/kanban.tsx
@@ -203,7 +203,7 @@ const Kanban = (props: Props) => {
             await Promise.all(awaits)
             await mutator.changeViewCardOrder(props.board.id, activeView.id, activeView.fields.cardOrder, cardOrder, description)
         })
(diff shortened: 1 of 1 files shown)


sbishel:
@mgdelacroix - Any idea why this one works, https://github.com/mattermost/focalboard/pull/4726

But this one fails?

mgdelacroix:
@sbishel there is a branch with the same name in `mattermost-server` for this one, so the CI instead of downloading the pre-mono-repo version, downloads the branch one which doesn't work. Either we wait until the server branch is merged and deleted (and some time more for the github cache to clear) or one of the branches needs to be deleted and reuploaded with a different name

mgdelacroix (approved):
LGTM! :+1: 

I've run both `npm run check` and `npm run test` and both seem to be working, so with that and the fact that the change is very small I'd say we can merge and fix the CI on `main` if something fails

mgdelacroix:
CI on webapp `main` run successfully

PR #4741, opened 2023-05-17
Backport fixes for import/export attachments and fixes

sbishel:
Backport of https://github.com/mattermost/mattermost-server/pull/22915

#### Summary
Initially intended to add attachment files to board exports (and handle the import of same). However, several other issues were discovered and fixed as well.

Duplicate Block (Card) - created the fileInfo object but did not duplicate the file. So all cards duplicated would share a single file.
Duplicate Board - it tried to copy files, but failed because it was only looking for files based on earlier path defined as `teamID/boardID/fileName`. It did not use fileInfo to look up file paths at all.

Both of there were fixed and made to share code along with updating during import.

Also added code to be able to retrieve the file path if the fileInfo row exists or use old file path if not. 

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-51281

#### Release Note
```release-note
Adds attachments to import/export
```

### DIFF
--- a/server/app/boards.go
+++ b/server/app/boards.go
@@ -185,8 +185,13 @@ func (a *App) DuplicateBoard(boardID, userID, toTeam string, asTemplate bool) (*
 	}
 
 	// copy any file attachments from the duplicated blocks.
-	if err = a.CopyCardFiles(boardID, bab.Blocks); err != nil {
-		a.logger.Error("Could not copy files while duplicating board", mlog.String("BoardID", boardID), mlog.Err(err))
+	err = a.CopyAndUpdateCardFiles(boardID, userID, bab.Blocks, asTemplate)
+	if err != nil {
+		dbab := model.NewDeleteBoardsAndBlocksFromBabs(bab)
+		if err = a.store.DeleteBoardsAndBlocks(dbab, userID); err != nil {
+			a.logger.Error("Cannot delete board after duplication error when updating block's file info", mlog.String("boardID", bab.Boards[0].ID), mlog.Err(err))
+		}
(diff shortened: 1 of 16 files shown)


server/app/boards.go
```diff
@@ -185,8 +185,13 @@ func (a *App) DuplicateBoard(boardID, userID, toTeam string, asTemplate bool) (*
...
 
 	// copy any file attachments from the duplicated blocks.
-	if err = a.CopyCardFiles(boardID, bab.Blocks); err != nil {
-		a.logger.Error("Could not copy files while duplicating board", mlog.String("BoardID", boardID), mlog.Err(err))
+	err = a.CopyAndUpdateCardFiles(boardID, userID, bab.Blocks, asTemplate)
+	if err != nil {
+		dbab := model.NewDeleteBoardsAndBlocksFromBabs(bab)
+		if err = a.store.DeleteBoardsAndBlocks(dbab, userID); err != nil {
+			a.logger.Error("Cannot delete board after duplication error when updating block's file info", mlog.String("boardID", bab.Boards[0].ID), mlog.Err(err))
+		}
```
mgdelacroix, line 193: nit: we can use a different `err` variable not to return the inner error instead of the copy one right after this block
```suggestion
		if dErr := a.store.DeleteBoardsAndBlocks(dbab, userID); dErr != nil {
			a.logger.Error("Cannot delete board after duplication error when updating block's file info", mlog.String("boardID", bab.Boards[0].ID), mlog.Err(dErr))
		}
```

mgdelacroix (approved):
LGTM! :rocket:

PR #4788, opened 2023-06-21
Updates import to read with a scanner

mgdelacroix:
#### Summary
This PR updates the import reading logic to use a scanner. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53232

### DIFF
--- a/server/app/import.go
+++ b/server/app/import.go
@@ -26,7 +26,8 @@ const (
 )
 
 var (
-	errBlockIsNotABoard = errors.New("block is not a board")
+	errBlockIsNotABoard  = errors.New("block is not a board")
+	errSizeLimitExceeded = errors.New("size limit exceeded")
 )
 
 // ImportArchive imports an archive containing zero or more boards, plus all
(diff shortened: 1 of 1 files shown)


server/app/import.go
```diff
@@ -20,8 +20,9 @@ import (
 )
 
 const (
-	archiveVersion  = 2
-	legacyFileBegin = "{\"version\":1"
+	archiveVersion    = 2
+	legacyFileBegin   = "{\"version\":1"
+	importMaxFileSize = 1024 * 1024 * 70
```
sbishel, line 25: Where is this file size coming from? Should this be from the configuration setting?

mgdelacroix: This comes directly from the Mattermost [slack import service](https://github.com/mattermost/mattermost/blob/master/server/platform/services/slackimport/slackimport.go#L78), but I'm happy to make it configurable

mgdelacroix: After taking a look at how the configuration is generated, making this easy to configure from the plugin's perspective is a bit more involved than expected. We've agreed on reusing MM's approach and we can revisit and expand it later if necessary

sbishel (approved):
Agree, its not worth the extra work to make it configurable, matching channels value seems reasonable to me.

esarafianou (approved):

PR #4815, opened 2023-07-21
Adds a database migration to restore the fileinfos that are deleted

mgdelacroix:
#### Summary
This migration restores wrongly deleted `FileInfo`s for boards.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53714


sbishel (approved):
Much simpler...LGTM

server/services/store/sqlstore/migrations/000040_fix_fileinfo_soft_deletes.up.sql
```diff
@@ -0,0 +1,8 @@
+{{if .plugin}}
+  UPDATE FileInfo
+     SET DeleteAt = 0
+   WHERE CreatorId = 'boards'
+     AND DeleteAt != 0;
```
lieut-data, line 5: To confirm, does this mean we undelete everything boards has ever deleted? And that we'll figure out what /should/ be deleted in a separate PR?

lieut-data (approved):

server/services/store/sqlstore/migrations/000040_fix_fileinfo_soft_deletes.up.sql
sbishel: Correct. Prior to the fix a few weeks ago, no "boards" fileInfos were deleted, so it is essentially moving that fix to here. It was a forward only fix, as it was a low security issue.

PR #4820, opened 2023-07-24
Sanitize user following config for ShowFullName and ShowEmailAddress

enahum:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53190

### DIFF
--- a/server/api/users.go
+++ b/server/api/users.go
@@ -89,6 +89,18 @@ func (a *API) handleGetUsersList(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	ctx := r.Context()
+	session := ctx.Value(sessionContextKey).(*model.Session)
+	isSystemAdmin := a.permissions.HasPermissionTo(session.UserID, model.PermissionManageSystem)
+
+	for _, user := range users {
+		if user.ID == session.UserID {
+			user.Sanitize(map[string]bool{})
+		} else {
+			a.app.SanitizeProfile(user, isSystemAdmin)
+		}
+	}
+
 	usersList, err := json.Marshal(users)
 	if err != nil {
 		a.errorResponse(w, r, err)
@@ -170,6 +182,7 @@ func (a *API) handleGetMe(w http.ResponseWriter, r *http.Request) {
 		user.Permissions = append(user.Permissions, model.PermissionCreatePost.Id)
 	}
 
+	user.Sanitize(map[string]bool{})
 	userData, err := json.Marshal(user)
 	if err != nil {
 		a.errorResponse(w, r, err)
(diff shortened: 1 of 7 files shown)


jespino (approved):
LGTM

enahum:
@mgdelacroix any idea why the tests are failing?

esarafianou:
Is there a way to check from the plugin if the requesting user is the System Admin? Because currently the fix is indeed correct but it also applies to System Admins whereas it shouldn't. These two settings don't apply to system admins.

mgdelacroix:
@enahum tests are failing because we're calling `GetConfig` on a mock without setting the expectation first:

```
2023-07-24T14:17:06.8596890Z     mattermostauthlayer_test.go:22: Unexpected call to *mocks.MockServicesAPI.GetConfig([]) at /home/runner/work/focalboard/focalboard/focalboard/server/services/store/mattermostauthlayer/mattermostauthlayer_test.go:22 because: there are no expected calls of the method "GetConfig" for that receiver
2023-07-24T14:17:06.8597086Z --- FAIL: TestGetBoardsBotID (0.00s)
```

I'd suggest to create the config directly as you're only going to use those two properties, so there is no need to fill the rest of the values

enahum:
@mgdelacroix thanks.

Any idea on how to check if the user is an admin?

mgdelacroix:
@esarafianou @enahum the best path forward is to do the same we're doing in `mattermost`, which is to sanitize the profile at the API level, as there's we have the permissions service available: https://github.com/mattermost/mattermost/blob/master/server/channels/api4/user.go#L281-L285

mgdelacroix:
@enahum same as we do [in channels](https://github.com/mattermost/mattermost/blob/master/server/channels/web/context.go#L122), we can use the [permissions service](https://github.com/mattermost/focalboard/blob/main/server/services/permissions/permissions.go) to check if the user is an admin through the `HasPermissionTo` method. Here is [an example of use](https://github.com/mattermost/focalboard/blob/main/server/api/teams.go#L350)

server/api/users.go
```diff
@@ -170,6 +182,7 @@ func (a *API) handleGetMe(w http.ResponseWriter, r *http.Request) {
 		user.Permissions = append(user.Permissions, model.PermissionCreatePost.Id)
 	}
 
+	user.Sanitize(map[string]bool{})
```
mgdelacroix, line 185: As the user is fetching themselves, shouldn't we sanitize this profile with all the options set or at least with those coming from the config?

enahum: as far as I'm aware, this is the same as channels, for the same user we just remove the password and MFA secret

mgdelacroix (approved):

esarafianou (approved):
LGTM

PR #4951, opened 2023-12-08
Fix broken webapp unit tests

Dripcoding:
#### Summary
Fix broken jest unit tests for the web app. Run 'npm run test' to run all tests.

#### Ticket Link
N/A

Fixes #4944

PR #5008, opened 2024-05-29
[Refactor]: updated dependency for webapp

Rajat-Dabade:
Fixes https://mattermost.atlassian.net/browse/MM-58301


webapp/src/properties/updatedTime/__snapshots__/updatedTime.test.tsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`properties/updatedTime should match snapshot 1`] = `
   <div
     class="UpdatedTime octo-propertyvalue octo-propertyvalue--readonly"
   >
-    June 15, 2021, 4:22 PM
```
M-ZubairAhmed, line 8: is this change ok? do we want to have it with `at` now?

M-ZubairAhmed (changes requested):

webapp/src/properties/updatedTime/__snapshots__/updatedTime.test.tsx.snap
Rajat-Dabade: This is because of the `Intl` dependency. But somehow it was failing in `main` branch as well. Not sure why. But as it is a non-blocker change, will look into this in subsequent PR.

M-ZubairAhmed: Sounds good

M-ZubairAhmed (approved):

webapp/src/properties/updatedTime/__snapshots__/updatedTime.test.tsx.snap
marianunez: @Rajat-Dabade there was a PR opened to fix webapp tests here a while back https://github.com/mattermost/focalboard/pull/4951 - take a look, maybe it fixes the broken tests

Rajat-Dabade: @marianunez So after a little deeper investigation I found that this was failing because of bumping up the node's version from 16.10.0 to 21.11 in nvmrc to match it with the latest MM webapp. 

cc: @M-ZubairAhmed

hmhealey (approved):

Rajat-Dabade:
The GitHub workflow is failing due to the old go version i.e. 1.19, The Version will be bumped up in this PR: https://github.com/mattermost/focalboard/pull/5009

harshilsharma63 (approved):

PR #5009, opened 2024-05-30
[Refactor]: updated dependency for focalboard server

Rajat-Dabade:
Fixes: https://mattermost.atlassian.net/browse/MM-58301

### Dependency Update:
- Updated the Mattermost server dependency from `github.com/mattermost/mattermost-server/v6/public` to `github.com/mattermost/mattermost/server/public`.

### Go Version Upgrade:
- Bumped the Go version from 1.19 to 1.21.

### GitHub Workflow Updates:
- Updated Go dependencies in GitHub workflows.

### Code Cleanup:

- Removed unnecessary code files and their related implementations across Focalboard:
	- cloud.go: cloud limits are no longer applied as part of plugins, also removed the server-side code which was using cloud limits.
	- limit.go: Removed for the same reason as cloud.go.
	- insight.go: Removed as insights on boards are no longer supported.
	- statistics.go: Removed as statistics routes are no longer supported.


mattermost-plugin/server/boards/boards_service_api.go
```diff
@@ -81,4 +80,5 @@ func (bs *boardsServiceAPI) DuplicateBoard(boardID string, userID string,
 }
 
 // Ensure boardsServiceAPI implements product.BoardsService interface.
-var _ product.BoardsService = (*boardsServiceAPI)(nil)
+// Commented out this becuase we are not using boards as a product.
+// var _ product.BoardsService = (*boardsServiceAPI)(nil)
```
marianunez, line 84: If no longer used, should we just delete this whole file that is implementing `boardsServiceAPI`? Is this reference at all in the plugin? Seems that this was for the Mattermost server to use as part of MPA which is no longer needed

mattermost-plugin/server/boards/boardsapp_test.go
```diff
@@ -85,9 +85,9 @@ func TestSetConfiguration(t *testing.T) {
 
 	t.Run("test boards feature flags", func(t *testing.T) {
 		featureFlags := &model.FeatureFlags{
-			TestFeature:        "test",
-			TestBoolFeature:    boolTrue,
-			BoardsFeatureFlags: "hello_world-myTest",
+			TestFeature:     "test",
+			TestBoolFeature: boolTrue,
+			// BoardsFeatureFlags: "hello_world-myTest",
```
marianunez, line 90: If no longer needed, let's remove and not leave commented out

mattermost-plugin/server/boards/configuration.go
```diff
@@ -76,13 +76,16 @@ func (b *BoardsApp) OnConfigurationChange() error {
...
-	enableShareBoards := false
+	enableShareBoards := true // TODO: made enableShareBoards to default true as mmconfi.ProductSettings.EnablePublicSharedBoards need to be rechecked.
 	if mmconfig.PluginSettings.Plugins[PluginName][SharedBoardsName] == true {
 		enableShareBoards = true
 	}
-	if mmconfig.ProductSettings.EnablePublicSharedBoards != nil {
-		enableShareBoards = *mmconfig.ProductSettings.EnablePublicSharedBoards
-	}
+
+	// TODO: shift it to PluginSetting rather than ProductSettings.
```
marianunez, line 84: Seems that is being done above with checking `mmconfig.PluginSettings.Plugins[PluginName][SharedBoardsName]` ? We may just need to remove overriding the setting with the product settings

mattermost-plugin/server/boards/configuration_test.go
```diff
@@ -55,7 +55,7 @@ func TestOnConfigurationChange(t *testing.T) {
 	basePlugins[PluginName][SharedBoardsName] = true
 
 	baseFeatureFlags := &serverModel.FeatureFlags{
-		BoardsFeatureFlags: "Feature1-Feature2",
+		// BoardsFeatureFlags: "Feature1-Feature2", // TODO: add feature flag for boards
```
marianunez, line 58: We ne longer need a feature flag given that it will not be embedded in the product. The feature flag was a way to disable the product within the MPA. Given that now it is a uploaded plugin, it can be disabled by disabling the plugin. 

We can remove this feature flag.

server/api/api.go
```diff
@@ -96,7 +94,6 @@ func (a *API) RegisterRoutes(r *mux.Router) {
 	a.registerBoardsRoutes(apiv2)
 	a.registerBlocksRoutes(apiv2)
 	a.registerContentBlocksRoutes(apiv2)
-	a.registerStatisticsRoutes(apiv2)
```
marianunez, line 99: Who was using this statistics? Are we certain it's no longer needed?

server/app/blocks_test.go
```diff
@@ -199,11 +199,7 @@ func TestIsWithinViewsLimit(t *testing.T) {
 	t.Run("within views limit", func(t *testing.T) {
 		th.Store.EXPECT().GetLicense().Return(fakeLicense)
 
-		cloudLimit := &mmModel.ProductLimits{
-			Boards: &mmModel.BoardsLimits{
-				Views: mmModel.NewInt(2),
-			},
-		}
+		cloudLimit := &mmModel.ProductLimits{}
```
marianunez, line 202: We should remove all related to ProductLimits. The `isWithinViewsLimit` check below is all commented out and returning `true` . We can remove that as well.

server/app/cloud.go
```diff
@@ -4,94 +4,18 @@
...
-	return false
-}
-
 // SetCloudLimits sets the limits of the server.
 func (a *App) SetCloudLimits(limits *mmModel.ProductLimits) error {
 	oldCardLimit := a.CardLimit()
 
 	// if the limit object doesn't come complete, we assume limits are
 	// being disabled
 	cardLimit := 0
```
marianunez, line 18: There is no need to set or handle any cloud limits, we can remove this whole file

server/model/board_insights.go
```diff
@@ -4,12 +4,11 @@ import (
 	"encoding/json"
 	"io"
 
-	mmModel "github.com/mattermost/mattermost-server/v6/model"
+	mmModel "github.com/mattermost/mattermost/server/public/model"
 )
 
 // BoardInsightsList is a response type with pagination support.
 type BoardInsightsList struct {
```
marianunez, line 11: Boards insights seems that was also just used for Insights that has been deprecated. We should be able to delete this whole file as well.

webapp/src/components/viewHeader/viewHeader.tsx
```diff
@@ -127,8 +127,9 @@ const ViewHeader = (props: Props) => {
             return true
         }
 
-        setShowViewLimitDialog(true)
-        return false
+        // TODO (clean-up): as limit are removed we have to clean up the following code.
+        setShowViewLimitDialog(false)
```
marianunez, line 131: Yes, we need to remove all limits dialogs

webapp/src/components/viewHeader/viewHeader.tsx
```diff
@@ -127,8 +127,9 @@ const ViewHeader = (props: Props) => {
...
+        return true
```
marianunez, line 132: This should probably remain as `false`?

marianunez (changes requested):
Great work @Rajat-Dabade! Let some comments below and there is still more limits and insights code we can delete.

mattermost-plugin/server/boards/boards_service_api.go
Rajat-Dabade: Okay, I have Checked and seems like this file no longer has any references in the focalboard. Removed it.

mattermost-plugin/server/boards/configuration.go
Rajat-Dabade: Luckily `ProductSettings` was already deprecated.

mattermost-plugin/server/boards/configuration_test.go
Rajat-Dabade: Okay, removed it.

server/api/api.go
Rajat-Dabade: Nice catch, we need this. Reverting the changes back.

server/app/blocks_test.go
Rajat-Dabade: Sure.

server/app/cloud.go
Rajat-Dabade: The only function in use from this file is `GetUsedCardsCount`, but I don't think this function belongs here so I will be moving the function to the `statistics.go` file. And will delete `cloud.go`.

server/model/board_insights.go
Rajat-Dabade: Done.

webapp/src/components/viewHeader/viewHeader.tsx
Rajat-Dabade: This whole function should be removed as we no longer need the limit dialog.

webapp/src/components/viewHeader/viewHeader.tsx
Rajat-Dabade: Done.

server/app/statistics.go
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package app
+
+func (a *App) GetUsedCardsCount() (int, error) {
+	return a.store.GetUsedCardsCount()
+}
```
marianunez, line 8: Was this migrated from another file?

server/services/store/sqlstore/data_migrations.go
```diff
@@ -868,8 +869,17 @@ func (s *SQLStore) doesDuplicateCategoryBoardsExist() (bool, error) {
 }
 
 func (s *SQLStore) runMySQLDeDuplicateCategoryBoardsMigration() error {
-	query := "DELETE FROM " + s.tablePrefix + "category_boards WHERE id NOT IN " +
-		"(SELECT * FROM ( SELECT min(id) FROM " + s.tablePrefix + "category_boards GROUP BY user_id, board_id ) as data)"
+	validatedTablePrefix := s.tablePrefix
```
marianunez, line 872: nit: doesn't seem there was any validation done on this, could we remove that in the var name so it's not misleading?
```suggestion
	tablePrefix := s.tablePrefix
```

server/services/store/sqlstore/migrate.go
```diff
@@ -68,7 +68,7 @@ func (s *SQLStore) getMigrationConnection() (*sql.DB, error) {
 	}
 	*settings.DriverName = s.dbType
 
-	db := sqlstore.SetupConnection("master", connectionString, &settings)
+	db, _ := sqlUtils.SetupConnection(s.logger, "master", connectionString, &settings, 5) // TODO: Need to find a way to get DBPingAttempts parameter from server
```
marianunez, line 71: Do we still need to get the DBPingAttempts configuration from the server?

webapp/src/components/cardDetail/cardDetail.tsx
```diff
@@ -116,8 +114,7 @@ const CardDetail = (props: Props): JSX.Element|null => {
     saveTitleRef.current = saveTitle
     const intl = useIntl()
 
-    const clientConfig = useAppSelector<ClientConfig>(getClientConfig)
-    const newBoardsEditor = clientConfig?.featureFlags?.newBoardsEditor || false
```
marianunez, line 120: Not sure why this is under feature flags but seems that this is some config setting different than Boards Product feature flag?

webapp/src/components/onboardingTour/tourTipRenderer/tourTipRenderer.tsx
```diff
@@ -33,19 +31,17 @@ type Props = {
...
     const onboardingTourStarted = useAppSelector(getOnboardingTourStarted)
     const onboardingTourCategory = useAppSelector(getOnboardingTourCategory)
     const onboardingTourStep = useAppSelector(getOnboardingTourStep)
-    const disableTour = clientConfig?.featureFlags?.disableTour || false
 
     if (props.showForce) {
         isOnboardingBoard = true
     }
 
-    const showTour = !disableTour && isOnboardingBoard && onboardingTourStarted && onboardingTourCategory === props.category
```
marianunez, line 48: We do want to keep checking this setting given that this will not show the onboarding tour in Boards if it was disabled in the server

webapp/src/route.tsx
```diff
@@ -28,12 +26,11 @@ function FBRoute(props: RouteProps) {
     const loggedIn = useAppSelector<boolean|null>(getLoggedIn)
     const me = useAppSelector<IUser|null>(getMe)
     const myConfig = useAppSelector(getMyConfig)
-    const clientConfig = useAppSelector<ClientConfig>(getClientConfig)
 
     let redirect: React.ReactNode = null
 
     // No FTUE for guests
-    const disableTour = me?.is_guest || clientConfig?.featureFlags?.disableTour || false
```
marianunez, line 36: Same here, we want to still check this

marianunez (commented):
Looking great @Rajat-Dabade 

Left some minor comments but one thing we do want to double check is that some settings are coming through the `clientConfig.FeatureFlags` which we do want to keep. The only FeatureFlag we don't want to check any longer is the one where it determined if Boards was running as a product in the server or not (the old MPA model)

server/app/statistics.go
Rajat-Dabade: Yes from `cloud.go`.

server/services/store/sqlstore/migrate.go
Rajat-Dabade: We can add the configuration for `dbPingAttemps` in `config.json` in focalboard server. This can do the job.

server/services/store/sqlstore/data_migrations.go
Rajat-Dabade: Done.

webapp/src/components/cardDetail/cardDetail.tsx
Rajat-Dabade: What I remember is that @jespino was working on new boardsEditor, not sure whether it was public or not.

webapp/src/components/onboardingTour/tourTipRenderer/tourTipRenderer.tsx
Rajat-Dabade: Okay will be reverting this code back also for fetching `FeatureFlag` from server, but as of now we are not getting any flag from the server with the name `disableTour`.

server/services/store/sqlstore/data_migrations.go
marianunez: Are you missing a push of the changes? I still see the same

Rajat-Dabade: It's already there, it's a part of this commit https://github.com/mattermost/focalboard/pull/5009/commits/ae9413a2421036f4ce6e6b96b473e27802fe4d2d

marianunez: I see it now, it was a caching issue with Github

webapp/src/components/cardDetail/cardDetail.tsx
marianunez: I think for the purpose of this update, we should leave as is

marianunez (approved):
Aside the minor revert of the boardsEditor setting, everything LGTM. Great work @Rajat-Dabade 🎉

webapp/src/components/cardDetail/cardDetail.tsx
marianunez: Yes, let's not change that for now as it is not in scope and we don't have context yet on that new editor effort

Rajat-Dabade:
Reverted the changes for BoardEditor setting back. Thanks 🙌 @marianunez

harshilsharma63 (approved):

PR #5015, opened 2024-06-12
MM-58502 Error on crossing block limit on client side

Rajat-Dabade:
Fixes: https://mattermost.atlassian.net/browse/MM-58502

This PR implements a client-side check for the block limit on the card. Instead of failing silently without any error message, the front end will display an error if the block limit is exceeded.

User flow: 
- An error will be shown if the user crosses the block limit on the client side. 
- Error will persist if the user loses the focus from the block and no API call will be made. 
- If users close the card with an error then the data will be lost. 

https://github.com/mattermost/focalboard/assets/26198062/1abc8038-7005-4204-b808-c8ae3d3b19e6

I also updated the `mysql/mysql-server` docker version because version `5.7.12` was deprecated.

### DIFF
--- a/webapp/src/components/content/textElement.tsx
+++ b/webapp/src/components/content/textElement.tsx
@@ -31,12 +31,8 @@ const TextElement = ({block, readonly}: Props): JSX.Element => {
 
     const textChangedHandler = (text: string): void => {
         setBlockTitle(text)
-        const textSize = new Blob([text]).size
-        if (textSize > BlockTitleMaxRunes) {
-            setIsError(true)
-        } else {
-            setIsError(false)
-        }
+        const textSize = text.length
+        setIsError(textSize > BlockTitleMaxRunes)
     }
 
     const handleBlur = (text: string): void => {
--- a/webapp/src/components/content/textElement.tsx
+++ b/webapp/src/components/content/textElement.tsx
@@ -1,7 +1,9 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
-import React from 'react'
+import React, {useState} from 'react'
 import {useIntl} from 'react-intl'
+
+import cx from 'classnames'
 
 import {ContentBlock} from '../../blocks/contentBlock'
 import {createTextBlock} from '../../blocks/textBlock'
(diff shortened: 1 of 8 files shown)


webapp/src/components/content/textElement.tsx
```diff
@@ -11,26 +13,56 @@ import {MarkdownEditor} from '../markdownEditor'
...
+const TextElement = ({block, readonly}: Props): JSX.Element => {
     const intl = useIntl()
 
+    const [isError, setIsError] = useState<boolean>(false)
+    const [blockTitle, setBlockTitle] = useState(block.title)
+
+    const textChangedHandler = (text: string): void => {
+        setBlockTitle(text)
+        const textSize = new Blob([text]).size
+        if (textSize > BlockTitleMaxRunes) {
```
harshilsharma63, line 35: nit: can be simplified to 

```
setIsError(textSize > BlockTitleMaxRunes)
```

webapp/src/components/content/textElement.tsx
```diff
@@ -11,26 +13,56 @@ import {MarkdownEditor} from '../markdownEditor'
...
+
+const TextElement = ({block, readonly}: Props): JSX.Element => {
     const intl = useIntl()
 
+    const [isError, setIsError] = useState<boolean>(false)
+    const [blockTitle, setBlockTitle] = useState(block.title)
+
+    const textChangedHandler = (text: string): void => {
+        setBlockTitle(text)
+        const textSize = new Blob([text]).size
```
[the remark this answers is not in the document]
hmhealey: I'm pretty sure we just check `text.length`. I'm not 100% sure, but that seems to account for unicode characters taking up more than one byte
Screenshot 2024-06-13 at 5 48 29 PM

webapp/src/components/content/textElement.tsx
```diff
@@ -11,26 +13,56 @@ import {MarkdownEditor} from '../markdownEditor'
...
-const TextElement = (props: Props): JSX.Element => {
-    const {block, readonly} = props
+const BlockTitleMaxBytes = 65535 // Maximum size of a TEXT column in MySQL
+const BlockTitleMaxRunes = BlockTitleMaxBytes / 4 // Assume a worst-case representation
+
+const TextElement = ({block, readonly}: Props): JSX.Element => {
     const intl = useIntl()
 
+    const [isError, setIsError] = useState<boolean>(false)
+    const [blockTitle, setBlockTitle] = useState(block.title)
```
hmhealey, line 30: Why store `blockTitle` in a `useState`? Couldn't we just use `props.title` where that's currently used?

webapp/src/components/content/textElement.tsx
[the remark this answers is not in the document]
Rajat-Dabade: I suppose we can use `text.length` directly. Initially, I thought this would be for file uploads like images on the block, but file uploads use different APIs. I'll make those changes. Nice catch 🙌.

webapp/src/components/content/textElement.tsx
Rajat-Dabade: Initially, the `onblur` event for the input field made an API call that replaced the new block.title with the old one. The problem with this approach was that if the API call returned an error, the block.title would revert to the old data, causing the newly typed data to be lost. To ensure the new data is retained even after an error, the `blockTitle` state was introduced.

webapp/src/components/content/textElement.tsx
Rajat-Dabade: Done.

harshilsharma63 (approved):

hmhealey (approved):

yasserfaraazkhan (approved):
verified uploading plugin to 9.10.0 server. The expected error is display.

PR #5022, opened 2024-07-04
stop logging errors for api error responses

lieut-data:
#### Summary
From a cursory glance, when a real error occurs, we already log something at the application layer, so there's no reason to echo every api error response as an error log too. Errors should be actionable, but these logs were not, e.g.:

> access denied to templates

or:

> category ID specified in input does not exist for user

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-59325

### DIFF
--- a/server/api/api.go
+++ b/server/api/api.go
@@ -178,7 +178,6 @@ func (a *API) userIsGuest(userID string) (bool, error) {
 // Response helpers
 
 func (a *API) errorResponse(w http.ResponseWriter, r *http.Request, err error) {
-	a.logger.Error(err.Error())
 	errorResponse := model.ErrorResponse{Error: err.Error()}
 
 	switch {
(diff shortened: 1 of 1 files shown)


Rajat-Dabade (approved):
LGTM 👍

lieut-data:
@Rajat-Dabade, the test failure here seems spurious -- do you know if it's a known issue?

marianunez:
@lieut-data we've separated the Boards as a plugin to it's own repo at: https://github.com/mattermost/mattermost-plugin-boards This will be the repo we will be supporting and releasing Boards from moving forward while Focalboard will remain as community supported. 

Do you want to move this PR there so your changes can make the next plugin release?

Rajat-Dabade:
> @Rajat-Dabade, the test failure here seems spurious -- do you know if it's a known issue?

@lieut-data  This has something to do with MySQL docker image deprecated version. Bumping up the image version might resolve the issue.

lieut-data:
Aha, thank you @marianunez! The new repo was a blindspot for me. I'll copy the PR over there: should I keep it here as well?

marianunez:
> should I keep it here as well?

@lieut-data Is this relevant for Focalboard as a standalone application or just for plugin logs in MM Server? It it's the latter, then I wouldn't think it's relevant to keep here.

lieut-data:
> > should I keep it here as well?
> 
> @lieut-data Is this relevant for Focalboard as a standalone application or just for plugin logs in MM Server? It it's the latter, then I wouldn't think it's relevant to keep here.

I guess from a log quality perspective, the changes might be relevant, but as we're not intending to keep the standalone and plugin in sync, I'll just close this out.

Issue #4843, opened 2022-09-08
Bug: Empty cards are created when no changes are made


hmhealey:
## Steps to reproduce the behavior

1. Click on the "+ New" button to make a new card
2. Immediately close the modal without making any changes

## Expected behavior

A card isn't created automatically unless the user enters some text or makes some changes. It feels too easy to accidentally create cards that you didn't mean to.

## Screenshots (optional)

![Screen Shot 2022-09-08 at 11 32 24 AM](https://user-images.githubusercontent.com/3277310/189163917-2134205c-c3bb-42fa-9c00-53e6b86870af.png)

## Edition and Platform

 - Edition: Mattermost Boards on community.mattermost.com
 - Version: v7.2.1

ryanrudzitis:
I’ll take a look at this!

Lokendra-sinh:
@jav974 can I take it up?

jav974:
> @jav974 can I take it up?

@Lokendra-sinh what do you mean ? i submitted a PR so we'll see

hmhealey:
@jav974 I can assign you to this to make it clear that your PR is the fix for this. I don't know when we'll be able to get to reviewing it since there's no one actively maintaining this repo, but we've been looking for a maintainer for it. Once we find someone, I'm sure they'll be happy that there's already a fix for this waiting

Gaurav-yadav-03:
Hi, I’d like to work on this. Please assign me.

aniirathod:
Is this issue is not solve yet ?

PR #5041, opened 2024-10-02
[GH-4843] Deletes newly created card after close if no interaction were made in

jav974:
#### Summary

After creating a new card, if user does not type into any field or interact with any of the different options and closes the dialog, the card gets removed immediately.

#### Ticket Link

Fixes https://github.com/mattermost-community/focalboard/issues/4843

### DIFF
--- a/webapp/cypress/integration/createBoard.ts
+++ b/webapp/cypress/integration/createBoard.ts
@@ -161,9 +161,10 @@ describe('Create and delete board / card', () => {
         }
 
         // Create empty card in last group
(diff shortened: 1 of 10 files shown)


jav974:
The implementation differs from the discussion on the ticket, in that the card gets created serverside as usual, but it gets removed if the modal is closed and no interaction were made.
The card is not destroyed if user types something in a field, then deletes it, or adds a new property (even empty) or attachment.

/claim #4843

